PR #4, opened 2018-05-09
feat: prepare for @zendeskgarden npm publish

jzempel:
* [x] **BREAKING CHANGE?** 

## Description

Package location has moved to the npm registry under the `@zendeskgarden` scope. See the README for installation/upgrade details.

## Detail

Additionally:

* apply license header to all JS files
* fill-out README npm versions & dependency checks
* rename `yarn build:styleguide` to `yarn build:demo` (more straightforward)

## Checklist

      designer as a reviewer)
  component


.travis.yml
```diff
@@ -9,22 +9,15 @@ cache:
...
-  - npm config set //zdrepo.jfrog.io/zdrepo/api/npm/npm/:username $ARTIFACTORY_USERNAME
-  - npm config set //zdrepo.jfrog.io/zdrepo/api/npm/npm/:email $ARTIFACTORY_EMAIL
-  - npm config set //zdrepo.jfrog.io/zdrepo/api/npm/npm/:always-auth true
-
 install: yarn --frozen-lockfile
 
 script:
   - yarn lint
-  - yarn test:all
+  - yarn test:all --runInBand
```
austin94: Is there a reason we moved the `--runInBand` up to Travis? `yarn test` is what the majority of development would be using. I think having that abstracting it out could make it more difficult once we enable `--coverage`.

utils/travis/publish-gh-pages.js
```diff
@@ -1,24 +1,15 @@
...
-      console.log(error);
-      throw error;
-    }
 
-    console.log('GH-Pages upload complete.');
-  }
-);
+const ghPages = require('gh-pages');
+
+ghPages.publish('demo', {
```
austin94: I know we originally decided to remove the `add: true` option for ghPages, but I think we forgot about the "pre-publish" steps we do sometime to shared changes.

Without the `add` option those pre-published files will be wiped out. I think leaving those files around will be helpful.

.travis.yml
jzempel: I guess it depends on what you're doing whether you're using `yarn test` or `yarn test:all`. But when running `test:all` its preferable to take advantage of a multi-core computer (18s with `--runInBand`; 8s without). I'm sure we can finesse `--coverage` when we get to that.

utils/travis/publish-gh-pages.js
jzempel: Ok, let's leave that as a conversation following this PR.

Issue #10, opened 2018-05-11
how to use with nextjs


alidcastano:
## Expectations
Should work after configuring css modules:

```
const withCss = require('@zeit/next-css')
module.exports = withCss({
  cssModules: true
})
```

## Reality

Getting CSS compilation error for View components. For the Container components I'm getting `Element is not defined error.`

## Steps to Reproduce

Any next starter with Garden.

austin94:
👋 Hi @alidcastano, thanks for opening this issue.

The last time I did any development with `next.js` was a long time ago, but if I remember correctly, it was _uniquely_ difficult to get external CSS (and css-modules) working correctly.

I haven't used their `next-css` package since it's been introduced so I'm going to do some research and post my findings here.

austin94:
After doing some research into this I found a variety of issues.

1. Due to our usage of css-modules a user would need to configure `withCss` to run against `/node_modules/` as well as local imports
1. The `ThemeProvider` component in `@zendeskgarden/react-theming` includes some configuration that will fail in a node environment

## CSS-Modules

I was able to get this working correctly by following the [official with-next-css example](https://github.com/zeit/next.js/tree/canary/examples/with-next-css). You can use the `webpack` config option to update the loaders to allow it to resolve for `/node_modules/` as well as local files.

This should fix the `.c-btn is not defined` error you would have been seeing in the Views.

## Theming issues

Within our `ThemeProvider` component we are embedding our own `StyleSheetManger` component from `styled-components`. This should ideally only be provided by the consumer as a wrapper around the root React component.

Additionally, we are referencing some browser only APIs like `document.head` and `Element`. This is whats causing the `Element is not defined...` errors.

@ryanseddon what do you think?

alidcastano:
Hey @austin94, what exactly did you edit? I wasn't able to get it to work. E.g. I tried getting the ccs rule and doing `rule.include =/node_modules\/@zendeskgarden\/css/`.

In regards to SSR, I'd love to see this addressed since it's something I'll be needing for my project. But for now I've started out with create-react-app. 

Thanks!

ryanseddon:
Not making assumptions about running in a browser context is definitely something we should do. 

I've created a card on our internal backlog to look at this, haven't prioritised it yet.

sunesimonsen:
I think it makes sense to create a FAQ with entries like: Running Garden React Components with CSS Modules, ...

austin94:
> I think it makes sense to create a FAQ

[...]

PR #19, opened 2018-05-17
[EXPERIMENT] feat(build): build with webpack and remove css-loader dependency

austin94:
* [x] **BREAKING CHANGE?** 

## Problem

Currently we only build the react-components with a simple babel-transform. This requires all consumers to configure their own webpack loader for CSS modules. This requirement means that consumers that are using SSR technologies like `next.js` to configure css-modules (which is non-trivial) and those using `create-react-app` to eject.

This isn't ideal because it raises the bar of entry and adds several "gotchas" for tooling like Jest and any combination of tooling that isn't necessarily document-able.

## Possible Solutions

While trying to solve this solution I broke it into 2 main problems: 
* removing the need for a non-standard css-loader webpack change
* reduce bundle size

### Removing CSS-Loader configuration

We currently require a custom webpack config to enable `css-modules`. This allows us to not worry about clashing class names. 

Rather than having the css-modules handled in the consuming application, I propose that we **compile CSS-modules with webpack and include a `styles.css` file for consumers to include however they wish.**

This concept is used in a few other libraries including https://reactstrap.github.io/ (without css-modules though, but same concept).

Pros:
* Works with standard webpack configs. No ejecting from create-react-app 🎉 
* Can be added in a variety of ways, including with a standard `<link>` tag
* Wont cause any weirdness with Jest, other tooling, or server-side-rendering
  * Would be an initial step towards helping with #10 
* Now that the CSS is split for each component there should be no duplication in styles

Cons:
* Requires an additional import for each packages styling
* For packages that rely on other react packages `(Select -> { Menus, TextFields })` we would bundle all of these styles. This may cause some limited duplication, but due to the compact CSS size isn't terrible.

An example usage would be:

```
// Once at the root of the application (along with other polyfills and 3rd party styling)
import '@zendeskgarden/react-buttons/dist/styles.css';

import { Button } from '@zendeskgarden/react-buttons';
```

### Reduce Bundle Size

With the introduction of webpack and bundling, the size of the packages is always a concern. To help address this I've added the following:

* Marking all `/node_modules/` packages as `external`
  * This removes all items from the bundle and will be handled in the consuming bundler
* Previously we were embedding the entire `package.json` which for every package just to get the current version 👻 
  * To remove this dependency I've added in a webpack global with the current package version

## Detail

This PR is split into several commits:
* `feat(build): update build to remove css-loader dependency`
  * is the bulk of the build changes relating to the new build
  * update styleguidist code examples to consume `styles.css` for devDependencies used in examples
* `feat(build): remove full package.json import`
  * Use the new `PACKAGE_VERSION` webpack global
  * A lot of snapshot test updates due to use consuming the dist rather than src for each dependency

With these changes it's now trivial to use us in create-react-app and the bundle size is working out well even for areas with multiple CSS dependencies.

screen shot 2018-05-16 at 10 32 02 pm

## Standing Questions

What type of bundle do we want to provide? commonjs, umd, full bundle with all dependencies for browser usage from a CDN?

Is this something we want to pursue?

## Checklist

      designer as a reviewer)
  component


sunesimonsen:
I'll have a look later today.

austin94:
@sunesimonsen and I had an offline talk and found a pretty cool solution that will make this all a lot simpler.

We have historically not used `style-loader` due to it injecting the styles into the head by default. I had thought that this was only configurable at build time, but after some research I think we can allow consumers to determine where it gets injected at runtime!

style-loader has an [insertInto](https://github.com/webpack-contrib/style-loader#insertinto) option that can take a function. This is handled at runtime and even allows injection into a shadow-root for view encapsulation if needed. I think a good solution with this is check if a global function is provided (`window.__GARDEN_STYLE_INJECTOR__` or something similar) and use that do determine placement, but default to the standard `<head>` placement.

This allows the create-react-app usage without the additional CSS import.  Also simplifies a lot of the webpack config.

Thoughts?

sunesimonsen:
I think it sounds great. Could the global variable be less obscure 😜 It should be enough to namespace it with Garden.

jzempel:
Getting @austin94 to use a python [dunder](https://www.geeksforgeeks.org/dunder-magic-methods-python/). Priceless. My work is done here :stuck_out_tongue_winking_eye:

sunesimonsen:
Hehe just because everyone is making similar decisions doesn't make them less crazy. But I guess it is consistently crazy.

PR #21, opened 2018-05-21
feat(build): move to webpack build

austin94:
* [x] **BREAKING CHANGE?** This will change all consumption and import paths

## Problem

The current `@zendeskgarden/react-*` components have a simplistic build that only performs a babel transpilation.  Due to our requirement on css-modules, this means that consumers can't use these packages "out-of-the-box" as they might expect.

| Pros | Cons |
|--------------|--------------|
| Maximum flexibility | Consumers have to create a "non-standard" webpack config to handle CSS Modules |
|  Consumers can fully handle how css is injected, bundled, or chunked in their application. | This means that create-react-app users can't use Garden by default |
|  | Server-side-rendering is hard to accomplish with next.js and other technologies |
|  | Consumers have to create custom configs for testing |
|  | No browser bundles for non-webpack usages |

## Solutions

The end-goal of these changes are to provide a build process that allows a wider range of consumers (create-react-app and browser) to consume Garden, while still allowing advanced customizations like where/how to inject or chunk styling with webpack or rollup.

All solutions will include:
* A standard `commonjs` build with all dependencies marked as "external" (not included in the bundle)
* A `umd` bundle with all dependencies included and minified for browser usage
  * i.e. `<script src="https://unpkg.com/@zendeskgarden/react-theming/dist/bundle.min.js"></script>`
* Rather than including the entire `package.json` in the bundles 👻 use a `PACKAGE_VERSION` global provided by webpack

### Solution 1: Include css-modules with bundle and inject with style-loader

This solution uses `style-loader` to inject styles automatically for the consuming application.

| Pros | Cons |
|--------------|--------------|
| A single import includes component and styling | Unable to allow style chunking |
| Garden dependencies automatically include their styling | Redundant use of style-loader may bloat some smaller packages |
|  | Only able to customize injection point with a "hacky" global function |
|  | Not 100% usable in all server side rendering scenarios |

### 🏅 🏆 Solution 2: Bundle all css-modules in a style-sheet for consumption

This solution transpiles the css into a minified `styles.css`. This allows the consuming application to include it however it likes.

Used in common component libraries like:
* [Reactstrap](https://reactstrap.github.io/)
* [Polaris](https://polaris.shopify.com/)

Example:

```jsx
// Include once in your project
import '@zendeskgarden/react-buttons/dist/styles.css';
// Use like normal
import { Button } from '@zendeskgarden/react-buttons';
```

| Pros | Cons |
|--------------|--------------|
| Allows full customization of style inclusion | Requires an additional import for each package |
| Works with create-react-app |  |
| Aligns well with browser usage (just use a `<link>`) |  |
| Is more flexible if we eventually include multiple entry points |  |

## Detail

After doing several implementations for both of these solutions I personally think the `Solution 2` is a good path forward. Providing the most flexibility for some of our unique style-injection requirements within our products (web widget) and still working with `create-react-app`.

The commits are broken down to allow an easier view of changes:
* `feat(build): move to webpack build`
  * All of the build changes that will be interesting for reviewers
* `Update template files to match new style`
  * Change our `yarn new` template to include the new styling
* `Update shared package dependencies to be peerDependencies`
  * Move some of our dependencies to peerDependencies (styled-components, theming)
* `Update all PACKAGE_VERSION constants, README imports, and snapshot tests`
  * Update all of our code examples and snapshot tests now that styled-components injection is working 🎉 

Working in create-react-app:
screen shot 2018-05-17 at 12 38 38 pm

Browser only usage:
screen shot 2018-05-20 at 9 39 58 am

## Checklist

      designer as a reviewer)
  component


austin94:
@louisscruz from our offline conversation, could you take a look at the webpack additions in the first commit? Interested in seeing if these suggested changes align well with Support consumption.

utils/build/webpack.base.js
```diff
@@ -0,0 +1,88 @@
...
+const webpack = require('webpack');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
+const packageManifest = require(path.resolve('package.json'));
+
+const options = {
+  mode: 'production',
+  entry: path.resolve('src', 'index.js'),
+  devtool: 'source-map',
```
ryanseddon: Do we distribute the sourcemap?

utils/build/webpack.base.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+  plugins: [
+    /**
+     * Allows the `data-garden-version` attribute to be applied to all
+     * views without importing entire package.json into the bundle
+     */
+    new webpack.DefinePlugin({
+      PACKAGE_VERSION: JSON.stringify(packageManifest.version)
+    }),
+    new webpack.BannerPlugin(`Copyright Zendesk, Inc.
```
ryanseddon: You could use `trim` and format this a little nicer.

```js
    new webpack.BannerPlugin(`
Copyright Zendesk, Inc.

Use of this source code is governed under the Apache License, Version 2.0
found at http://www.apache.org/licenses/LICENSE-2.0
    `.trim()),
```

utils/scripts/build.sh
```diff
@@ -2,4 +2,6 @@
 set -x
 set -e
 
-gulp --gulpfile ../../utils/build/gulpfile.build.js --cwd ./
+rimraf dist
+webpack --config ../../utils/build/webpack.commonjs.js --hide-modules
+webpack --config ../../utils/build/webpack.umd.js --hide-modules
```
ryanseddon: We can trigger multiple configs in one calls

```bash
webpack --config ../../utils/build/webpack.commonjs.js  --config ../../utils/build/webpack.umd.js --hide-modules
```

packages/buttons/package.json
```diff
@@ -9,23 +9,29 @@
...
     "prop-types": "^15.6.1",
+    "react": "^0.14.0 || ^15.0.0 || ^16.0.0",
+    "react-dom": "^0.14.0 || ^15.0.0 || ^16.0.0",
     "styled-components": "^3.2.6"
   },
-  "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0 || ^16.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0 || ^16.0.0"
+  "devDependencies": {
+    "@zendeskgarden/css-buttons": "^6.0.0",
```
ryanseddon: So is the idea that due to the proposed change this needs to be explicitly included in your app css now hence it becoming a devDep?

packages/menus/README.md
```diff
@@ -9,20 +9,28 @@ This package includes components relating to menus in the
...
-<Menu
-  onChange={selectedKey => console.log(`Item selected: "${selectedKey}"`)}
-  trigger={({ ref }) => <button ref={ref}>Simple Example</button>}
->
-  <Item key="item-1">Item 1</Item>
-  <Item key="item-2">Item 2</Item>
-  <Item key="item-3">Item 3</Item>
-</Menu>;
+import { ThemeProvider } from '@zendeskgarden/react-theming';
+import { Menu, Item } from '@zendeskgarden/react-menus';
```
ryanseddon: Is it made clear that the consumer needs to import the CSS somewhere? 

These demos don't show that.

utils/build/webpack.base.js
austin94: With this setup yes. I don't see any downsides to providing it for the commonjs and UMD bundle. Do you?

utils/build/webpack.base.js
austin94: Fixed.

packages/buttons/package.json
austin94: The proposed change would remove `@zendeskgarden/css-buttons` from any dependency chain from the consumer.

They now only worry about importing the `react` styling (if there is any).

For buttons it would be:
* `import '@zendeskgarden/react-buttons/styles.css'`
  * This is a transformed version of `@zendeskgarden/css-buttons` with the CSS-modules already transformed for usage with the `react-buttons`
* `import { Button } from '@zendeskgarden/react-buttons'
  * Just like normal

The only reason it's a `devDependency` is for styleguidist and and the webpack build to create the `styles.css` file.

packages/menus/README.md
austin94: Nope. Forgot to add that to all of the README's.  This is now updated across all of the packages.

Outside the scope of this PR, but most other libraries include a `How to use with create-react-app` section. We might want to think of adding that to the root README with a general "always include CSS" message.

utils/build/webpack.umd.js
```diff
@@ -0,0 +1,62 @@
...
+  optimization: {
+    minimizer: [
+      new UglifyJsPlugin({
+        cache: true,
+        parallel: true,
+        sourceMap: true
+      })
+    ]
+  },
+  externals: {
```
louisscruz: It looks like some of the external names might be messed up. You might want to double check them.

utils/build/webpack.base.js
```diff
@@ -0,0 +1,92 @@
...
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
+const packageManifest = require(path.resolve('package.json'));
+
+const options = {
+  mode: 'production',
+  entry: path.resolve('src', 'index.js'),
+  devtool: 'source-map',
+  optimization: {
+    minimizer: [new OptimizeCSSAssetsPlugin({})]
```
jzempel: Configure with

```json
cssProcessorOptions: {
  discardComments: { removeAllButfirst: true }
}
```

austin94: Added.

package.json
```diff
@@ -4,7 +4,8 @@
   "version": "0.0.0",
   "scripts": {
     "build": "lerna run build --stream",
-    "build:demo": "lerna run build:demo --stream",
+    "build:analyze": "ANALYZE_BUNDLE=true yarn run build:single",
```
ryanseddon: This just exits and doesn't let me pick a package?

ryanseddon: Weird package issues, it works for me know

package.json
```diff
@@ -4,7 +4,8 @@
...
+    "build:demo": "yarn run build && lerna run build:demo --stream",
```
jzempel: minor nit: remove ~run~; so `yarn build`. Same with `yarn build:single` above.

austin94: Fixed.

jzempel:
@austin94 when it's time to cut a non-beta release, are there plans to update either/both `API.md` and `DEVELOPMENT.md` documentation with respect to these changes?

austin94:
@jzempel I was waiting to see if there was going to be any more comments/concerns that would change the process.

Now that everything seems clear I've updated the README and DEVELOPMENT docs to include the new build/consumption steps.

package.json
```diff
@@ -60,6 +62,7 @@
     "gulp-babel": "^7.0.1",
     "gulp-flatten": "^0.4.0",
```
jzempel: @austin94 you can remove the `gulp` dependencies, right?

austin94: Removed.

package.json
```diff
@@ -58,9 +58,6 @@
     "eslint-plugin-react": "^7.8.2",
     "fs-extra": "^6.0.1",
     "gh-pages": "^1.1.0",
-    "gulp": "^4.0.0",
-    "gulp-babel": "^7.0.1",
-    "gulp-flatten": "^0.4.0",
```
jzempel: `del` is in this set, too, right?

austin94: Good catch, removed.

PR #22, opened 2018-05-21
feat(tabs): add onChange event for selection

austin94:
## Description

`react-tabs` was one of the first packages created during the migration to the new architecture. Because of this, it's missing the `onChange` event that most of the newer components have (Menu, Select, Pagination, etc.) . Another issue is that the first example in the README uses an `onChange` event that doesn't exist yet 😢 

This PR introduces an `onChange` event for consumers that only want to control the currently selected Tab (which is most people).

## Checklist

      designer as a reviewer)
  component


eddiesholl:
Looks great, definitely a nice improvement for consumers 👍

PR #23, opened 2018-05-23
chore(build): update styleguidist to only build recently changed packages

austin94:
## Description

Currently the travis builds on `master` are running between 12-18 minutes per run. That isn't the best.

The bulk of time is from us building a separate `react-styleguidist` documentation site for every package on each run.  We could drastically decrease the build time by only building the styleguide for packages changed in the most recent commit.

## Detail

This PR introduces a new `build:demo_changed` yarn command that performs the `build:demo` command with the `--since HEAD^` lerna command to build across all of the updated packages.

This will continue to work as expected due to:
* we are using the gh-pages `add` option to not overwrite old docs
* if there are any interdependencies that change in other documentation areas (i.e. buttons that are used in other package examples) lerna will recognize the relations and build everything.

Depending on how many packages are touched in a commit, this should decrease the build times to around 4-5 minutes.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "scripts": {
     "build": "lerna run build --stream",
     "build:demo": "lerna run build:demo --stream",
-    "build:demo_changed": "yarn build:demo --since HEAD^",
     "build:single": "utils/scripts/scoped-npm-command.js --script build",
     "format": "yarn format:package_json && yarn format:js && yarn format:markdown",
     "format:js": "prettier --write 'packages/**/!(dist)/*.js' && prettier --write utils/**/*.js",
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "lerna run build --stream",
     "build:demo": "lerna run build:demo --stream",
+    "build:demo_changed": "yarn build:demo --since HEAD^",
     "build:single": "utils/scripts/scoped-npm-command.js --script build",
     "format": "yarn format:package_json && yarn format:js && yarn format:markdown",
     "format:js": "prettier --write 'packages/**/!(dist)/*.js' && prettier --write utils/**/*.js",
--- a/utils/travis/publish-gh-pages.sh
+++ b/utils/travis/publish-gh-pages.sh
@@ -2,5 +2,5 @@
 set -x
 set -e
 
-yarn build:demo --since
+yarn build:demo --since HEAD^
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "lerna run build --stream",
     "build:demo": "lerna run build:demo --stream",
+    "build:demo_changed": "yarn build:demo --since HEAD^",
```
jzempel: Probably don't need this. I think the command reads better inline in the shell script.

utils/travis/publish-gh-pages.sh
```diff
@@ -2,5 +2,5 @@
 set -x
 set -e
 
-yarn build:demo
+yarn build:demo_changed
```
jzempel: Just use `yarn build:demo --since` here. (Wouldn't `--since` be a better choice than `--since HEAD^`?)

package.json
austin94: Makes sense. Removed.

utils/travis/publish-gh-pages.sh
austin94: By default it will run against anything since the last tag https://github.com/lerna/lerna#--since-ref, which should be fine!

We originally had this in the publish script, but we changed it to `HEAD^`. Not sure what the reasoning was for that though.

Fixed.

jzempel: On the publish side, we need to prevent inadvertent publish attempts on packages that may have been excluded (for one reason or another) from the `yarn run tag` version updates. In this case, we want to be more aggressive about picking everything up for the demo build.

austin94:
@jzempel just realized there might be an issue with only using `--since`.  Since we are displaying the package version in the sidebar, this needs to be built.

With `--since` only it will ignore the version bump in the `chore(release)` commit. This means that the version will become stale.

With `--since HEAD^` it will recognize that version bump and perform another build.  Thoughts?

jzempel:
@austin94 couple things here.

* we should get rid of the sidebar version. The npm version is now prominently displayed in the header.
* now that I'm playing with it (via `yarn lerna updated`) I think you're right – it needs `--since HEAD^` to pick up anything at all.

austin94:
> we should get rid of the sidebar version. The npm version is now prominently displayed in the header.

I just removed the version from the sidebar. And update the `--since` inclusion.

PR #25, opened 2018-05-25
fix(grid): update CSS dependency to remove global styling

austin94:
## Description

There was some global styling leakage that was fixed in `@zendeskgarden/css-grids`. This PR bumps the dependency to include the fixes.

## Detail

Pre-published at: http://garden.zendesk.com/react-components/grid

(looks like gh-pages might be down globally, give it some time if it doesn't resolve)

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/grid/package.json
+++ b/packages/grid/package.json
@@ -16,7 +16,7 @@
     "start": "../../utils/scripts/start.sh"
   },
   "dependencies": {
-    "@zendeskgarden/css-grid": "^0.1.1",
+    "@zendeskgarden/css-grid": "^0.1.3",
     "@zendeskgarden/react-theming": "^1.0.0",
     "classnames": "^2.2.5",
     "prop-types": "^15.6.1",


packages/grid/package.json
```diff
@@ -16,7 +16,7 @@
     "start": "../../utils/scripts/start.sh"
   },
   "dependencies": {
-    "@zendeskgarden/css-grid": "^0.1.1",
+    "@zendeskgarden/css-grid": "^0.1.3",
```
dclowd9901: Isn't this already latest 0.1.x?

> Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4
>Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

austin94: In the current build consumers could use a frozen lockfile or a cache (like within lotus) which wouldn't update this dependency since it's still within the range. Still possible, but would have to rely on people to upgrade. By bumping the version it ensures that any consumers are on the correct fix.

With the new webpack build, this will be a dev-dependency and will be bundled within the packages `styles.css`.  Since that would only be created at build time, bumping this dependency would force a build and a new publish that includes the fixes.

dclowd9901: Cool, so that said, it might be a bit clearer to just make this a static version?

austin94: From our offline talk on the flight deck (sorry, that's the coolest Github comment I've gotten to post and just wanted to say it 😄) since we currently have open dependency values for all of our internal lerna dependencies it might make sense to leave everything open to keep it consistent.

I'll make an internal issue to track the comment though and we can revisit during sprint planning and if it gets picked up we will make the changes to all packages, not just react-grid.

PR #26, opened 2018-05-25
chore(lint): add jsx-a11y to linting and update failing accessibility rules

austingreendev:
## Description

I recently found the [jsx-a11y eslint plugin](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) and thought it would be a good bump for us to make sure we are being as accessible as possible.

It has some really cool rules and is descriptive about failures. Here is an example of a failure in one of our test files:

screen shot 2018-05-25 at 10 19 26 am

> why would we want accessibility in our tests?

Since we are normally testing how consumers use our components I think we should be testing everything in an accessible way, not just the files we are building.  Let me know if this is off-base and I'll ignore them.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/avatars/src/__snapshots__/Avatar.spec.js.snap
+++ b/packages/avatars/src/__snapshots__/Avatar.spec.js.snap
@@ -8,6 +8,7 @@ exports[`Avatar Sizes applies large correctly if provided 1`] = `
   size="large"
 >
   <img
+    alt="test"
     src="test"
   />
 </figure>
@@ -21,6 +22,7 @@ exports[`Avatar Sizes applies small correctly if provided 1`] = `
   size="small"
 >
   <img
+    alt="test"
     src="test"
   />
 </figure>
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -32,6 +34,7 @@
     "no-invalid-this": "off",
     "no-unused-expressions": ["error", { "allowShortCircuit": true }],
     "react/jsx-key": "off",
-    "react/display-name": "off"
+    "react/display-name": "off",
+    "jsx-a11y/label-has-for": "off"
```
austin94: Since we use the `FieldContainer` in many areas it's a pain to disable this everywhere. Don't think it will cause too much of an issue.

jzempel:
@austin94 looks like some snapshots need updating. https://travis-ci.org/zendeskgarden/react-components/builds/383808794

Issue #37, opened 2018-06-12
Compatibility with react@15.3.2


luis-almeida:
## Expectations

We just started trying to use zendeskgarden's react-components and we started with `react-menus`.
Since it lists `react@15` as a [peer dependency](https://github.com/zendeskgarden/react-components/blob/master/packages/menus/package.json#L33) we weren't expecting to run into compatibility issues.

## Reality

Apologies if I'm missing something obvious but here's what I'm getting when adding a `<Menu />` component in our application:
<img width="1217" alt="screen shot 2018-06-11 at 14 52 05" src="https://user-images.githubusercontent.com/1159544/41280506-66215f82-6e2f-11e8-8832-d49888b3577e.png">

## Steps to Reproduce

I've managed to reproduce the error with the basic example found in the documentation page:
https://stackblitz.com/edit/react-uadcwt
Upgrading to `react@16` fixes the issue so it seems to be related to the version of react in use.

###### Fine Print

* Component: menus
* Browsers: all

austin94:
We have some consumers still on `v15` so I think this might be an issue specifically with Menus.

I was able to spin up a `create-react-app` with Buttons and Tabs, but the error is definitely visible if you attempt to use the Menu.

![screen shot 2018-06-12 at 9 11 10 am](https://user-images.githubusercontent.com/4030377/41302942-b23d9e26-6e20-11e8-9f4e-53ab55d4ed99.png)

I'll investigate today and update once I find anything.

PR #38, opened 2018-06-12
fix(menus|select): remove fragment usage to support React v15.X

austin94:
## Description

After researching #37 I found some assorted issues in relation to React `v15.X` support and the Menu/Select components.

#### 1. react-portal dropped support for React `v14.X`

While double checking the dependencies, it looks like `react-portal` dropped their v14 support. Unrelated to the issues experienced above, but I updated our peerDependencies to reflect that in the affected components. All others support `v14` as expected.

#### 2. react-popper and ~~React.Fragments~~ (root cause)

Within our positioning library, react-popper, there is an option to not wrap the trigger/popover with a containing element (https://github.com/FezVrasta/react-popper/tree/v0.x#tag-proptypesoneoftypeproptypesstring-proptypesbool).

Since React `v16` supports [Fragments](https://reactjs.org/docs/fragments.html), this was working as expected in the Lotus, Admin Centre, Web Widget, and our local documentation. But, when ran in `v15.X` it would blow up as it was returning an array of elements.

#### 3. react-portal `appendTo` body usage in `v15.X`

We also had some "appendToBody" logic within our MenuContainer component. This would behave differently between React versions if we manually provided the body element. I've slightly modified the default Portal state to be as expected in `v15`

## Detail

The quick fix, was to re-enable the `tag` property in the react-popper Manager. This wraps the trigger/popover with containing `<div>` for the Menu/Select components. This fixes the usages in React 15.

The unfortunate side-effect of this is that many of our `css-modules` stylings require sibling selectors, with no nesting, to properly manage margining.  This issue is specific to our React implementation so I'm not sure if it's wise to modify the `forms` styling to fix this.

As a "quick" fix I've added in some logic to dyanmically add in the `marginTop` styling based on the rendering order of the `Select` component (menus is not affected).

Working Menu in React `v15.X`:

Closes https://github.com/zendeskgarden/react-components/issues/37

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/containers/MenuContainer.js
+++ b/packages/menus/src/containers/MenuContainer.js
@@ -10,6 +10,7 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Manager, Popper, Target } from 'react-popper';
 import { Portal } from 'react-portal';
+import Fragment from 'render-fragment';
 import {
   SelectionContainer,
   ControlledComponent,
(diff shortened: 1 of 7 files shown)


packages/menus/src/containers/MenuContainer.js
```diff
@@ -575,10 +575,10 @@ class MenuContainer extends ControlledComponent {
             );
 
             if (appendToNode) {
-              menu = <Portal node={appendToNode}>{menu}</Portal>;
+              return <Portal node={appendToNode}>{menu}</Portal>;
```
ryanseddon: So trying this code with react@15.x.x and supplying the `appendToNode={document.body}` prop wipes out everything in body and renders only the menu? Am I doing something silly?

### Gif
appendtonode

packages/select/src/elements/SelectField.js
```diff
@@ -76,15 +84,39 @@ export default class SelectField extends ControlledComponent {
               }
 
               if (hasType(child, Hint)) {
-                return cloneElement(child, getHintProps(child.props));
+                this.hasHintRendered = true;
+
+                let hintProps = child.props;
+
+                if (this.hasSelectRendered) {
+                  hintProps = { style: { marginTop: '8px', ...hintProps } };
```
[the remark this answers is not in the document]
austin94: > Probably worth pulling this out and just spreading to the places used

Refactored this into a local function

> Also I don't think you meant to spread the remaining props into the style object right?

Definitely not. Now fixed.

packages/menus/src/containers/MenuContainer.js
austin94: This was something I noticed as well, it looks like a quirk of the `react-portal` library.

In React 16, since they are using the provided portals, it will append to the body if you explicitly pass it through.  While in React 15 it will completely replace it. This difference sucks, but actually makes sense. React docs even explicitly say to _not_ apply directly to the body.

If we don't provide the `node` prop it is correctly placed within a div relative to the `body` in both 15 and 16 so I removed the previous default.

packages/select/src/elements/SelectField.js
```diff
@@ -52,11 +52,21 @@ export default class SelectField extends ControlledComponent {
     };
   };
 
+  applyMarginStyling = otherProps => ({ style: { marginTop: '8px' }, ...otherProps });
```
jzempel: This does not capture the nuance provided by https://github.com/zendeskgarden/css-components/blob/master/packages/forms/src/_text/_label.css which accounts for `--sm` sized fields, handles `--inline`, and special-casing for `.c-txt__label` followed by a `.c-txt__hint`.

Although it's not generally advisable, I think we should take a serious look at [render-fragment](https://www.npmjs.com/package/render-fragment) in order to provide a first-class styling experience for React 16 (allowing React 15 users to manipulate the layout in a more manual fashion). That approach would get us through this somewhat hacky `has[?]Rendered` code in the short-to-medium term.

austin94: Even though `render-fragment` will create different DOM based on the version, I think you're right that it's going to be the easiest way forward.

I'll make sure to include a warning in the documentation as well for this with example of correct styling. The good news is that consumers can continue to style this with fela/css-modules however they want to get the desired layout.

austin94:
@jzempel I just updated this with the `render-fragment` usage. Let me know what you think.

austin94:
> when we deprecate/remove support for React 15

Lets wait until we decide if (when) to do that. There are going to be several changes that we do so I think this one will fall under a larger bucket.  If you feel strongly though I can add something in.

austin94:
@luis-almeida do you see any issues with this implementation? 

Since you're still on React 15 you _may_ have to tweak some vertical margins when using the Hint/Message component with the Select, but the Menus implementation should be the same.

luis-almeida:
I think it's fair having to tweak some styling when on React 15.
I'm not familiar enough with the codebase but it looks good to me! 👍 

Thanks for looking into this issue so quickly!

PR #42, opened 2018-06-14
feat(docs): include changelog in all documentation areas

austin94:
## Description

After some discussion in the `#garden` channel and enough people reaching out thinking that we "don't version or have any changelogs" I think it might be best to surface that information somewhere in the react documentation.

This PR adds an additional `View Changelog` button to the sidebar and includes the `CHANGELOG.md` file with the Github Markdown styling (since this is outside the "Examples" components in react-styleguidist we are on our own with styling here).

## Detail

Pre-published to the Modal component: https://garden.zendesk.com/react-components/modals/

Example:

screen shot 2018-06-13 at 5 41 26 pm

## Checklist

      designer as a reviewer)
  component


utils/styleguide/TableOfContentsRenderer/index.js
```diff
@@ -29,69 +32,118 @@ const BadgeContainer = styled.div`
...
-                    location.search = '';
-                  } else {
-                    location.search = '?isRtl';
-                  }
-                }}
-              >
-                {isRtl ? 'Disable RTL' : 'Enable RTL'}
-              </Button>
-            }
+const ChangelogButton = styled(Button)`
```
ryanseddon: Could then changelog components be in another file that this component imports?

allisonacs:
Heyo @austin94! I've got a few suggestions to improve the usability. 

1. Let's use a toggle for "Enable RTL." It's actually one of the rare times when the use of a toggle is appropriate: it's an instant switch, that requires no "save" action. It also feels less daunting as a toggle. 
2. For the link to the changelog, a button doesn't really make sense either: we're not submitting an action, we're just linking to a modal. Can we use a link here instead? Nice and simple :)

PR #44, opened 2018-06-19
fix: Stop double focus of button group

### LINKED ISSUE #43: ButtonGroup double focus
## Expectations

A `<ButtonGroup>` should never allow focus on more than one item.

## Reality

With some amount of effort, you can break focus styling for buttons.

## Steps to Reproduce

1. Mouse down on the last button within a `<ButtonGroup>`
1. Drag off of the button and then mouse up
1. Switch to a new tab and then return

Result: both the first and last button will style with a focus state.

double-focus

###### Fine Print

* Component: 
* Browsers: all

ryanseddon:
## Description

Stop browser double focusing the buttons in side a `<ButtonGroup />` when following the reproduction steps in #43.

~~Not sure on my solution and this hasn't got tests yet but wanted to put it up and see what you think.~~

## Detail
buttongroup-focus

closes #43 

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/buttons/src/containers/ButtonGroupContainer.spec.js
+++ b/packages/buttons/src/containers/ButtonGroupContainer.spec.js
@@ -76,5 +76,15 @@ describe('ButtonGroupContainer', () => {
         expect(button).toHaveProp('tabIndex', -1);
       });
     });
+
+    it('moves focus to the ButtonGroupView if a button receives focus', () => {
+      const btn = findButtons(wrapper)
+        .at(0)
+        .simulate('focus');
+
+      expect(btn.getDOMNode().dataset.focused).toBeTruthy();
+      expect(btn.getDOMNode()).not.toEqual(document.activeElement);
+      expect(findButtonGroup(wrapper).getDOMNode()).toEqual(document.activeElement);
+    });
   });
 });
--- a/packages/buttons/src/containers/ButtonGroupContainer.spec.js
+++ b/packages/buttons/src/containers/ButtonGroupContainer.spec.js
@@ -78,13 +78,13 @@ describe('ButtonGroupContainer', () => {
     });
 
     it('moves focus to the ButtonGroupView if a button receives focus', () => {
-      const btn = findButtons(wrapper)
-        .at(0)
-        .simulate('focus');
+      const focusMock = jest.fn();
 
-      expect(btn.getDOMNode().dataset.focused).toBeTruthy();
-      expect(btn.getDOMNode()).not.toEqual(document.activeElement);
-      expect(findButtonGroup(wrapper).getDOMNode()).toEqual(document.activeElement);
+      findButtons(wrapper)
+        .at(0)
+        .simulate('focus', { target: { parentNode: { focus: focusMock } } });
+
+      expect(focusMock).toHaveBeenCalled();
     });
   });
 });
--- a/packages/buttons/src/containers/ButtonGroupContainer.spec.js
+++ b/packages/buttons/src/containers/ButtonGroupContainer.spec.js
@@ -7,8 +7,12 @@
 
 import React from 'react';
 import { mount } from 'enzyme';
+import closest from 'dom-helpers/query/closest';
 
 import ButtonGroupContainer from './ButtonGroupContainer';
+
+jest.mock('dom-helpers/query/closest');
+closest.mockImplementation(() => ({ focus: jest.fn() }));
 
 describe('ButtonGroupContainer', () => {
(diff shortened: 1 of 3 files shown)


packages/selection/src/containers/SelectionContainer.js
```diff
@@ -206,8 +212,14 @@ export class SelectionContainer extends ControlledComponent {
...
           if (typeof focusedKey === 'undefined') {
             let selectedIndex = this.keyIndexMap[this.getControlledState().selectedKey];
-
-            if (typeof selectedIndex === 'undefined') {
+            const activeElement = document.activeElement;
+            const activeElementIndex = [...activeElement.parentNode.children].indexOf(
+              activeElement
+            );
+
+            if (activeElement !== -1) {
```
austin94: Should you be comparing `activeElementIndex` here instead?

packages/selection/src/containers/SelectionContainer.js
```diff
@@ -169,6 +169,12 @@ export class SelectionContainer extends ControlledComponent {
 
   onFocusSelectionModelChange = ({ newSelection }) => {
     const focusedKey = this.indexKeyMap[newSelection];
+    const activeElement = document.activeElement;
```
austin94: Could we use the `activeElement` `dom-helper` utility here for cross-browser support? https://www.npmjs.com/package/dom-helpers

This helps cover the weird use-cases in IE11 with returning from another tab (probably some other things as well).

ryanseddon:
> this hasn't got tests yet but wanted to put it up and see what you think.

Was planning on doing tests today, just wanted feedback on the approach and if you can think of a better way to do it?

packages/selection/src/containers/SelectionContainer.js
```diff
@@ -169,6 +169,12 @@ export class SelectionContainer extends ControlledComponent {
...
+    const browserSelectionIndex = [...activeElement.parentNode.children].indexOf(activeElement);
```
austin94: Another thing I just thought of is that this has the assumption that all "selectable" items are children of the parentNode. This isn't always the case. Since you can spread these render-props to a variety of areas it may or may not be in the correct placement.

austin94:
After looking through this some more, I think this would solve the ButtonGroup example, but would fail in areas that don't have selectable items that are direct siblings of each other.

I'm also not seeing this occur in Firefox. Or in components that are not ButtonGroup in the other browsers.  https://garden.zendesk.com/react-components/tabs/#!/Tabs/1 is working for example.

Could you look into if it's just a focus issue since these components are using `<buttons>` internally and are able to grab focus?

ryanseddon:
This is working as expected now across all components that utilise SelectionContainer.

The spec around activeElement focus is fairly ambiguous so I think browser vendors interpret it differently to each other.

ryanseddon:
So after some internal discussion we've agreed that this isn't the right approach, will keep this open and push the change when I have it.

ryanseddon:
@austin94 I added a test to cover this code, I did however look at the dom properties instead as enzyme's `simulate()` doesn't trigger a render. If there's a better way let me know.

packages/buttons/src/containers/ButtonGroupContainer.spec.js
```diff
@@ -76,5 +76,15 @@ describe('ButtonGroupContainer', () => {
         expect(button).toHaveProp('tabIndex', -1);
       });
     });
+
+    it('moves focus to the ButtonGroupView if a button receives focus', () => {
```
austin94: Active element detection is hard to test within Enzyme 😢 

A common approach that I use is to only mock the expected functionality. The `onFocus` composition that we have on each button is only calling `target.parentNode.focus()`. How focus is handled after that point is a browser concern and not really something that is "testable", especially with jsdom.  I suggest just ensuring that we call the correct method with a mock and calling it a day.

An example is:

```
it('moves focus to the ButtonGroupView if a button receives focus', () => {
      const focusMock = jest.fn();

      findButtons(wrapper)
        .at(0)
        .simulate('focus', { target: { parentNode: { focus: focusMock } } });

      expect(focusMock).toHaveBeenCalled();
    });
```

ryanseddon: ok that makes sense, will do that 👍

ryanseddon:
> Didnt realize that dom-helpers had that utility!

Me either, it's not documented either. I was originally using `Element#closest` which works in all browsers but IE11 so did a search on the dom-helpers codebase.

PR #47, opened 2018-06-27
feat(autocomplete): introduce @zendeskgarden/react-autocomplete package

austin94:
## Description

This PR (finally) adds the `@zendeskgarden/react-autocomplete` package.

Due to the complexity of autocompletes in general and the wide variety of implementations that are within the products (ticket props, CCs, followers, ticket fields, etc.) this PR focuses on providing the interaction and accessibility patterns in a render prop container: `AutocompleteContainer`

It is relatively simple to create 2 patterns with this component:

* `AutoComplete`  ✨ **NEW** ✨ 
  * Acts as a "filterable select" this component is used through the admin settings area and was not originally available in Garden
* `MutliSelect`
  * Same interaction as before, but more flexible due to the different implementations that are needed.
  * Specifically, CC's (@clairekrogers )

All interactions are currently accessible and working with both chromevox and voiceover as well.

## Detail

Pre-published at: https://garden.zendesk.com/react-components/autocomplete/

## Checklist

      designer as a reviewer)
  component


packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,540 @@
...
+    }
+  };
+
+  getItemId = key =>
+    typeof key === 'undefined' ? '' : `${this.getControlledState().id}--item-${key}`;
+
+  getTagId = key =>
+    typeof key === 'undefined' ? '' : `${this.getControlledState().id}--tag-${key}`;
+
+  getMenuId = () => `${this.getControlledState().id}--menu`;
```
ryanseddon: So I think you had the intention of using this but it's not used in the code base.

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,540 @@
...
+        this.triggerMousedDown = true;
+      }),
+      onMouseUp: composeEventHandlers(onMouseUp, () => {
+        this.triggerMousedDown = false;
+      }),
+      onClick: composeEventHandlers(onClick, () => {
+        this.openDropdown();
+        this.inputRef.focus();
+      }),
+      ...other
```
ryanseddon: According the the authoring practices we need to add some more stuff here:

```
role: 'combobox',
'aria-expanded': isOpen, // may nee to pass strings as false will just remove this
'aria-owns': this.getMenuId(),
'aria-haspopup': 'listbox'
```

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,540 @@
...
+  getMenuProps = ({ onMouseDown, onMouseUp, ...other } = {}) => {
+    return {
+      onMouseDown: composeEventHandlers(onMouseDown, () => {
+        this.menuMousedDown = true;
+      }),
+      onMouseUp: composeEventHandlers(onMouseUp, () => {
+        this.menuMousedDown = false;
+        this.inputRef.focus();
+      }),
+      ...other
```
ryanseddon: Same here for the above changes we need an `id` & `role` applied

```
id: this.getMenuId(),
role: 'listbox'
```

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,540 @@
...
+const MenuWrapper = styled.div`
+  z-index: ${props => props.zIndex};
+`;
+
+class AutocompleteContainer extends ControlledComponent {
+  static propTypes = {
+    /**
+     * Appends the menu to the provided element
+     */
+    appendToNode: PropTypes.instanceOf(Element),
```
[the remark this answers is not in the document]
austin94: For this to be used in SSR it would have to be [pony-filled](https://ponyfoo.com/articles/polyfills-or-ponyfills) (my new vocab for the day) anyway.

Would it make more sense to just use `PropTypes.node` rather than a custom validator (since it wouldn't match in SSR anyway)?

packages/autocomplete/src/containers/AutocompleteContainer.js
austin94: Was looking at the wrong pattern for listbox when I implement this 😱 

I'm now fully matching the pattern at https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html

packages/autocomplete/src/containers/AutocompleteContainer.js
austin94: This is now used now that I'm following the correct W3C design doc.

It's used in the `getInputProps` for `aria-owns` as well as `getMenuProps` for applying the ID

packages/autocomplete/src/containers/AutocompleteContainer.js
austin94: Now added.

austin94:
Just pushed all of the a11y comments from @ryanseddon. It seems chromevox and voiceover do a good job of "guessing" which is why it was still working before.

The biggest improvement is it now lists the correct number of available menu items i.e. "On menu item 1 of 26"

austin94:
Testing is now added for the `AutocompleteContainer` and `requiredParam` utility. Was able to get 100% coverage as well.

screen shot 2018-06-29 at 2 16 23 pm

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,563 @@
...
+    /**
+     * The z-index of the popper.js placement container
+     */
+    zIndex: PropTypes.number
+  };
+
+  static defaultProps = {
+    placement: 'bottom-start',
+    eventsEnabled: true,
+    zIndex: 1000
```
jzempel: Does it make sense to expose and pull this default prop from `MenuContainer`?

packages/autocomplete/src/containers/AutocompleteContainer.spec.js
```diff
@@ -0,0 +1,582 @@
...
+
+import React from 'react';
+import { KEY_CODES } from '@zendeskgarden/react-selection';
+import { mountWithTheme } from '@zendeskgarden/react-testing';
+import AutocompleteContainer from './AutocompleteContainer';
+
+/**
+ * Mocks popper.js calls within react-popper due to virtual testing environment
+ */
+jest.mock('popper.js', () => {
```
jzempel: Should this repeated mock be moved into `@zendeskgarden/react-testing`?

packages/autocomplete/src/examples/autocomplete.md
```diff
@@ -0,0 +1,175 @@
...
+                open: isOpen,
+                focused: state.isFocused || isOpen,
+                select: true,
+                inputRef: ref => {
+                  this.wrapperRef = ref;
+                  triggerRef(ref);
+                }
+              })}
+            >
+              {!isOpen && <span>{state.value}</span>}
```
jzempel: There is a styling problem with this `<span>` that pixel-jumps from value to placeholder.

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
...
+  text-align: center;
+`;
+
+const SpacedTag = styled(Tag)`
+  margin: 2px;
+`;
+
+const InlineInput = styled(Input)`
+  flex-grow: 1;
+  margin-left: 2px;
```
jzempel: This doesn't layout correctly for RTL. You might need `margin: 0 2px;`

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
...
+
+This examples includes an example of "copy-to-clipboard" functionality for tags.
+
+```jsx
+const { MenuView, Item, AddItem, Separator } = require('@zendeskgarden/react-menus');
+const { FauxInput, Input, TextGroup, Label } = require('@zendeskgarden/react-textfields');
+const { Tag, Close } = require('@zendeskgarden/react-tags');
+const { Anchor } = require('@zendeskgarden/react-buttons');
+const { KEY_CODES, FieldContainer } = require('@zendeskgarden/react-selection');
+
```
jzempel: Should we be concerned that the following styling section isn't represented by the CSS components?

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
+The `MultiSelect` layout operates similarly to the `Autocomplete`, but must
+manage the interaction between the Input and it's collection of Tags.
+
+This examples includes an example of "copy-to-clipboard" functionality for tags.
```
jzempel: grammar ;)

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
...
+                          onKeyDown: e => {
+                            if (tagFocusedKey !== undefined) {
+                              if (
+                                e.keyCode === KEY_CODES.DELETE ||
+                                e.keyCode === KEY_CODES.BACKSPACE
+                              ) {
+                                deleteTag(tagFocusedKey);
+                              }
+
+                              // copy-to-clipboard functionality
```
jzempel: It would also be nice to show `cmd-A` functionality that selects all tags. See Gmail for the ability to delete all tags at once.

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
...
+                          innerRef: inputRef,
+                          value: state.inputValue,
+                          onChange: e => {
+                            setState({ inputValue: e.target.value });
+                          },
+                          onKeyDown: e => {
+                            if (tagFocusedKey !== undefined) {
+                              if (
+                                e.keyCode === KEY_CODES.DELETE ||
+                                e.keyCode === KEY_CODES.BACKSPACE
```
jzempel: The backspace to focus + another backspace to delete is tedious. We should be deleting on backspace as we did previously. An implementation of `cmd-Z` would be required for convenient undo of an unintended action.

Or am I missing a hard a11y requirement here?

packages/autocomplete/src/examples/multiselect.md
```diff
@@ -0,0 +1,304 @@
...
+                        })
+                      )}
+                    />
+                  </FauxInput>
+                );
+              }}
+            >
+              {({ getMenuProps, getItemProps, placement, focusedKey }) => {
+                return (
+                  <MenuView
```
jzempel: Something is wrong with the initial top placement of this menu.

screen shot 2018-06-29 at 3 35 17 pm

packages/autocomplete/src/examples/autocomplete.md
```diff
@@ -0,0 +1,175 @@
...
+
+          setState(newState);
+        }}
+        trigger={({ getTriggerProps, getInputProps, triggerRef, inputRef, isOpen }) => {
+          return (
+            <FauxInput
+              {...getTriggerProps({
+                open: isOpen,
+                focused: state.isFocused || isOpen,
+                select: true,
```
jzempel: I'm not exactly sure how this will work in the new container world, but we agreed that a `<select>` should retain mouse control for collapsing an expanded menu with our previous https://garden.zendesk.com/react-components-archive/#!/MultiSelect. We should model that here.

packages/autocomplete/src/examples/multiselect.md
ryanseddon: I guess the trouble there might be conveying what happened to a screen read if it never focuses the tag and just delete it

jzempel: A problem mitigated somewhat by cursoring-to the tag before hitting delete.

sunesimonsen:
@austin94 is there a plan for making high level abstractions, or is this the required work to get anything going:

```js
const { MenuView, Item, AddItem, Separator } = require('@zendeskgarden/react-menus');
const { TextGroup, Label, FauxInput, Input } = require('@zendeskgarden/react-textfields');
const { FieldContainer } = require('@zendeskgarden/react-selection');

const NoItemsMessage = styled.div`
  margin: 16px;
  text-align: center;
`;

const stringContains = (original, comparison) => {
  const formattedOriginal = original.replace(/ /g, '').toLocaleLowerCase();
  const formattedComparison = comparison.replace(/ /g, '').toLocaleLowerCase();

  return formattedOriginal.indexOf(formattedComparison) !== -1;
};

const getMatchingMenuItems = (searchValue, selectedValue, getItemProps, focusedKey) => {
  const menuItems = state.natoPhonetics
    .filter(phonetic => stringContains(phonetic, searchValue))
    .map(phonetic => (
      <Item
        {...getItemProps({
          key: phonetic,
          focused: focusedKey === phonetic,
          checked: selectedValue === phonetic,
          'aria-selected': selectedValue === phonetic
        })}
      >
        {phonetic}
      </Item>
    ));

  return menuItems.length === 0 ? <NoItemsMessage>No items found</NoItemsMessage> : menuItems;
};

initialState = {
  value: 'Default value',
  inputValue: '',
  selectedKeys: {},
  isFocused: false,
  isOpen: false,
  natoPhonetics: [
    'Alfa',
    'Bravo',
    'Charlie',
    'Delta',
    'Echo',
    'Foxtrot',
    'Golf',
    'Hotel',
    'India',
    'Juliett',
    'Kilo',
    'Lima',
    'Mike',
    'November',
    'Oscar',
    'Papa',
    'Quebec',
    'Romeo',
    'Sierra',
    'Tango',
    'Uniform',
    'Victor',
    'Whiskey',
    'X-ray',
    'Yankee',
    'Zulu'
  ]
};

<FieldContainer>
  {({ getLabelProps, getInputProps: getFieldInputProps }) => (
    <TextGroup>
      <Label {...getLabelProps()}>Basic Autocomplete</Label>
      <AutocompleteContainer
        isOpen={state.isOpen}
        onSelect={selectedKey => {
          let natoPhonetics = state.natoPhonetics;

          if (selectedKey === 'add-item') {
            natoPhonetics = natoPhonetics.slice();
            natoPhonetics.push(state.inputValue);
            selectedKey = state.inputValue;
          }

          setState({ value: selectedKey, natoPhonetics, inputValue: '' });
        }}
        onStateChange={newState => {
          let inputValue = state.inputValue;

          if (typeof newState.isOpen !== 'undefined' && !newState.isOpen) {
            newState.inputValue = '';
          }

          setState(newState);
        }}
        trigger={({ getTriggerProps, getInputProps, triggerRef, inputRef, isOpen }) => {
          return (
            <FauxInput
              {...getTriggerProps({
                open: isOpen,
                focused: state.isFocused || isOpen,
                select: true,
                inputRef: ref => {
                  this.wrapperRef = ref;
                  triggerRef(ref);
                }
              })}
            >
              {!isOpen && <span>{state.value}</span>}
              <Input
                {...getInputProps(
                  getFieldInputProps({
                    bare: true,
                    innerRef: inputRef,
                    value: state.inputValue,
                    onChange: e => {
                      setState({ inputValue: e.target.value });
                    },
                    placeholder: state.value,
                    onFocus: () => {
                      setState({ isFocused: true });
                    },
                    onBlur: () => {
                      setState({ isFocused: false });
                    },
                    style: !isOpen
                      ? { opacity: 0, height: 0, minHeight: 0, width: 0, minWidth: 0 }
                      : {}
                  })
                )}
              />
            </FauxInput>
          );
        }}
      >
        {({ getMenuProps, getItemProps, placement, focusedKey }) => {
          const menuItems = getMatchingMenuItems(state.inputValue, state.value, getItemProps, focusedKey);

          const addItemProps =
            state.inputValue.length === 0
              ? { disabled: true }
              : getItemProps({ key: 'add-item', focused: focusedKey === 'add-item' });

          return (
            <MenuView
              {...getMenuProps({
                placement,
                animate: true,
                style: {
                  width: this.wrapperRef ? this.wrapperRef.getBoundingClientRect().width : 0
                }
              })}
            >
              <div style={{ height: '200px', overflowY: 'auto' }}>
                {menuItems}
              </div>
              <Separator />
              <AddItem {...addItemProps}>Add item</AddItem>
            </MenuView>
          );
        }}
      </AutocompleteContainer>
    </TextGroup>
  )}
</FieldContainer>;
```

packages/autocomplete/src/examples/multiselect.md
austin94: > A problem mitigated somewhat by cursoring-to the tag before hitting delete

That was the original intention and is the current behavior in all multi-select/tag collections in Support currently.

A deletion is an action performed against an item and traditionally you have to be cursored to an item to perform an action.

_Side Notes_: It's important to get the example right for this, but this isn't something that is enforced by the `AutocompleteContainer` logic. So there is flexibility for either way.

packages/autocomplete/src/examples/multiselect.md
austin94: > It would also be nice to show cmd-A functionality that selects all tags

This functionality isn't used anywhere in the products (that I am aware of). Also, not sure what the accessibility logic would be for this since you would then have to have the add/remove logic with shift+ctrl combinations as well.

Since this wasn't in the original I think this is out-of-scope for this iteration.

packages/autocomplete/src/examples/multiselect.md
austin94: This is currently being address in the SplitButton PR https://github.com/zendeskgarden/react-components/pull/48

Looks like a regression happened in how chrome/firefox renders `translatex` placements with our animations nested within it. No version changes have occurred in popper.js or react-popper

austin94:
@sunesimonsen 

> is there a plan for making high level abstractions

I would say yes, but would be a follow-on PR as needed.

Some issues that we have seen in the past with the high-abstraction for MultiSelect is:

- Very few people used it
  - Nearly every implementation had to be heavily customized for it to work correctly
  - For a majority of implementations in Support is just flat out didn't work for their use-case
- This example code is actually around the same length/complexity of the "high-abstraction" element in the previous components https://garden.zendesk.com/react-components-archive/#!/MultiSelect/1
  - Most of this logic is abstracted into a backend anyway

packages/autocomplete/src/containers/AutocompleteContainer.example.md
```diff
@@ -0,0 +1,18 @@
...
+
+The examples below show two common layouts using assorted Garden packages:
+
+- [@zendeskgarden/react-textfields](/react-components/textfields/)
+- [@zendeskgarden/react-menus](/react-components/menus/)
+- [@zendeskgarden/react-tags](/react-components/tags/)
+
+### WARNING
+
+This package comes _"with some assembly required"_
```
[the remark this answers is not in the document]
austin94: Explanation added.

packages/autocomplete/src/containers/AutocompleteContainer.js
austin94: Yes, but unfortunately `react-styleguidist` doesn't recognized imported propTypes and defaultProps. There was a PR added for this, but seems like something they don't want to support.

packages/autocomplete/src/containers/AutocompleteContainer.spec.js
austin94: It looks like the `jest.mock()` call has to occur inline in the component. When imported and ran it doesn't seem to mock correctly.

_I tried this when we originally created the `react-testing` library_

packages/autocomplete/src/examples/multiselect.md
austin94: Good catch. Now fixed.

packages/autocomplete/src/examples/multiselect.md
austin94: I would say "yes" for the long-term, but out of the 4 implementations that are currently looking to use this component, only 1 (ticket props) is going to be able to match this look exactly.

Definitely think we need to eventually have some standardization around this in the future.

packages/autocomplete/src/examples/multiselect.md
austin94: Grammar is fixed. I also rewrote the intro to be more clear.

packages/autocomplete/src/examples/autocomplete.md
austin94: This is now fixed. Just missed this interaction when bringing over the new functionality.

sunesimonsen:
@austin94 don't get me wrong, I think this abstraction looks very useful to build other components. I would just never use it directly in UI code as it has all of its wires handing out. I would always hide this behind some kind of facade.

> I would say yes, but would be a follow-on PR as needed.

Great. I would not ask for it to be added as part of this PR.

> Nearly every implementation had to be heavily customized for it to work correctly

I would be interested it what implementation we are talking here, is it the MultiSelect or the AutoComplete component? Are we talking re-implementing existing features or add new ones.

> For a majority of implementations in Support is just flat out didn't work for their use-case

I find it really surprising if that's the case. In Guide we manage to build almost everything new with high abstraction components, and I haven't heard any complaints about it. On the contrary people seems happy about the consistency and the ease this approach brings. I agree that some times it is necessary to go more custom and for that the render-prop pattern is great for building a new custom component for your use-case.

> This example code is actually around the same length/complexity of the "high-abstraction" element in the previous components https://garden.zendesk.com/react-components-archive/#!/MultiSelect/1

I think that is because the old MultiSelect is not high enough an abstraction. There is no reason why you should handle the `onKey` yourself. In this new world where you can dive down to the render-prop abstraction, I would probably also default the look of the completed tokens.

About the AutoComplete, it seems like it is almost a `Select` component without a label, so I expect to find a high-abstraction for that that is similar to the one for `Select`.

But hey, great work on the foundation, it seems solid. I'm looking forward to see things build on top of it.

Kind regards
Sune

packages/autocomplete/src/examples/multiselect.md
jzempel: Fair enough. Agreed out-of-scope.

packages/autocomplete/src/examples/multiselect.md
jzempel: Ok. Will look at that PR for the fix – we also lost animation on popping the menu up (it's still there on down).

packages/autocomplete/src/examples/multiselect.md
austin94: After sitting down with Allison, we decided to add in the "auto-delete" functionality for the example.  This now more closely matches the interaction in the new Gmail multiselect.

Consumers still have all the customization to go either way, but the example is just a little more opinionated now.

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -0,0 +1,568 @@
...
+    const previousFocusedKey = this.isControlledProp('focusedKey')
+      ? prevProps.focusedKey
+      : prevState.focusedKey;
+
+    /**
+     * We must programatically scroll the newly focused element into view.
+     * Side-effect of the `aria-activedescendant` accessibility strategy.
+     */
+    if (typeof focusedKey !== 'undefined' && focusedKey !== previousFocusedKey) {
+      const itemNode = document.getElementById(this.getItemId(focusedKey));
```
ryanseddon: If we push through the [document context branch](https://github.com/zendeskgarden/react-components/compare/zendeskgarden:e192dc6...zendeskgarden:04610b8) I have we'll need to update this too

austin94: I agree. Whichever solution we go with we will have to make sure that it will work with all of the components, not just select/menu.

Are you ok with handling this in a follow-on task since it seems out-of-scope for this?

ryanseddon: For posterity, #49 has been merged so this can use the `getDocument` util now.

austin94: Rebased, updated all the dependencies, and added in the `getDocument` utility

PR #48, opened 2018-07-02
feat(splitbutton): add ChevronButton view and SplitButton example

austin94:
## Description

This PR adds the necessary visual components and examples for a SplittButton component.

Rather than make `react-menus` a dependency of `react-buttons` I have opted to include the necessary visual elements (chevron button) and provided an example usage.

Some reasons for this decision:

- A large majority of consumers are using `react-buttons` without a SplitButton. The menus styling and popper/portal dependencies are not cheap enough to force this bundle size on people that just want a button.
- The API complexity is similar in # of lines as well as complexity between providing all customization needed with a high-abstraction element.
- By not coupling these dependencies it allows these packages to be versioned freely.

### Button ref transclusion

Found some interesting issues with forwarding refs in nested `styled-components`. A workaround is to provide a `buttonRef` to retrieve the reference for all buttons (rather than `innerRef` provided by `styled-components`).

This is "fixed" with the React `forwardRef` API, but is limited to `^v16.4` 😢 

### Menu Regression

While testing this components I realized a regression has occurred with the animation of a menu when it is placed to the top of it's trigger. This isn't easily reproducible and occurs more frequently in firefox/safari.  To help mitigate this I've now disabled our `animate` prop when any top-placement is applied.

I will create a separate bug to track this in the future.

## Detail

Pre-published at: https://garden.zendesk.com/react-components/buttons/#splitbutton

(screen-reader example shows both assistive and non-assistive examples for interacting with the Menu)

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/views/__snapshots__/MenuView.spec.js.snap
+++ b/packages/menus/src/views/__snapshots__/MenuView.spec.js.snap
@@ -1,5 +1,173 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`MenuView Animation should be disabled if any top placement is provided 1`] = `
+.c1.c1 {
+  position: relative;
+}
+
+.c1:focus {
+  outline: none;
+}
+
+.c0 {
+  margin-bottom: 4px;
+}
+
+<MenuView
+  placement="top"
+>
+  <MenuView__MenuWrapper
+    placement="top"
+  >
+    <div
+      className="c0"
+    >
+      <MenuView__StyledMenuView
+        placement="top"
+      >
+        <ul
+          className="c-menu c-menu--up c-arrow--b c1"
+          data-garden-id="menus.menu_view"
+          data-garden-version="version"
+        />
+      </MenuView__StyledMenuView>
+    </div>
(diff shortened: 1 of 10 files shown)


packages/menus/src/views/__snapshots__/MenuView.spec.js.snap
```diff
@@ -948,7 +1116,7 @@ exports[`MenuView Renders renders custom top animation correctly 1`] = `
         placement="top"
       >
         <ul
-          className="c-menu c-menu--up is-open c-arrow--b c1"
+          className="c-menu c-menu--up c-arrow--b c1"
```
jzempel: I'm still seeing strange placement/behavior in Chrome.

jul-02-2018 12-32-21

austin94: > is-active state as expected

This is now added to the example.

> I'm still seeing strange placement/behavior in Chrome.

Maybe I didn't publish the new example, could you run locally? I'm currently not seeing this in Chrome, Firefox, Safari, or IE11

PR #53, opened 2018-07-09
fix(ssr): remove browser related usages to enable ssr

austin94:
## Description

I was playing around with some SSR use-cases and found a few more areas that are using a browser context when they shouldn't be.

This PR:
- removes the browser related usages (defaultProps and propTypes)
- adds a new examples directory with the first entry being a next.js project that consumes Garden

## Detail

Example Garden page that is server rendered: 

nextjs-example

Relates to #10 

## Checklist

      designer as a reviewer)
  component


packages/select/src/containers/SelectContainer.js
```diff
@@ -15,7 +15,7 @@ export default class SelectContainer extends ControlledComponent {
     /**
      * Appends the menu to the provided element
      */
-    appendToNode: PropTypes.instanceOf(Element),
+    appendToNode: PropTypes.any,
```
ryanseddon: Nice, this check never actually worked most of the time. So good to just be super flexible in what it accepts.

examples/nextjs/pages/_document.js
```diff
@@ -0,0 +1,48 @@
...
+
+import React from 'react';
+import Document, { Head, Main, NextScript } from 'next/document';
+import { ServerStyleSheet } from 'styled-components';
+import { IdManager } from '@zendeskgarden/react-selection';
+
+import BedrockStyles from '@zendeskgarden/css-bedrock/dist/index.css';
+import ChromeStyles from '@zendeskgarden/react-chrome/dist/styles.css';
+import ButtonStyles from '@zendeskgarden/react-buttons/dist/styles.css';
+import MenuStyles from '@zendeskgarden/react-menus/dist/styles.css';
```
ryanseddon: I could be wrong but I think the idiomatic way to add styles to a nextjs app is via the `withCSS` plugin:

This example should have a `next.config.js` file in the root.

```js
const withCSS = require('@zeit/next-css');

module.exports = withCSS();
```

Then in `_document.js` you can just add `<link rel="stylesheet" href="/_next/static/style.css" />`

Then in your index or layout file you can just do nameless imports for the styles needed.

austin94: I stole this from one of the examples in their repo. 

I think the one you linked to is better though. Will implement in that style.

PR #55, opened 2018-07-09
feat(tables): introduce @zendeskgarden/react-tables package

austin94:
🚧 There is some weird focus treatment on the OverflowButton. Will need to address 🚧 

## Description

This PR introduces the `@zendeskgarden/react-tables` package. Previously, we baked [react-virtualized](https://bvaughn.github.io/react-virtualized/#/components/List) into our table, which was overkill. This led to a lot of complexity and ritual for the majority of implementations.

These new components are just functional wrappers around a native `<table>` implementation. These components are able to be used in a virtual environment if necessary though.

Additionally, the checkbox selection columns and overflow menus are _NOT_ bundled within this package.  Examples are shown how to consume these Garden packages and use them with Tables.

## Detail

Pre-published at: https://garden.zendesk.com/react-components/tables/

screen shot 2018-07-05 at 4 17 03 pm

## Checklist

      designer as a reviewer)
  component


packages/tables/README.md
```diff
@@ -0,0 +1,56 @@
...
+ */
+<ThemeProvider>
+  <Table>
+    <Caption>Your Unsolved Tickets</Caption>
+    <Head>
+      <Row header>
+        <HeaderCell>Subject</HeaderCell>
+        <HeaderCell>Requester</HeaderCell>
+        <HeaderCell>Requested</HeaderCell>
+        <HeaderCell>Type</HeaderCell>
```
ryanseddon: It'd be good to add the `scope` attribute here to show best practice with table a11y. 

```jsx
<Row header>
  <HeaderCell scope="col">Subject</HeaderCell>
  <HeaderCell scope="col">Requester</HeaderCell>
  <HeaderCell scope="col">Requested</HeaderCell>
  <HeaderCell scope="col">Type</HeaderCell>
</Row>
```

austin94: TIL https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Advanced#Tables_for_visually_impaired_users

Didn't realize how useful `scope` was. I'll add this to all of the examples.

packages/tables/src/examples/selectable-table.md
```diff
@@ -0,0 +1,93 @@
...
+            }
+          }}>
+          <Label
+            hidden
+            indeterminate={isSelectAllIndeterminate(state.selectedRows, state.rows)}>
+            Select all tickets
+          </Label>
+        </Checkbox>
+      </Cell>
+      <Cell scope="col">Long Truncated Title</Cell>
```
jzempel: Should be `<HeaderCell>`

packages/tables/src/views/OverflowButton.js
```diff
@@ -0,0 +1,64 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import TableStyles from '@zendeskgarden/css-tables';
+import { retrieveTheme } from '@zendeskgarden/react-theming';
+import { composeEventHandlers } from '@zendeskgarden/react-selection';
+
+const COMPONENT_ID = 'tables.overflow_button';
+
+/**
+ * Accepts all `<button>` props
```
jzempel: May render correctly as a `<div role="button">` – although CSS should probably provide resets.

utils/styleguide/setup.js
```diff
@@ -29,3 +29,10 @@ global.Col = styled(Col)`
   margin-top: 4px;
   margin-bottom: 4px;
 `;
+
+/** Add aditional global for packages that override these components (Table) */
+global.SudoGrid = {
```
jzempel: Uh, `<Layout.Grid>`? :wink:

packages/tables/src/examples/overflow-menu.md
```diff
@@ -0,0 +1,98 @@
...
+const StyledCaption = styled(Caption)`
+  font-size: ${zdFontSizeBeta};
+  font-weight: ${zdFontWeightSemibold};
+  margin-bottom: ${zdSpacingSm};
+`;
+
+const OverflowMenu = () => (
+  <Menu
+    onChange={selectedKey => alert(selectedKey)}
+    placement="bottom-end"
```
jzempel: Menu positioning should be corrected (for all table densities) as seen in https://garden.zendesk.com/react-components-archive/#!/Table

packages/tables/src/examples/overflow-menu.md
```diff
@@ -0,0 +1,98 @@
...
+    trigger={({ ref, isOpen }) => (
+      <OverflowButton
+        innerRef={ref}
+        active={isOpen}
+        onBlur={e => {
```
jzempel: Focus state should not be seen with mouse-only interaction.

packages/tables/src/examples/selectable-table.md
austin94: Fixed.

packages/tables/src/views/OverflowButton.js
austin94: This is now fixed. I opened an issues against [css-tables](https://github.com/zendeskgarden/css-components/issues/79) for us to address there.

utils/styleguide/setup.js
austin94: Fixed.

packages/tables/src/examples/overflow-menu.md
austin94: > Menu positioning should be corrected

There was some positioning mixups in my example. The examples now correctly:

- Apply the popperModifier to use the `viewport` as it's boundary element (not the `<tbody>`)
- Applies the necessary margin changes

packages/tables/src/examples/overflow-menu.md
austin94: This is now fixed in the example.

packages/tables/src/examples/overflow-menu.md
```diff
@@ -1,3 +1,10 @@
+Overflow menus are achieved by using the `Menu` component available in
+the [@zendeskgarden/react-menus](https://garden.zendesk.com/react-components/menus/)
+package.
+
+Based on `Table` positioning and other implementation specific details you may need
+to apply manual positioning against the `Menu` to ensure a standard look and feel.
```
jzempel: :+1:

packages/tables/src/views/OverflowButton.js
```diff
@@ -0,0 +1,71 @@
...
+      [TableStyles['is-hovered']]: props.hovered,
+      [TableStyles['is-active']]: props.active,
+      [TableStyles['is-focused']]: props.focused
+    })
+})`
+  /**
+   * Must provide a width that is equivalent to the :before pseudo
+   * element to ensure correct Popper.js positioning
+   **/
+  min-width: 28px;
```
jzempel: Is this still necessary after reset removal of the button padding?

austin94: Yes, this is necessary regardless of the type of element.

Since the only `:before` pseudo element is what is getting the width/height of the icon, Popper isn't able to understand this and applies relative to the edge, which throws of the horizontal alignment.

Rather than require all implementations this allows it to "just work" with the Menu component. Since this is implementation specific I feel comfortable leaving it on the React side for now.

austin94: Resolves this on the css-tables side. Bumped the dependency to pull in the fix.

PR #56, opened 2018-07-13
fix(autocomplete|menus|selection|theming): guard against `getDocument` being undefined

ryanseddon:
## Description

This guards against `getDocument` being undefined to avoid throwing errors. 

The following scenario where you have <3.0.0 of `react-theming` installed but have a package that utilises `getDocument` added to your project it'll throw on trying to call `undefined` as a function.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/autocomplete/src/containers/AutocompleteContainer.js
+++ b/packages/autocomplete/src/containers/AutocompleteContainer.js
@@ -156,7 +156,7 @@ class AutocompleteContainer extends ControlledComponent {
      * Side-effect of the `aria-activedescendant` accessibility strategy.
      */
     if (typeof focusedKey !== 'undefined' && focusedKey !== previousFocusedKey) {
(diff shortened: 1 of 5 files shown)


austin94:
Word of warning before we merge. We need to update the travis credentials since the [eslint issue](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes) cycled the credentials.

jzempel:
@austin94 done :point_up:

luis-almeida:
Worth noticing that `<3.1.1` of `react-theming` won't throw but the functionality is broken as `getDocument` still returns undefined.

I just spend some time debugging this issue but let me know if I'm missing something. I believe `peerDependency` on `react-theming` might have to be updated (?)

austin94:
Good catch, the peerDependencies need to be updated with the correct range. I'll make a PR for this today.

PR #62, opened 2018-07-19
feat(tooltips): remove wrapping div from TooltipContainer

austin94:
* [x] **BREAKING CHANGE** 

## Description

The current `TooltipContainer` wraps the trigger element in a `<div>` to provide a consistent `ref` to Popper.js

This is bad because it limits the amount of areas that a tooltip can be used. A `<div>` may not be a valid element for a specific dom nesting, it may miss-align other items, etc.

Bottom-line, it isn't very flexible.

## Detail

This PR removes the wrapping `<div>` from `TooltipContainer` and provides a `ref` callback to the `trigger` render-prop.  This can be applied to any element without a wrapper being necessary.

This is a **breaking change** to `TooltipContainer`

For the high-abstraction `Tooltip` element I've moved the `<div>` wrapper to be automatically applied, which shouldn't break any current usages and supports the "easiest" usecase.

Originally brought up by @thebritican with his new Rich Text Editor implementation.

Currently available under `@zendeskgarden/react-tooltips@beta` in NPM

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/tooltips/src/containers/TooltipContainer.example.md
+++ b/packages/tooltips/src/containers/TooltipContainer.example.md
@@ -14,8 +14,8 @@ All state is handled internally in the component.
 const { Button } = require('@zendeskgarden/react-buttons/src');
 
 <TooltipContainer
-  trigger={({ getTriggerProps }) => (
-    <Button {...getTriggerProps()}>Hover or Focus to trigger tooltip</Button>
+  trigger={({ getTriggerProps, ref }) => (
+    <Button {...getTriggerProps({ buttonRef: ref })}>Hover or Focus to trigger tooltip</Button>
   )}
 >
   {({ getTooltipProps, placement }) => (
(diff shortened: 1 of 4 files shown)


packages/tooltips/src/containers/TooltipContainer.example.md
```diff
@@ -163,7 +165,7 @@ initialState = {
             isVisible
             appendToBody
             placement="top-start"
-            trigger={({ getTriggerProps }) => <TriggerDiv {...getTriggerProps()} />}
+            trigger={({ getTriggerProps, ref }) => <TriggerDiv {...getTriggerProps({ innerRef: ref })} />}
```
jzempel: Plz fix this and later lines for `markdownlint`. https://travis-ci.org/zendeskgarden/react-components/builds/406001331#L1044

austin94: Weird, my pre-commit hooks didn't run. Will re-apply them. Fixed for now.

austin94:
I'm going to sit on this until we can get some confirmation that this addresses all of the new usages that prompted us to make the change.

PR #66, opened 2018-07-24
[ACCESSIBILITY] Global changes to pass AXE/Lighthouse Audits

austin94:
**TL;DR**: We have improved from an aggregate lighthouse accessibility score of 45% to 98.5% across **ALL** packages (with the remaining failures being lighthouse miss-fires) 🎉 

* [X] **BREAKING CHANGES?**

⚠️ There are several breaking changes in this PR. This is just a collection of all changes that I completed. This should (and will) be broken up into several smaller PRs that can address the changes in more depth. ⚠️ 

## Tooling Used
- [Google Lighthouse Audits](https://developers.google.com/web/tools/lighthouse/)
  - The google Lighthouse accessibility audit is a good first-pass for accessibility health
- [AXE Auditing Tool](https://www.deque.com/axe/)
  - Chrome extension for more involved accessibility testing
- [Mac Voiceover](https://www.apple.com/accessibility/mac/vision/)
  - Nothing beats actually testing with a screen reader

All three tools were used for the tests/assumptions performed during this process.

Here is a link to the lighthouse exports after the changes shown in this PR: https://drive.google.com/open?id=11bOPsxBof8gMa8Md0ZXhKzvMg0PHyVYq

These files can be viewed with https://googlechrome.github.io/lighthouse/viewer2x/

## Description

Accessibility is important.  All components within this repository are built to the [W3C Design Patterns](https://www.w3.org/TR/wai-aria-practices/#aria_ex) spec; so this means that everything is 100% accessible, right?

**WRONG** 😢 (this is [react-buttons](https://garden.zendesk.com/react-components/buttons/)) 

old-lighthouse-audit

The majority of component implementations are accessible, but our examples are currently lacking some good best practices. Some of this is a combination of color-contrast issues and customizations on top of react-styleguidist, but the worst offenders currently are the handwritten examples and some invalid assumptions about menu accessibility.

Currently all examples are hovering around the 40-60% lighthouse score.  This PR is a collection of several breaking API changes, styleguidist improvements, and example updates. These will be split into several individual PRs, but I felt a background PR would help.

## Detail

I've split this PR into 3 commits:

* Breaking (and non-breaking) API Changes
  * Turns out sticking to the W3C spec doesn't provide _everything_. I'll go into depth in the individual PRs
  * Examples:
    * Can't apply two IDs to `aria-describedby`
    * `aria-controls` and `aria-owns` can only to reference elements that are rendered already (duh)
    * .etc

* Example updates
  * Assorted `aria-label` and documentation changes

* Styleguidist upgrades
  * Tried to use Garden colors throughout
  * Migrated propTypes table to use the Garden table component

Before:
old

After:
new

## Learnings

Even thought we have static accessibility linting with`jsx-a11y` a large chunk of our codebase lives within the Markdown examples and isn't reachable with this tool.  We should perform more frequent accessibility audits or find some additional ways of auditing this in an automated fashion.

## Checklist

      designer as a reviewer)
  component


allisonacs:
Happy to see this happen @austin94! Less "do as I say," more "do as I do."

austin94:
> Nice, so these all came out of running the styleguide through the chrome dev tools audit panel

Unfortunately no. The overall scores did, but the majority of corrections were guided by the AXE plugin. It seems to be a lot more picky about things than the lighthouse audit and helped find some of the larger holes.

packages/autocomplete/src/containers/AutocompleteContainer.js
```diff
@@ -340,7 +338,6 @@ class AutocompleteContainer extends ControlledComponent {
       role,
       'aria-autocomplete': 'list',
       'aria-haspopup': 'true',
-      'aria-owns': this.getMenuId(),
```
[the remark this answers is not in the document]
austin94: No it does not. Since we are moving focus to the menu it seems to not be required? I ran AXE across all of the states in the menu and the only thing it picked up was us appending it to the body in a "non-landmark region" which would be something a consumer would customize.

packages/selection/src/containers/FieldContainer.spec.js
```diff
@@ -48,15 +48,26 @@ describe('FieldContainer', () => {
...
+        <FieldContainer id={CONTAINER_ID}>
+          {({ getInputProps }) => (
+            <div>
+              <input {...getInputProps({ 'data-test-id': 'input' }, { isDescribed: false })} />
+            </div>
+          )}
+        </FieldContainer>
       );
+
+      expect(findInput(wrapper)).toHaveProp('aria-describedby', null);
```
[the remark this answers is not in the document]
austin94: Unfortunately no. The syntax above would expect the prop to have `undefined` as a value, but is null as provided.

If we were to provide `undefined` through the render-prop it would actually show as `undefined` in the DOM. With `null` React is removing it from the attribute list.

packages/selection/src/containers/FieldContainer.spec.js
```diff
@@ -48,15 +48,26 @@ describe('FieldContainer', () => {
...
 
       expect(input).toHaveProp('id', `${CONTAINER_ID}--input`);
       expect(input).toHaveProp('aria-labelledby', `${CONTAINER_ID}--label`);
-      expect(input).toHaveProp(
-        'aria-describedby',
-        `${CONTAINER_ID}--hint ${CONTAINER_ID}--message`
+      expect(input).toHaveProp('aria-describedby', `${CONTAINER_ID}--hint`);
+    });
+
+    it('excludes aria-describedby if option is provided', () => {
```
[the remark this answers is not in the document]
austin94: Fixed. Was going to save this for the separate PR, but depending on what we view as the most common scenario, we might have this default to false.

Updated for now.

packages/radios/src/views/Label.example.md
```diff
@@ -1,38 +1,40 @@
...
-      </Radio>
-    </Col>
-    <Col md={4}>
-      <Radio disabled>
-        <Label>Disabled Label</Label>
-      </Radio>
-    </Col>
-  </Row>
-</Grid>
+<div role="group" aria-label="State Usage Example">
```
jzempel: Can these attributes apply to `<Grid>`? In other words, make `role="group"` automatic with the component and apply `aria-label` manually.

packages/tables/src/examples/overflow-menu.md
```diff
@@ -31,7 +34,7 @@ const OverflowMenu = ({ isHeader = false }) => (
         enabled: false
       },
       offset: {
-        fn: (data) => {
+        fn: data => {
```
jzempel: hopefully grabbed by rebase on #65

packages/textfields/src/views/__snapshots__/Input.spec.js.snap
```diff
@@ -2,6 +2,7 @@
 
 exports[`Input renders RTL styling 1`] = `
 <input
+  aria-invalid={false}
```
jzempel: Isn't `{false}` in these (and following) snaps unexpected?

utils/styleguide/ArgumentsRenderer/index.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import Argument from 'react-styleguidist/lib/rsg-components/Argument/ArgumentRenderer';
+import { zdFontSizeEpsilon, zdFontWeightSemibold } from '@zendeskgarden/css-variables';
+
+const ArgumentsContainer = styled.div`
+  margin-bottom: 8px;
```
jzempel: Use `xs` spacing from css-variables.

utils/styleguide/ArgumentsRenderer/index.js
```diff
@@ -0,0 +1,43 @@
...
+  font-size: inherit;
+`;
+
+const Heading = styled.div`
+  margin-bottom: 4px;
```
jzempel: Use `xxs` spacing from css-variables.

utils/styleguide/github-hljs-theme.css
```diff
@@ -0,0 +1,90 @@
+.hljs {
```
jzempel: If you can't access variables, it would be good to comment this file with selected color names from https://garden.zendesk.com/css-components/utilities/color/.

packages/radios/src/views/Label.example.md
austin94: These are no added to the `<Grid>`

packages/tables/src/examples/overflow-menu.md
austin94: Fixed with a rebase

packages/textfields/src/views/__snapshots__/Input.spec.js.snap
austin94: We are defaulting this to `false` to match the "default" state (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute).

Without any validation this would still be false.  AXE didn't seem to find any issues with this.

utils/styleguide/ArgumentsRenderer/index.js
austin94: Fixed

utils/styleguide/ArgumentsRenderer/index.js
austin94: Fixed

utils/styleguide/github-hljs-theme.css
austin94: I've added the equivalent color values as comments. I also updated all of the colors to match an equivalent in the palette so it should be a little clearer now.

packages/textfields/src/views/__snapshots__/Input.spec.js.snap
jzempel: But shouldn't it be `aria-invalid="false"` rather than `{false}`? I haven't seen JSX syntax in snapshots before.

jzempel: nvmd. Bogus expectations on my part here. 😳

ryanseddon:
> AXE plugin

Oh weird I thought from [this post](https://www.deque.com/blog/google-selects-deques-axe-chrome-devtools/) that lighthouse actually internally uses AXE in the audit?

austin94:
> Oh weird I thought from this post that lighthouse actually internally uses AXE in the audit?

I think it does, it at least linked to the dequeue links that AXE provides, but it seems to be a sub-set of the audits that the standalone plugin uses.

> So next steps is to introduce smaller PRs for these changes?

Yes.  Probably 3-4

austin94:
Closing this now that all areas have been merged separately.

PR #76, opened 2018-07-26
feat(tooltips): add zIndex prop to TooltipContainer

liaujianjie:
🚧 work-in-progress

* [ ] **BREAKING CHANGE?** 

## Description

Add `zIndex` prop to `TooltipContainer`

## Detail

The management of z-indexes on Support's lotus is quite funky and if `appendToBody` is true, the tooltip will be hidden behind the main ember app. Introducing a `zIndex` prop fixes this problem.

The `zIndex` props defaults to `1000` like in `MenuContainer`.

## Checklist

      designer as a reviewer)
  component


liaujianjie:
@austin94 This PR should be good to go. Just one thing: how would we write unit tests for `z-index`?

packages/tooltips/src/containers/TooltipContainer.js
```diff
@@ -90,13 +92,18 @@ class TooltipContainer extends ControlledComponent {
...
+     */
+    zIndex: PropTypes.number
   };
 
   static defaultProps = {
     placement: 'top',
     eventsEnabled: true,
-    delayMilliseconds: 500
+    delayMilliseconds: 500,
+    zIndex: 1000
```
jzempel: Unlike the menu component (and so that this is not a breaking change) the default should be `'auto'` – which is the CSS default. Note that the PropType needs to be flexible.

jzempel: Or, perhaps, no default is best. Either way, the PropType must accept valid CSS values (auto, inherit, etc)

packages/tooltips/src/containers/TooltipContainer.js
```diff
@@ -90,13 +92,18 @@ class TooltipContainer extends ControlledComponent {
      * @param {Function} renderProps.getTriggerProps - Props to be spread onto the trigger element
      * @param {Function} renderProps.isVisible - Whether the Tooltip is currently visible
      */
-    trigger: PropTypes.func
+    trigger: PropTypes.func,
+    /**
+     * The z-index of the popper.js placement container
+     */
+    zIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(['auto', 'inherit', 'unset'])])
```
jzempel: Missing `initial` – but I'm not sure we get too fussy about specifying props to this level of detail. We may prefer a simpler `PropTypes.oneOfType([PropTypes.number, PropTypes.string])`. But I'll let @austin94 weigh-in to say for sure.

liaujianjie: I'll change this after Austin weighs in. Thanks for taking a look!

ryanseddon: Yeah no need to specify exact string values `string` is enough

packages/tooltips/src/containers/TooltipContainer.js
```diff
@@ -90,13 +92,18 @@ class TooltipContainer extends ControlledComponent {
...
   };
 
   static defaultProps = {
     placement: 'top',
     eventsEnabled: true,
-    delayMilliseconds: 500
+    delayMilliseconds: 500,
+    zIndex: 'auto'
```
ryanseddon: The browser will default the zIndex for us no need to do it here.

packages/tooltips/src/elements/Tooltip.js
```diff
@@ -58,7 +58,11 @@ export default class Tooltip extends Component {
     /** Passes options to [Popper.JS Instance](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#new-popperreference-popper-options) */
     popperModifiers: PropTypes.object,
     size: PropTypes.oneOf([SIZE.SMALL, SIZE.MEDIUM, SIZE.LARGE, SIZE.EXTRA_LARGE]),
-    type: PropTypes.oneOf([TYPE.LIGHT, TYPE.DARK])
+    type: PropTypes.oneOf([TYPE.LIGHT, TYPE.DARK]),
+    /**
+     * The z-index of the popper.js placement container
+     */
+    zIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(['auto', 'inherit', 'unset'])])
```
ryanseddon: Same as above just `PropTypes.string` is enough

austin94:
Sorry for the delay, just got back from PTO ⛰ 

> Just one thing: how would we write unit tests for z-index

There are a few ways. The easiest is probably going to be an assertion with out styled-components extension for enzyme.

That allows us to use some common assertion syntax like `expect(...).toHaveStyle('z-index', 'foo')`

Lets meet up in person to figure out the best way forward with this (and to get me back up to speed).

liaujianjie:
Happy to contribute, thanks for all the help! 😄

PR #80, opened 2018-07-26
chore: add pendo snippet

jzempel:
* [ ] **BREAKING CHANGE?** 

## Description

Add pendo snippet to Garden site for test tracking of on-page component usage.

## Detail

Validated working snippet via gh-pages deploy and https://app.pendo.io/features by "Manage Features In-App" against a Garden URL (i.e. https://garden.zendesk.com/react-components/autocomplete/).


utils/styleguide/styleguide.base.config.js
```diff
@@ -40,14 +40,46 @@ const defaultStyleguideConfig = {
         {
           async: '',
           src: `//www.googletagmanager.com/gtag/js?id=${googleTrackingId}`
+        },
+        {
+          async: '',
+          src: '//static-staging.zdassets.com/customer_analytics_integration/garden_dev/cai.min.js'
```
ryanseddon: Can we just do straight up https?

jzempel: Just staying consistent w the other protocol-less URL.

ryanseddon: I think we should just make them all https no need to ever load this over http.

PR #89, opened 2018-08-02
fix(menu): Don't render Menu when it's closed

ryanseddon:
## Description

If you tried to make dynamic prop changes to `Menu` it would throw due to Popper.js trying to get the computed styles of `null`.

## Detail

Simple fix here is to just not render the Menu when the menu is closed.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/containers/MenuContainer.js
+++ b/packages/menus/src/containers/MenuContainer.js
@@ -499,97 +499,96 @@ class MenuContainer extends ControlledComponent {
               );
             }}
           </Target>
-          <Popper
-            placement={this.convertGardenToPopperPlacement()}
-            eventsEnabled={eventsEnabled}
-            modifiers={popperModifiers}
-          >
-            {({ popperProps, scheduleUpdate }) => {
-              const popperPlacement = popperProps['data-placement'];
-              const outOfBoundaries = popperProps['data-x-out-of-boundaries'];
+          {isOpen && (
+            <Popper
+              placement={this.convertGardenToPopperPlacement()}
+              eventsEnabled={eventsEnabled}
+              modifiers={popperModifiers}
+            >
+              {({ popperProps, scheduleUpdate }) => {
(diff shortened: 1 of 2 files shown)


packages/menus/src/containers/MenuContainer.js
```diff
@@ -499,97 +499,96 @@ class MenuContainer extends ControlledComponent {
...
-                        onStateChange={newState => {
-                          /**
-                           * We only care if `selectedKey` is involved with the state change
-                           */
-                          if (Object.prototype.hasOwnProperty.call(newState, 'selectedKey')) {
-                            let newSelectedKey = newState.selectedKey;
-
-                            if (newSelectedKey === undefined) {
-                              newSelectedKey = selectedKey;
+          {isOpen && (
```
zillding: I believe this is the only change? 😂

ryanseddon: Yep the rest was prettier reformatting, best viewed with `?w=1`

packages/menus/src/containers/MenuContainer.js
```diff
@@ -499,97 +499,96 @@ class MenuContainer extends ControlledComponent {
...
-            modifiers={popperModifiers}
-          >
-            {({ popperProps, scheduleUpdate }) => {
-              const popperPlacement = popperProps['data-placement'];
-              const outOfBoundaries = popperProps['data-x-out-of-boundaries'];
-
-              if (!isOpen) {
-                return false;
-              }
-
```
ryanseddon: This was deleted too @zillding

packages/menus/src/containers/MenuContainer.js
zillding: 💯 thanks, Ryan~

ryanseddon:
@jzempel coveralls has got confused due to the prettier format its complaining about new stuff that isn't actually new

PR #97, opened 2018-08-08
feat(menus): add maxHeight prop for scrollable menus

austin94:
## Description

While converting the `Menu` package I forgot to include a `maxHeight` prop to easily create scrollable Menus.

This should be something that doesn't require us to eject into a render-prop container.

## Detail

Adds a new `maxHeight` prop

screen shot 2018-08-08 at 2 25 59 pm

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/elements/Menu.js
+++ b/packages/menus/src/elements/Menu.js
@@ -7,6 +7,7 @@
 
 import React, { Children, cloneElement, isValidElement } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import { ControlledComponent, IdManager } from '@zendeskgarden/react-selection';
 import { hasType } from '@zendeskgarden/react-utilities';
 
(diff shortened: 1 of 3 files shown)


packages/menus/src/elements/Menu.js
```diff
@@ -18,6 +19,18 @@ import MediaItem from '../views/items/media/MediaItem';
 import NextItem from '../views/items/NextItem';
 import PreviousItem from '../views/items/PreviousItem';
 
+const ScrollableMenuView = styled(MenuView)`
+  /* stylelint-disable */
+  ${({ maxHeight }) =>
+    maxHeight
+      ? `
+    max-height: ${maxHeight};
+    overflow-y: scroll;
```
ryanseddon: this should probably just be `overflow: auto`

packages/menus/src/elements/Menu.spec.js
```diff
@@ -42,10 +42,11 @@ describe('Menu', () => {
   let wrapper;
   let onChangeSpy;
 
-  const basicExample = ({ onChange } = {}) => (
+  const basicExample = ({ onChange, maxHeight } = {}) => (
     <Menu
       data-test-id="menu"
       onChange={onChange}
+      maxHeight={maxHeight}
```
ryanseddon: You've added the prop here but aren't actually testing it?

packages/menus/src/elements/Menu.js
austin94: Fixed.

packages/menus/src/elements/Menu.spec.js
austin94: Weird, my tests didn't get added. Yay rebasing.

I've now updated with some tests that should get coveralls to stop complaining.

PR #109, opened 2018-08-12
feat(buttons|selection): allow customization of selected accessibility value

austin94:
This is a continuation of #66 

## Description

During the accessibility sweep I found an issue with the `aria-selected` value that we use by default in our `SelectionContainer` component.  There are certain usages where that selected attribute might be different (aria-pressed, aria-checked).

This PR introduces a customization option for the `getContainerProps` method for use in those instances.

I also fixed some assorted accessibility issues found in our react-buttons package including missing `aria-label`'s and adding `aria-pressed` into `ButtonGroupContainer`.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/buttons/src/containers/ButtonGroupContainer.js
+++ b/packages/buttons/src/containers/ButtonGroupContainer.js
@@ -101,7 +101,8 @@ export default class ButtonGroupContainer extends ControlledComponent {
         {({ getContainerProps, getItemProps }) =>
           render({
             getGroupProps: props => getContainerProps(this.getGroupProps(props)),
-            getButtonProps: props => getItemProps(this.getButtonProps(props)),
+            getButtonProps: props =>
+              getItemProps(this.getButtonProps(props), { selectedAriaKey: 'aria-pressed' }),
             focusedKey,
             selectedKey
           })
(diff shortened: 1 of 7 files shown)


austin94:
Also, looks like the deprecated `commands` option in lerna got added back in somehow. This fixes that deprecation warning as well.

lerna.json
```diff
@@ -1,7 +1,7 @@
 {
   "npmClient": "yarn",
   "version": "independent",
-  "commands": {
+  "command": {
```
jzempel: Yikes, just discovered that the 3.0 lerna bump around this is volatile. Given https://github.com/lerna/lerna/tree/master/commands/publish#deprecated-options, we'll need to reconfigure to use the `version` command (instead of publish). This will also necessitate a change to this line: https://github.com/zendeskgarden/react-components/blob/master/utils/travis/publish-packages.js#L16

jzempel: See https://github.com/zendeskgarden/css-components/commit/7ba1e3648130089d762ed1552e88a1bfeb6e1dfa for details. Except note that the `"push": "false"` still isn't working as expected 😬

jzempel: ...just tested and discovered that `"push": false` (no quotes) does the trick. Allowing you to check the commit/tags before sending off via `git push --follow-tags`. Oh, lerna.

PR #113, opened 2018-08-16
chore(lerna): ignore non-publish files

austin94:
## Description

Now that we are hitting a good development cadence in these packages, there have been a lot of package updates recently.  Most of these are necessary, but many have been caused by markdown and file changes that don't affect the distributed files.

In lerna v3 (or earlier and we missed it 😢) the `ignoreChanges` option does wonders for limiting the global affect documentation changes have on the repo.

I am also looking at ways to limit the `devDependency` relationship of internal packages since these are mostly used in the documentation. 

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,9 @@
   "npmClient": "yarn",
   "version": "independent",
   "command": {
+    "publish": {
+      "ignoreChanges": ["*.md", "*.spec.js", "*.snap", "styleguide.config.js"]
+    },
     "version": {
       "push": false,
       "allowBranch": "master",


lerna.json
```diff
@@ -2,6 +2,9 @@
   "npmClient": "yarn",
   "version": "independent",
   "command": {
+    "publish": {
+      "ignoreChanges": ["*.md", "*.spec.js", "**/__snapshots__/**", "styleguide.config.js"]
```
jzempel: note, this is actually a `version` option (`publish` is a soon-to-deprecate alias). However, the (not super great) docs recommend applying `ignoreChanges` to the config root. https://github.com/lerna/lerna/blob/master/commands/version/README.md#--ignore-changes

utils/styleguide/styleguide.base.config.js
```diff
@@ -174,7 +174,31 @@ const defaultStyleguideConfig = {
...
+        '@zendeskgarden/react-selection': path.resolve('..', 'selection', 'src'),
+        '@zendeskgarden/react-tables': path.resolve('..', 'tables', 'src'),
+        '@zendeskgarden/react-tabs': path.resolve('..', 'tabs', 'src'),
+        '@zendeskgarden/react-tags': path.resolve('..', 'tags', 'src'),
+        '@zendeskgarden/react-testing': path.resolve('..', 'testing', 'src'),
+        '@zendeskgarden/react-textfields': path.resolve('..', 'textfields', 'src'),
+        '@zendeskgarden/react-theming': path.resolve('..', 'theming', 'src'),
+        '@zendeskgarden/react-toggles': path.resolve('..', 'toggles', 'src'),
+        '@zendeskgarden/react-tooltips': path.resolve('..', 'tooltips', 'src'),
+        '@zendeskgarden/react-utilities': path.resolve('..', 'utilities', 'src')
```
jzempel: Please add a reminder for this to the https://github.com/zendeskgarden/react-components/blob/master/packages/.template/README.md TODO comment section.

lerna.json
jzempel: Also, is this right for `README.md` which is part of a package and should probably affect a version patch if changed.

utils/styleguide/styleguide.base.config.js
austin94: Added

lerna.json
austin94: > This option is best specified as root lerna.json configuration, both to avoid premature shell evaluation of the globs and to share the config with lerna diff and lerna changed:

Great points. Added.

> Also, is this right for README.md

I have mixed feelings about this. If it's just a markdown change this is _technically_ a documentation change. If we need to publish we can do a force tag locally with the `--no-ignore-changes`? Not sure how often this will be

jzempel: Yup. Fine w me.

PR #119, opened 2018-08-20
fix(tooltips): fix rendering stutter in Safari

austin94:
## Description

Since tooltips was one of our first usages of Popper.js the visibility logic was slightly different than Menus, Select, and other usages within the codebase.  This difference causes a stutter effect for some renderings in Safari.

## Detail

Previously, we handled visibility with a custom `aria-hidden` value to toggle the visibility on/off. This was originally for an accessibility requirement that was shown to not be needed in the recent Axe/Lighthouse walkthrough that I recently completed.

This PR:
- Toggles visibility with a ternary rather than a css:hidden trick
- Adds accessibility changes that was originally surfaced in #66 
- Updates tests to work with new logic

Thanks @zillding for finding this one!

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/tooltips/README.md
+++ b/packages/tooltips/README.md
@@ -31,7 +31,7 @@ import { Tooltip } from '@zendeskgarden/react-tooltips';
  * Place a `ThemeProvider` at the root of your React application
  */
 <ThemeProvider>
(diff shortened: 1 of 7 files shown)


austin94:
> as long as it works across browsers

Tested and working as expected without a stutter in all browsers

PR #128, opened 2018-08-23
fix(lerna): update lerna publish to use built-in command

austin94:
## Description

With the changes to `lerna --since` that broke our publish process in Travis. This is a good excuse to upgrade these commands to a more "standard" lerna configuration.  

Now that we no-longer have a custom `npm publish` process I've updated the travis command to use:

`yarn lerna publish from-git --no-verify-access --yes`

- `from-git`
  - Will only publish if correctly tagged by Lerna (so no commit message check necessary)
- `--no-verify-access`
  - This is a little sketchy, but it seems just setting the NPM _authToken doesn't allow Lerna to know a user is "logged in". If a user doesn't have access it would just fail on publish (same as before)
- `--yes`
  - Skip confirmation prompts since we're in CI

## Checklist

      designer as a reviewer)
  component


.eslintrc.json
```diff
@@ -7,17 +7,17 @@
     "plugin:jsx-a11y/recommended",
     "prettier"
   ],
+  "settings": {
```
austin94: Added due to some warnings after the ESLINT upgrade

Issue #137, opened 2018-08-31
Menu doesn't close when an outside click hits an iframe


sunesimonsen:
## Expectations

Click inside iframes closes the menu:

![iframe click works](https://user-images.githubusercontent.com/90802/44903027-c53aa300-ad0b-11e8-8a10-f5900cded9e5.gif)

## Reality

It doesn't:

![iframe click](https://user-images.githubusercontent.com/90802/44903098-ec917000-ad0b-11e8-98ba-5364ad44d865.gif)

## Steps to Reproduce

  1. Open a menu
  2. Click in an iframe
  3. The menu is not closed

###### Fine Print

* Component: menus
* Browsers: all

sunesimonsen:
I also reproduced it on the styleguide just to isolate the problem:

![iframe click styleguide](https://user-images.githubusercontent.com/90802/44904920-d20dc580-ad10-11e8-8fa3-b871288dfb58.gif)

ryanseddon:
Thanks @sunesimonsen I know exactly why this isn't working.

Right now we rely on an event that is attached to the `document` so when clicking the iframe it never triggers that event and never closes, I'll have a play around but I think we can get away with just triggering the close when the menu is blurred. 

I'll put up a PR shortly.

sunesimonsen:
This is how it was done in the old solution https://github.com/zendeskgarden/react-components-archive/blob/master/src/core/RelativePositionedPopup/index.js#L118-L166

But if it is possible to make a solution that works with blur that would also be great. The usual problem is that you need to listen for mousedown on everything and prevent default to avoid unwanted blur events.

sunesimonsen:
I just tried the new solution in Knowledge Client and it unfortunately don't work. The reason that the listener on the documents isn't using [event capturing](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) like the implementation mentioned about and on the PR, so if the iframe calls `event.stopPropagation()` this won't work.

I can try to give it a shot if you like.

PR #141, opened 2018-09-03
fix(menus): Use blur to close menu when clicking away

ryanseddon:
Fixes: #137 

## Description

If you click outside the menu when open and your cursor happens to be in a different document context the menu won't close.

Rather than check if the event occurred outside the container we just attach a blur event.

## Detail

With blur event:
close_menu_blur

## Todo
* [ ] This doesn't close the menu if you click the button again, ideas welcome to guard against that.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/containers/MenuContainer.example.md
+++ b/packages/menus/src/containers/MenuContainer.example.md
@@ -240,9 +240,7 @@ initialState = {
         placement="end"
         onChange={selectedKey => setState({ selectedKey })}
         trigger={({ getTriggerProps, triggerRef, isOpen }) => (
-          <button {...getTriggerProps({ ref: triggerRef, active: isOpen })}>
-            Heavily customized menu
-          </button>
+          <button {...getTriggerProps({ ref: triggerRef })}>Heavily customized menu</button>
         )}
       >
         {({ getMenuProps, menuRef, placement, getItemProps, focusedKey }) => (
--- a/packages/menus/src/containers/MenuContainer.js
+++ b/packages/menus/src/containers/MenuContainer.js
@@ -146,6 +146,7 @@ class MenuContainer extends ControlledComponent {
       selectedKey: undefined,
       focusOnOpen: undefined,
       closedByBlur: undefined,
+      closingWithClick: undefined,
       defaultFocusedIndex: 0
     };
   }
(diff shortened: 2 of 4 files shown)


sunesimonsen:
I tested the solution out locally playing around with the different menus in the styleguide, it seems to work really well and is quite simple 👍

ryanseddon:
@sunesimonsen I've pushed 3651746 which handles clicking on the trigger twice and correctly closes it and the blur event gets ignored.

sunesimonsen:
Thanks I'll look at it later today.

sunesimonsen:
@ryanseddon the fix seems to work 👍 Another approach would be to use `preventDefault` in the mousedown, that also avoids the blur.

austin94:
It looks like this doesn't currently work with the `Inline Search` example in the styleguide? Might need to be a tweak in the example.

I also found an invalid prop addition in an example from when I did the accessibility sweep. Just pushed a commit to address that in this PR.

sunesimonsen:
Mhh @austin94 is right this might be challenging for menus containing inputs. Alternatively we can use the same approach as the old library.

ryanseddon:
> It looks like this doesn't currently work with the Inline Search example in the styleguide? Might need to be a tweak in the example.

Bummer I missed this one, I don't think this option is viable then.

Only other solution I can think of now is treat this like a modal and give menu an transparent backdrop that handles the outside click logic. Thoughts?

sunesimonsen:
I think an overlay would be a bad idea, I think we should copy the old solution I linked to in the issue https://github.com/zendeskgarden/react-components-archive/blob/master/src/core/RelativePositionedPopup/index.js#L118-L166 it works really great, I'm yet to see it fail. The only edge case is if and iframe appears while the menu is open, but I don't think that would ever happen.

Anifacted:
> Only other solution I can think of now is treat this like a modal and give menu an transparent backdrop that handles the outside click logic. Thoughts?

While this could work, it would effectively mean that all UI elements below the level of the current popover would be unclickable. Or the user would have to click once to dismiss and then again to trigger the target item. Not sure that this would be the desired behavior?
It would also be adding a bunch of different background elements in between the triggers making up the chain of menus etc, which would be slightly obscuring the markup.

I also lean towards the document(s)-listener approach that @sunesimonsen refers to with the logic for detecting nesting as introduced here: https://github.com/zendeskgarden/react-components-archive/pull/426

sunesimonsen:
Thanks that sounds great. We can also try to help if you prefer. But I would guess this change requires a pretty deep understanding of the implementation, so maybe it would be easier for you.

Kind regards 
Sune

PR #144, opened 2018-09-07
fix(modals): FocusJailContainer correctly preserves logical tab order

ryanseddon:
## Description

Currently if an item not in the tab flow (tabIndex=-1) of FocusJailContainer is programmatically focused pressing tab will jump to first tabbable item and not the next logical item.

## Detail

To fix the issue I've rewritten the logic that intercepts TAB to only fire when it's the last item or the first when holding shift down.

So the browser will handle tabbing and changing focus on all other elements within FocusJail and the logic will intercept to keep it within container for first/last items.

### Before:
focusjail_before

### After:
focusjail_after

## Checklist

### DIFF
--- a/packages/modals/src/containers/FocusJailContainer.spec.js
+++ b/packages/modals/src/containers/FocusJailContainer.spec.js
@@ -8,7 +8,7 @@
 import React from 'react';
 import { mountWithTheme } from '@zendeskgarden/react-testing';
 import { KEY_CODES } from '@zendeskgarden/react-selection';
-import FocusJailContainer from './FocusJailContainer';
+import { FocusJailContainer } from './FocusJailContainer';
 
 jest.useFakeTimers();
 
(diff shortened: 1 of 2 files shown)


packages/modals/src/containers/FocusJailContainer.spec.js
```diff
@@ -96,27 +95,44 @@ describe('FocusJailContainer', () => {
...
 
       it('focuses last element if tab and shift key is pressed', () => {
         wrapper.simulate('keydown', { keyCode: KEY_CODES.TAB, shiftKey: true });
-        jest.runOnlyPendingTimers();
 
         expect(focusElementSpy).toHaveBeenCalledTimes(2);
         expect(focusElementSpy).toHaveBeenLastCalledWith(findInput(wrapper).getDOMNode());
       });
+
+      it("doesn't intercept tab key if not the first or last tabbable item", () => {
```
austin94: Could you run a `yarn format`? Not sure why the `"` isn't getting caught in the linting

ryanseddon: heh yarn format changed that to double quotes I originally had an escape `\'` with single quotes

jzempel: Go ahead and merge, then! (Or use a curly single quote ;)

PR #145, opened 2018-09-10
fix(menus): Change handleOutsideMouseDown to handle when clicking on iframes

ryanseddon:
Fixes: #137 

## Description

Take two, as @sunesimonsen suggested, we had a pretty good solution in our previous menu component to handle closing the menu when clicked outside of the containers.

## Detail

This adds a mousedown event to any same origin iframe documents and triggers the toggle function to see if it should close or not.

## Checklist

### DIFF
--- a/packages/menus/src/containers/MenuContainer.js
+++ b/packages/menus/src/containers/MenuContainer.js
@@ -151,15 +151,15 @@ class MenuContainer extends ControlledComponent {
   }
 
   componentDidMount() {
-    const doc = getDocument ? getDocument(this.props) : document;
-
-    doc.addEventListener('mousedown', this.handleOutsideMouseDown);
+    this.getDocuments().forEach(doc => {
+      doc.addEventListener('mousedown', this.handleOutsideMouseDown);
+    });
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


packages/menus/src/containers/MenuContainer.js
```diff
@@ -195,18 +195,38 @@ class MenuContainer extends ControlledComponent {
     }
   }
 
+  getDocuments = () => {
+    const doc = getDocument ? getDocument(this.props) : document;
+    const iframes = doc.querySelectorAll('iframe');
+
+    return [...iframes].reduce(
```
austin94: This _should_ be good, but lets check that this is transpiled into IE11 compatible syntax.

sunesimonsen:
> Very clean. Thanks @sunesimonsen for the guidance on this one!

🤝

packages/menus/src/containers/MenuContainer.js
ryanseddon: Yeah babel handles this fine transforms it to:

```js
return [].concat(_toConsumableArray(iframes)
```

Tested this already in all browsers in the checkbox marked in the description. So IE11 and edge are all good.

PR #149, opened 2018-09-13
fix(menus): Lazy attach auto close handlers

sunesimonsen:
Fixes: #137

## Description

Take three, we had a pretty good solution in our previous menu component to handle closing the menu when clicked outside of the containers. This PR mirrors the functionality even closer. 

## Detail

* Attaches the click outside handler to all documents when a menu opens.
* Detaches the click outside handler when the menu closes or the component is unmounted.
* Using the event capturing phase to beat `event.stopPropagation()`

## Checklist

### DIFF
--- a/packages/menus/src/containers/MenuContainer.js
+++ b/packages/menus/src/containers/MenuContainer.js
@@ -152,13 +152,13 @@ class MenuContainer extends ControlledComponent {
 
   attachClickOutsideHandler() {
     this.getDocuments().forEach(doc => {
-      doc.addEventListener('mousedown', this.handleOutsideMouseDown, true);
+      doc.addEventListener('mousedown', this.handleOutsideMouseDown, true /* use capture */);
     });
   }
 
   detachClickOutsideHandler() {
     this.getDocuments().forEach(doc => {
-      doc.removeEventListener('mousedown', this.handleOutsideMouseDown, true);
+      doc.removeEventListener('mousedown', this.handleOutsideMouseDown, true /* use capture */);
     });
   }
 


packages/menus/src/containers/MenuContainer.spec.js
```diff
@@ -254,17 +254,41 @@ describe('MenuContainer', () => {
...
+      it('closes the menu', () => {
+        wrapper.simulate('click');
+
+        expect(findMenu(wrapper)).not.toExist();
+      });
+
+      it('removes click outside event listener', () => {
+        const removeEventListenerSpy = jest.fn();
+
+        document.removeEventListener = removeEventListenerSpy;
```
luis-almeida: Should we restore the original method not to temper with future tests?
Not sure what's the cleanest way do this with `jest` but with `sinon` it looks something like this:
``` js
const removeEventListenerSpy =  sinon.spy(document, 'removeEventListener');
// ... assert
removeEventListenerSpy.restore();
```

sunesimonsen: It is a good point, I just copied it from the old test. I'll just fix that.

luis-almeida: The update looks good 👍

packages/menus/src/containers/MenuContainer.js
```diff
@@ -150,18 +150,22 @@ class MenuContainer extends ControlledComponent {
     };
   }
 
-  componentDidMount() {
+  attachClickOutsideHandler() {
     this.getDocuments().forEach(doc => {
-      doc.addEventListener('mousedown', this.handleOutsideMouseDown);
+      doc.addEventListener('mousedown', this.handleOutsideMouseDown, true);
```
jzempel: It's always friendly to provide a quick inline comment reminder for boolean parameters: i.e. `doc.addEventListener(..., true /* useCapture*/);

austin94:
Are you able to link this locally within Guide to ensure that it works? We can also publish a `beta` tag if you need to test that way. Thanks for helping with this one!

ryanseddon:
> Coverage decreased (-0.07%)

Seriously I don't get coveralls to me you've added tests yet almost every pr it slightly declines

austin94:
Perfect, I'll take it through the release cycle.

> I'll try to see how hard it is to link into the guide-chrome, but my guess is that it is not easy

With yarn it's gotten a lot easier. Just `yarn build` in the repo and then you can `yarn add ~/Repos/react-components/packages/menus` to install it without worrying about duplicated dependencies.

sunesimonsen:
I know how to link stuff. But it needs to get through 3 steps of bundling, so that was what I thought might be challenging 😜

sunesimonsen:
Thanks 💯

austin94:
Just published to the registry

```
 - @zendeskgarden/react-menus: 3.6.2 => 3.6.3
 - @zendeskgarden/react-select: 2.4.3 => 2.4.4
```

Cheers 🥂

PR #152, opened 2018-09-14
fix(select): remove array render to support React <16

### LINKED ISSUE #150: Select does not work on React versions below 16
The Select examples from the Styleguide should render on React 15.6.

## Reality

```
Uncaught Error: FieldContainer.render(): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.
```

It returns an array, because `FieldContainer.render()` (at https://github.com/zendeskgarden/react-components/blame/master/packages/selection/src/containers/FieldContainer.js#L86) calls the anonymous render prop function in `SelectField` (at https://github.com/zendeskgarden/react-components/blob/master/packages/select/src/elements/SelectField.js#L63-L92) which returns `React.Children.map(...)`.

React.Children.map() returns an array.

For comparison, the TextField's render prop function wraps the array returned by `React.children.map()` in a `<TextGroup>`, which avoids this problem. TextField works fine in React 15.6.

###### Fine Print

* Component: SelectField
* Browsers: all

austin94:
Closes #150 

## Description

Due to the ordering of our `render()` method in the `SelectField` component, we were unintentionally returning an array. This was breaking in React version < 16.

This PR updates the render method to now match the `TextField` component and remove the array.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/select/src/elements/SelectField.js
+++ b/packages/select/src/elements/SelectField.js
@@ -58,15 +58,15 @@ export default class SelectField extends ControlledComponent {
     this.selectRef = undefined;
 
     return (
-      <SelectGroup {...otherProps}>
-        <FieldContainer>
-          {({
-            getLabelProps: getFieldLabelProps,
-            getInputProps: getFieldInputProps,
-            getHintProps,
-            getMessageProps
-          }) =>
-            Children.map(children, child => {
+      <FieldContainer>
+        {({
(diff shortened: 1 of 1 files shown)


austin94:
@aravindet could you `yarn add` this locally into your codebase to ensure that this fixes the issue? I ran it locally in `v15.4`, but just want to double-check.

PR #159, opened 2018-09-17
fix(autocomplete): remove event default prevention for all right arrow presses

austin94:
## Description

@albertldlan found a bug in our event default prevention logic in `AutocompleteContainer`. Regardless of if we performed an action from a `RIGHT` keyCode, we were calling `preventDefault`. 

This was causing all `RIGHT` keyboard navigation in an input to fail.

## Checklist


austin94:
> You'll need this on KEY_CODES.LEFT for RTL, correct?

Correct. Just pushed an additional fix.

austin94:
Looking through the logic it seems to be mirrored correctly. Lets look through this offline to check the logic.

PR #161, opened 2018-09-18
fix(IdManager): apply custom prefix to all ID generation

### LINKED ISSUE #157: Textfield-Checkbox ID collision
## Expectations

HTML IDs should not collide. Clicking on a label should reliably focus the input next to it.

Also it should be possible to use different packages at slightly different versions (as long as peer dependencies are satisfied).

## Reality

I have a form with some text inputs and checkboxes as siblings. The rendered DOM contains:

```html
<input data-garden-id="textfields.input" data-garden-version="3.3.6"
  ... id="garden-0--input" ... >

...

<input type="checkbox" data-garden-id="checkboxes.input" data-garden-version="3.3.5"
  ... id="garden-0--input" ... >
```

As you can see, the `data-garden-version` differs by a patch increment and the id (`garden-0--input`) clashes.

The root cause seems to be that each package has `react-selection` as a dependency, and this brings in two copies of `IdManager`. As the Id Manager uses a module-local variable to keep sequential IDs, this is guaranteed to clash.

## Steps to Reproduce

1. Install two versions of textfields and checkbox so that they bring in different versions of react-selection.
2. Put a checkbox (using `<Checkbox>`) and a text input (using `<TextField>`) on the same page so they are rendered at the same time.
3. Clicking the label of the second one (in DOM order) will focus the first input.

## Suggestions

1. Use UUID
2. Use a different suffix/prefix in the IDs for each package. E.g. `garden-0--checkbox-input`.
3. Make the ID counter global (window.idCounter?)
4. Put the ID counter in some sort of Provider
...

###### Fine Print

* Component: Checkbox, Textfields, Radio?
* Browsers: all

austin94:
Closes #157 

## Description

To help reduce the number of duplicate IDs that are found within duplicated `react-selection` packages that use `IdManager` I have added in scoped ID prefixes for all containers and elements.

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/autocomplete/src/containers/AutocompleteContainer.js
+++ b/packages/autocomplete/src/containers/AutocompleteContainer.js
@@ -133,7 +133,7 @@ class AutocompleteContainer extends ControlledComponent {
 
     this.state = {
       isOpen: false,
-      id: IdManager.generateId(),
+      id: IdManager.generateId('garden-autocomplete-container'),
       focusedKey: undefined,
       tagFocusedKey: undefined
     };
--- a/packages/buttons/src/containers/ButtonGroupContainer.js
+++ b/packages/buttons/src/containers/ButtonGroupContainer.js
@@ -53,7 +53,7 @@ export default class ButtonGroupContainer extends ControlledComponent {
     this.state = {
       focusedKey: undefined,
       selectedKey: undefined,
-      id: IdManager.generateId()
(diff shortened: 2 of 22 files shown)


austin94:
> is this a breaking change?

Since all components allow consumers to override all of the these IDs this shouldn't affect anyones usage. Also, these IDs are uniquely generated so no-one should be referencing them in any way.

PR #162, opened 2018-09-18
feat(accessibility): add isDescribedBy flag to FormContainer and update global accessibility values

austin94:
No breaking changes with this one 🎉  Wraps up #66 

Everything should be backwards compatible and act as a single `feat` release.

## Description

This PR closes up the remaining work in #66 and removes the last of the AXE auditing failures.

## Detail

This PR includes assorted DOC changes, aria-labels for examples, and some wording tweaks, but the majority of changes are:

* `AutocompleteContainer`
  * Remove invalid `aria-owns` value
* `MenuContainer`
  * Remove invalid `aria-controls` value
* `FieldContainer`
  * Now optionally include the `aria-describedby` values based on an optional customization
* `Checkbox | Radio | Ranges | Textfields | Toggles`
  * Use the new `FieldContainer` with updates

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/checkboxes/src/elements/Checkbox.spec.js
+++ b/packages/checkboxes/src/elements/Checkbox.spec.js
@@ -40,7 +40,7 @@ describe('Checkbox', () => {
   });
 
   it('applies container props to Message', () => {
-    expect(wrapper.find(Message)).toHaveProp('id', `${CHECKBOX_ID}--message`);
+    expect(wrapper.find(Message)).toHaveProp('role', 'alert');
   });
 
   it('applies no props to any other element', () => {
--- a/packages/autocomplete/src/containers/AutocompleteContainer.js
+++ b/packages/autocomplete/src/containers/AutocompleteContainer.js
@@ -318,12 +318,10 @@ class AutocompleteContainer extends ControlledComponent {
   };
 
   getItemId = key =>
-    typeof key === 'undefined' ? '' : `${this.getControlledState().id}--item-${key}`;
+    typeof key === 'undefined' ? null : `${this.getControlledState().id}--item-${key}`;
 
   getTagId = key =>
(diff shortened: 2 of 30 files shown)


packages/checkboxes/src/elements/Checkbox.spec.js
```diff
@@ -40,7 +40,7 @@ describe('Checkbox', () => {
   });
 
   it('applies container props to Message', () => {
-    expect(wrapper.find(Message)).toHaveProp('id', `${CHECKBOX_ID}--message`);
+    expect(wrapper.find(Message)).toHaveProp('role', 'alert');
```
ryanseddon: role alert will interrupt the screen reader and read that message when it appears is that what we want?

ryanseddon: But that isn't always the case right? I can not supply a prop and use it to just display ancillary message that may not be validation related.

ryanseddon: Could you do some logic there that if there is no `validation` prop or it's value is none don't apply the role alert?

austin94: We could conditionally apply the `getMessageProps` based on that value. Or would that be too abstract?

ryanseddon: Hmm not sure I think consumers would be used to using propGetter pattern.

austin94: Good point. I think it would be best to start the deprecation process on this one. In the most recent commit I've:

- Added a deprecation warning to `getMessageProps`
- Updated testing to cover deprecation warning
- Remove all internal uses of `getMessageProps` in the form related components.

PR #167, opened 2018-09-26
feat(loaders): introduce Spinner loader

ryanseddon:
* [ ] **BREAKING CHANGE?** 

## Description

This PR introduces the the `<Spinner />` component that uses SVG animations rather than our original gif-based option.

Some features of the loader:
- Delayed rendering to avoid quick UI flashes for common loading times (750ms & customizable)
- Able to inherit size and color based on parent css values (color & font-size)

This steps through the pre-computed frames captured from the AE export.

## Detail

spinner_loader

## Checklist

      designer as a reviewer)


packages/loaders/src/utils/spinner-coordinates.js
```diff
@@ -0,0 +1,206 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export const dPathFrames = [
```
jzempel: In the case of this brute force method, wouldn't it be cleaner/more elegant to extract the keyframe values from https://github.com/zendeskgarden/css-components/blob/jzempel/loaders/packages/loaders/src/_keyframes.css#L92-L148 and apply them to a `<circle>` element (like this SVG: https://github.com/zendeskgarden/css-components/blob/jzempel/loaders/packages/loaders/src/svg/spinner.svg)? In order to work for IE11 the values would need to be applied to attributes of the `<circle>` (rather than CSS properties). I think you'd end up with tighter, simpler to understand code.

packages/loaders/src/Dots.js
```diff
@@ -104,7 +106,14 @@ export default class Dots extends React.Component {
     const dotThreeFrame = this.retrieveFrame(2 / 3);
 
     return (
-      <StyledSVG fontSize={size} color={color} {...other}>
+      <StyledSVG
+        fontSize={size}
+        color={color}
+        width="80"
+        height="72"
+        data-garden-id={COMPONENT_ID}
```
austin94: Did you refactor this to be `dataGardenId` in `styled-elements`?

packages/loaders/src/utils/spinner-coordinates.js
jzempel: p.s. take a look at https://github.com/zendeskgarden/css-components/blob/jzempel/loaders/demo/loaders/index.js#L103-L116 to see how I converted the values you're using below to the keyframe data seen in the `css-components` branch.

packages/loaders/src/Dots.js
ryanseddon: yeah so we can re-use `<StyledSvg />` component for both loaders I pass down the id from here.

packages/loaders/src/utils/spinner-coordinates.js
ryanseddon: of course 🤦‍♂️  that'll shrink it a bit!

packages/loaders/src/utils/spinner-coordinates.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export const strokeWidthFrames = {
```
austin94: [NIT] if these are constants could we go with ALL CAPS? `STROKE_WIDTH_FRAMES`

packages/loaders/src/styled-elements.js
```diff
@@ -23,42 +23,62 @@ StyledCircle.propTypes = {
...
+  cy: 40,
+  r: 34,
+  fill: 'none',
+  stroke: 'currentColor',
+  strokeLinecap: 'round',
+  strokeWidth: ({ strokeWidthValue }) => strokeWidthValue,
+  strokeDasharray: ({ dasharrayValue }) => `${dasharrayValue} 250`,
+  transform: props => props.transform
+})`
+  transform-origin: center center;
```
jzempel: Does it make sense to identify malleable attributes via `will-change` here? If not, then we should probably remove `will-change` from the Dots circle above.

packages/loaders/src/styled-elements.js
```diff
@@ -23,42 +23,62 @@ StyledCircle.propTypes = {
   transform: PropTypes.string
 };
 
-const COMPONENT_ID = 'loaders.dots';
+export const SpinnerCircle = styled.circle.attrs({
```
jzempel: Should rename above `StyledCircle` to `DotsCircle` for alignment and clarity.

packages/loaders/src/styled-elements.js
```diff
@@ -23,42 +23,62 @@ StyledCircle.propTypes = {
...
   color: PropTypes.string,
   fontSize: PropTypes.any
 };
 
-export const StyledSVG = ({ children, fontSize, ...other }) => {
+export const StyledSVG = ({ children, fontSize, width, height, ...other }) => {
   return (
-    <StyledSvg fontSize={fontSize} {...other}>
+    <StyledSvg fontSize={fontSize} width={width} height={height} {...other}>
       <g fill="currentColor">{children}</g>
```
jzempel: This `<g>` element should not be applied to the spinner's SVG – it's only applicable to the grouping of Dots.

packages/loaders/src/utils/spinner-coordinates.js
```diff
@@ -0,0 +1,93 @@
...
+  75: 62,
+  76: 53,
+  78: 47,
+  80: 37,
+  81: 27,
+  83: 23,
+  85: 16,
+  90: 7,
+  98: 1
+};
```
jzempel: If these frames end up working well for us, we should revisit the Dots loader and pull keyframes from https://github.com/zendeskgarden/css-components/blob/jzempel/loaders/packages/loaders/src/_keyframes.css accordingly. May not be necessary for this PR, but should be done as a follow-on.

packages/loaders/src/utils/spinner-coordinates.js
```diff
@@ -0,0 +1,93 @@
...
+  31: 4,
+  36: 3,
+  48: 2,
+  56: 3,
+  68: 4,
+  80: 5,
+  91: 6
+};
+
+export const rotationFrames = {
```
jzempel: Why are so many more values needed here than used by https://github.com/zendeskgarden/css-components/blob/jzempel/loaders/packages/loaders/src/_keyframes.css#L114-L136?

packages/loaders/src/Spinner.example.md
```diff
@@ -0,0 +1,137 @@
...
+- Delays the initial render of the component to reduce UI shifting
+  in normal loading conditions
+- Includes accessibility attributes to identify as an `indeterminate progress bar`
+
+### Basic Usage
+
+```jsx
+const { zdColorBlue500 } = require('@zendeskgarden/css-variables');
+
+<Spinner color={zdColorBlue500} size="50px" />;
```
jzempel: [nit] please update all examples from current `50px` to `48px` for base-4 consistency :wink:

packages/loaders/src/utils/spinner-coordinates.js
jzempel: btw, we're seeing much smoother movement (easy to observe when slowed down in the demo) from the keyframe values under the `jzempel/loaders` branch of `css-components`. Have you tried those exact values along with your interpolation logic here?

packages/loaders/src/Spinner.example.md
```diff
@@ -0,0 +1,137 @@
...
+        <Col md={6}>
+          <RangeField>
+            <Label>
+              Duration {state.duration}
+              ms
+            </Label>
+            <Range
+              value={state.duration}
+              onChange={event => setState({ duration: parseFloat(event.target.value) })}
+              min={1250}
```
jzempel: Please add more realistic min/max values here. The default is, indeed, 1250ms. Extremes would be twice as fast and/or slow.

packages/loaders/src/utils/spinner-coordinates.js
ryanseddon: I tried these and I'm getting some jitter towards the end I'm pretty sure this was the same on my values too. See [test case](https://jsbin.com/kecosapiye/1/edit?js,output).

kapture 2018-10-15 at 16 02 30

Seems the rotation linear progression is too slow for the dasharray linear changes any ideas?

packages/loaders/src/Spinner.js
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  computeFrames = frames => {
+    const { duration } = this.props;
+    const { totalFrames } = this.state;
+
+    return Object.entries(frames).reduce((acc, item, index, arr) => {
+      const [frame, value] = item;
+      const [nextFrame, nextValue] = arr[index + 1] || ['100', arr[0][1]];
+      const diff = nextFrame - frame - 1;
```
jzempel: I think this should be `const diff = nextFrame - frame;`. Currently, `computeFrames` produces out-of-order values in the array.

packages/loaders/src/utils/spinner-coordinates.js
jzempel: Along with modifying the diff calculation in `computeFrames`, I did a fair bit of fussing with the keyframe values. I'm getting much better results with the following:

```js
const ROTATION_FRAMES = {
  0: -90,
  8: -81,
  36: -30,
  41: -18,
  44: -8,
  48: 0,
  55: 22,
  63: 53,
  64: 62,
  66: 67,
  68: 78,
  69: 90,
  71: 99,
  73: 112,
  74: 129,
  76: 138,
  78: 159,
  79: 180,
  81: 190,
  83: 207,
  84: 221,
  86: 226,
  88: 235,
  90: 243,
  99: 270
};

const DASHARRAY_FRAMES = {
  0: 0,
  13: 2,
  26: 13,
  53: 86,
  58: 90,
  63: 89,
  64: 88,
  66: 86,
  68: 83,
  69: 81,
  71: 76,
  73: 70,
  74: 62,
  76: 58,
  78: 47,
  79: 37,
  81: 31,
  83: 23,
  84: 16,
  88: 10,
  89: 7,
  98: 1,
  99: 0
};
```

See modified [test case](https://jsbin.com/facogisudu/edit?js,output).

packages/loaders/src/utils/spinner-coordinates.js
```diff
@@ -0,0 +1,64 @@
...
+  0: 6,
+  14: 5,
+  26: 4,
+  36: 3,
+  46: 2,
+  58: 3,
+  70: 4,
+  80: 5,
+  91: 6,
+  100: 6
```
jzempel: Shouldn't need this `100: 6`.

packages/loaders/src/utils/spinner-coordinates.js
jzempel: Result...

spinner

PR #175, opened 2018-10-05
fix(testing): fix enzyme options passing

ssidorchik:
* [ ] **BREAKING CHANGE?** 

## Description

The Enzyme [`mount`](https://github.com/airbnb/enzyme/blob/9d2ebb7b35197739c3f30a3a0d9a6c21dd6ee95a/packages/enzyme/src/mount.js#L9), [`render`](https://github.com/airbnb/enzyme/blob/9d2ebb7b35197739c3f30a3a0d9a6c21dd6ee95a/packages/enzyme/src/render.js#L18), [`shallow`](https://github.com/airbnb/enzyme/blob/9d2ebb7b35197739c3f30a3a0d9a6c21dd6ee95a/packages/enzyme/src/shallow.js#L9) helpers accept only 2 arguments. So, this fix spreads the passed `enzymeOptions` into the second one before `context` and `childContextTypes` properties.

If `context` or `childContextTypes` is passed through the options, should be there any warning that these properties will be overridden?

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/menus/src/containers/MenuContainer.spec.js
+++ b/packages/menus/src/containers/MenuContainer.spec.js
@@ -132,6 +132,8 @@ describe('MenuContainer', () => {
     findTrigger(wrapper).simulate('click');
 
     expect(wrapper.find('Portal').length).toBeGreaterThan(0);
+
+    wrapper.unmount();
   });
 
   it("doesn't render Popper element if menu is not open", () => {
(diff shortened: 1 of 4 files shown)


ssidorchik:
This one seems to be tricky, since it doesn't fail when runs on its own. In Enzyme [docs](https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md) it's mentioned that tests may affect each other if they are using the same DOM. And it is suggested to use `unmount` for cleanup. I tried to do this for the first test and it fixes the failing test. But this doesn't look like the solid solution. @austin94 what do you think about running `unmount` for every test in the `afterEach` block?

PR #178, opened 2018-10-08
chore(docs): add CodeSandbox template example

austin94:
## Description

This PR adds documentation linking to a new CodeSandbox template: https://codesandbox.io/s/1qw90o82pl

Please take a look and let me know if there is anything else that should be added. It is currently using `create-react-app@v2` so everything should be as up-to-date as possible.

## Detail

screen shot 2018-10-08 at 3 35 06 pm

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,10 @@ components are packaged and published individually, but combined under
 this single repository. Components rely on [Garden
 CSS](https://github.com/zendeskgarden/css-components) for styling.
 
+Try out the Garden React Components in a mock product environment with CodeSandbox
+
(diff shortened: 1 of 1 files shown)


ryanseddon:
Hooray for codesandbox! We should definitely make this into a kitchensink example and just include all the the things.

austin94:
> We should definitely make this into a kitchensink example and just include all the the things.

I've updated the code example to include all of our current component packages. I've also pre-added the scoped `styles.css` for each in the `global-styles.js` file.

PR #187, opened 2018-10-15
feat(chrome): add new CollapsibleSubNavItem component

austin94:
🚧 WIP, still trying to get animation working correctly and some API tweaks 🚧 

## Description

This PR adds a new `CollapsibleSubNavItem` component.

## Detail

screen shot 2018-10-15 at 3 03 29 pm

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
+++ b/packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
@@ -1,0 +1,75 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import classNames from 'classnames';
+import ChromeStyles from '@zendeskgarden/css-chrome';
+import { retrieveTheme } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.collapsable_sub_nav_item';
+const PANEL_COMPONENT_ID = 'chrome.collapsable_sub_nav_item_panel';
+
+import SubNavItem from './SubNavItem';
+
+/** Accepts all `<div>` props */
+const StyledSubNavItemHeader = styled(SubNavItem).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  'aria-expanded': props => props.expanded,
+  className: props =>
+    classNames(ChromeStyles['c-chrome__subnav__item--header'], {
+      [ChromeStyles['is-expanded']]: props.expanded
+    })
+})`
+  ${props => retrieveTheme(COMPONENT_ID, props)};
+`;
+
+StyledSubNavItemHeader.propTypes = {
+  expanded: PropTypes.bool
+};
+
+/** Accepts all `<div>` props */
+const StyledSubNavPanel = styled.div.attrs({
+  'data-garden-id': PANEL_COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  className: props =>
+    classNames(ChromeStyles['c-chrome__subnav__panel'], {
+      [ChromeStyles['is-hidden']]: props.hidden
+    })
+})`
+  ${props => retrieveTheme(PANEL_COMPONENT_ID, props)};
+`;
+
+StyledSubNavPanel.propTypes = {
+  hidden: PropTypes.bool
+};
+
+/**
+ * Accepts all `<button>` props
+ */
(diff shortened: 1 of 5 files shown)


packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
```diff
@@ -0,0 +1,75 @@
...
+  expanded: PropTypes.bool
+};
+
+/** Accepts all `<div>` props */
+const StyledSubNavPanel = styled.div.attrs({
+  'data-garden-id': PANEL_COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  className: props =>
+    classNames(ChromeStyles['c-chrome__subnav__panel'], {
+      [ChromeStyles['is-hidden']]: props.hidden
```
jzempel: Does this need `aria-hidden`? What's your thinking on extracting some of this logic into an [Accordion](https://www.w3.org/TR/wai-aria-practices/#accordion) container?

austin94: > Does this need aria-hidden

Since it has `display: none` it _shouldn't_ need it, but I've added it just in case 😄 

> What's your thinking on extracting some of this logic into an Accordion container?

Lets discuss some more offline. Normally an Accordion widget is in relation between several related panels & collapsable areas. Not really sure how this will all piece together in SubNav

jzempel: The WAI-ARIA pattern linked above applies to a single combination of an accordion `header` and its `panel` – exactly the pattern we have here (obviously, a Chrome subnav may have multiples of these). At a minimum, we should consider adding the expected aria `role` and `aria-controls` attributes.

packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
```diff
@@ -14,6 +14,7 @@ import { retrieveTheme } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'chrome.collapsable_sub_nav_item';
 const PANEL_COMPONENT_ID = 'chrome.collapsable_sub_nav_item_panel';
+const SUB_NAV_ITEM_HEIGHT_PX = 38;
```
jzempel: Item height is variable – can wrap to multiple lines. I was hoping this measurement could be obtained and applied post-render.

packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
```diff
@@ -14,6 +14,7 @@ import { retrieveTheme } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'chrome.collapsable_sub_nav_item';
 const PANEL_COMPONENT_ID = 'chrome.collapsable_sub_nav_item_panel';
```
jzempel: sp. _collapsible_

packages/chrome/src/views/subnav/CollapsibleSubNavItem.js
austin94: If we are animating the `max-height` over time, how would we calculate that total height without having a render flash? Lets discuss offline.

jzempel: On post-render, iterate over the children and get `element.offsetHeight + top-margin + bottom-margin` for each. Apply the value to the panel as `style="max-height: [value]"`.

austin94: I just pushed an update that adds a new `AccordionContainer` component. This abstracts all of the accessibility logic required for the [W3C accordion pattern](https://www.w3.org/TR/wai-aria-practices/#accordion)

I also calculated the panel `maxHeight` using a slightly different approach. Lets discuss offline.

packages/chrome/src/containers/AccordionContainer.js
```diff
@@ -0,0 +1,101 @@
...
+    };
+  }
+
+  getHeaderId = () => `${this.getControlledState().id}-header`;
+
+  getPanelId = () => `${this.getControlledState().id}-panel`;
+
+  getHeadingProps = ({ role = 'heading', ...other } = {}) => {
+    return {
+      role,
```
jzempel: @austin94 per https://www.w3.org/TR/wai-aria-practices/#accordion, the `role="heading"` also needs a value for `aria-level`. Can this be addressed so that the consumer can feed in that value (and in the demo so that it passes the axe audit)?

austin94: Voiceover was smart enough to auto-detect the level, didn't realize that AXE was complaining.

This is now a required attribute and prop for the `AccordionContainer` and the `CollapsibleSubNavItem`.

austin94:
> Yessss! This is looking good. Please just remove the @zendeskgarden/svg-icons duplicate from "dependencies" and we should be good to go!

Looks like I erroneously added it as a dependency when we first created the package. Now removed.

PR #201, opened 2018-11-07
feat(loaders): Add duration prop to dots loader

### LINKED ISSUE #147: Rethink loading icon speed API
## Expectations

## Reality

## Steps to Reproduce

Try it

## Suggestions

Either:

###### Fine Print

* Component: loaders/dots
* Browsers: all

ryanseddon:
* [x] **BREAKING CHANGE?** 

Fixes #147 

## Description

Previously the dots loader had a velocity prop that didn't really work as expected when changing the value. To align with the spinner loader this PR adds a `duration` prop which accepts a number in ms, default is 1250ms, to either slow down or speed up the animation.

## Detail

This is a breaking change as it removes the `velocity` prop in favour of the `duration` prop.

It also heavily refactors the code of the dots loader to follow the same approach used in the spinner loader.

Unfortunately due to our IE11 support we have to do these animations via js which are fairly CPU intensive compared to css animations. 

Pre published: https://garden.zendesk.com/react-components/loaders/#!/Dots/3

If anyone has any optimisations I can do I will be forever grateful.

## Checklist

      designer as a reviewer)
  component


ryanseddon:
Ok I've made the dots loader considerably worse, i'm gonna close this and move along.

PR #208, opened 2018-11-22
fix(loaders): competing calls to requestAnimationFrame

sunesimonsen:
## Description

We currently have situations where we get multiple requestAnimationFrame callbacks queued up and componentWillUnmount will only cancel the latest one. This results in settings the state on an unmounted component which React warns about.

This fix uses the more conventional way of using requestAnimationFrame, where you start a new animation frame at the end of the last one.

## Detail

screen shot 2018-11-22 at 13 26 49

screen shot 2018-11-22 at 13 27 21


sunesimonsen:
Thanks 👌

PR #215, opened 2018-11-28
feat: add new `breadcrumbs` component

jzempel:
* [ ] **BREAKING CHANGE?** 

## Description

Adds a new `@zendeskgarden/react-breadcrumbs` component package.

## Detail

Demo pre-published for review at https://garden.zendesk.com/react-components/breadcrumbs/

/cc @austin94 😉 

## Checklist

      designer as a reviewer)
  component


jzempel:
Y'all reviewers. The "next" svg icons look low to me. I wonder if we need to vertical center `text-bottom` (w pixel offset) vs `middle`. Could be due to demo showing without any prominent descenders.

ginnywood:
@jzempel good point, they do look low; I don't think it's about descenders.

.travis.yml
```diff
@@ -13,6 +13,7 @@ install: yarn --frozen-lockfile
 
 script:
   - yarn lint
+  - yarn format
```
austin94: @jzempel What prompted this change?  

I would be concerned if Travis was testing and deploying code that was transformed by a `format` without it being visible in Github.

jzempel: We seem to continue to let non-formatted code slip through (misconfigured husky?). The travis build will fail if the git tree isn't clean after test.

austin94: Oh that's perfect 💯

jzempel:
@allisonacs @ginnywood the demo page has been updated with the latest CSS code: https://garden.zendesk.com/react-components/breadcrumbs/

packages/breadcrumbs/src/views/BreadcrumbView.js
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveTheme } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'breadcrumbs.breadcrumb_view';
```
ryanseddon: 🙌 the theming table works and handled this just fine.

packages/breadcrumbs/src/containers/BreadcrumbContainer.js
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Identical to children
+     */
+    render: PropTypes.func
+  };
+
+  getContainerProps = ({ role = 'navigation', ...other } = {}) => {
+    return {
+      role,
+      'aria-label': 'Breadcrumb navigation',
```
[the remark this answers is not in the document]
austin94: > Should we default this like role in the argument destructuring to make it more obvious to teams that they'll need to translate this?

Not sure how clean this would be since we would be destructuring a hyphenated value?

> Should we mention in the readme example that we hard code an english value?

Definitely.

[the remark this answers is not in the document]
jzempel: @ryanseddon I added copy to the element example about expected override. But I don't think I can move `'aria-label'` into the destructured parameter list as it's an invalid property name.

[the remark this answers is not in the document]
ryanseddon: oh yeah good point

packages/breadcrumbs/src/elements/Breadcrumb.js
```diff
@@ -15,31 +16,31 @@ import Item from '../views/Item';
...
+    return Children.map(items, (item, index) => {
+      let retVal;
       const itemProps = {
         current: index === total - 1,
         key: index
       };
 
-      if (item.type === Item) {
-        retVal.push(cloneElement(item, itemProps));
+      if (hasType(item, Item)) {
```
ryanseddon: You could drop the `retVal` and just do a ternary:

```js
return hasType(item, Item) ? (
  cloneElement(item, itemProps)
) : (
  <Item {...itemProps}>{item}</Item>
);
```

jzempel: Of course. Good call.

jzempel:
@ryanseddon I think everything is ready for final review here – thanks!

Issue #211, opened 2018-11-26
Invariant violation error when trying to use `<MediaInput />`


vaevictis:
## Expectations

When importing `MediaInput` in a React Component, following [this trivial example](https://garden.zendesk.com/react-components/textfields/#mediainput) I expected to get a nice looking fake input with little icons on both sides. 

## Reality
Instead, I got a lousy `Invariant Violation` error:
> An error has occurred. Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `TestComponent`.

## Steps to Reproduce

  1.
  1.
  1.

###### Fine Print

* Component: `{MediaInput _et al_}` from `@zendeskgarden/react-textfields`
* Browsers: Chrome

PR #225, opened 2018-12-05
chore: add export testing to all packages

austin94:
## Description

This PR adds root export testing to all packages. This is to help reduce incidents where we forget to add a new component/utility to the root export. Similar to #211 

## Detail

These tests use a `glob` utility to help with finding the components.  Since each package is unique it was difficult to add with the standard `readdir` node methods.

A few of the packages (loaders/tooltips) needed some custom logic to match the expected exports, but overall the method is consistent.

### DIFF
--- a/packages/autocomplete/src/index.spec.js
+++ b/packages/autocomplete/src/index.spec.js
@@ -1,0 +1,27 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import glob from 'glob';
+import * as rootIndex from './';
+
+describe('Index', () => {
+  it('exports all components and utilities', done => {
+    glob('**/!(index|*.spec).js', { cwd: __dirname }, (er, files) => {
+      const mappedFiles = files
+        .map(entry =>
+          entry
+            .replace(/\.js$/u, '')
+            .split('/')
+            .pop()
+        )
+        .sort();
+
+      expect(Object.keys(rootIndex).sort()).toEqual(mappedFiles);
+      done();
+    });
+  });
+});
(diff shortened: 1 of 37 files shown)


packages/autocomplete/src/index.spec.js
```diff
@@ -0,0 +1,27 @@
...
+          entry
+            .replace(/\.js$/u, '')
+            .split('/')
+            .pop()
+        )
+        .sort();
+
+      expect(Object.keys(rootIndex).sort()).toEqual(mappedFiles);
+      done();
+    });
```
jzempel: Can't this basic `glob` structure be abstracted out to `react-testing` or `react-utilties`?

austin94: I've abstracted this into `react-testing` and added the appropriate customizations to the few packages that needed it.

austin94:
> Along with the common code abstraction, consider whether it makes sense to remove the associated PR checklist item from

I've removed this checklist item and also added a new `index.spec.js` to the `.template` component.

austin94:
> Along with the common code abstraction, consider whether it makes sense to remove the associated PR checklist item from

I've removed this checklist item and also added a new `index.spec.js` to the `.template` component.

packages/testing/src/utils/matchExports.js
```diff
@@ -0,0 +1,49 @@
...
+} = {}) {
+  return new Promise((resolve, reject) => {
+    glob(globPath, { cwd }, (error, files) => {
+      if (error) {
+        reject(error);
+      }
+
+      const mappedFiles = fileMapper(files);
+
+      expect(keys).toEqual(mappedFiles);
```
ryanseddon: should we really have the `expect` hidden away in another module? Can we have this method just return mapped files and then in each test we do the `expect` there.

austin94: Addressed in #228

PR #227, opened 2018-12-06
chore(build): Avoid running deploy stage 3 times

ryanseddon:
## Description

Seems my "speed up pr" made the deploy stage run 3 times in each `TEST_GROUP` job when merging to master! This change hopefully adds a separate deploy stage that will wait for the other jobs to finish and then run.

### TL;DR;

`--ignore-scripts` avoids building every package and speeds up the build greatly. ~3min builds now.

### More Info

The main speed up change is passing `--ignore-scripts` to the yarn install to avoid the `postinstall` step from triggering AFAIK this is only needed for dev. 

Before running `yarn test:all` I need to call `lerna bootstrap` to run but we also pass `--ignore-scripts` to lerna as it'll bypass the long `yarn build` being called for each package.

Instead of splitting this into multiple jobs and the overhead of spinning up 3 vms I've moved my learnings back into the original file and only changed it slightly.


ryanseddon:
@austin94 yes the shell script called in the deploy stage calls `yarn build:demo` which does a `yarn build && lerna run build:demo --stream"

PR #228, opened 2018-12-06
feat: refactor matchExports to getExports

austin94:
* [x] **BREAKING CHANGE**

## Description

My original implementation of `matchExports` included a nested `expect` 😢and wasn't the most readable.

This PR adds a `BREAKING CHANGE` to `react-testing` by removing the `matchExports` utility with a more generalized `getExports`. 

I've also updated all necessary tests to use asyn/await as necessary.

## Checklist

      designer as a reviewer)
  component


utils/test/jest.setup.js
```diff
@@ -15,3 +15,6 @@ import 'jest-styled-components';
 import Adapter from 'enzyme-adapter-react-16';
 import Enzyme from 'enzyme';
 Enzyme.configure({ adapter: new Adapter() });
+
+// Enable async/await
+import '@babel/polyfill';
```
ryanseddon: Do we need this if we're running in a node version that supports it natively?

austin94: From what I could tell, yes 😭 The jest environment is slightly different

austin94: Would continually run into `regeneratorRuntime` errors

PR #233, opened 2018-12-11
chore(build): Add concurrency flag to lerna runs to speed up slow builds

ryanseddon:
## Description

#227 sped up PR build times by ~50% this should speed up merge builds by ~50% but we'll need to see the timings from travis which we can't test, locally this runs fine but actual speed up times won't be known until this is merged.

## Detail

lerna defaults to 4 threads when running commands so i've upped the `concurrency` flag to 8, locally this speeds up the styleguidist build by ~50% 800s vs 360s on my local machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ script:
   - yarn format
   - yarn lerna bootstrap --ignore-scripts
   - yarn test:all --coverage --runInBand
+  - echo "Testing concurrency changes will revert this"
+  - time yarn build:demo
 
 after_success: yarn coveralls < demo/coverage/lcov.info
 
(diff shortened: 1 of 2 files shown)


ryanseddon:
Running a build with this change I got total build time 18 min 41 sec with `yarn build:demo` being time at about 15 mins.

https://travis-ci.org/zendeskgarden/react-components/builds/466338136#L1744

ryanseddon:
Did another one with `concurrency` set to 2 and it was about the same so this PR is not doing much. I think our best bet in the future will be to cache the demo folder and do conditional builds based on whats changed only.

austin94:
@ryanseddon We used to do conditional builds, but as interdependencies between packages in the docs examples kept growing the benefits were reduced.

PR #234, opened 2018-12-11
feat: introduce new Autocomplete component

### LINKED ISSUE #51: There are no high-abstraction implementations for the autocomplete package
## Expectations

Many packages include high-abstraction "element" implementations for the most common usages. 

I would expect that the autocomplete package at minimum have a high-abstraction implementation for the `Autocomplete` example.

There may also be a need for a `MutliSelect` abstraction in the future, but it is used less often.

These components would probably have a similar API to the [Select and SelectField](https://garden.zendesk.com/react-components/select/#selectfield) components.

## Reality

There is currently no high-abstraction elements in the `@zendeskgarden/react-autocomplete` package.

austin94:
Closes #51 

## Description

This PR introduces a simplified, high-abstraction element for common `Autocomplete` patterns.  The basic API is:

```jsx
<Autocomplete
  label="Basic Autocomplete"
  placeholder="Filter values"
  selectedValue={state.selectedValue}
  onChange={selectedValue => setState({ selectedValue })}
  options={[
    {
      value: 'option-1',
      label: 'Option 1'
    },
    {
      value: 'option-2',
      label: 'Option 2'
    },
    {
      value: 'option-3',
      label: 'Option 3'
    }
  ]}
/>;
```

This component also abstracts out much of the logic around filtering and rendering, but allows deeper customization through the `renderOption` and `renderDropdown` render-props as well as the ability to override the default text comparison with a custom comparator.

## Checklist

      designer as a reviewer)
  component


utils/test/jest.config.js
```diff
@@ -26,8 +26,9 @@ module.exports = {
     '\\.(svg)$': '<rootDir>/utils/test/svg-mock.js'
   },
   collectCoverageFrom: [
-    '<rootDir>/packages/*!(.template)/src/**/*.{js,jsx}',
+    '<rootDir>/packages/*/src/**/*.{js,jsx}',
```
austin94: My addition of the `index.spec.js` tests broke our code coverage 😢 This re-enables that feature

ryanseddon:
I think we should copy the same way we've done radios/checkboxes, rather than pass a prop for label, hint & message we have the children inside the `<Autocomplete />` component.

e.g.

```jsx
<Autocomplete
  placeholder="Filter values"
  selectedValue={state.selectedValue}
  onChange={selectedValue => setState({ selectedValue })}
  options={[...]}
>
  <Label>Basic Autocomplete</Label>
  <Hint>Autocomplete hint</Hint>
  <Message validation="error">Error</Message>
</Autocomplete>
```

This way they can still pass `validation` to `<Message>` and it's up to the consumer to explicitly include them.

austin94:
> I think we should copy the same way we've done radios/checkboxes, rather than pass a prop for label, hint & message we have the children inside the `<Autocomplete />` component.

I originally tried this, but had trouble getting everything to play nice together since we are stitching the `FieldContainer` and `AutocompleteContainer`s together.

Also, with the example above, would the `Autocomplete` container have to conditionally render the "input" depending on the type of elements provided? Not sure creating separate `AutocompleteField` and `Autocomplete` components would help much since this is already such an opinionated component.

packages/autocomplete/src/elements/Autocomplete.js
```diff
@@ -0,0 +1,270 @@
...
+                  }
+                });
+
+                if (disabled) {
+                  triggerProps = { disabled: true, small, validation };
+                }
+
+                return (
+                  <FauxInput {...triggerProps}>
+                    {!isOpen && <span>{optionDictionary[selectedValue]}</span>}
```
jzempel: The behavior of this span isn't working well.

* should be seeing an i-beam style `text` cursor across the entire width and height of the select (except for the dropdown cursor strike area)
* a click on the autocomplete should allow immediate typing (as it does with `examples/autocomplete.md`)
* need to figure out the 1px bump occurring on `isOpen`

The action should feel a little more like the **ContentEditable &lt;div&gt; Dropdown** on https://garden.zendesk.com/css-components/forms/dropdown/

packages/autocomplete/src/elements/Autocomplete.js
```diff
@@ -0,0 +1,270 @@
...
+      dictionary[option.value] = option.label;
+
+      return dictionary;
+    }, {});
+
+    return (
+      <FieldContainer>
+        {({ getLabelProps, getInputProps: getFieldInputProps, getHintProps }) => (
+          <TextGroup>
+            {label && <Label {...getLabelProps()}>{label}</Label>}
```
jzempel: A `Label` click should be focusing the field (it does in the `example/autocomplete.md`).

packages/autocomplete/src/elements/Autocomplete.js
```diff
@@ -0,0 +1,270 @@
...
+  overflow-y: auto;
+  max-height: ${props => props.maxHeight || 'inherit'};
+`;
+
+const StyledInput = styled(Input)`
+  ${props =>
+    !props.isOpen &&
+    `
+    && {
+      opacity: 0;
```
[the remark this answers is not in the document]
austin94: You're unable to programmatically `.focus()` an element with `display: none` :(

packages/autocomplete/src/elements/Autocomplete.js
austin94: > should be seeing an i-beam style text cursor across the entire width and height of the select (except for the dropdown cursor strike area)

This is now styled correctly.

> a click on the autocomplete should allow immediate typing (as it does with examples/autocomplete.md)

This is working as expected for me. Can you help reproduce this?

packages/autocomplete/src/elements/Autocomplete.js
austin94: I've updated this as well as handling the `onMouseEnter` and `onMouseLeave` functionality.

ryanseddon:
> Also, with the example above, would the Autocomplete container have to conditionally render the "input" depending on the type of elements provided? Not sure creating separate AutocompleteField and Autocomplete components would help much since this is already such an opinionated component.

Fair enough maybe a good compromise would be to allow passing through a string or node to label, hint & message props.

packages/autocomplete/src/elements/Autocomplete.js
```diff
@@ -0,0 +1,305 @@
...
+  }
+`;
+
+const StyledFauxInput = styled(FauxInput)`
+  cursor: text;
+`;
+
+const StyledValueWrapper = styled.div`
+  display: inline-block;
+  vertical-align: middle;
```
jzempel: this component needs a `margin-left: -1px` (reversed for RTL) in order to prevent the visual shift that is occurring.

packages/autocomplete/src/elements/Autocomplete.example.md
```diff
@@ -0,0 +1,89 @@
...
+  return dictionary;
+}, {});
+
+initialState = {
+  selectedValue: options[0].value
+};
+
+<Autocomplete
+  label="Basic Autocomplete"
+  placeholder={optionsMap[state.selectedValue]}
```
jzempel: replacing the value with a placeholder on menu expansion feels weird to me. It's not consistent with a true `<select>` dropdown and it's not consistent across browsers (placeholder disappears in IE). The value should remain intact and be replaced on keyboard entry. The placeholder should be whatever is displayed if a value is cleared.

jzempel:
@allisonacs @austin94 we need to continue to think about the normative gesture here. Consider the tab action through native form elements. On `<input>` fields, we get automatic range selection of any existing value (allowing for efficient replacement) and on `<textarea>` the cursor is located at the end of the existing value.

In most custom controls where the cursor is placed at the front of the value, we have the option of a quick `cmd-A` in order to reset the value.

Since neither range selection or `cmd-A` are available to us in this context, I wonder about the value of placing the cursor at the end and allowing `backspace` keycode to remove the value and expand the menu?

I'm not entirely sure, but the action feels off atm. I wonder if a more complete demo (mocking up additional form fields) would be illuminating?

allisonacs:
> Since neither range selection or `cmd-A` are available to us in this context, I wonder about the value of placing the cursor at the end and allowing `backspace` keycode to remove the value and expand the menu? I'm not entirely sure, but the action feels off atm. It's not intuitive when I see the front-positioned blinking cursor that I can't interact with the value via keyboard. I wonder if a more complete demo (mocking up additional form fields) would be illuminating?

I'd be happy to look at a more fully formed demo. From the way the constraints were explained to me, what Austin and I worked out felt like a reasonable compromise between expected behavior and technical feasibility, but I agree it's still a little awk. 

Placing the cursor at the end and only initiating a replacement/menu via a backspace is… weird also. What if I attempt to type more in addition to the current string? It would just disappear, leaving only the newly typed thing—but there'd be no indication that that's what would happen. If we prevent you from typing when there's a selected item, that's also odd—it would feel broken.

But yea, I'm happy to feel this out with a more comprehensive demo.

jzempel:
yeah @allisonacs that backspace idea is clearly not viable. I'm wondering about our ability to detect a `String` vs an `Object` here? If it's possible, we could range select on a `String` (just like a native `<input>` element) and we could focus (no cursor?) on an `Object` (just like a native `<select>` dropdown).

I'm also wondering about using `Object` autocomplete as a justification for the design. Shouldn't an object-style dropdown (i.e. emoji picker) just be handled via our `<Select>` component? Why is visible text entry interesting for that case? Plz enlighten me :wink:

allisonacs:
> yeah @allisonacs that backspace idea is clearly not viable. I'm wondering about our ability to detect a `String` vs an `Object` here? If it's possible, we could range select on a `String` (just like a native `<input>` element) and we could focus (no cursor?) on an `Object` (just like a native `<select>` dropdown).

I … think this would work, just need to see it? 

> I'm also wondering about using `Object` autocomplete as a justification for the design. Shouldn't an object-style dropdown (i.e. emoji picker) just be handled via our `<Select>` component? Why is visible text entry interesting for that case? Plz enlighten me 😉

Esp. in the case of an emoji picker, it's nice to be able to filter by typing, for example, "smile" to narrow down the emojis to all the emojis whose descriptions include "smile." Same kind of thing you see if you use the native apple emoji picker (`cmd + ctrl + space`). Another example would be showing avatars instead of people's names (though I think in that case people should be using the multi-select in order to show a tag).

austin94:
@jzempel @allisonacs Lets all have an offline discussion with some of this interaction today. Feels like we're going down a path that might require a separate PR rather than trying to wrap everything into this.

austin94:
> Oops, please update the WARNING text on the demo page, which currently indicates there is no Autocomplete visual component.

Updated!

ryanseddon:
image

Adding a message prop the spacing looks off, also if I set the validation of the autocomplete that should filter through to the message too correct?

packages/autocomplete/src/elements/Autocomplete.js
```diff
@@ -0,0 +1,312 @@
...
+                const { children: menuChildren, ...otherMenuProps } = props;
+
+                return (
+                  <MenuView {...otherMenuProps}>
+                    <StyledMenuOverflow maxHeight={maxHeight}>{menuChildren}</StyledMenuOverflow>
+                  </MenuView>
+                );
+              }}
+            </AutocompleteContainer>
+            {message}
```
ryanseddon: So we should be wrapping this in `<Message>` like we do with `<Hint>` and `<Label>` correct? We should also forward the validation prop here too

austin94: Correct. Missed this in the first run-through. Now fixed.

austin94:
> Adding a message prop the spacing looks off, also if I set the validation of the autocomplete that should filter through to the message too correct?

@ryanseddon This is now corrected. Thanks for catching that.

PR #235, opened 2018-12-12
BREAKING CHANGE: upgrade styled-components to v4 and React v16

austin94:
🚧 This is going to be a WIP for a while 🚧 

## Description

This PR upgrades styled-components to it's most recent version, which also necessitates us moving to a minimum version of `React v16.3`.  This PR has several moving parts which I will breakdown below.

### Testing

The testing story for `styled-components v4` and `enzyme` is almost non-existent at this point. `jest-styled-components@next` is making some progress here, but we may be blocked for the foreseeable future until this is fixed. 

### New Peer Dependencies

This PR updates the following dependencies:

* `@zendeskgarden/react-theming` - lerna doesn't updated peerDependencies for us
* `react`
* `react-dom`
* `styled-components`
* `react-fragment` - removed in favor of React.fragment
* `react-portal` - removed in favor of React.portal

### Breaking API Changes

With the move to `React.forwardRef` in `React v16.3` all components now receive a standard `ref` or use `React.forwardRef` to accomplish the same behavior. This removes all `innerRef` and `buttonRef` usages throughout the codebase.

### Styled-Components Deprecation warnings and breakages (most of the changes)

In v4 the `.attrs()` method has changed dramatically https://github.com/styled-components/styled-components/issues/2270 This requires us to re-write all usages (literally every view component).

In the long-run this will be much more performant for us, but we have to recognize that when extending one of our internal components with `styled(FooBar).attrs()` we must manually pass through the original className.


packages/buttons/src/elements/ButtonGroup.js
```diff
@@ -93,7 +93,7 @@ export default class ButtonGroup extends ControlledComponent {
   };
 
   render() {
-    const { children, ...otherProps } = this.props; // eslint-disable-line no-unused-vars
+    const { children, onStateChange, ...otherProps } = this.props; // eslint-disable-line no-unused-vars
```
ryanseddon: What's this prop restructured for I can't see it used below?

packages/menus/src/containers/MenuContainer.js
```diff
@@ -509,6 +508,7 @@ class MenuContainer extends ControlledComponent {
       id,
       defaultFocusedIndex
     } = this.getControlledState();
+    const defaultNode = getDocument(this.props).body;
```
ryanseddon: Could we just assign `getDocument(this.props).body` as the default value for `appendToNode`?

packages/notifications/src/content/Paragraph.js
```diff
@@ -18,7 +17,7 @@ const COMPONENT_ID = 'notifications.paragraph';
 const Paragraph = styled.div.attrs({
```
ryanseddon: Should this be a function call `styled.div.attrs props => {`?

packages/notifications/src/content/Title.js
```diff
@@ -18,7 +17,7 @@ const COMPONENT_ID = 'notifications.title';
 const Title = styled.div.attrs({
```
ryanseddon: Same here function call?

packages/tables/src/examples/draggable-table.md
```diff
@@ -97,10 +97,7 @@ class DraggableExample extends React.Component {
           <Droppable droppableId="droppable">
             {(provided, droppableSnapshot) => {
               return (
-                <Body
-                  innerRef={provided.innerRef}
-                  isDraggingOver={droppableSnapshot.isDraggingOver}
-                >
+                <Body ref={provided.innerRef} isDraggingOver={droppableSnapshot.isDraggingOver}>
```
ryanseddon: Is `<Droppable />` still calling the prop `innerRef`?

packages/buttons/src/elements/ButtonGroup.js
austin94: styled-components has a new props white-list that they allow. Unless we destructure the `onStateChange` value out there are warnings.

packages/menus/src/containers/MenuContainer.js
austin94: Didn't think that was possible since `defaultProps` are static?

packages/notifications/src/content/Paragraph.js
austin94: That's only necessary if we are dynamically changing the classes based on prop changes. This always has a static className so it's not necessary.

packages/tables/src/examples/draggable-table.md
austin94: `innerRef` is the value provided by the render-prop in `<Droppable />` so nothing on our end. It's applied to the correct `ref` so we should be good.

packages/menus/src/containers/MenuContainer.js
ryanseddon: Ah I didn't mean defaultProps but a default value to the destructure value like we do with the `render` prop e.g.

```js
 render() {
    const {
      children,
      render = children,
      trigger,
      eventsEnabled,
      popperModifiers,
      appendToNode = getDocument(this.props).body,
      zIndex
    } = this.props;
    //...

    return createPortal(menu, appendToNode);
  }
```

packages/buttons/src/elements/ButtonGroup.js
ryanseddon: Ah good to know, thanks.

vinceprofeta:
@austin94 is there an update for this?

austin94:
Closing in favor of #342 which is the non-exploratory PR for this migration. Thanks for the comments!

agrohs:
Would be great to have some sample documentation of all of the changes needed to move from the old Garden components that have been deprecated to the new ones!

austin94:
@agrhos I've created https://github.com/zendeskgarden/react-components/issues/372 to track this. Thanks for the suggestion!

PR #243, opened 2019-01-04
chore(styleguide): Use auto completion for selecting the package to run

sunesimonsen:
## Description

This PR replaces the current prompt for starting the styleguide with one that auto-completes the package name.

## Detail

autocomplete


sunesimonsen:
@austin94 hehe I did it 5 times and got annoyed and fixed it. Then I found that the real thing I wanted to fix already worked 😂

PR #249, opened 2019-01-10
 feat(autocomplete): introduce Multiselect component

austin94:
## Description

This PR introduces a new high-abstraction element, `Multiselect`.  This implementation is very similar to `Autocomplete`, but with the original tweaks present in the `AutocompleteContainer` example.

## Checklist

      designer as a reviewer)
  component


allisonacs:
do you have this somewhere i can play with it?

packages/autocomplete/src/elements/Multiselect.js
```diff
@@ -0,0 +1,449 @@
...
+const StyledMenuOverflow = styled.div`
+  overflow-y: auto;
+  max-height: ${props => props.maxHeight || 'inherit'};
+`;
+
+const StyledInput = styled(Input)`
+  && {
+    flex-grow: 1;
+    margin: 2px;
+    width: 60px;
```
ryanseddon: Should this be `flex-basis: 60px` since I assume this is meant to be a minimum width kind of thing?

austin94: Correct. I also had some specificity weirdness that I've corrected as well.

austin94:
> do you have this somewhere i can play with it?

@allisonacs I've pre-published to https://garden.zendesk.com/react-components/autocomplete/#multiselect it should be the same interaction as the original example built with the render-prop container.

allisonacs:
IMO, there's an odd interaction in the keyboard nav. If i use the arrow keys to move back through the list (indicated by focus), and then delete a tag, the cursor returns to the end of the input and starts flashing again. Instead, I'd expect it to focus on the tag prior to the left of the item just deleted. 

https://cl.ly/0ce3045bab23

jzempel:
Agreed with @allisonacs (or focus right, depending on RTL).

jzempel:
Please add support here for https://github.com/zendeskgarden/react-components/pull/253

austin94:
> Please add support here for #253

@jzempel with this being a high-abstraction component would this be necessary? Users aren't able to combine this with a ButtonGroup or it's equivalent.

jzempel:
Yes, we want the API to be consistent for all fields. Also, this gets us away from pre-imagining all valid use cases.

allisonacs:
@austin94 I've been chatting with a few folks and came across a use case where the component needs to be able to show a selection/tag which is in error. I can't remember if we had that in the old components—I feel like we must have—but if we don't, let's make sure we get that functionality in here as well (it will be necessary alongside the "add" capability mentioned by JZ above).

austin94:
@jzempel @ryanseddon @allisonacs This is now at a place that matches the original functionality provided in the examples that used the `AutocompleteContainer`.  (I can run locally for you to re-evaluate)

From all of the conversations over the past few weeks, it's obvious that this style of API doesn't fit _every_ possible use case.  But, as an intermediary step, this PR should help those that are stuck with the most common implementations.

Some of this ultra-flexibility, like Tag validation, async indicators, and paginated scrolling, should be addressed in a different API/approach that is more flexible than the path we have gone down with this package.

packages/autocomplete/src/elements/Multiselect.js
```diff
@@ -0,0 +1,449 @@
...
+`;
+
+const StyledMenuOverflow = styled.div`
+  overflow-y: auto;
+  max-height: ${props => props.maxHeight || 'inherit'};
+`;
+
+const StyledInput = styled(Input)`
+  && {
+    flex-basis: 60px;
```
jzempel: What is this `flex-basis` about? How is the multiselect handling both `small` and default sizing?

packages/autocomplete/src/elements/Multiselect.js
```diff
@@ -0,0 +1,449 @@
...
+    maxHeight: PropTypes.string,
+    disabled: PropTypes.bool,
+    options: PropTypes.array.isRequired,
+    noOptionsMessage: PropTypes.string,
+    renderOption: PropTypes.func,
+    renderDropdown: PropTypes.func,
+    renderShowMore: PropTypes.func,
+    optionFilter: PropTypes.func,
+    closeOnSelect: PropTypes.bool
+  };
```
jzempel: Please add support for `small` and `focusInset`

packages/autocomplete/src/elements/Multiselect.js
austin94: Addressing the `small` sizing in the comment below

packages/autocomplete/src/elements/Multiselect.js
austin94: Forgot about these two 😓 Both are now supported

packages/autocomplete/src/elements/Multiselect.js
jzempel: Then use `min-width`. I don't think `flex-basis` is doing what you think it is. https://gedd.ski/post/the-difference-between-width-and-flex-basis/

austin94: There were conflicts with the default width of `100%` that led me down that road. Correcting `width` in combination with `min-width` helped resolve this.

PR #254, opened 2019-01-22
feat(ranges): introduce MultiThumbRange component

austin94:
## Description

This PR adds a new `MultiThumbRange` component. It follows the [Two-Thumb Slider W3C Pattern](https://www.w3.org/TR/wai-aria-practices/#slidertwothumb).

Due to the complexity of the drag functionality and it's tight coupling to how we structure our sliders visually, I've chosen to not break this component into separate Container and View components.

@jzempel @ryanseddon any thoughts on this? If more customization is needed in the future we can re-evaluate the split.

## Detail

screen shot 2019-01-22 at 11 24 18 am

Pre-published at https://garden.zendesk.com/react-components/ranges

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -26,7 +26,7 @@
   },
   "rules": {
     "prettier/prettier": "error",
-    "indent": ["error", 2],
+    "indent": ["error", 2, { "SwitchCase": 1 }],
     "sort-imports": "off",
     "valid-jsdoc": "off",
     "no-invalid-this": "off",
--- a/packages/ranges/src/elements/MultiThumbRange.js
+++ b/packages/ranges/src/elements/MultiThumbRange.js
@@ -272,15 +272,49 @@ class MultiThumbRange extends Component {
     }
   };
 
+  calculateMinPosition = minDistancePx => {
+    const { min, max, minValue, maxValue } = this.props;
+    const { railWidthPx } = this.state;
+
+    let boundedMinValue = minValue;
+
+    if (boundedMinValue < min) {
+      boundedMinValue = min;
+    } else if (boundedMinValue > maxValue) {
+      boundedMinValue = maxValue;
+    } else if (boundedMinValue > max) {
+      boundedMinValue = max;
+    }
+
+    return ((boundedMinValue - min) / (max - min)) * (railWidthPx - minDistancePx);
+  };
+
+  calculateMaxPosition = minDistancePx => {
+    const { min, max, maxValue, minValue } = this.props;
+    const { railWidthPx } = this.state;
+
+    let boundedMaxValue = maxValue;
+
+    if (boundedMaxValue > max) {
+      boundedMaxValue = max;
+    } else if (boundedMaxValue < minValue) {
+      boundedMaxValue = minValue;
+    } else if (boundedMaxValue < min) {
+      boundedMaxValue = min;
+    }
+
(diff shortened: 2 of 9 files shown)


.eslintrc.json
```diff
@@ -26,7 +26,7 @@
   },
   "rules": {
     "prettier/prettier": "error",
-    "indent": ["error", 2],
+    "indent": ["error", 2, { "SwitchCase": 1 }],
```
jzempel: This is the existing `eslint-config` rule: https://github.com/zendeskgarden/eslint-config/blob/master/rules/stylistic-issues.js#L58. Can just be deleted.

austin94: I'm not sure what's going on with our custom config, but it seems prettier is fighting the config on this one and can't seem to align on anything. 

Lets meet offline to walk through this.

ryanseddon:
Seeing some weird behaviour if I get silly with min/max values
multi-slider

packages/ranges/src/elements/MultiThumbRange.js
```diff
@@ -0,0 +1,362 @@
...
+
+const StyledThumb = styled.div.attrs({
+  'data-test-id': 'thumb',
+  className: props =>
+    classNames(RangeStyles['c-range__slider__track__rail__thumb'], {
+      [RangeStyles['is-focused']]: props.isFocused
+    })
+})``;
+
+class MultiThumbRange extends PureComponent {
```
ryanseddon: Technically we still support react 0.14.0 and `PureComponent` came in v15.3.0.

packages/ranges/src/elements/MultiThumbRange.js
```diff
@@ -0,0 +1,362 @@
...
+  addRangeResizeEvents = () => {
+    window.addEventListener('resize', this.onWindowResize);
+  };
+
+  removeRangeResizeEvents = () => {
+    window.removeEventListener('resize', this.onWindowResize);
+  };
+
+  addDragEvents = () => {
+    document.addEventListener('mousemove', this.onDocumentMouseMove);
```
ryanseddon: Should we use `getDocument` here?

packages/ranges/src/elements/MultiThumbRange.js
austin94: I've updated this to use the standard `Component`. I was already checking for extra renders to help with `onChange` management so no other changes were needed.

packages/ranges/src/elements/MultiThumbRange.js
austin94: Updated

.eslintrc.json
jzempel: If there are plugin rules unintentionally overriding our config, then I think the `extends` section needs to be reordered, like so:

```json
{
  "extends": [
    "prettier",
    "@zendeskgarden",
    "@zendeskgarden/eslint-config/plugins/notice.js",
    "plugin:react/recommended",
    "plugin:jest/recommended",
    "plugin:jsx-a11y/recommended"
  ]
}
```

based on https://eslint.org/docs/user-guide/configuring#extending-configuration-files

austin94:
> Seeing some weird behaviour if I get silly with min/max values

@ryanseddon I've added in some bounded values on in our render. I'm not able to get propType validators to consistently flag issues, but the rendered thumbs won't jump out of the page at least.

Since these values would have to be manually set and the `onChange` callback is already bounding the values I think this should be good.

PR #255, opened 2019-01-24
fix(ranges): add analytics attributes to MultiThumbRange

austin94:
## Description

The original release of `MultiThumbRange` was missing the analytics attributes we use within Pendo.

This PR adds them!

## Checklist

      designer as a reviewer)
  component

### DIFF
--- a/packages/ranges/src/elements/MultiThumbRange.js
+++ b/packages/ranges/src/elements/MultiThumbRange.js
@@ -16,11 +16,15 @@ import { KEY_CODES } from '@zendeskgarden/react-selection';
 import { withTheme, isRtl, getDocument } from '@zendeskgarden/react-theming';
 import RangeStyles from '@zendeskgarden/css-forms/dist/range.css';
 
+const COMPONENT_ID = 'ranges.multi_thumb_range';
+
 /**
  * These Styled components are not exported with the other Views due to their logic
  * being more tightly coupled with this specific implemenation.
  */
 const StyledSlider = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
   'data-test-id': 'slider',
   className: props =>
     classNames(RangeStyles['c-range__slider'], {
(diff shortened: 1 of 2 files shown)


packages/ranges/src/elements/MultiThumbRange.js
```diff
@@ -16,11 +16,15 @@ import { KEY_CODES } from '@zendeskgarden/react-selection';
...
 
+const COMPONENT_ID = 'ranges.multi_thumb_range';
+
 /**
  * These Styled components are not exported with the other Views due to their logic
  * being more tightly coupled with this specific implemenation.
  */
 const StyledSlider = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
```
ryanseddon: This should be added to all views right? Otherwise I won't be able to theme it like I can with the single range.

austin94: Originally thought we would wan't these theme-able, but you're right, it should be possible.

I've added unique keys to each styled component as well as the `retrieveTheme` injectable.

packages/ranges/src/elements/MultiThumbRange.js
```diff
@@ -13,14 +13,21 @@ import classNames from 'classnames';
...
 import debounce from 'lodash.debounce';
 import { KEY_CODES } from '@zendeskgarden/react-selection';
-import { withTheme, isRtl, getDocument } from '@zendeskgarden/react-theming';
+import { retrieveTheme, withTheme, isRtl, getDocument } from '@zendeskgarden/react-theming';
 import RangeStyles from '@zendeskgarden/css-forms/dist/range.css';
 
+const SLIDER_COMPONENT_ID = 'ranges.multi_thumb_range.slider';
+const TRACK_COMPONENT_ID = 'ranges.multi_thumb_range.track';
+const RAIL_COMPONENT_ID = 'ranges.multi_thumb_range.rail';
+const THUMB_COMPONENT_ID = 'ranges.multi_thumb_range.thumb';
```
ryanseddon: Now my second question is this will now break the auto theming table id docs as it won't find these, I'll pull down this PR and poke around my script to get it working

ryanseddon: Resolved in  5dbd11b

PR #261, opened 2019-01-28
feat(loaders): introduce Skeleton loader

austin94:
## Description

This PR introduces the new `Skeleton` loader component.  The component is built fully with styled-components and includes 2 main animations:

* The initial fade
* A recurring sweep for the linear-gradient

## Detail

You can view this component live at https://garden.zendesk.com/react-components/loaders/#!/Skeleton

For `polished` to work correctly in jest, I had to increase the specificity of our `@zendeskgarden/css-*` mock.  This change should help us have more accurate snapshot tests as we create more pure `styled-components` implementations.

## Checklist

      designer as a reviewer)
  component


packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,111 @@
...
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: inline-block;
+  position: relative;
+
+  animation: ${fadeInAnimation} 700ms linear; /* stylelint-disable-line */
+  opacity: ${props => (props.dark ? '0.15' : '0.1')};
+
+  border-radius: ${zdSpacingXxs};
+  background-color: ${props => (props.dark ? zdColorWhite : zdColorGrey800)};
```
ryanseddon: You could simplify this animation a bit and set an rgba.

```js
import { rgba } from 'polished';

const StyledSkeleton = styled.div`
  background-color: ${props => (props.dark ? rgba(zdColorWhite, 0.1) : rgba(zdColorGrey800, 0.15))};
`;
```

You could then drop setting opacity and just have the animation set what the 100% keyframe is:

```js
const fadeInAnimation = props => keyframes`
 100% {
    opacity: ${props.dark ? '0.15' : '0.1'};
  }
`;

File: packages/loaders/src/Skeleton.js [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+  width: ${props => props.width};
+  height: ${props => props.height};
+  overflow: hidden;
+  line-height: 1;
+
+  &::before {
+    position: absolute;
+    top: 0;
+    left: -1200px;
```
ryanseddon: Whats's the reasoning for a 1000px pseudo element and positioning it 200px further out? Timings?

Since you're setting this value you don't need to repeat it in the animation:

```js
const fadeInAnimation = props => keyframes`
 100% {
    left: 100%;
  }
`;
```

works the same.

ryanseddon: demo here https://jsbin.com/powecanuwe/2/edit

packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,111 @@
...
+  ${props => retrieveTheme(COMPONENT_ID, props)};
+`;
+
+/**
+ * Loader used to create Skeleton objects
+ */
+function Skeleton({ width, height, style, dark }) {
+  return (
+    <StyledSkeleton dark={dark} width={width} height={height} style={style}>
+      &nbsp;
```
ryanseddon: Does the `&nbsp` need to added here? Is this to correctly get line-height and font-size to work?

packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,111 @@
...
+  &::before {
+    position: absolute;
+    top: 0;
+    left: -1200px;
+    animation: ${skeletonAnimation} 1.5s ease-in-out 300ms infinite;
+    width: 1000px;
+    height: 100%;
+    content: '';
+    /* stylelint-disable */
+    background-image: linear-gradient(
```
ryanseddon: another strategy is we could set this on the div and animation the `background-position` to sweep it through?

packages/loaders/package.json
```diff
@@ -19,7 +19,9 @@
     "start": "../../utils/scripts/start.sh"
   },
   "dependencies": {
-    "classnames": "^2.2.5"
+    "@zendeskgarden/css-variables": "^5.2.4",
```
jzempel: Update to latest 6.1.0.

packages/loaders/src/Skeleton.example.md
```diff
@@ -0,0 +1,54 @@
...
+  background-color: ${props => (props.isDarkMode ? zdColorKale800 : zdColorWhite)};
+  color: ${props => (props.isDarkMode ? zdColorWhite : zdColorBlack)};
+  padding: 18px;
+`;
+
+<Grid>
+  <Row alignItems="start">
+    <StyledCol md={8} isDarkMode={state.isDarkMode}>
+      <XXXL>
+        {state.isLoading ? <Skeleton dark={state.isDarkMode} /> : 'There. You see Lord Vader.'}
```
jzempel: How are we thinking about this strategy when the line wraps? (Shrink your browser to see.) I can understand recommending multiple `<Skeleton>` elements, but the `font-size` vs. `line-height` isn't really taken into account.

packages/loaders/src/Skeleton.js
austin94: > Whats's the reasoning for a 1000px pseudo element and positioning it 200px further out? Timings?

It's for timing and the size of the transparent area.

I've updated the keyframe

packages/loaders/src/Skeleton.js
austin94: This helps ensure the line-height and font-size are accurate.

packages/loaders/src/Skeleton.js
austin94: I'll look into this.

packages/loaders/src/Skeleton.example.md
austin94: I think this is something that we will have to provide in the design documentation/recommendations for this.

austin94:
@ginnywood This has been updated to match the final designs. Please let me know if anything else sticks out to you.

packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,118 @@
...
+`;
+
+/* eslint-disable */
+const StyledSkeleton = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: inline-block;
+  position: relative;
+  animation: ${fadeInAnimation} 750ms linear; /* stylelint-disable-line */
```
jzempel: `stylelint-disable` shouldn't be needed here (I was able to remove without issue).

packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,118 @@
...
+  border-radius: ${zdSpacingXxs};
+  background-color: ${props => (props.dark ? rgba(zdColorWhite, 0.2) : rgba(zdColorGrey800, 0.1))};
+  width: ${props => props.width};
+  height: ${props => props.height};
+  overflow: hidden;
+  line-height: 0.6; /* stylelint-disable-line */
```
jzempel: `stylelint-disable` shouldn't be needed here (I was able to remove without issue).

packages/loaders/src/Skeleton.js
```diff
@@ -0,0 +1,118 @@
...
+  `
+        : `
+    left: -1800px;
+    animation: ${skeletonAnimation} 1.5s ease-in-out 300ms infinite;
+  `}
+
+    width: 1000px;
+    height: 100%;
+    content: '';
+    /* stylelint-disable */
```
jzempel: `stylelint-disable/enable` shouldn't be needed here (I was able to remove without issue).

packages/loaders/src/Skeleton.js
austin94: It seems `stylelint` updated their [globby](https://github.com/sindresorhus/globby) dependency. This requires all globs to be wrapped in `"`.

Updating our `lint:css` script triggered the warnings. I removed 2 rules that continually get flagged due to prettier and styled-components usage. This should be now be accurate.

packages/loaders/src/Skeleton.js
austin94: Resolved with comment above.

packages/loaders/src/Skeleton.js
austin94: Resolved with comment above.

austin94:
> Approved, but with the caveat that snapshot tests need to be fixed for CI. Thanks!

I've updated the snapshots. I'm going to try and figure out how to run the local tests in the `.lintstagedrc` file now that it's faster

ginnywood:
@austin94 per our conversation just now, let's change the dark mode gradient to match our chrome: `kale-700`. The alpha values can stay the same—0.0, 0.4, 0.0. Thanks!

austin94:
@ginnywood this is corrected in #274

PR #268, opened 2019-02-13
feat(buttons): introduce external navigation option to Anchor component

austin94:
## Description

After seeing some work from @jpandl19 I realized that we still hadn't added in an `external` option for our `Anchor` components to denote when a resource opens in a new window and/or navigates to a 3rd party link.

## Detail

This PR introduces an `external` prop to the `Anchor` component that displays the `new-window` SVG icon.

LTR:
screen shot 2019-02-13 at 12 42 12 pm

RTL:
screen shot 2019-02-13 at 12 42 25 pm

This is currently not based off any designs, just what exists in the products. @allisonacs @ginnywood I could use some guidance on sizing (currently 0.75em) and other visual treatments.

## Checklist

      designer as a reviewer)
  component


packages/buttons/src/views/Anchor.js
```diff
@@ -34,30 +36,51 @@ const StyledAnchor = styled.a.attrs({
...
+      {({ getFocusProps, keyboardFocused }) => (
+        <StyledAnchor
+          {...getFocusProps({
+            ...other,
+            focused: focused || keyboardFocused
+          })}
+        >
+          {rtl && external && <StyledNewWindowIcon />}
+          {children}
+          {!rtl && external && <StyledNewWindowIcon />}
```
ryanseddon: If the page has `direction` set then that should handle the correct position of the icon right?

packages/buttons/src/views/Anchor.js
```diff
@@ -34,30 +36,51 @@ const StyledAnchor = styled.a.attrs({
   ${props => retrieveTheme(COMPONENT_ID, props)};
 `;
 
+const StyledNewWindowIcon = styled(NewWindowIcon)`
```
ryanseddon: Do we need to add any aria attributes here to make sure this doesn't get screen reader confusion?

allisonacs:
@austin94 is this prepublished? I don't see it on our pages rn. Without that, I can't properly review the appearance of the icon when using .75em, nor see how word wrapping works, etc. etc. Basically I can't review it without playing with it! 

I assume you're sizing it with ems so that it will scale alongside the text it appears in-line with, right? Seems like there are two possible solutions: 
1. Do it the way you're doing it, sizing the icon relative to the type size
2. Define a static size for the icon. _Maybe_ be able to pass in a `small` or `large` prop depending on the type size of the link itself.

Although the graphic designer in me would like to control the appearance as much as possible, the pragmatist in me says we'll never be able to really scope it right all the time.

Once you get this pre-published I can look at whether or not .75em is the right measurement.

packages/buttons/src/views/Anchor.example.md
```diff
@@ -12,12 +12,12 @@ If you need a `<button>` that has anchor styling, use the
   <Anchor href="https://www.zendesk.com" target="_blank">
     atmosphere
   </Anchor>, for the sky was a dead black, and the stars had ceased to twinkle. By the same
-  <Anchor href="https://www.zendesk.com" target="_blank">
+  <Anchor href="https://www.zendesk.com" target="_blank" external>
     illusion which lifts the horizon
   </Anchor>
-  of the sea to the level of the spectator on a hillside, the sable cloud beneath was dished out, and
+  {' '} of the sea to the level of the spectator on a hillside, the sable cloud beneath was dished out, and
```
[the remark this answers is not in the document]
austin94: This is a side-effect of the prettier MD formatting. It's just to add a single space.

packages/buttons/src/views/Anchor.js
austin94: From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/direction)

> The direction CSS property sets the direction of text, table columns, and horizontal overflow

Testing locally it didn't affect the ordering of the SVG

packages/buttons/src/views/Anchor.js
austin94: > Do we need to add any aria attributes here to make sure this doesn't get screen reader confusion?

I couldn't find any consensus on this in the WAI-ARIA docs and other resources. Personally, I feel that since the screen-reader already says the url location and whether it will open in a new tab/window through the `target` attribute that it could be ok.

> Somehow that seems to be getting stripped in the (SVGO?) build process – and it means that this icon is focusable in IE (which we don't want)

I've updated our styleguidist and webpack builds to directly reference the `@zendeskgarden/svg-icons/.svgo.yml` file. Is that ok to do?

packages/buttons/src/views/Anchor.js
```diff
@@ -34,30 +36,51 @@ const StyledAnchor = styled.a.attrs({
...
+  width: 0.75em;
+  height: 0.75em;
+  vertical-align: middle;
+
+  ${props => (isRtl(props) ? `margin-right: ${zdSpacingXxs};` : `margin-left: ${zdSpacingXxs};`)}
```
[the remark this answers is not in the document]
austin94: > The use of margin here means that the `default` cursor is showing between the link and the icon in some browsers. It should be `pointer` all the way through.

Moving to padding corrected this.

> Also, the padding should be expressed as `em`s in order to space consistently at extreme sizes.

This is now calculated in `em`'s.

packages/buttons/src/views/Anchor.example.md
```diff
@@ -12,12 +12,12 @@ If you need a `<button>` that has anchor styling, use the
   <Anchor href="https://www.zendesk.com" target="_blank">
```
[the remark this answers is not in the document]
austin94: I'll update to local URLs in the example.

packages/buttons/src/views/Anchor.example.md
```diff
@@ -12,12 +12,12 @@ If you need a `<button>` that has anchor styling, use the
...
   </Anchor>, for the sky was a dead black, and the stars had ceased to twinkle. By the same
-  <Anchor href="https://www.zendesk.com" target="_blank">
+  <Anchor href="https://www.zendesk.com" target="_blank" external>
     illusion which lifts the horizon
   </Anchor>
-  of the sea to the level of the spectator on a hillside, the sable cloud beneath was dished out, and
+  {' '} of the sea to the level of the spectator on a hillside, the sable cloud beneath was dished out, and
   the car seemed to float in the middle of an
-  <Anchor href="https://www.zendesk.com" target="_blank">
+  <Anchor href="https://www.zendesk.com" target="_blank" external>
```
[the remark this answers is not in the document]
austin94: I've added some new sizing examples

austin94:
> is this prepublished? I don't see it on our pages rn

@allisonacs it got overridden with the most recent publish. We _really_ need some branch-specific builds 😄 

It should be deployed now at https://garden.zendesk.com/react-components/buttons/#anchor

austin94:
@jzempel it looks like the SVG icon being used is getting clipped once styleguidist is built. This isn't reflected when being ran locally.

Must be something with the `.svgo` configuration. @allisonacs I'm going to un-publish until this is resolved.

packages/buttons/src/views/Anchor.js
ryanseddon: Yeah fair enough not sure what lead me to believe that.

packages/buttons/src/views/Anchor.js
jzempel: I wouldn't recommend referencing that config directly. Use only what you need. For example, we do want `svg-icons` to be pretty printed for ease of consumer inspection. But we don't want that in our build.

jzempel: Your updates lost the `viewBox` – that's why the SVG isn't sizing properly.

packages/buttons/src/views/Anchor.js
```diff
@@ -34,30 +35,52 @@ const StyledAnchor = styled.a.attrs({
   ${props => retrieveTheme(COMPONENT_ID, props)};
 `;
 
+const StyledNewWindowIcon = styled(NewWindowIcon)`
+  margin-bottom: -0.075em;
+  width: 0.75em;
+  height: 0.75em;
+  vertical-align: baseline;
```
jzempel: Remove as `vertical-align: baseline;` is the default.

utils/build/webpack.base.js
```diff
@@ -74,11 +74,7 @@ found at http://www.apache.org/licenses/LICENSE-2.0
           {
             loader: '@svgr/webpack',
             options: {
-              svgoConfig: {
-                plugins: {
-                  removeViewBox: false
-                }
-              }
+              externalConfig: path.resolve('../../node_modules/@zendeskgarden/svg-icons/.svgo.yml')
```
jzempel: replace with...

```json
                svgoConfig: {
                  plugins: [
                    { removeViewBox: false },
                    {
                      addAttributesToSVGElement: {
                        attributes: [
                          {
                            focusable: false
                          }
                        ]
                      }
                    }
                  ]
                }
```

The rest of the config has already been handled by `svg-icons` and shouldn't be duplicated here. Also, the SVGR config should be externalized and shared between webpack and styleguidist.

austin94: I've added a local `.svgo.yml` file with this configuration

.svgo.yml
```diff
@@ -0,0 +1,5 @@
+plugins:
+  - removeViewBox: false
+  - addAttributesToSVGElement:
+    - attributes:
```
jzempel: Incorrect yaml here. Remove the dash to get rid of the console errors.

austin94: Weird, http://www.yamllint.com/ listed that as a "correct" format. Dashes are now removed and the console errors are no longer visible

.svgo.yml
```diff
@@ -0,0 +1,5 @@
+plugins:
+  removeViewBox: false
+  addAttributesToSVGElement:
+    attributes:
+      focusable: false
```
jzempel: @austin94 please check your commits – this yaml doesn't do anything. Ensure that you see `viewBox="0 0 12 12"` and `focusable="false"` are on the SVG that's being produced in source. The following syntax (pulled from `svg-icons`) does exactly that:

```yaml
plugins:
  - addAttributesToSVGElement:
      attributes:
        - focusable: false
  - removeViewBox: false
```

austin94: This is now corrected. Not sure why it was still referencing the older config.

austin94:
@allisonacs @ginnywood lets get together in person to double-check the sizing on this icon.

packages/buttons/src/views/Anchor.js
```diff
@@ -34,30 +35,56 @@ const StyledAnchor = styled.a.attrs({
...
+const StyledNewWindowIcon = styled(NewWindowIcon)`
+  margin-bottom: -0.085em;
+  padding-left: 0.25em;
+  width: 0.85em;
+  height: 0.85em;
+
+  ${props =>
+    isRtl(props) &&
+    `
+    transform: scaleX(-1);
```
jzempel: This approach to RTL is not reliable. Consider applying `dir: rtl` to the SVG or use `background` for left/right positioning.

austin94: The background approach didn't work due to our use of `currentColor` in the SVGs.  I've added a new approach.

packages/buttons/src/views/Anchor.example.md
```diff
@@ -9,23 +9,56 @@ If you need a `<button>` that has anchor styling, use the
...
 </div>
 ```
 
+### External Anchor Sizing
+
+```jsx
+const { SM, MD, LG, XL, XXL } = require('@zendeskgarden/react-typography');
+
+const AnchorExample = () => (
+  <>
```
jzempel: Remove?

packages/buttons/src/views/Anchor.js
```diff
@@ -31,33 +32,69 @@ const StyledAnchor = styled.a.attrs({
       [ButtonStyles['is-selected']]: selected
     })
 })`
+  ${props =>
+    props.external &&
+    `
+    &&& {
```
jzempel: Is this really necessary?

packages/buttons/src/views/Anchor.example.md
austin94: Since the AnchorExample component is technically returning an array: `['An`, React.createElement(Anchor, ...), 'looks like this']` it's required to wrap the items in a fragment (or a span)

packages/buttons/src/views/Anchor.js
austin94: I was able to remove a single level of specificity, now `&&`. Not sure how/why our CSS classes are so specific.

packages/buttons/src/views/Anchor.example.md
austin94: I actually just removed this example for something a little clearer. No longer needed.

packages/buttons/src/views/Anchor.example.md
```diff
@@ -9,23 +9,55 @@ If you need a `<button>` that has anchor styling, use the
...
```
jzempel: So I guess you wouldn't need these now.

PR #274, opened 2019-02-19
fix(loaders): increase alpha of skeleton animation

austin94:
## Description

The alpha value that I used for the background animation in the `Skeleton` component was off by a few percent.

This PR corrects the animation color.

(Reviewed locally by @ginnywood)

### DIFF
--- a/packages/loaders/src/Skeleton.js
+++ b/packages/loaders/src/Skeleton.js
@@ -12,7 +12,7 @@ import { rgba } from 'polished';
 import { retrieveTheme, isRtl } from '@zendeskgarden/react-theming';
 import {
   zdColorGrey800,
-  zdColorKale800,
+  zdColorKale700,
   zdColorWhite,
   zdSpacingXxs
 } from '@zendeskgarden/css-variables';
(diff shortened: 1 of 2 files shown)


packages/loaders/src/Skeleton.js
```diff
@@ -75,9 +74,9 @@ const StyledSkeleton = styled.div.attrs({
     /* stylelint-disable */
     background-image:
       linear-gradient(
-        -45deg,
+        45deg,
```
jzempel: Doesn't this need to be spun to `-45deg` to retain expected orientation for RTL?

packages/loaders/src/Skeleton.js
```diff
@@ -75,9 +74,9 @@ const StyledSkeleton = styled.div.attrs({
     /* stylelint-disable */
```
jzempel: Please either fix the stylelint config to allow this exception globally or use the disable syntax that's indicating exactly why you need the exception here. Ex: `stylelint-disable function-comma-space-after */

packages/loaders/src/Skeleton.js
```diff
@@ -75,9 +74,9 @@ const StyledSkeleton = styled.div.attrs({
...
     background-image:
       linear-gradient(
-        -45deg,
+        45deg,
         ${rgba(zdColorWhite, 0)},
-        ${props => rgba(props.dark ? zdColorKale800 : zdColorGrey100, 0.5)},
+        ${props => (props.dark ? rgba(zdColorKale800, 0.4) : rgba(zdColorWhite, 0.6))},
         ${rgba(zdColorWhite, 0)}
```
jzempel: Both this and the previous one can just be `transparent`, right?

PR #276, opened 2019-02-27
feat: export components as mjs files

### LINKED ISSUE #198: Export ES6 modules
This is a nice project 👏 

It would be ace if you could include the ES6 modules in the exported assets and reference them in the `package.json` file.

That would make it possible to use the ES6 module directly in other projects, rather than relying on only the commonjs version.

li-kai:
## Description

This uses webpack instead of rollup to export files, which brings benefits such as:
1. faster, parallel building of umd, cjs and mjs file formats
1. smaller file sizes as [webpack bootstrap code](https://gist.github.com/li-kai/3a265a11c2cbe9a1b1ce22930713d7fb#file-index-webpack-js-L7) is not included (see here for difference in output: https://gist.github.com/li-kai/3a265a11c2cbe9a1b1ce22930713d7fb) for all file types

Disadvantage: Maintenance overhead of rollup, because styleguidist is still using webpack, so webpack dependencies still remain.

## Detail

How this works:
1. Replace webpack parts with rollup
1. Combine the 3 different webpack configs into one rollup config, toggled with environment variables such as `TARGET:node` and `TARGET:web`.
1. Change styleguidist's webpack config to use mjs files
1. Disable `babel-plugin-styled-components` in styleguidist, as it was causing problems.

closes #198 

## Checklist

      designer as a reviewer)
  component

## Risks

CSS bundled could be different, because zendeskgarden css is not bundled as css-modules. However, from my observation of a few css outputs, they seem to be the same.


ryanseddon:
Wow so we're gonna see on average a 50% drop in file size just by switching to rollup? 100% onboard with just that if that's the case I'm testing locally and 

Some notes so far:
- I think something is up with the umd builds they seem way to small considering they should bundle all deps needed for that package.
- You're missing the `yarn build:analyze` support where we could inspect and visualise what was being used. Look like rollup-plugin-analyzer might be something we could use.

jzempel:
@li-kai closing due to lack of resolution. Please re-open if you want to explore further. Thanks!

PR #284, opened 2019-03-05
fix(notifications): update package.json to pass build

austin94:
## Description

It looks like #283 caused the build to break due to the `package.json` being formatted.  I don't know why this didn't fail in the PRs Travis run. I thought that prettier-package-json moved it for you since it stayed alphabetical 🤷‍♂️ 

We will need to force apply a `patch` version to `react-menus` to ensure it gets deployed correctly.


austin94:
@jzempel Weird, that should fail the PR Travis run then ☹️

PR #289, opened 2019-03-14
feat(loaders): introduce Inline component

austin94:
## Description

This is the v1 of a "vertical space neutral" typing indicator component.  @allisonacs could you help me with the naming here? Not sure if we need to generalize this name for other uses in the future.

## Detail

This component uses the CSS pulse animation from the original self-service implementation with the following changes:

* The SVG circles are now defaulted to 0.2 opacity to avoid a "flash of unstyled content"
* It now supports RTL mode via our `ThemeProvider`
* Size and color are now calculated in CSS-in-JS so can be provided via props

**Default size:** 16px
**Default color:** grey-600

### LTR

TypingIndicatorLTR

### RTL

TypingIndicatorRTL

## Accessibility

For loading indicators, the `aria-busy` and `aria-live` attributes must be applied to containing elements and have to be managed by consumers (same as the other loading components).

## Checklist

      designer as a reviewer)
- ~:nail_care: view component styling is based on a Garden CSS
      component~
- [x] :guardsman: includes new unit and snapshot tests

### DIFF
--- a/packages/loaders/src/TypingIndicator.example.md
+++ b/packages/loaders/src/TypingIndicator.example.md
@@ -1,7 +1,18 @@
 ```jsx
 const { zdColorGreen600, zdColorRed600, zdColorBlue600 } = require('@zendeskgarden/css-variables');
 
-<Grid>
+<Grid aria-busy="true" aria-live="polite">
+  <Row>
+    <p>
+      All areas that contain these loaders must include the{' '}
+      <code>[aria-busy="true",aria-live="polite"]</code>
+      attributes for
+      <a href="https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy">
+        an accessible experience
+      </a>
+      .
+    </p>
+  </Row>
   <Row>
     <Col md>
       <TypingIndicator />
--- a/packages/loaders/src/TypingIndicator.example.md
+++ b/packages/loaders/src/TypingIndicator.example.md
@@ -1,0 +1,28 @@
+```jsx
+const { zdColorGreen600, zdColorRed600, zdColorBlue600 } = require('@zendeskgarden/css-variables');
+
+<Grid>
+  <Row>
+    <Col md>
+      <TypingIndicator />
+    </Col>
+    <Col md>
+      <TypingIndicator size={32} />
+    </Col>
+    <Col md>
+      <TypingIndicator size={64} />
+    </Col>
+  </Row>
+  <Row>
+    <Col md>
+      <TypingIndicator color={zdColorGreen600} />
+    </Col>
+    <Col md>
+      <TypingIndicator size={32} color={zdColorRed600} />
+    </Col>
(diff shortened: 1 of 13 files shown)


austin94:
Pre-published for review at https://garden.zendesk.com/react-components/loaders/

packages/loaders/src/TypingIndicator.js
```diff
@@ -0,0 +1,89 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { keyframes } from 'styled-components';
+import { isRtl } from '@zendeskgarden/react-theming';
+import { zdColorGrey600 } from '@zendeskgarden/css-variables';
+
+const COMPONENT_ID = 'loaders.spinner';
+
+const pulseAnimation = keyframes`
+  0% {
```
ryanseddon: can combine `0%` and `100%` here.

```css
0%, 100% {
 opacity: .2;
}
```

packages/loaders/src/TypingIndicator.js
```diff
@@ -0,0 +1,89 @@
...
+
+TypingIndicator.propTypes = {
+  /* Width of the loader in px */
+  size: PropTypes.number,
+  color: PropTypes.string
+};
+
+TypingIndicator.defaultProps = {
+  size: 16,
+  color: zdColorGrey600
```
ryanseddon: Should this be `inherit` to match the other loaders?

packages/loaders/src/TypingIndicator.example.md
```diff
@@ -0,0 +1,28 @@
...
+  </Row>
+  <Row>
+    <Col md>
+      <TypingIndicator color={zdColorGreen600} />
+    </Col>
+    <Col md>
+      <TypingIndicator size={32} color={zdColorRed600} />
+    </Col>
+    <Col md>
+      <TypingIndicator size={64} color={zdColorBlue600} />
```
ryanseddon: Should we update the demo to include the `aria-busy` and `aria-live` attributes?

packages/loaders/src/TypingIndicator.js
austin94: I didn't realizing that we were defaulting that with the other components.

@allisonacs we believe the for `Dots` it should be `blue-500` by default and for this inline loader, it should be `grey-600`, correct? If that's the case I think we should just default these otherwise we will have a wide variety of loading colors.

It's still customizable. Just with a better default.

austin94: After talking with @allisonacs we decided to have a more opinionated default with this specific loader.  The `inherit` value can still be applied for the use-cases that require it.

jzempel: @austin94 @allisonacs we should've made the API consistent with other loaders. Let's revisit this minor detail at some point.

PR #293, opened 2019-03-19
fix(loaders): default Inline loader to use inherit color

austin94:
## Description

Per our conversation today, the `Inline` loader should have a default color of `inherit`.

## Checklist

      designer as a reviewer)
      component
- [x] :guardsman: includes new unit and snapshot tests


austin94:
> Nice, do you need to remove the css-variables dep too?

This is still used in the `Skeleton` loader

PR #294, opened 2019-03-22
feat(autocomplete): make it a controlled component

Anifacted:
- [ ] **BREAKING CHANGE?** 

## Description

- Makes it possible to fully control the Autocomplete component's state from the outside, similarly to how the Menu component works.

- Also adds support for the `children` prop, which can be used to display a custom value while the component is not open.

## Checklist

      designer as a reviewer)
      component
- [ ] :guardsman: includes new unit and snapshot tests

### DIFF
--- a/packages/autocomplete/src/elements/Autocomplete.example.md
+++ b/packages/autocomplete/src/elements/Autocomplete.example.md
@@ -26,6 +26,37 @@ initialState = {
 />;
 ```
 
+### Controlled Example
+
+```jsx
+initialState = {
+  selectedValue: 'option-1',
+  isOpen: false
+};
+
+<Autocomplete
+  label="Controlled Autocomplete"
+  selectedValue={state.selectedValue}
+  isOpen={state.isOpen}
+  onChange={selectedValue => setState({ selectedValue })}
+  onStateChange={setState}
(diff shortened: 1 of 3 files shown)


Anifacted:
Thanks, @austin94.

I have a question: Our use-case requires us to define a custom value for the dropdown when it isn't open. Just like you can specify `children` for a Select component.
The `optionDictionary[selectedValue]` pattern doesn't really work well for dynamic data, since it requires the selected piece of data to be present in the `options` prop at all times, even (especially) when the component is not open.

I made it work via `children`, simply by adding this conditional:

```js
<StyledValueWrapper>
  {children || optionDictionary[selectedValue]}
</StyledValueWrapper>
```

Should I push this change to this PR, or should we pursue a different pattern in your opinion?

We're in a bit of a hurry BTW :)

austin94:
I think that change makes sense. Feel free to push it. 

A lot of the decisions in these components were for too limited of a use case. Would love to run some newer APIs by you when we’re in town next week.

Anifacted:
@austin94 I pushed the commit for adding support for the `children` prop 👍

Anifacted:
Cool, thanks! 👍

PR #303, opened 2019-04-05
chore(testing): remove snapshot testing

austin94:
## Description

In preparation for us moving to styled-components v4 we must remove our existing snapshot testing.

## Detail

- [x] @zendeskgarden/react-autocomplete
- [x] @zendeskgarden/react-avatars
- [x] @zendeskgarden/react-breadcrumbs
- [x] @zendeskgarden/react-buttons
- [x] @zendeskgarden/react-checkboxes
- [x] @zendeskgarden/react-chrome
- [x] @zendeskgarden/react-grid
- [x] @zendeskgarden/react-loaders
- [x] @zendeskgarden/react-menus
- [x] @zendeskgarden/react-modals
- [x] @zendeskgarden/react-notifications
- [x] @zendeskgarden/react-pagination
- [x] @zendeskgarden/react-radios
- [x] @zendeskgarden/react-ranges
- [x] @zendeskgarden/react-select
- [x] @zendeskgarden/react-selection
- [x] @zendeskgarden/react-tables
- [x] @zendeskgarden/react-tabs
- [x] @zendeskgarden/react-tags
- [x] @zendeskgarden/react-testing
- [x] @zendeskgarden/react-textfields
- [x] @zendeskgarden/react-theming
- [x] @zendeskgarden/react-toggles
- [x] @zendeskgarden/react-tooltips
- [x] @zendeskgarden/react-typography
- [x] @zendeskgarden/react-utilities

## Checklist

- [ ] ~:ok_hand: design updates are Garden Designer approved (add the
      designer as a reviewer)~
- [ ] ~:nail_care: view component styling is based on a Garden CSS
      component~
- [ ] ~:globe_with_meridians: Styleguidist demo is up-to-date (`yarn start`)~
- [ ] ~:arrow_left: renders as expected with reversed (RTL) direction~
- [ ] ~:wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver~
- [ ] ~:guardsman: includes new unit and snapshot tests~
- [ ] ~:memo: tested in Chrome, Firefox, Safari, Edge, and IE11~

PR #304, opened 2019-04-08
chore(testing): migrate limited packages to react-testing-library

austin94:
## Description

While removing our snapshot tests in #303 our test coverage dropped a few percentages.  It turned out that relying on snapshots actually didn't test a whole lot things 😬 

We've been discussing the merits of having more "integration" focused testing over the past few weeks.  Enzyme is a great testing library, but many times we find our selves testing how we think our components work rather than how they actually get placed within the DOM.

`react-testing-library` is a popular testing library that can help us test our components in more "real-world" scenarios.

From their [Guiding Principles](https://testing-library.com/docs/guiding-principles)...

> The more your tests resemble the way your software is used, the more confidence they can give you.

This is a great, short read. It does a much better job of explaining the long-term benefits that this change could provide than I ever could. Please give it a look.

## Detail

`react-testing-library` is able to be used side-by-side with enzyme. There isn't an immediate need for us to migrate all of our tests at this time.

During the snapshot removal, `react-chrome` and `react-loaders` saw the most significant decrease in coverage. This PR increases their coverage to 100% with the following changes:

1. Include `react-testing-library` package
1. Create a custom `garden-test-utils` module mapping within Jest. This provides custom `render` and `renderRtl` methods that wrap all examples in our `ThemeProvider` component.
1. Include the `jest-dom` package to have DOM specific jest matchers/utilities
1. Use the `toHaveStyleRule` assertion provided by `jest-styled-components` (already a dependency) that also supports this library.

## Finale

I actually uncovered several bugs in each of the packages while making this move 🎉 I personally feel like this testing strategy will help us keep issues like this from happening in the first place 😄 

* `react-chrome`
  * `NavItemText` and `SubNavItemText`
    * Both of these packages had a `wrap` prop. 
    * This is a valid attribute for React elements and would have been passed through to its respective DOM element.  
    * Also, us requiring a `bool` prop-type triggers a ReactDOM error due to it expecting a different type
    * Due to this never working I believe this prop change **shouldn't require a breaking change**
* `react-loaders`
  * `Skeleton`
    * This component used the `width` and `height` props to calculate other styling values.
    * These props are also valid React element attributes and would be passed to its DOM element.
    * This fix doesn't change our prop structure.

## Checklist

      designer as a reviewer)~
      component~


packages/chrome/src/views/Chrome.spec.js
```diff
@@ -6,21 +6,20 @@
  */
 
 import React from 'react';
-import { shallow } from 'enzyme';
-import { shallowWithTheme } from '@zendeskgarden/react-testing';
+import { render, renderRtl } from 'garden-test-utils';
```
ryanseddon: Only issue with this is it create a sense that this is an external package and might be a bit of misdirection as to where it comes from.

packages/loaders/src/Dots.spec.js
```diff
@@ -6,43 +6,228 @@
...
+
+  describe('Animation', () => {
+    it('updates animation after request animation frame', () => {
+      const { container } = render(<Dots data-test-id="dots" />);
+
+      jest.runOnlyPendingTimers();
 
-    it('clears delay timeout as component is unmounted', () => {
-      const wrapper = mount(<Dots />);
+      expect(container.firstChild.firstChild).toMatchInlineSnapshot(`
```
ryanseddon: You can just do a `querySelector` to get the first `g` elem.

```js
container.querySelector('g')
```

packages/loaders/src/Dots.spec.js
```diff
@@ -6,43 +6,228 @@
...
     });
 
-    it('does not render LoadingPlaceholder if delayMS is 0', () => {
-      const wrapper = mount(<Dots delayMS={0} />);
+    it('updates animation after request animation frame with negative bound velocity', () => {
+      const { container } = render(<Dots velocity={-1.1} />);
+
+      jest.runOnlyPendingTimers();
+
+      expect(container.firstChild.firstChild).toMatchInlineSnapshot(`
```
ryanseddon: same for all the uses of `.firstChild.firstChild` I think querySelector is clearer as to what you're trying to find.

packages/chrome/src/views/Chrome.spec.js
austin94: I agree that this may become confusing, but I think this abstraction isn't premature.  From their [setup docs](https://testing-library.com/docs/react-testing-library/setup#configuring-jest-with-test-utils) this seems to be a common practice and we already do this with our enzyme test utilities.

With the file being the same name as the import I hope it would be discoverable as well.

utils/test/garden-test-utils.js
```diff
@@ -0,0 +1,37 @@
...
+const RtlProvider = ({ children }) => {
+  return <ThemeProvider rtl>{children}</ThemeProvider>;
+};
+
+RtlProvider.propTypes = {
+  children: PropTypes.node
+};
+
+const customLtrRender = (ui, options) => render(ui, { wrapper: LtrProvider, ...options });
+const customRtlRender = (ui, options) => render(ui, { wrapper: RtlProvider, ...options });
```
jzempel: Just curious why under `utils` instead of the `testing` package? Is this not considered generally useful for other components that render under our theme provider?

austin94: Since this is the first implementation of this testing library I don't think it's ready for us to be shared as a consumable for other teams. Also, no other teams use `react-testing-library` at this time.

I'd like to have us use this in some more advanced scenarios before we decide it's stable enough to share.

sunesimonsen:
Conceptually I think this is a really good change. I generally think snapshot testing is just a way to avoid writing real tests and it became popular because it is easy. Snapshot tests is usually really brittle because they are not focused and attach themselves to too much detail.

I like the switch to DOM testing as well. We have been using this way of testing in the guide-client for some time now and I think it is much better. The big benefit is that you test again an API that is forever stable.

While I understand the choice of react-testing-library based on popularity. I do think it is a pitty that you are not evaluating https://unexpected.js.org/unexpected-reaction/ and https://unexpected.js.org/unexpected-dom/ which in my very biased opinion is just a lot better. 

But overall great decision.

PR #308, opened 2019-04-19
fix(tooltips): allow focusable content within tooltips

austin94:
## Description

Currently, if a user clicks or keyboard focuses the triggering element the tooltips dismisses itself when a user tries to interact with any link, button, or focusable element in the tooltip body.

## Detail

This PR introduces some additional `onFocus` and `onBlur` event handlers for the tooltip body to allow elements within it to receive focus via mouse and keyboard regardless of trigger focus state.

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/tooltips/src/containers/TooltipContainer.spec.js
+++ b/packages/tooltips/src/containers/TooltipContainer.spec.js
@@ -188,7 +188,7 @@ describe('TooltipContainer', () => {
       expect(findTooltip(wrapper)).toHaveLength(1);
     });
 
-    it('should close tooltip open if blurred', () => {
+    it('should close tooltip if blurred', () => {
       findTrigger(wrapper).simulate('mouseenter');
       jest.runOnlyPendingTimers();
       wrapper.update();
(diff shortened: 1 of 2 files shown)


packages/tooltips/src/containers/TooltipContainer.spec.js
```diff
@@ -175,6 +175,36 @@ describe('TooltipContainer', () => {
...
+
+      findTrigger(wrapper).simulate('blur');
+      findTooltip(wrapper).simulate('focus');
+      jest.runOnlyPendingTimers();
+      wrapper.update();
+
+      expect(findTooltip(wrapper)).toHaveLength(1);
+    });
+
+    it('should close tooltip open if blurred', () => {
```
claricen: // minor: remove "open" from test description

austin94:
> This also looks like an accessibility win, yes?

I think in general we should be avoiding having actionable content within tooltips, but yes! This should have some marginal improvements on the a11y side

PR #312, opened 2019-04-23
fix(autocomplete): Do not always trigger dropdown closing on blur

Anifacted:
- [ ] **BREAKING CHANGE?** 

## Description

Currently, when blurring the input, a re-render is triggered when the `isOpen` prop going from `true` to `false`. This should only happen when `isOpen` is `true` already, since side-effects may be attached to this logic. This PR fixes that.

## Checklist

      designer as a reviewer)
      component


austin94:
This was deployed under `@zendeskgarden/react-autocomplete@0.10.8`

PR #315, opened 2019-04-25
feat(dropdowns): introduce new Dropdowns package

### LINKED ISSUE #239: Add `isLoading` property to `Autocomplete` and `MutliSelect` components
## Expectations

Using our `<Dots />` loader is a common use-case within our filterable components, Autocomplete and MultiSelect.

Providing an opinionated placement for the loading/async state would be helpful for consumers. 

Original feature request by @sunesimonsen 

###### Fine Print

* Component: `@zendeskgarden/react-autocomplete`
* Browsers: all

austin94:
## Description

This is it! Here is the new `react-dropdowns` package. It consolidates `react-selection`, `react-menus`, and `react-autocomplete` into a single package (multiselect will be added in a future minor release).

This API has the following new dependency requirements:

- `react@^16.8`
- `react-dom@^16.8`

These new requirements allow us to use the newer Context and Hook features to provide a more developer friendly experience.

This API should be flexible enough to ensure that **developers no longer have to consistently use our render-prop containers** 🎉 .

## Example

```js
<Dropdown onSelect={item => alert(item)}>
    <Trigger refKey="innerRef">
      <Button>Default Menu</Button>
    </Trigger>
    <Menu>
       <Item value="option-1">Option 1</Item>
       <Item value="option-2">Option 2</Item>
       <Item value="option-3">Option 3</Item>
     </Menu>
</Dropdown>
```

## Detail

The new API provides the [Downshift](https://github.com/downshift-js/downshift) and [PopperJS](https://popper.js.org/) libraries through a shared Context.  Due to Downshifts use of control props and the [prop getter pattern](https://github.com/downshift-js/downshift#prop-getters) we retain the same level of flexibility as our original containers.

If there are still advanced areas that need customization then they can use `Downshift` directly.

## Testing

There is currently `100%` line and `~95%` branch coverage for this package. I also implemented the tests in `react-testing-library` to match our new style.

## Known Issues

There is still some pixel shifting in the `Autocomplete` component when transitioning to the `input` view.  I will look into this some more.

Closes #239 

CC @vinceprofeta

## Checklist

      designer as a reviewer)
      component
- [x] :memo: tested in Chrome, Firefox, Safari, Edge, and IE11 (and mobile since this is a rewrite)


README.md
```diff
@@ -1,4 +1,5 @@
 # Garden React Components [![Build Status](https://img.shields.io/travis/zendeskgarden/react-components/master.svg?style=flat-square)](https://travis-ci.org/zendeskgarden/react-components) [![Dependency Status](https://img.shields.io/david/dev/zendeskgarden/react-components.svg?style=flat-square)](https://david-dm.org/zendeskgarden/react-components?type=dev) [![Coverage Status](https://img.shields.io/coveralls/github/zendeskgarden/react-components/master.svg?style=flat-square)](https://coveralls.io/github/zendeskgarden/react-components) <!-- markdownlint-disable -->
```
austin94: This file got formatted by prettier, even thought it's not necessary I personally like the table layout being "correct" 😄 Let me know if this is something I need to revert

__mocks__/popper.js.js
```diff
@@ -0,0 +1,35 @@
+/**
```
austin94: This is a global mock for `popper.js`. It now removes all `document.createRange` errors that were visible while running tests.

packages/dropdowns/src/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,185 @@
...
+            onOpen && onOpen(changes.isOpen, stateAndHelpers);
+          }
+
+          if (Object.prototype.hasOwnProperty.call(changes, 'selectedItem')) {
+            if (selectedItems) {
+              const { itemToString } = stateAndHelpers;
+              const existingItemIndex = selectedItems.findIndex(item => {
+                return itemToString(item) === itemToString(changes.selectedItem);
+              });
+              const updatedSelectedItems = selectedItems.slice();
```
ryanseddon: `Array#from` would be clearer here that you want a new copy and not a reference.

packages/dropdowns/src/Menu/Menu.js
```diff
@@ -0,0 +1,122 @@
...
+    nextItemsHashRef,
+    popperReferenceElementRef,
+    downshift: { isOpen, getMenuProps }
+  } = useDropdownContext();
+
+  // Reset Downshift refs on every render
+  itemIndexRef.current = 0;
+  nextItemsHashRef.current = {};
+  previousItemRef.current = undefined;
+  previousIndexRef.current = undefined;
```
ryanseddon: How come you need to do this?

packages/dropdowns/src/Menu/Menu.js
```diff
@@ -0,0 +1,122 @@
...
+    <MenuContext.Provider value={{ itemIndexRef }}>
+      <Popper placement={popperPlacement} modifiers={popperModifiers} eventsEnabled={eventsEnabled}>
+        {({ ref, style, scheduleUpdate, placement: currentPlacement }) => {
+          let computedStyle = menuStyle;
+
+          // Calculate custom width if ref is provided from Select or Autocomplete
+          if (popperReferenceElementRef.current) {
+            computedStyle = {
+              ...menuStyle,
+              width: popperReferenceElementRef.current.getBoundingClientRect().width
```
ryanseddon: Should we check if `getBoundingClientRect` is available this will break SSR right?

packages/dropdowns/src/Menu/Menu.js
```diff
@@ -0,0 +1,122 @@
...
+
+Menu.propTypes = {
+  popperModifiers: PropTypes.object,
+  eventsEnabled: PropTypes.bool,
+  zIndex: PropTypes.number,
+  style: PropTypes.object,
+  /**
+   * All valid [Popper.JS Placements](https://popper.js.org/popper-documentation.html#Popper.placements)
+   */
+  placement: PropTypes.string,
```
ryanseddon: Are we dropping the logical placement names now?

packages/dropdowns/src/Select/Select.js
```diff
@@ -0,0 +1,106 @@
...
+import useFieldContext from '../utils/useFieldContext';
+
+const VALIDATION = {
+  SUCCESS: 'success',
+  WARNING: 'warning',
+  ERROR: 'error'
+};
+
+/**
+ * Applies state and a11y attributes to it's children. Must be nested within a `<Field>` component.
```
ryanseddon: `it's` -> `its`

packages/dropdowns/src/Trigger/Trigger.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+import React, { useRef, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Reference } from 'react-popper';
+import { StyledInput } from '../styled';
+import useDropdownContext from '../utils/useDropdownContext';
+
+/**
+ * Applies state and a11y attributes to it's children. Must be nested within a `<Dropdown>` component.
```
ryanseddon: `it's` -> `its`

packages/dropdowns/src/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,185 @@
+/**
```
carvallegro: Why not create functors instead of declaring the functions inside the component? There is a few of them and they're not small, it makes it harder to understand the logic

packages/dropdowns/src/Fields/Field.js
```diff
@@ -0,0 +1,35 @@
...
+import PropTypes from 'prop-types';
+import { StyledField } from '../styled';
+import useDropdownContext from '../utils/useDropdownContext';
+
+export const FieldContext = createContext(undefined);
+
+/**
+ * Accepts all `<div>` props
+ */
+function Field(props) {
```
carvallegro: Should we use only the arrow or the function notation for consistency?

packages/dropdowns/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "styled-components": "^3.2.6"
+  },
+  "devDependencies": {
+    "@zendeskgarden/css-arrows": "^3.1.1",
+    "@zendeskgarden/css-forms": "^7.0.5",
+    "@zendeskgarden/css-menus": "^9.0.5",
+    "@zendeskgarden/react-theming": "^3.2.1",
+    "@zendeskgarden/svg-icons": "^6.1.1"
```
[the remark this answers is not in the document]
austin94: I removed it from the examples, but left an import. Removed.

packages/dropdowns/src/Dropdown/Dropdown.js
austin94: Neat. These both make shallow copies so it should be equivalent. Added.

packages/dropdowns/src/Menu/Menu.js
austin94: I guess it would depend on the SSR environment. I've added another check for `getBoundingClientRect`

packages/dropdowns/src/Menu/Menu.js
austin94: This is correctly handled in the code, I just forgot to bring over the correct propTypes. Updated to reflect the RTL-aware placements

packages/dropdowns/src/Menu/Items/Item.spec.js
```diff
@@ -0,0 +1,205 @@
...
+      </Dropdown>
+    );
+
+    expect(() => {
+      render(<Example />);
+    }).toThrow('All Item components require a `value` prop');
+  });
+
+  it('does not throw if value is not provided while item is disabled', () => {
+    console.error = jest.fn(); // eslint-disable-line no-console
```
louisscruz: This is affecting other tests too, right? Maybe a `beforeEach` + `afterEach` should be used so that any new tests written won't unknowingly rely on this?

packages/dropdowns/src/Menu/Menu.js
austin94: Since the Downshift action `selectItemAtIndex` is index based, we keep the ref of the most recent (should only ever be 1) `PreviousItem`.

The `previousItemRef` is actually left over from a previous implementation. I've now removed it.

packages/dropdowns/src/Select/Select.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Focus trigger when Menu is closed
+    if (!isOpen && previousIsOpenRef.current) {
+      triggerRef.current && triggerRef.current.focus();
+    }
+    previousIsOpenRef.current = isOpen;
+  }, [isOpen]);
+
+  const { ...selectProps } = getToggleButtonProps({
```
louisscruz: Does this get function already return a new object? Maybe it's safe to just `const selectProps = getToggleButtonProps(...)`?

packages/dropdowns/src/examples/menu.md
```diff
@@ -0,0 +1,229 @@
+## Basic Menu Layout
+
+```jsx
+const { Button } = require('@zendeskgarden/react-buttons/src');
+const { IconButton, Icon } = require('@zendeskgarden/react-buttons/src');
+const SettingsIcon = require('@zendeskgarden/svg-icons/src/16/gear-stroke.svg').default;
```
louisscruz: Any reason `require` is used throughout?

packages/dropdowns/src/Menu/Items/Item.spec.js
austin94: I've updated all `console.error` tests including those in other packages. These are now mocked correctly.

__mocks__/popper.js.js
jzempel: I removed this locally and am not seeing any errors.

README.md
jzempel: I find it less pretty, but it wasn't super pretty to begin with.

README.md
```diff
@@ -14,34 +15,35 @@ CSS](https://github.com/zendeskgarden/css-components) for styling.
 See the individual package README for the React component you would like
 to install.
 
-Package | Version | Dependencies
```
jzempel: I think we could consider removing `autocomplete`, `menus`, and `selects` from this table as a move toward encouraging the behavior (use dropdowns, instead) we desire. Although, I guess `multiselect` isn't represented, yet.

packages/dropdowns/README.md
```diff
@@ -0,0 +1,107 @@
...
+    </Menu>
+  </Dropdown>
+</ThemeProvider>;
+```
+
+For all components within the `react-dropdowns` package, the menu layouts and
+implementations are interchangeable.
+
+Whether you're making a `Select`, `Autocomplete`, or a traditional `Menu` the `<Menu>` implementation
+will adapted to it's consumer.
```
jzempel: *its

packages/dropdowns/README.md
```diff
@@ -0,0 +1,107 @@
...
+The customizations available within this can be broken into two groups: _placement / positioning_
+and _dropdown state_
+
+### Placement / Positioning
+
+Internally, the `<Dropdown>` component uses [PopperJS](https://popper.js.org/popper-documentation.html)
+for its positioning calculations.
+
+The `<Menu>` component accepts all customizations regarding placement, boundaries, overflows,
+etc. via the `popperModifiers` prop.
```
jzempel: Is it consistent to name this `popperModifiers` while using `downshiftProps` below?

packages/dropdowns/src/examples/menu.md
austin94: This is a side-effect of styleguidist. Just specific to our examples.

packages/dropdowns/src/examples/menu.md
```diff
@@ -0,0 +1,229 @@
...
+const GroupIcon = require('@zendeskgarden/svg-icons/src/16/user-group-stroke.svg').default;
+const ClipboardSvg = require('@zendeskgarden/svg-icons/src/12/clipboard-list-stroke.svg').default;
+const BoxSvg = require('@zendeskgarden/svg-icons/src/12/box-3d-stroke.svg').default;
+const DatabaseSvg = require('@zendeskgarden/svg-icons/src/12/database-stroke.svg').default;
+
+const StyledItemWrapper = styled.div`
+  display: flex;
+
+  li:not(:last-child) {
+    border-right: 1px solid #e9ebed;
```
jzempel: Can you use the CSS variable `zdColorGrey200` here?

packages/dropdowns/src/styled/field/StyledInput.js
```diff
@@ -0,0 +1,41 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import styled, { css } from 'styled-components';
+import TextStyles from '@zendeskgarden/css-forms/dist/text.css';
+
+const hiddenStyling = css`
```
jzempel: Much of this styling appears superfluous for what you're trying to achieve.

packages/dropdowns/src/styled/field/StyledInput.js
```diff
@@ -0,0 +1,41 @@
...
+  border-color: initial;
+  background: 0 center;
+  width: 1px;
+  color: transparent;
+  font-size: inherit;
+  border-image: initial;
+`;
+
+const StyledInput = styled.input.attrs({
+  className: classNames(TextStyles['c-txt__input'], TextStyles['c-txt__input--bare'])
```
jzempel: add `vertical-align: baseline` to prevent pixel bumping.

packages/dropdowns/README.md
```diff
@@ -0,0 +1,107 @@
...
+npm install @zendeskgarden/react-dropdowns
+
+# Peer Dependencies - Also Required
+npm install react react-dom prop-types styled-components@3.4.10 @zendeskgarden/react-theming
+```
+
+## Requirements
+
+This package uses several modern React features like [Context](https://reactjs.org/docs/context.html)
+and [Hooks](https://reactjs.org/docs/hooks-intro.html) to provide a more consistent developer experience.
```
jzempel: "to provide a more consistent developer experience" doesn't really mean anything to me – consistent compared to what? Maybe "to provide an enhanced component API" instead.

README.md
austin94: I've removed menus and select from the table. Lets leave `autocomplete` until the multiselect is moved over.

austin94: This reminds me. We will need to add in the deprecation warnings to the docs and packages, but that can come after the PR.

packages/dropdowns/README.md
austin94: `downshiftProps` actually get spread onto the `Downshift` element. This is just a 1-to-1 prop match with PopperJS since framework agnostic.

packages/dropdowns/README.md
austin94: I've updated this to your suggestion.

packages/dropdowns/src/Dropdown/Dropdown.js
austin94: There is a lot of interaction with the props and items provided via context.  Removing the transforms out of the component was making it personally hard for me to follow the logic.

I'll look into breaking this logic up some more.

packages/dropdowns/src/Fields/Field.js
austin94: I was mixed with this in a few components, but I'd like to have function notation for functional components.  Even thought we use them interchangeably within classes they do perform differently.

Within the React docs it seems to follow the `function notation` whenever you don't have to worry about scoping. https://reactjs.org/docs/components-and-props.html#function-and-class-components

packages/dropdowns/src/styled/field/StyledInput.js
austin94: I've seen this reset used in a few libraries. [react-select](https://react-select.com/home) is one that comes to mind.

I've removed a few of the more extreme resets, but it should all be hidden by default.

__mocks__/popper.js.js
austin94: The `console.errors` don't trigger test failures, but you can view them in CI or if you look back through the test runner.

sunesimonsen:
🔥👏

PR #323, opened 2019-05-03
fix(dropdowns): improve performance of Menu updates

### LINKED ISSUE #321: Dropdown component performance issues
The new dropdown package is great!
Though, It does have a performance issue. Looks like a memory leak to me.

## Expectations

Unchanged high framerate and performance over time when hovering menus.

## Reality

dropdown-perf

Decreasing framerate while hovering more and more items in the same dropdown. This performance problem is "reset" when the menu is closed and reopened.

I have not run analysis in detail, but profiling in Chrome could maybe reveal the source of the problem.

## Steps to Reproduce

  1. Open a dropdown menu (E.g. the Select Usage example in the Styleguide: https://garden.zendesk.com/react-components/dropdowns/#!/Select%20usage/3)
  2. Keep hovering items until the highlighted backdrop starts lacking behind

### Fine Print

* Component: dropdowns
* Version: 0.1.1
* Browsers: Only tested in Chrome

austin94:
## Description

There are some performance issues with the original dropdown implementation, specifically with how we handle scheduled updates within PopperJS.

## Detail

The specific issues that are fixed include:

- `eventsEnabled` prop was enabled when dropdown was closed
  - This was causing popper to re-position as a user was scrolling, even when the menu were closed. This isn't needed. It now disables this option while the menu is closed.
- Constantly scheduling updates
  - I originally placed a `scheduleUpdate()` call within the Popper renderer. This causes an infinite loop of renders 💀 
  - This isn't as noticeable in some instances since Popper is using requestAnimationFrame for its positioning logic, but can become noticeable with a lot of state updates in Downshift.
  - There is now logic to only run this once per render while the dropdown is open. We are unable to only call this when the `Menu` transitions from closed -> open due to the ability to change the number of items and/or the size of the `Menu`. This ensures that it is always placed correctly.

Closes #321 

## Open Question

@ryanseddon I remember you working with some positioning quirks in PopperJS when it came to animation. There are some visual "jerks" that can be rarely seen in the non-vertical placements. I think this is more of an issue with our hidden styling, but could use some help figuring this one out.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/packages/dropdowns/src/Menu/Menu.js
+++ b/packages/dropdowns/src/Menu/Menu.js
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { createContext } from 'react';
+import React, { createContext, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Popper } from 'react-popper';
 import { withTheme, isRtl } from '@zendeskgarden/react-theming';
@@ -37,6 +37,18 @@ const Menu = props => {
     popperReferenceElementRef,
     downshift: { isOpen, getMenuProps }
   } = useDropdownContext();
+  const scheduleUpdateRef = useRef(undefined);
+
+  useEffect(() => {
+    /**
+     * Recalculate popper placement while open to allow animations to complete.
+     * This must be ran every render to allow for the number of items to change
+     * and still be placed correctly.
+     **/
+    if (isOpen) {
+      scheduleUpdateRef.current && scheduleUpdateRef.current();
+    }
+  });
 
   // Reset Downshift refs on every render
   itemIndexRef.current = 0;
@@ -49,9 +61,16 @@ const Menu = props => {
 
   return (
     <MenuContext.Provider value={{ itemIndexRef }}>
-      <Popper placement={popperPlacement} modifiers={popperModifiers} eventsEnabled={eventsEnabled}>
+      <Popper
+        placement={popperPlacement}
+        modifiers={popperModifiers}
+        // Disable position updating on scroll events while menu is closed
+        eventsEnabled={isOpen && eventsEnabled}
+      >
         {({ ref, style, scheduleUpdate, placement: currentPlacement }) => {
           let computedStyle = menuStyle;
+
+          scheduleUpdateRef.current = scheduleUpdate;
 
           // Calculate custom width if ref is provided from Select or Autocomplete
           if (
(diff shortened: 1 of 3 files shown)


packages/dropdowns/src/Menu/Menu.js
```diff
@@ -37,6 +37,18 @@ const Menu = props => {
...
+
+  useEffect(() => {
+    /**
+     * Recalculate popper placement while open to allow animations to complete.
+     * This must be ran every render to allow for the number of items to change
+     * and still be placed correctly.
+     **/
+    if (isOpen) {
+      scheduleUpdateRef.current && scheduleUpdateRef.current();
+    }
```
sunesimonsen: Either:

```js
if (isOpen && scheduleUpdateRef.current) {
  scheduleUpdateRef.current();
}
```

or 

```js
isOpen && scheduleUpdateRef.current && scheduleUpdateRef.current();
```

packages/dropdowns/src/Menu/Menu.js
```diff
@@ -49,10 +61,17 @@ const Menu = props => {
 
   return (
     <MenuContext.Provider value={{ itemIndexRef }}>
-      <Popper placement={popperPlacement} modifiers={popperModifiers} eventsEnabled={eventsEnabled}>
+      <Popper
+        placement={popperPlacement}
+        modifiers={popperModifiers}
+        // Disable position updating on scroll events while menu is closed
+        eventsEnabled={isOpen ? eventsEnabled : false}
```
sunesimonsen:
```suggestion
        eventsEnabled={isOpen && eventsEnabled}
```

Notice this statement is not exactly the same, as it might return `undefined`, but if eventsEnabled is `Proptype.bool` it should be fine.

packages/dropdowns/src/Menu/Menu.js
sunesimonsen: Mhh thinking about it I think we should keep it the way it is, that will allow more code to be added to this if statement later.

packages/dropdowns/src/Menu/Menu.js
```diff
@@ -64,9 +83,7 @@ const Menu = props => {
             };
           }
 
-          if (isOpen) {
-            scheduleUpdate();
```
ryanseddon: So the horizontal display issue is happening as this has been removed now. 

The `computedStyles` below gives popper a smaller div to position that doesn't reflect the actual size of the menu and its contents, scheduleUpdate in popper uses rAF to get the coords before paint but now that we use `useEffect` it triggers after paint/layout so the first time the menu is opened it's in the wrong position, after that it's fine.

If we change the `computedStyles` to be:

```js
computedStyles = {
  visibility: 'hidden',
  position: 'absolute'
}
```

It gets in a better position but is still off horizontally on first render.

ryanseddon: Actually if we just drop the `position: absolute` this works fine now.

austin94: If we only leave `visibility: hidden` then the invisible menu will block interactions of content below it when closed. This also removes the animation.

I'll work on this some more locally.

austin94: @jzempel and I were able to resolve this locally. The trick was to apply this positioning against the element that popper is positioning against, rather than the `Menu` directly.

I also apply the `animate` prop (`is-open` className) on open to enable the animations start during initial opening.

PR #329, opened 2019-05-08
Upgrade tables, grid, tooltips, and buttons to React 16 and styled-components v4

austin94:
## Description

Similar to #327 this migrates several packages to the `react-16` branch and is split by commit.

This also includes some `useState` hooks to simplify some basic state. I've specifically not moved to the `useKeyboardFocus` hooks here since that's more of a re-write than a conversion.

### DIFF
--- a/packages/buttons/src/views/Anchor.js
+++ b/packages/buttons/src/views/Anchor.js
@@ -16,22 +16,21 @@ import NewWindowIcon from '@zendeskgarden/svg-icons/src/12/new-window-stroke.svg
 
 const COMPONENT_ID = 'buttons.anchor';
 
-export const StyledAnchor = styled.a.attrs({
+export const StyledAnchor = styled.a.attrs(props => ({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  className: ({ danger, disabled, focused, hovered, active, selected }) =>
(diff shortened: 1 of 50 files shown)


packages/tooltips/src/containers/TooltipContainer.js
```diff
@@ -235,13 +230,13 @@ class TooltipContainer extends ControlledComponent {
               );
 
               if (appendToBody) {
-                return <Portal>{tooltip}</Portal>;
+                return ReactDOM.createPortal(tooltip, document.body);
```
ryanseddon: You could just destructure the import so you only need `createPortal`

packages/buttons/src/views/Anchor.js
```diff
@@ -60,9 +59,9 @@ export const StyledNewWindowIcon = styled(NewWindowIcon)`
 /**
  * Accepts all `<a>` props
  */
-const Anchor = props => {
-  // eslint-disable-next-line
-  const { focused, external, children, theme, ...other } = props;
+const Anchor = React.forwardRef((props, ref) => {
+  // eslint-disable-next-line no-unused-vars
+  const { focused, external, children, theme, ...other } = props; // eslint-disable-line react/prop-types
```
ryanseddon: Why is theme destructured here?

austin94: Great question. Not sure why it was like this. I've removed it.

PR #336, opened 2019-05-13
Upgrade tags, tabs, selection, notifications, and pagination to React 16

austin94:
## Description

More React 16 upgrades

### DIFF
--- a/packages/pagination/src/elements/Pagination.spec.js
+++ b/packages/pagination/src/elements/Pagination.spec.js
@@ -6,14 +6,10 @@
  */
 
 import React from 'react';
-import { mountWithTheme } from '@zendeskgarden/react-testing';
+import { render, fireEvent } from 'garden-test-utils';
 import { KEY_CODES } from '@zendeskgarden/react-selection';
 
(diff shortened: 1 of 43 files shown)


packages/pagination/src/elements/Pagination.spec.js
```diff
@@ -69,184 +65,172 @@ describe('Pagination', () => {
...
-      wrapper
-        .find('ul')
-        .children()
-        .forEach(child => {
-          expect(child).toHaveProp('data-test-id', CUSTOM_PROP_VALUE);
-        });
+      for (let index = 0; index < container.firstChild.children.length; index++) {
+        const child = container.firstChild.children[index];
+
+        expect(child).toHaveAttribute('data-test-id', CUSTOM_PROP_VALUE);
```
ryanseddon: You could just spread the children into an array and keep the similar forEach approach

```js
[...container.firstChild.children].forEach(child => {
  expect(child).toHaveAttribute('data-test-id', CUSTOM_PROP_VALUE);
});
```

PR #337, opened 2019-05-15
Upgrade menus, dropdowns, modals, and autocomplete

austin94:
## Description

Here is the last grouping of packages for the React 16 upgrade 🎉 

This also includes the removal on enzyme related utilities in `react-testing`


packages/testing/src/index.js
```diff
@@ -5,7 +5,4 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-export { default as mountWithTheme } from './utils/mountWithTheme.js';
-export { default as renderWithTheme } from './utils/renderWithTheme.js';
-export { default as shallowWithTheme } from './utils/shallowWithTheme.js';
 export { default as getExports } from './utils/getExports.js';
```
jzempel: Does it make sense to move this one to `react-utilities` and dump the testing package?

packages/modals/src/containers/FocusJailContainer.spec.js
```diff
@@ -130,11 +128,11 @@ describe('FocusJailContainer', () => {
...
               </div>
             )}
-          </FocusJailContainer>,
-          { enzymeOptions: { attachTo: document.body } }
+          </FocusJailContainer>
         );
+
         focusElementSpy.mockClear();
-        wrapper.simulate('keydown', { keyCode: KEY_CODES.TAB });
+        fireEvent.keyDown(container.firstChild, { keyCode: KEY_CODES.TAB });
```
ryanseddon: You could do `getByTestId('container')` to more explicit about which item the event is firing on.

packages/autocomplete/src/elements/Autocomplete.spec.js
```diff
@@ -6,118 +6,104 @@
...
+    expect(getByTestId('label').textContent).toBe(label);
   });
 
   it('renders aria-label if provided', () => {
     const ariaLabel = 'Test Aria Label';
-    const wrapper = mountWithTheme(<Autocomplete aria-label={ariaLabel} options={[]} />);
+    const { container } = render(<Autocomplete aria-label={ariaLabel} options={[]} />);
 
-    expect(wrapper.find(FauxInput)).toHaveProp('aria-label', ariaLabel);
+    expect(container.firstChild.firstChild).toHaveAttribute('aria-label', ariaLabel);
```
ryanseddon: since it's just DOM you could do `container.querySelector('label')`

packages/testing/src/index.js
austin94: This is really custom to just us. I would vote to move this to our `garden-test-utils` module mapping along with `react-testing-library`.

packages/autocomplete/src/elements/Autocomplete.spec.js
austin94: This is actually testing for an `aria-label` value so there is no semantic `<label>` to query for. This is the same way we get to the `<FauxInput>` in other tests.

packages/testing/src/index.js
jzempel: Yes, even better.

austin94: I have moved this utility and removed the `react-testing` package 🎉

PR #345, opened 2019-05-28
fix(build): remove duplicate classNames from webpack build

austin94:
## Description

@louisscruz and @bmson were able to find some CSS collisions between our `react-menus` and `react-dropdowns` package.

Depending on how CSS is handled within a consuming application this may or may not be affecting all consumers at this point.

Due to our split lerna build we have to add some more advanced options to generate a unique hash for CSS modules to be unique.

## Detail

Our current webpack config includes the follow CSS-modules customization:

```js
{
        test: /\.css?$/u,
        include: /@zendeskgarden\/css/u,
        use: [
          MiniCssExtractPlugin.loader,
          {
            loader: 'css-loader',
            options: {
              modules: true,
              localIdentName: '[name]__[local]___[hash:base64:5]'
            }
          }
        ]
      }
```

The `localIdentName`, specifically `hash:base64:5` was being duplicated. 
This has been creating common hashes in the computed CSS.

After looking through the [loader-utils](https://github.com/webpack/loader-utils#interpolatename) provided by Webpack the `hash:base64:5` is calculated by the import structure and module name.

https://github.com/webpack-contrib/css-loader/issues/413 was the fix for this. By adding `context: __dirname` this helps create a unique hash for each package.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/utils/build/webpack.base.js
+++ b/utils/build/webpack.base.js
@@ -93,6 +93,7 @@ found at http://www.apache.org/licenses/LICENSE-2.0
             loader: 'css-loader',
             options: {
               modules: true,
+              context: __dirname, // Solves CSS name clashes in multiple packages https://github.com/webpack-contrib/css-loader/issues/413
               localIdentName: '[name]__[local]___[hash:base64:5]'
             }
           }


austin94:
Note: this will require a force publish `PATCH` for every package to make sure that all CSS is re-generated.

PR #351, opened 2019-06-03
chore(forms): add deprecation notices to checkboxes, radios, textfields, and toggles

austin94:
## Description

This PR adds deprecation warnings to the root import and documentation of `react-checkboxes`, `react-radios`, `react-textfields`, and `react-toggles`.

## Checklist

      designer as a reviewer)~
      component~
- [ ]~ :guardsman: includes new unit tests~

### DIFF
--- a/packages/checkboxes/README.md
+++ b/packages/checkboxes/README.md
@@ -3,6 +3,13 @@
 This package includes components relating to checkboxes in the
 [Garden Design System](https://zendeskgarden.github.io/).
 
+## DEPRECATION WARNING
+
+This package has been deprecated in favor of the API provided in the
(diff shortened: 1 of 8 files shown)


sunesimonsen:
I think we should definitely remove these packages, but is there any reason for removing them from the repository so quickly. It could take time for users to migrate off the existing packages. It could be that they are caught on React 15, or something else that makes it is headache to move forward this quickly. They can of cause use the old released packages, but what if a bugfix if required. I do understand that the maintenance burden is growing, but you kind of called that on yourself.  I think we should keep the existing packages around for a couple of months before we remove them from the repository, but that is just my feeling (as a user).

Notice I don't know when you are planing on removing the implementation.

PR #352, opened 2019-06-04
BREAKING feat(avatars): upgrade Avatars to new badge styling

austin94:
- [x] **BREAKING CHANGE**

## Description

This PR implements the new breaking changes to match in https://garden.zendesk.com/css-components/avatars/

It is a breaking change

## Checklist

      designer as a reviewer)
      component


packages/avatars/src/Avatar.example.md
```diff
@@ -1,22 +1,50 @@
-For accessibility ensure that all children include an `alt` description.
+For accessibility ensure that all children images include an `alt` description.
+
+For implementations that use the `badge` prop ensure that information is ready
```
jzempel: *read

packages/avatars/src/Avatar.js
```diff
@@ -5,50 +5,65 @@
...
+      {children}
+      <FormattedBadge />
+    </StyledAvatar>
+  );
+};
 
 Avatar.propTypes = {
   /** Applies system styling */
-  system: PropTypes.bool,
+  isSystem: PropTypes.bool,
```
jzempel: is it really important to change this property name here? If so, then it will need to be noted in the BREAKING CHANGE commit merge message.

austin94: For boolean props we need to watch out for valid attribute names. We haven't done this in the past. If we follow the `isFoo` pattern consistently we can avoid this.

We should probably do this system wide as we move to CSS-in-JS, but I thought this would be a good place to start.

packages/avatars/src/Avatar.example.md
```diff
@@ -26,82 +54,93 @@ For accessibility ensure that all children include an `alt` description.
...
+const StyledTextAvatar = styled(Avatar)`
+  background-color: ${zdColorSecondaryAzure600};
+`;
+
+<Grid>
   <Row>
     <Col md={4}>
-      <Avatar size="small">
-        <img src="images/jason.png" alt="Example Avatar" />
+      <Avatar status="active" badge="9+">
```
jzempel: Oops, all statuses throughout this example need to either be `away` or `available` – no `active`.

packages/avatars/src/Avatar.js
```diff
@@ -5,50 +5,65 @@
...
+    computedStatus = STATUS.ACTIVE;
+  }
+
+  const FormattedBadge = () => {
+    if (status === undefined) {
+      return null;
+    }
+
+    if (computedStatus === STATUS.AWAY) {
+      return <StyledBadge />;
```
jzempel: The `away` status should not be rendering a badge `figcaption` at all.

packages/avatars/src/Avatar.js
```diff
@@ -5,50 +5,65 @@
...
-`;
+const Avatar = ({ isSystem, size, status, children, badge, ...other }) => {
+  let computedStatus = status;
+
+  if (status === STATUS.AVAILABLE && badge !== undefined) {
+    computedStatus = STATUS.ACTIVE;
+  }
+
+  const FormattedBadge = () => {
+    if (status === undefined) {
```
jzempel: Logic still isn't right. Change a status on the example page from `available` to `away` – the figcaption should not render.

austin94: Oh I was misunderstanding the requirements for this one. We should be good now 😄 

I've also added an additional test case to test this logic.

PR #353, opened 2019-06-07
chore(lerna): update versioning flow

austin94:
## Description

This PR modifiers our versioning process to handle the following:

- Allow a publisher to select a new semantic version for all packages
- Generate a consolidated changelog based on pull requests that contain customizable PR labels
- Allow modification of this generated changelog diff
- Publish this changelog diff as a `draft` Github release
  - This gives us a chance to "publish" the release when necessary
  - Since releases trigger notifications this is the "standard" way of handling it 🤷‍♂ 
- Insert the changelog diff into the root CHANGELOG and publish it to master
  - This step uses `[skip ci]` to avoid a Travis run

## Detail

There is a new `version.sh` script and some related utilities. I have also pinned our version in lerna to `v5.0.0` to allow the first major bump to be v6!

## TODO
- It will still be useful to remove some of the semantic formatting in our commit messages, but I would like to wait to add that as it will add some complexity to this process. We should be able to add another node script into `version.sh` to handle this formatting


lerna.json
```diff
@@ -1,6 +1,18 @@
 {
+  "version": "5.0.0",
   "npmClient": "yarn",
-  "version": "independent",
+  "npmClientArgs": ["--no-lockfile"],
```
jzempel: Is this desired? Why adding now?

renovate.json
```diff
@@ -27,5 +27,6 @@
       }
     }
   ],
-  "reviewers": ["ryanseddon", "jzempel", "austin94"]
+  "labels": ["PR: Internal :bouquet:"],
+  "reviewers": ["zendeskgarden/maintainers"]
```
jzempel: Does this work now? In the past, renovate was unable to notify groups (due to limited API permissions).

renovate.json
```diff
@@ -27,5 +27,6 @@
       }
     }
   ],
-  "reviewers": ["ryanseddon", "jzempel", "austin94"]
+  "labels": ["PR: Internal :bouquet:"],
```
jzempel: `:seedling:`

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
+
+> **Tags:**
+> - :boom: Breaking Change
+> - :rocket: New Feature
+> - :bug: Bug Fix
+> - :memo: Documentation
+> - :bouquet: Internal
```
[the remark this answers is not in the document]
austin94: I've also updated the PR label to match

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+
+_Note: Gaps between patch versions are faulty, broken or test releases._
+
+See [individual changelogs](https://github.com/zendeskgarden/react-components/tree/930d97a0c923a98aa821a475c2557ceda22bec2c)
+for pre `v6.x` changes.
+
+<!-- DO NOT MODIFY BELOW THIS COMMENT -->
+<!-- insert-new-changelog-here -->
```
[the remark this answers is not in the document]
austin94: I will just add this into the v6 release changelog

lerna.json
austin94: I noticed this in the babel `lerna.json` since we are .gitignoring these already this is equivalent.

renovate.json
austin94: > In the past, renovate was unable to notify groups

Didn't know this was an issue. I will revert.

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+> **Tags:**
+> - :boom: Breaking Change
+> - :rocket: New Feature
+> - :bug: Bug Fix
+> - :memo: Documentation
+> - :bouquet: Internal
+
+_Note: Gaps between patch versions are faulty, broken or test releases._
+
+See [individual changelogs](https://github.com/zendeskgarden/react-components/tree/930d97a0c923a98aa821a475c2557ceda22bec2c)
```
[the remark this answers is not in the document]
austin94: I will add the individual changelogs to the v6 release

utils/scripts/github-release.js
```diff
@@ -0,0 +1,48 @@
...
+async function githubRelease() {
+  const changelog = await getStdin();
+
+  const response = await fetch(
+    'https://api.github.com/repos/zendeskgarden/react-components/releases',
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `token ${process.env.GITHUB_AUTH}`
```
jzempel: Let's align on `$(git config github.token)` per long-standing practice. See https://github.blog/2008-10-11-local-github-config/.

utils/scripts/version.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+set -e
+
+if [[ -z "$GITHUB_AUTH" ]]; then
```
jzempel: `$(git config github.token)`

PR #354, opened 2019-06-08
fix: update all react-theming peerDependencies

austin94:
## Description

Update all peer dependencies for React 16 move.


austin94:
Going to merge this as it also fixes the build on master.

jzempel:
@austin94 I meant to ask... Is this a fundamental problem with `lerna` or is there something we need to do with `yarn run tag` in order to avoid this kind of thing with every future major version update?

austin94:
Peer dependencies are a weird beast https://github.com/lerna/lerna/issues/375

I'm not sure this will be able to be automated even if lerna added the functionality. There is the possibility that we could only support the new major `^v7.0.0` or a package, or a range `^v6.0.0 | v7.0.0`. There is also the case where we may use only a specific feature of a library like styled-components and want to pin it at a minimum of that minor.

Since there is a lot of business logic that determines the most optimum range I think it would have to remain manual.

Issue #365, opened 2019-06-13
`Icon` component does not allow styling via the `className` prop


austin94:
## Expectations

I should be able to wrap https://github.com/zendeskgarden/react-components/blob/master/packages/buttons/src/views/icon-button/Icon.js with styled-components and style the element.

## Reality

The `Icon` component does not allow styling via the `className` prop. This should be allowed.
### Fine Print

- Component: react-buttons
- Version: v5 & v6
- Browsers: all

PR #366, opened 2019-06-14
fix(buttons): Pass className to <Icon />

giacomorebonato:
## Description

Hi, this is my first Garden PR. Happy to learn and to adjust anything :)
Passes down the className prop to the <Icon /> component, for using it with styled components, fela and other CSS in JS libraries.

## Detail

https://github.com/zendeskgarden/react-components/issues/365

### DIFF
--- a/packages/buttons/src/views/icon-button/Icon.spec.js
+++ b/packages/buttons/src/views/icon-button/Icon.spec.js
@@ -29,4 +29,14 @@ describe('Icon', () => {
 
     expect(container.firstChild).toHaveClass('is-rotated');
   });
+
+  it('passes className prop', () => {
+    const { container } = render(
+      <Icon className="test">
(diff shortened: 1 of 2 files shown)


packages/buttons/src/views/icon-button/Icon.spec.js
```diff
@@ -29,4 +29,14 @@ describe('Icon', () => {
...
   });
+
+  it('passes className prop', () => {
+    const { container } = render(
+      <Icon className="test">
+        <svg />
+      </Icon>
+    );
+
+    expect(container.firstChild.classList[1]).toBe('test');
```
austin94: We have a jest matcher to help with these class assertions. Maybe give this a try:

```suggestion
    expect(container.firstChild).toHaveClass('test');
```

PR #370, opened 2019-06-20
feat(loaders): Replace ScheduleContainer with useSchedule hook

### LINKED ISSUE #147: Rethink loading icon speed API

ryanseddon:
- [x] **BREAKING CHANGE?** 

## Description

Replace `ScheduleContainer` with new shiny `useSchedule` hook. 

This is breaking due to heavy internal changes and removing the `velocity` prop on the `Dots` loader in favour of the aligned `duration` prop.

## Detail

The new approach of `useSchedule` heavily reduces the code compared to the original.

This PR also fixes #147 by removing the `velocity` prop in favour of `duration` to match the other loaders, this is a breaking change.

- [x] Spinner
- [x] Dots

## Checklist

      designer as a reviewer)~
      component~


packages/loaders/src/Dots.js
```diff
@@ -5,112 +5,86 @@
...
 
-  static defaultProps = {
-    size: 'inherit',
-    color: 'inherit',
-    velocity: 0.05,
-    delayMS: 750
-  };
+  useEffect(() => {
+    // gah user agent sniffing! This is so we only run the effect in IE11
+    if (/MSIE \d|Trident.*rv:/u.test(navigator.userAgent) && delayComplete) {
```
ryanseddon: Before you banish me for user agent sniffing here me out, we can support CSS based animations in all browsers except IE11 *but* we can query the computed value of the animation in IE11 and update the transform attribute to enable the animation to work just fine.

This simplifies the code greatly and also allows us to remove the `velocity` prop in favour of the `duration` prop without loss of perf or fps.

If you have a nicer way of detecting svg animation support let me know! I tried listening to `animationiteration` event on mount but that made for a weird first loop in the good browsers where it would do the CSS animation and the attribute transform.

jzempel: @ryanseddon is there something that could be cribbed from this? https://github.com/Modernizr/Modernizr/blob/master/feature-detects/svg/smil.js. i.e. `/SVGAnimate/.test(({}).toString.call(document.createElementNS('http://www.w3.org/2000/svg', 'animate')))`

packages/loaders/src/utils/animations.js
```diff
@@ -1,32 +1,97 @@
+/* stylelint-disable rule-empty-line-before */
```
jzempel: Prefer moving this directly adjacent to the code it is disabling.

ryanseddon: It's the whole file

packages/loaders/src/Dots.js
ryanseddon: Wasn't like I was a core contributor to that library or anything! Good idea I'll test this and see if we're good.

packages/loaders/src/utils/useCSSSVGAnimation.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { useEffect, useState } from 'react';
+
+export const useCSSSVGAnimation = () => {
```
ryanseddon: TL;DR; 

It injects an svg and uses translate on a `rect` to position it, it then uses the `document.elementFromPoint` method to see if the translate actually moved the element and whether css animation will work on svgs.

packages/loaders/src/Dots.js
```diff
@@ -5,112 +5,88 @@
...
+  const dotThree = useRef(null);
 
-  static defaultProps = {
-    size: 'inherit',
-    color: 'inherit',
-    velocity: 0.05,
-    delayMS: 750
-  };
+  useEffect(() => {
+    // gah user agent sniffing! This is so we only run the effect in IE11
```
jzempel: Sad comment no longer needed ;)

packages/loaders/src/utils/animations.js
jzempel: But doesn’t include the copyright header. Hopefully just above the keyframes

packages/loaders/src/utils/useCSSSVGAnimation.js
austin94: Is this logic borrowed from anywhere? Might need an attribution link in the header

packages/loaders/src/utils/useCSSSVGAnimation.js
```diff
@@ -0,0 +1,36 @@
...
+  const [canAnimateSVG, setAnimateSVG] = useState(true);
+
+  useEffect(() => {
```
austin94: Is this truly a side-effect? If this value doesn't change or update it might make more sense as a constant that can be imported than as a hook.

ryanseddon: It needs to mutate the dom and then query it so I'd say yes

packages/loaders/src/utils/useCSSSVGAnimation.js
ryanseddon: Yeah good catch it' adapted from this blog post will link to it in a comment https://eprev.org/2017/01/05/how-to-detect-if-css-transforms-are-supported-on-svg/

jzempel:
Given where we agreed to head with containers, should this just land in the `next` branch?

ryanseddon:
Yeah I'm all for that, I'll switch it to next and then maybe y'all can :+1: it again?

PR #369, opened 2019-06-19
feat(build): introduce Typescript and Storybook to the build process

austin94:
## Description

This PR focuses on adding two new technologies into our stack, [Typescript](https://www.typescriptlang.org/) and [Storybook](https://storybook.js.org/).

### Storybook

We've been talking about this for a while, but with the addition of Typescript, styleguidist is making less sense for our "visual unit test" documentation. Storybook will be able to act as a decent stop-gap until we are able to invest more time into a custom documentation solution.

This storybook config includes:

- `addon-notes` Allows us to include markdown notes with each story
- `addon-knobs` Allows us to include customizations for each story. These can be shared with a custom URL as well.
- `addon-storysource` Show the source of each story
- `addon-actions` Show actions performed within a story
- `addon-a11y` Show axe audits and color-blindness filters in preview
- `addon-contexts` Show custom `RTL` and `LTR` locale dropdown for our `ThemeProvider`

In the webpack config I have enabled the following:

- Our standard babel config
- All TS files are passed through the `babel-loader` to ensure IE11 is supported
- All files are passed through `eslint-loader` and `ts-loader` to show all type and lint errors during the storybook build 

storybookExample

storybook-example

#### Overall thoughts

Being able to write the examples in TS/JS rather than MD is a _much better_ experience. The prop tables and the Docs functionality that is in technical preview may help in the future as well.

### Typescript

The main goal of this addition is to include the `typings/index.d.ts` type declarations for the dropdowns package. This is automatically consumable for TS consumers and will benefit our standard users with an improvement in IDEs and tools that leverage the TS language server (mostly VS Code). 

The changes to introduce this were fairly minimal and included the following:

- Modify `eslint` to support linting TS file
- Prettier already handles TS well
- Add `ts-jest` to allow our jest environment to consume and type-check our tests
- Adding in all of our missing types for storybook, react, prop-types, etc.
- Adding `@babel/preset-typescript` to our standard babel config to allow TS in our standard build steps
- Add TS linting and testing steps to travis

## Try it at home

If you want to try these changes locally I have included some new npm scripts.

- `yarn start:storybook`
  - Starts storybook in development mode
- `yarn build:storybook`
  - Creates a static build of storybook. Added with our standard `build:demo` steps now
- `yarn ts`
  - Type-check all files
- `yarn ts:watch`
  - Type-check all files in watch mode (honestly this isn't needed often due to storybook and jest running typescript in their loaders)

**Example TS failure**

exampleTsFailure

[...]

PR #371, opened 2019-06-20
 feat(dropdowns): migrate react-dropdowns to Typescript

austin94:
## Description

This PR is a redo of #369 😄 

There have been some changes recently in the typescript parser that makes sticking with styleguidist more feasible than before.

For our components it's common for us to spread attributes to their semantic component. This is possible to type, but results in:

```ts
interface IButtonProps extends HTMLProps<HTMLButtonElement> {
  ourCustomProp: boolean;
}
```

For automated prop generation this would show all 1000+ valid props, including our custom ones.

The `react-docgen-typescript` module has added some helpers for this where I can filter out any props coming from `node_module` libraries (like React in this case).

## Detail

Same as #369, but without Storybook.

The first commit is only the TS additions to the build and test infrastructure. The second is migrating dropdowns to TS.

The examples and component/testing logic have been left unchanged in this PR.

## Checklist

      designer as a reviewer)
      component


austin94:
> Can you comment on the Travis script reordering

Yes! This is due to typescript needing the type definitions in the nested packages `node_modules`. Specifically downshift and react-popper for this package.

Moving the `lerna bootstrap` earlier in the process ensures that all the modules (and their types) are available.

utils/build/webpack.base.js
```diff
@@ -68,6 +71,16 @@ found at http://www.apache.org/licenses/LICENSE-2.0
           }
         ]
       },
+      {
+        test: /\.tsx?$/u,
+        exclude: /node_modules/u,
+        use: [
+          {
+            loader: 'babel-loader',
```
fodier: I’m curious @austin94, why do you prefer to use `babel-loader` (with `@babel/preset-typescript`) instead of `ts-loader` (or `ts-loader` + `babel-loader` without `@babel/preset-typescript`)? What would you lose?

And a related question: have you tried to emit declaration files at the same time as when compiling TS?

Glad to see this PR by the way 😃.

austin94: > why do you prefer to use babel-loader (with @babel/preset-typescript) instead of ts-loader (or ts-loader + babel-loader without @babel/preset-typescript)?

This matches the default webpack config that exists in `create-react-app` now.

> have you tried to emit declaration files at the same time as when compiling TS?

This got me thinking about the webpack ordering and I've reconsidered the `ts-loader` vs `@babel/preset-typscript` decision.

I've now updated the build to use `ts-loader -> babel-loader` and emit those declarations during the build step. This should increase the build speed over time as we move more packages over.

packages/dropdowns/src/Menu/Menu.tsx
```diff
@@ -5,21 +5,47 @@
...
-
-export const MenuContext = createContext();
+import {
+  GARDEN_PLACEMENT,
+  getPopperPlacement,
+  getRtlPopperPlacement
+} from '../utils/garden-placements';
+
+export interface IMenuContext {
+  itemIndexRef: React.MutableRefObject<number>;
```
ryanseddon: Could you import `MutableRefObject` like you've done for `CSSProperties` and `HTMLProps`?

utils/styleguide/styleguide.base.config.js
```diff
@@ -156,6 +168,30 @@ const defaultStyleguideConfig = {
           loader: 'babel-loader',
           options: babelOptions
         },
+        {
+          test: /\.tsx?$/u,
+          loaders: [
+            {
+              loader: 'eslint-loader'
```
ryanseddon: Why the separation here for tsx into two different loaders?

austin94: This loader uses `enforce: 'pre'` which I believe has to be enforced at a module rule level. We don't want to enforce that on `ts-loader` and `babel-loader`

PR #373, opened 2019-06-26
feat(datepickers): introduce Datepicker component

austin94:
## Description

Here is a first stab at the `<Datepicker>` component. Some highlights:

- Wraps any input (`<input />` or our `<Input />`)
- Uses [date-fns](https://date-fns.org/) for our date calculations
- Fully localized with `Intl.DateTimeFormat`
  - Localized day/week/month display
  - Localized date formats and parsing
  - Localized monday/sunday start days
- Styled with CSS-in-JS and includes an RTL layout

## Detail

I tried to abstract moment from as much of the API as I could. For this reason the component consumes and returns native `Date()` objects.

The only moment specific props are `dateFormat` and `locale` as they use the localization formats provided by moment.

By default the component also "non-strict" parsing of dates. This means that `Jan 3, 2019`, `January 03, 2019`, and `01/03/2019` are all valid dates. This is customizable, but the "non-strict" default should provide a more flexible UX.

## Accessibility

Since users have access to input while the datepicker is open there are intentionally no a11y messages or keyboard navigation.

When we eventually include ranges into the datepicker this logic would begin to fall apart so I have not included any keyboard navigation.

### Localization

Localization

## Checklist

      designer as a reviewer)
      component~


babel.config.js
```diff
@@ -20,5 +20,10 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     '@babel/proposal-object-rest-spread',
     'babel-plugin-styled-components'
-  ]
+  ],
```
austin94: With us using `data-test-*` IDs more in our components we should probably remove these attributes in our production bundles.

utils/build/webpack.base.js
```diff
@@ -67,7 +67,7 @@ found at http://www.apache.org/licenses/LICENSE-2.0
         use: [
           {
             loader: 'babel-loader',
-            options: babelOptions
+            options: { ...babelOptions, envName: 'production' }
```
austin94: Turns out setting `mode: production` in webpack doesn't set the global `NODE_ENV` variable to `production`. This change tells the babel-loader to use our `production` environment to remove the `data-test-*` attributes.

packages/dropdowns/package.json
```diff
@@ -23,6 +23,8 @@
     "@zendeskgarden/container-selection": "^0.2.3",
     "classnames": "^2.2.5",
     "downshift": "^3.2.7",
+    "moment": "^2.24.0",
```
sunesimonsen: I worry about the date picker and therefore moment being included in vendor bundles even if it is not referenced. 

Moment is huge https://bundlephobia.com/result?p=moment@2.24.0 especially with locale support on top. I would much prefer a new package that build on top of react-dropdowns that was focused around handling calendar/date/time.

packages/dropdowns/package.json
```diff
@@ -23,6 +23,8 @@
...
+    "polished": "^3.4.1",
```
sunesimonsen: This is by the way not small either https://bundlephobia.com/result?p=polished@3.4.1

utils/build/declarations.d.ts
```diff
@@ -24,3 +49,148 @@ declare module '@zendeskgarden/react-grid';
...
+  export const zdLineHeightXxl: string;
+  export const zdLineHeightXxxl: string;
+  export const zdSpacingXxs: string;
+  export const zdSpacingXs: string;
+  export const zdSpacingSm: string;
+  export const zdSpacing: string;
+  export const zdSpacingLg: string;
+  export const zdSpacingXl: string;
+  export const zdSpacingXxl: string;
+}
```
sunesimonsen: 😢

sunesimonsen:
This looks like a fantastic amount of effort, but I think it belongs in a new package.

packages/dropdowns/package.json
austin94: The original plan to include this with dropdowns was to reduce the amount of CSS bloat for the dropdown related styling. `react-popper` and the menu positioning/animation styling would be duplicated.

Is that the correct bundle size? That looks like the size of Moment.js with _all_ locales included. Minified it should be around 16KB which is still hefty.

sunesimonsen:
Depending on the amount of moment usage that is also https://date-fns.org/ which is much smaller.

packages/dropdowns/package.json
sunesimonsen: You might be right, about the size - it would be great if you check it.

I think it would be great if getting a styled react-popper would just be something this package would offer to other packages, as that seems to be something that would happen multiple times. That was exactly what the Popover was in the old components.

I just think that it is a mistake to mix these things, as I see the amount of calendar related functionality grow a lot.

sunesimonsen:
It might be that the localisation team has a say in how we format dates. And if you think about this in an open source context, I'm not sure that providing localised date formatting in the component reasonable. I would tent to lean on Intl by default and provide a hook.

I know I'm being a pain in the ass, but I guess that is my role. You can chose how much of it you want to listen to :-)

packages/dropdowns/package.json
sunesimonsen: Screenshot 2019-06-27 at 23 06 42

You need this as far as I can tell:

```
plugins: [
  new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)
]
```

https://github.com/moment/moment/issues/2416

packages/dropdowns/package.json
austin94: I've updated my import structure to have less of a bundle impact 

```js
import rgba from 'polished/lib/color/rgba';
```

This is already a common dependency with `react-loaders` so it shouldn't be a bundle hit.

utils/build/declarations.d.ts
austin94: This isn't needed for the code to compile, but with creating some of these CSS-in-JS areas it helped a lot. Once we include these values in `react-theming` it will no longer be necessary.

packages/dropdowns/package.json
jzempel: `polished` is going to be ubiquitous throughout `react-components` as we continue forward with our CSS-in-JS efforts. Not sure if it will make sense as a peer dependency?

packages/dropdowns/package.json
austin94: Resolving with the latest push.

austin94:
@zendeskgarden/maintainers I've worked this PR over some and force-pushed. Overall the visual styling and logic are the same as previous, but with these difference:

- This is now in a new `react-datepickers` package
  - Even though this leads to some duplicated styling and placement logic from `react-dropdowns` this is probably worth it due to the difference in the API
- I have migrated from `momentjs` to `date-fns@next`
  - `next` is their v2 implementation which includes the improved parsing and localization that we need
  - I'm normally skeptical of a pre-release version, but several other datepickers are using it so we should be ok
- Reformatted the documentation now that it's in it's own package

A great side-effect of using `date-fns` is that it locks us into the [Unicdoe TR35 date field symbols](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table) rather than moments custom formats.

@sunesimonsen I've also used the esm import structures for polished and date-fns so the size increase should be minimal.

sunesimonsen:
@austin94 thank you so much for taking my feedback into consideration. I'll take a look at it tonight.

packages/datepickers/src/examples/localization.md
```diff
@@ -0,0 +1,92 @@
...
+  Label: DropdownLabel,
+  Field: DropdownField,
+  Hint: DropdownHint
+} = require('@zendeskgarden/react-dropdowns/src');
+const arLocale = require('date-fns/locale/ar-SA');
+const enLocale = require('date-fns/locale/en-US');
+const enGbLocale = require('date-fns/locale/en-GB');
+const esLocale = require('date-fns/locale/es');
+const jaLocale = require('date-fns/locale/ja');
+const zhLocale = require('date-fns/locale/zh-CN');
```
sunesimonsen: I think it is really neat that the locales can just be imported this way. That means I can dynamically import the locale I need.

sunesimonsen:
I didn't look it through in details, but conceptually it seems really solid and close to done. The only thing I noticed was the localisation:

Screenshot 2019-06-29 at 20 07 15

Another way of formatting numbers and dates is with Intl (supported in most browsers).

Screenshot 2019-06-29 at 20 13 49
Screenshot 2019-06-29 at 20 12 27

I don't if that would be useful in any way, just wanted to mention it.

ryanseddon:
Can you update the initial readme to reflect the current state, still says you're using momentjs.

I know a few teams already use momentjs over date-fns mainly due to the fact that date-fns wasn't around or it didn't quite yet cover all the locales needed. Could we design this in a way that could work with either or library for handling dates?

packages/datepickers/src/utils/reducer.ts
```diff
@@ -0,0 +1,120 @@
...
+  value,
+  dateFormat,
+  locale
+}: {
+  value?: Date;
+  dateFormat: string;
+  locale: any;
+}) => (state: IDatepickerState, action: DatepickerAction): IDatepickerState => {
+  switch (action.type) {
+    case 'open':
```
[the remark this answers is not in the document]
austin94: Fixed!

packages/datepickers/src/Datepicker.tsx
```diff
@@ -0,0 +1,342 @@
...
+function parseInputValue({
+  inputValue,
+  locale,
+  dateFormat
+}: {
+  inputValue: string;
+  locale?: Locale;
+  dateFormat: string;
+}): Date {
+  const MINIMUM_DATE = new Date(1000, 0, 0);
```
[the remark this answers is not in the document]
austin94: With non-strict parsing the year 219 is valid as 0219 in certain locales. This is just a bottom bound to stop the picker from going biblical on people.

packages/datepickers/src/Datepicker.tsx
```diff
@@ -0,0 +1,342 @@
...
+    <DatepickerContext.Provider
+      value={{
+        state,
+        dispatch
+      }}
+    >
+      <Manager>
+        <Reference>
+          {({ ref }) => {
+            return React.cloneElement(React.Children.only(children as any), {
```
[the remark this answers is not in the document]
austin94: Is this true with `cloneElement`? I'm able to pass props and event through without issue.

The tests currently apply a `data-test-id` in this manner.

austin94:
> Could we design this in a way that could work with either or library for handling dates

@ryanseddon I've re-worked this some more. This component now:

- Uses `date-fns` for date calculation logic only. This is all internal usage so a consumer doesn't have to know or understand this.
- Uses the `Int.DateTimeFormat` utility for all formatting and display localization
  - This works down to IE11 and supports our locales better than `date-fns` with no bundle increase
  - There is no concept of "start of week" in the current Intl spec. I've created a utility that provides this info for our most used locales and regions. Otherwise it defaults to a Sunday start date.
- Date parsing is now using the `Date.parse()` method
  - This doesn't work well with localization, but is customizable with the `customParseDate()` prop
  - This allows consumers to use momentjs or date-fns for their parsing if desired

sunesimonsen:
@austin94 that is awesome, I think that strikes the right balance between bundle size, customizability and out of the box behavior 👌

austin94:
> Ideally, we'd see something very close to https://garden.zendesk.com/react-components/dropdowns/#autocomplete-usage when clicking into the field or on the associated label.

This was made easier in `react-dropdowns` because we had a reference to the containing boundary around the label, hint, message, and dropdown. This was because there is shared a11y between all of these elements.

For the datepicker it's only worry is the `input` so our standard comparison of checking `e.target` isn't possible here.  I've just pushed some changes which include some event massaging between `onMouseDown -> onMouseUp -> onClick`. It's gross, but with the event ordering from associated labels this might be the cleanest approach.

packages/datepickers/src/Datepicker.tsx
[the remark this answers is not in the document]
ryanseddon: Hmm maybe it was just me was testing on the pre-published site but cloning element should copy all the props so maybe ignore me.

packages/datepickers/src/Datepicker.tsx
```diff
@@ -0,0 +1,318 @@
...
+  /** Display an arrow to the reference element */
+  arrow?: boolean;
+  /**
+   * Passes options to [Popper.JS Instance](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#new-popperreference-popper-options)
+   */
+  popperModifiers?: any;
+  /**
+   * Show open animations
+   */
+  animate?: boolean;
```
jzempel: Should this show a `defaultProps` of `true`?

packages/datepickers/README.md
```diff
@@ -0,0 +1,45 @@
...
+npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
+```
+
+## Usage
+
+```jsx static
+/**
+ * Include datepickers styling at the root of your application
+ */
+import '@zendeskgarden/react-datepickers/dist/styles.css';
```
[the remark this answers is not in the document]
jzempel: Oh, still needed for menus. 😐

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  ${boldedStyling};
+
+  ${props => retrieveTheme('datepickers.header_label', props)};
+`;
+
+export const StyledCalendar = styled.div<{ isSmall?: boolean }>`
+  width: calc(${props => (props.isSmall ? zdSpacingLg : zdSpacingXl)} * 7);
```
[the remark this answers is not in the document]
austin94: Didn't realize this existed! I've brought it over

packages/datepickers/src/examples/min-max-dates.md
```diff
@@ -0,0 +1,36 @@
...
+  year: 'numeric'
+});
+
+const isInvalid = () => {
+  return isBefore(state.value, subDays(state.minValue, 1));
+};
+
+<Field>
+  <Label>Select a date</Label>
+  <Hint>Minimum date of today</Hint>
```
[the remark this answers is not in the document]
austin94: Added a max date to only allow selection within the next 7 days.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+export const StyledHeader = styled.div`
+  display: flex;
+
+  ${props => retrieveTheme('datepickers.header', props)};
+`;
+
+export const StyledHeaderPaddle = styled.div<{ isSmall: boolean }>`
+  width: ${props => (props.isSmall ? zdSpacingLg : zdSpacingXl)};
+  height: ${props => (props.isSmall ? zdSpacingLg : zdSpacingXl)};
```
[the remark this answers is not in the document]
austin94: I've refactored the sizing logic to a utility, rather than a `css` object. I'm now able to use it in all of the other size-aware components.

packages/datepickers/src/Datepicker.tsx
austin94: It looks like the props-parser for Typescript doesn't take `defaultProps` into account https://github.com/styleguidist/react-docgen-typescript/issues/154

I've added the `@defualt` JSDoc comments to make this visible for now

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  font-size: ${props => (props.small ? zdFontSizeSm : zdFontSizeMd)};
+  font-weight: ${props => (props.isToday && !props.isDisabled ? zdFontWeightBold : 'inherit')};
+  line-height: ${zdLineHeightMd};
+  color: ${props => {
```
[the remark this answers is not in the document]
austin94: I've cleaned up this logic and refactored the `color` and `background-color` logic to their own functions.

TIL you can perform `Refactor` actions within styled-components with the Typescript Language Service in VSCode. 😲

packages/datepickers/src/examples/default-usage.md
```diff
@@ -0,0 +1,27 @@
...
+```jsx
+const { Field, Label, Input } = require('@zendeskgarden/react-forms/src');
+
+initialState = {
+  value: new Date()
+};
+
+<Field>
+  <Label>Select a date</Label>
+  <Datepicker value={state.value} onChange={newDate => setState({ value: newDate })}>
```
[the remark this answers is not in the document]
austin94: > Clicking on a datepicker field should select the entry text for me.

I discussed this with @ginnywood during the implementation and we decided to preserve the "standard" input selection. This allows users to click into a Datepicker and immediately start editing a specific value. If the text was selected on click it would take several keyboard presses to perform an edit.

Example:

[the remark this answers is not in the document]
austin94: Also, if you click the label of the it will select the full-text. Same as a native form control.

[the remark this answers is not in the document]
austin94: > Date entry with no year defaults to 2001. I'd want it to fallback to current year.

This was a side-effect of using the `Date()` parse method. It is unique per browser (Safari defaults to 2000 for example). Since we're already using `date-fns` I've moved this logic to their `parse()` method and am allowing the [P, PP, PPP formats for entry](https://date-fns.org/v2.0.0-beta.2/docs/parse). This should include some additional flexibility.

PR #378, opened 2019-07-02
feat(breadcrumbs): Migrate package to container-breadcrumb

ryanseddon:
- [x] **BREAKING CHANGE?** 

## Description

Replace `BreadcrumbContainer` with `useBreadcrumb` hook.

## Detail

This PR deprecates `BreadcrumbContainer` by adding `console.warn` message in non production environments.

It also removes `BreadcrumbContainer` in `Breadcrumb` element in favour of `useBreadcrumb`.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/packages/breadcrumbs/src/elements/Breadcrumb.js
+++ b/packages/breadcrumbs/src/elements/Breadcrumb.js
@@ -5,11 +5,11 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { Component, Children, cloneElement } from 'react';
+import React, { Children, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import { hasType } from '@zendeskgarden/react-utilities';
+import { useBreadcrumb } from '@zendeskgarden/container-breadcrumb';
 
-import BreadcrumbContainer from '../containers/BreadcrumbContainer';
 import BreadcrumbView from '../views/BreadcrumbView';
 import List from '../views/List';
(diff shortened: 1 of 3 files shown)


packages/breadcrumbs/src/elements/Breadcrumb.js
```diff
@@ -18,42 +18,37 @@ import Item from '../views/Item';
...
+const renderItems = (items, getCurrentPageProps) => {
+  const total = Children.count(items);
+
+  return Children.map(items, (item, index) => {
+    const itemProps = {
+      current: index === total - 1,
+      key: index
+    };
+
+    return hasType(item, Item)
```
austin94: Is there a way for us to start using context rather than our `hasType` utility? If we're going to move to the hooks we should probably upgrade our usage here as well.

ryanseddon: yep was thinking that too, I will investigate and see if it's worth bringing in this PR too.

ryanseddon: So after digging into this I don't think it makes a whole lot of sense to use context. This current code is just checking if the child is already an `<Item />` and just passing it along and if not it wraps the child in an `<Item>`.

austin94: 👍 Sounds good. This is a unique instance since everything is a direct child.

PR #380, opened 2019-07-04
chore: Add deprecation docs and warning to all Container components

ryanseddon:
- [ ] **BREAKING CHANGE?** 

## Description

Before we move to the new container hooks internally we need to deprecate the exiting container components.

## Detail

Right now this warns in non-production environments but it will get a little noisy in tests. ~So I think I might push `!== production||test`~. Pushed in 7475bd7.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/packages/breadcrumbs/src/containers/BreadcrumbContainer.js
+++ b/packages/breadcrumbs/src/containers/BreadcrumbContainer.js
@@ -5,6 +5,16 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
+if (process.env.NODE_ENV !== 'production') {
+  /* eslint-disable no-console */
+  console.warn(
+    'Deprecation Warning: The `@zendeskgarden/react-breadcrumbs` BreadcrumbContainer has been deprecated. ' +
+      'It will be removed in an upcoming major release. Migrate to the ' +
+      '`@zendeskgarden/container-breadcrumb` package to continue receiving updates.'
+  );
+  /* eslint-enable */
+}
+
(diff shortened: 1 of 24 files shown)


ryanseddon:
This will be noisy regardless of direct usage of the containers since all high abstraction components use the containers. Any ideas? 

A possible direction could be to have a flag we look at on `process.env` that will suppress the warnings.

ryanseddon:
@austin94 yes see comment above even if they don't use it directly they'll see for any usage. 

It'd be the same for the dropdown deprecation just a lot more. Not sure how we communicate the deprecation in a less noisy way?

packages/breadcrumbs/src/containers/BreadcrumbContainer.js
```diff
@@ -5,6 +5,16 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
+if (process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test') {
```
austin94: Even though this would be noisy in our environment, I think removing `process.env.NODE_ENV !== 'test'` would be helpful.

Consumers should see these warnings in their tests if they're using a deprecated package.

ryanseddon:
@austin94 so we park this and just cherry pick from this branch as the CSS-in-JS conversion happens?

austin94:
I think that could work best. Otherwise it would be very noisy in the jest tests.

austin94:
Closing as this has been completedl

Issue #174, opened 2018-10-05
Give styled buttons access to keyboardFocused state


aravindet:
## Expectations

The Garden Button displays a focus shadow when it receives focus as a result of a keyboard interaction, but not when it receives focus as a result of a mouse interaction.

This is accomplished using the `KeyboardFocusContainer`, which provides a `keyboardFocused` flag to the render prop.

It is often necessary to add styling to the Garden button, for example to change its background color. In this scenario, it is also necessary to change the color of the focus shadow. One could do this with:

```js
const StyledButton = styled(Button)`
  background-color: red;
  :focus {
    box-shadow: 0 0 0 3px red;
  }
`;
```

but `:focus` does not have the same behaviour the actual button, which adds the `is-focused` class on `keyboardFocused || props.focused`.

There should be a way for styled buttons to override styling for specific button states, without having to reimplement the logic of how it gets into that state.

## Reality

There isn't.

## Suggestion

In Button:

```js
class Button {
  ...
  static classNames = ButtonStyles;
}
```

In StyledButton:

```js
const StyledButton = styled(Button)`
  background-color: red;
  &.${Button.classNames['is-focused']} {
     box-shadow: 0 0 0 3px red;
  }
`;
```

As CSS Components is also an open source project, class names like 'is-focused' are already a documented part of the Garden public API.

This approach is generic enough to handle all such use cases (other states and other components).

###### Fine Print

* Component: Buttons, but maybe others?
* Browsers: all

ryanseddon:
Within your component you can access props, so to accomplish changing the focus style when it's prop based and not css based you can do the following:

```js
const focusRing = 'box-shadow: 0 0 0 3px red;';
const StyledButton = styled(Button)`
  background-color: red;
  :focus {
    ${focusRing}
  }
  ${(props) => props.focused && focusRing;
`;
```

aravindet:
`keyboardFocused` is not a prop. The component structure is:

```
<StyledButton>
  <Button>
    <KeyboardFocusContainer>
      <button></button>
    </KeyboardFocusContainer>
  </Button>
</StyledButton>
```

`keyboardFocused` is an argument that the `KeyboardFocusContainer` passes to its render prop (an anonymous function that returns a `<button>`. I can't intercept this in the StyledButton.

aravindet:
Also, this is part a more generic problem.

[...]

PR #381, opened 2019-07-04
feat(theming): theme baseline for CSS-in-JS

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

Foundational theming work for upcoming CSS-in-JS migration. Key elements of the PR include the following:

* [**BREAKING CHANGE**](https://github.com/zendeskgarden/react-components/pull/381/files#diff-88f886ae93cea50aceea3844633ae74d). re: `retrieveTheme` -> `retrieveComponentStyles`... Arguably, the function was never technically retrieving the theme. However, with the introduction of a full-fledged `theme` object, it becomes important to clarify what this function is actually doing. _This constitutes the majority of the PR diff._
* New `PALETTE` and `DEFAULT_THEME` exports.
* Remove all dependencies on `css-variables` (except for `/example` code, which will probably need to be updated as a follow-on to v7) – replacing with `PALETTE` and `DEFAULT_THEME`. In some cases (i.e. `datepicker` & `typography`) CSS-in-JS component styling will need to be revisited to properly exploit the theming context.
* Collection of styleguidist updates to support a dark mode via `?dark` URL parameter. This will be helpful for ongoing stress testing of our theming object.

## Detail

**View published [theming.next](https://garden.zendesk.com/react-components/next/theming/) page as required pre-reading for PR review.** The page provides information and examples that serve an important backdrop to the updated code. (Contact @jzempel if page needs updating.)

Important CSS-in-JS patterns (or anti-patterns?) to consider and discuss prior to merge include:

* All relevant CSS-in-JS styling will go through the component's `theme` prop for borders, border-radius, spacing, fonts, line-height, etc. Furthermore, all color selections will be resolved via the `getColor` utility as it provides resolution for missing palette hues & shades. See [button CSS-in-JS explorations](https://github.com/zendeskgarden/react-components/blob/e0f9c2f2c038eca71806a3c207dc2878d44cef0d/packages/buttons/src/styled/styles.js#L23-L117) for detail.
* Use of `defaultProps { theme: defaultTheme }`. Now, more than ever, I think it's important that components can perform a viable render outside the context of a `ThemeProvider`. It's one thing to say we need a `ThemeProvider` for RTL. It's another to have failed `theme` object references while attempting to style a component. This approach is seen under the first example of https://www.styled-components.com/docs/advanced#theming. Additionally, the `withTheme` HOC indicates this is [expected](https://github.com/styled-components/styled-components/blob/7fb63b23716b42a31bdeb381abb7c3918dcbe76e/packages/styled-components/src/hoc/withTheme.js#L19). _Currently, we have a problem (failed test) with this approach along with the `renderRtl` utility in `garden-test-utils`._
* Recommend styling to a `.focus-visible` class that will be added to a styled element based on the current `focused` prop. This should improve the experience of extending our styled components and address https://github.com/zendeskgarden/react-components/issues/174. See [button CSS-in-JS exploration](https://github.com/zendeskgarden/react-components/blob/jzempel/css-in-js/packages/buttons/src/styled/StyledButton.js#L32) for details.
* I'm finding that collecting a majority of `size` and `color` styling under respective retrieve functions that receive props and return `css` helpers makes styled components quite a bit cleaner – especially when theming comes into play. See https://github.com/zendeskgarden/react-components/blob/jzempel/css-in-js/packages/buttons/src/styled/StyledButton.js for example.

## Checklist

      designer as a reviewer)
      component~


packages/theming/package.json
```diff
@@ -18,6 +18,10 @@
     "build:demo": "../../utils/scripts/build-demo.sh",
     "start": "../../utils/scripts/start.sh"
   },
+  "dependencies": {
+    "merge-anything": "^2.2.5",
```
jzempel: Used by `styled-components` See https://github.com/styled-components/styled-components/blob/7fb63b23716b42a31bdeb381abb7c3918dcbe76e/packages/styled-components/package.json#L67

packages/theming/src/examples/theme.md
```diff
@@ -0,0 +1,217 @@
...
+    const end = fn.indexOf('\n') - 2;
+
+    retVal = `${fn.substring(start, end)} => expression`;
+  }
+
+  return retVal;
+};
+
+<Well recessed>
+  <StyledTheme>{JSON.stringify(defaultTheme, replacer, '  ')}</StyledTheme>
```
ryanseddon: TIL about the stringify replace function I've also just passed `null` and never considered what it was for.

packages/theming/src/palette/index.js
```diff
@@ -0,0 +1,52 @@
+/**
```
austin94: Would it be possible to combine all the palettes into this file? When the contents of the file are less than the license header it might be too spread out 😄 

Also, you can use:

```js
import azure from './azure'; 
```

for your imports

packages/theming/src/theme/index.js
```diff
@@ -0,0 +1,36 @@
+/**
```
austin94: Similar to `palette`, could we combine this into a single file to help with readability? It will be large, but would be useful to understand the relations between the different values.

packages/theming/src/theme/index.spec.js
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import defaultTheme from '.';
+import palette from '../palette';
+
+describe('defaultTheme', () => {
```
austin94: Would this be a good use-case for a snapshot test?

packages/theming/src/utils/getColor.js
```diff
@@ -0,0 +1,87 @@
...
+ *  - `'__warning'` = `theme.colors.warningHue`
+ *  - `'__success'` = `theme.colors.successHue`
+ *  - `'__neutral'` = `theme.colors.neutralHue`
+ *  - `'__chrome'` = `theme.colors.chromeHue`
+ * @param {number} [props.shade=600] A hue shade.
+ * @param {Object} [props.theme] Context `theme` object.
+ *
+ * @component
+ */
+export default function getColor({ hue, shade = 600, theme } = {}) {
```
austin94: Would it be possible for this function to be used for retrieving the `foreground` and `background` colors? Might need an override without a `shade`, but I think it might be helpful for all color usages to go through this utility.

packages/theming/src/utils/withTheme.js
```diff
@@ -6,8 +6,15 @@
  */
 
 import { withTheme as styledWithTheme } from 'styled-components';
+import { default as theme } from '../theme';
 
 /** @component */
 export default function withTheme(WrappedComponent) {
+  if (WrappedComponent.defaultProps === undefined) {
```
austin94: Would this be best enforced at the component level? There might be use-cases where we provide a tweaked default theme for a specific component.

packages/theming/src/index.js
```diff
@@ -6,7 +6,10 @@
  */
 
 export { default as ThemeProvider } from './ThemeProvider/ThemeProvider';
+export { default as defaultTheme } from './theme';
```
austin94: As the theme and palette are constants, should they be exported as `DEFAULT_THEME` and `PALETTE`

packages/theming/README.md
```diff
@@ -14,43 +14,33 @@ npm install react react-dom prop-types styled-components @zendeskgarden/react-th
```
austin94: Should we remove the `retrieveComponentStyles` ID table in the theming docs? Not sure how useful that will be now that the CSS-in-JS usages won't be found. This would also require removing the logic from `styleguide.base.config.js`

packages/theming/src/ThemeProvider/ThemeProvider.js
```diff
@@ -8,13 +8,14 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { ThemeProvider as StyledThemeProvider } from 'styled-components';
+import merge from 'merge-anything';
+import { default as defaultTheme } from '../theme';
 
 const ThemeProvider = props => {
```
austin94: From our offline discussion about RTL not being propagated...

For us to use `defaultProps` properly we should move `rtl` and `document` into the `theme` object. This is a breaking change, but makes more sense having rtl as a truly themable value.

austin94: This would also allow us to just directly export `ThemeProvider` from `styled-components`.

packages/theming/src/ThemeProvider/ThemeProvider.js
```diff
@@ -8,13 +8,14 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { ThemeProvider as StyledThemeProvider } from 'styled-components';
+import merge from 'merge-anything';
+import { default as defaultTheme } from '../theme';
```
austin94:
```suggestion
import defaultTheme from '../theme';
```

packages/theming/src/ThemeProvider/ThemeProvider.js
```diff
@@ -8,13 +8,14 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { ThemeProvider as StyledThemeProvider } from 'styled-components';
+import merge from 'merge-anything';
```
austin94: Rather than having "magic" merges, would it be best to require users to spread the theme? An RTL example

```jsx
<ThemeProvider theme={{
    ...DEFAULT_THEME,
    rtl: true
}}>
   ...
</ThemeProvider>
```

jzempel: I'm not a huge fan of how this is turning out as I personally feel the `merge` is quite nice for the consumer. However, I'll keep pushing the changes through as it does simplify Garden's surface area. But we'll need to keep an ear out for how folks think this is affecting their use of the API.

packages/theming/src/utils/withTheme.js
jzempel: This just ensures that `withTheme` provides the same information to a Component as we've agreed will be provided to a StyledComponent (either through ThemeProvider context or via defaultProps). The code here would still allow providing

> a tweaked default theme for a specific component

I've added a test showing that `withTheme(<Component theme={theme} />)` works as expected.

PR #394, opened 2019-07-25
Update autocomplete dropdown example due to google translate issue

KaiminHuang:
## Description

The bug is originally captured by rollbar whitin Zendesk, after upgrading dropdown component @zendeskgarden/react-dropdowns": "6.2.1" .

The problem is that Google Translate replaces text nodes with <font> tags containing translations while React keeps references to the text nodes that are no longer in the DOM tree. 

React throws in the following cases:
- A text node is conditionally rendered and it's not the only child of its parent.
- A node before a text node is conditionally rendered

The easiest workaround is to wrap those text nodes with <span> so that nodes referenced by React will stay in the DOM tree even though their contents are replaced with <font> tags.

## Detail

- [Issue reported to React](https://github.com/facebook/react/issues/11538#issuecomment-390386520)

## Checklist
      designer as a reviewer)
      component

### DIFF
--- a/packages/dropdowns/src/examples/autocomplete.md
+++ b/packages/dropdowns/src/examples/autocomplete.md
@@ -2,6 +2,12 @@ The `<Dropdown>` package does not include any filtering logic. The simplest way 
 perform filtering is to control the `inputValue` and `onInputValueChange` props and
 conditionally render `<Item>`s as necessary.
 
+Due to a known issue caused by google translation extension, which replaces text node
+with `<font>` while React keeps references to the text node.
+Any conditionally rendered text node should be wrapped with `<span>`.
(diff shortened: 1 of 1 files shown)


packages/dropdowns/src/examples/autocomplete.md
```diff
@@ -2,6 +2,12 @@ The `<Dropdown>` package does not include any filtering logic. The simplest way
 perform filtering is to control the `inputValue` and `onInputValueChange` props and
 conditionally render `<Item>`s as necessary.
 
+Due to a know issue caused by google translation extension, which replaces text node
```
jzempel: *known

KaiminHuang: 👌

PR #396, opened 2019-07-29
fix(buttons): allow correct ref validation in PropTypes

austin94:
## Description

With the new `React.createRef` usages, some of our old "forward-ref" props have invalid prop-types. An example error:

```
// When supplying React.createRef() to `buttonRef`
Warning: Failed prop type: Invalid prop `buttonRef` of type `object` 
supplied to `IconButton`, expected `function`.
```

## Detail

This PR corrects the invalid prop-types for the `react-buttons` package. There are also other usages of `PropTypes.func` for refs in additional packages, but they are currently deprecated. I am excluding them from this PR.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/packages/buttons/src/views/icon-button/ChevronButton.js
+++ b/packages/buttons/src/views/icon-button/ChevronButton.js
@@ -45,8 +45,6 @@ ChevronButton.propTypes = {
   focusInset: PropTypes.bool,
   hovered: PropTypes.bool,
   active: PropTypes.bool,
-  /** Callback for reference of the native button element */
-  buttonRef: PropTypes.any,
   /** Rotates icon 180 degrees */
   rotated: PropTypes.bool
 };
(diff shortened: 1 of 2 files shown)


packages/buttons/src/views/icon-button/ChevronButton.js
```diff
@@ -46,7 +46,7 @@ ChevronButton.propTypes = {
   hovered: PropTypes.bool,
   active: PropTypes.bool,
   /** Callback for reference of the native button element */
-  buttonRef: PropTypes.func,
+  buttonRef: PropTypes.any,
```
dclowd9901: Is `any` really the appropriate choice here? It should only be an object or a function, right?

```
In the other hand, if you actually want to enforce the type of an actual ref, for example a prop that is forwarded to another component's ref, you have to allow both on object shaped by React.createRef() (The preferred way to create ref in React components, see type in react code here) and function types.

Object shape: PropTypes.shape({ current: PropTypes.instanceOf(Element) }) is for refs created using React.createRef()
PropTypes.func is for callback refs as used in the above question
Which gives the following PropType definition:

forwardRef: PropTypes.oneOfType([
  PropTypes.func, 
  PropTypes.shape({ current: PropTypes.instanceOf(Element) })
])
```

austin94: We have avoided the `shape` prop-type in other packages which is why I went with `any` here.

We ended up removing them as they actually don't do anything in the `v6+` world.

PR #397, opened 2019-07-30
 feat(datepickers): introduce DatepickerRange component

austin94:
## Description

This PR introduces the `DatepickerRange` component. To allow for additional flexibility in input and calendar placement, this component uses the following API:

```jsx
<DatepickerRange>
  <DatepickerRange.Start>
    <input />
  </DatepickerRange.Start>
  <DatepickerRange.End>
    <input />
  </DatepickerRange.End>
  <DatepickerRange.Calendar />
</DatepickerRange>
```

This type of composition allows consumers to determine which type of inputs will be used and how they can be laid out. In the provided example I use our `react-forms` and `react-grid` components for the dynamic layout.

## Checklist

      designer as a reviewer)
      component~


packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -104,10 +107,25 @@ export const StyledCalendar = styled.div<{ isSmall?: boolean }>`
...
 
+export const StyledRangeCalendar = styled.div`
+  display: flex;
+  overflow: auto;
+
+  ${StyledDatepicker} {
+    padding: 0;
+  }
+
+
```
jzempel: delete extra space

austin94:
> The interaction between the ranged date picker and the associated start/end inputs feels off to me

@ginnywood and I sat down and re-worked some of the focus strategy for this interaction. The component now alternates selection based on previous selection. It now also allows `ENTER` key presses to validate the date input.

ryanseddon:
I noticed a weird bug while testing the custom format example, but it turns out you're just missing the `locale` being set for the `dateFormatter` in the `custom-formats.md` example. 

It happens for me as it tries to switch to my locale on close which is dd/mm/yyyy but the example sets the initial `locale` to `en-US` of `mm/dd/yyyy`.

datepicker

packages/datepickers/src/Datepicker/Datepicker.tsx
```diff
@@ -328,7 +328,8 @@ Datepicker.defaultProps = {
   refKey: 'ref',
   animate: true,
   eventsEnabled: true,
-  zIndex: 1000
+  zIndex: 1000,
+  locale: 'en-US'
```
ryanseddon: Should we have a default? Like the issue mentioned in the PR if a user forgets to pass the locale in a custom formatter to match this default it'll have some confusing UX if they don't use the US format.

packages/datepickers/src/DatepickerRange/DatepickerRange.tsx
```diff
@@ -0,0 +1,146 @@
...
+  minValue: PropTypes.instanceOf(Date),
+  maxValue: PropTypes.instanceOf(Date),
+  onChange: PropTypes.func,
+  formatDate: PropTypes.func,
+  customParseDate: PropTypes.func,
+  small: PropTypes.bool
+};
+
+DatepickerRange.defaultProps = {
+  locale: 'en-US',
```
ryanseddon: Same issue as above, I don't think we should have an explicit default

packages/datepickers/src/DatepickerRange/components/Calendar.tsx
```diff
@@ -0,0 +1,65 @@
...
+    startValue,
+    endValue,
+    onChange
+  } = useDatepickerRangeContext();
+
+  return (
+    <StyledRangeCalendar
+      data-garden-id="datepickers.range"
+      data-garden-version={PACKAGE_VERSION}
+      data-test-id="range-calendar"
```
ryanseddon: Do we ship these test-ids to our published packages?

packages/datepickers/src/DatepickerRange/components/Month.tsx
```diff
@@ -0,0 +1,303 @@
...
+            if (!isSelected) {
+              dispatch({ type: 'HOVER_DATE', value: date });
+            }
+          }}
+          data-test-id="day"
+          data-test-previous={isPreviousMonth}
+          data-test-selected={!isInvalidDateRange && isSelected}
+          data-test-disabled={isDisabled}
+          data-test-today={isCurrentDate}
+          data-test-hidden="false"
```
ryanseddon: What's the purpose of all of these for testing only?

packages/datepickers/src/DatepickerRange/utils/datepicker-range-reducer.ts
```diff
@@ -0,0 +1,320 @@
...
+
+      return {
+        ...state,
+        endInputValue,
+        hoverDate: undefined,
+        previewDate
+      };
+    }
+    case 'CLICK_DATE':
+      // console.log(state.isStartFocused, state.isEndFocused);
```
ryanseddon: delete

packages/datepickers/src/DatepickerRange/components/Calendar.tsx
austin94: https://github.com/zendeskgarden/react-components/blob/master/babel.config.js#L26 removes all `data-test*` attributes from the `production` bundle. This should only be visible locally and in our tests.

packages/datepickers/src/DatepickerRange/components/Month.tsx
austin94: With us no longer using `enzyme` we aren't able to assert against props. This allows us to ensure the states are being applied without having to assert against a specific styled-components style.  This is the same process as the `Datepicker` PR.

packages/datepickers/src/DatepickerRange/utils/datepicker-range-reducer.ts
austin94: removed

packages/datepickers/src/Datepicker/Datepicker.tsx
austin94: This originally came up in the `#garden` channel. In some specific locale and browser combinations, `Intl.DateTimeFormat` will use the system locale by default.

This works for formatting, but browsers are only able to natively parse in short, medium, and long `en-US` locales. (There are some slight variations between browsers, but this is mostly consistent).

Without the default we would have the Datepicker showing in `danish` for example, but unable to parse anything until MomentJS or date-fns is provided. By locking it to `en-US` it ensures that the date is always able to be parsed.

PR #398, opened 2019-07-31
fix(buttons): white-space/word-wrap styling in the Anchor component

luis-almeida:
`white-space: pre;` blocks `word-wrap: break-word;` from taking effect in both Chrome and Firefox. 
As an example, the following string in the screenshot is actually "one line" but the text within the link is not wrapping properly:

Screenshot 2019-07-31 at 15 07 20

`white-space: pre-wrap;` seems to work in all browsers.

### DIFF
--- a/packages/buttons/src/views/Anchor.js
+++ b/packages/buttons/src/views/Anchor.js
@@ -35,7 +35,7 @@ export const StyledAnchor = styled.a.attrs(props => ({
     props.external &&
     `
     && {
-      white-space: pre;
+      white-space: pre-wrap;
       word-wrap: break-word;
     }
   `}


luis-almeida:
@jzempel, @austin94 thanks for looking into this PR.

I'm ware that we can fix it on our end. Please do consider addressing this issue though as we had a couple of bugs flagged after switching to garden `Anchor`s component and I think others might do as well.

Here's another example:

| before             |  after |
:-------------------------:|:-------------------------:
Screenshot 2019-08-01 at 13 46 30  |  Screenshot 2019-08-01 at 13 47 06

One can argue that these only have a visual impact but since we support so many different locales this is actually likely to happen with longer links and be worse than an orphaned SVG on a new line:

Screenshot 2019-08-01 at 14 01 06

PR #399, opened 2019-07-31
feat(buttons): CSS-in-JS refactor

### LINKED ISSUE #174: Give styled buttons access to keyboardFocused state

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

- [BREAKING CHANGES](https://github.com/zendeskgarden/react-components/pull/399/files#diff-88f886ae93cea50aceea3844633ae74d)
  - no more `styles.css` dist
  - removed `ButtonGroupView` export
  - removed `Icon` export
- Restructured source
  - `src/components` – high-level package components
  - `src/styled` – styled-components variants (formerly `views`)
- All button CSS contained within components and themable! (no more CSS module dependencies)
  - button styling is enhanced per design requirements /cc @ginnywood 
- Enhanced `getColor` to make shade adjustments to non-palette `hue` values

## Detail

closes #174 via `.focus-visible` class styling that can be targeted for `:focus` CSS overrides (although updated (theming)[https://github.com/zendeskgarden/react-components/pull/381] should be considered instead).

## Checklist

      designer as a reviewer)
      component~


BREAKING_CHANGES.md
```diff
@@ -1,5 +1,16 @@
...
 
+## @zendeskgarden/react-buttons
+
+- no longer packages a `styles.css` dist; CSS is self-contained
+- removed `ButtonGroupView`; use new `SplitButton` component instead
+- removed `Icon`; just use `IconButton` with a `rotated` prop
+
+## @zendeskgarden/react-loaders
+
+- `<Dots />` loader no longer has the `velocity` prop
```
ryanseddon: Do you need to rebase next this should be merged already?

jzempel:
@ryanseddon looks like this PR is weighing-in at about 59.3K vs. 60.3K for the combined (45.8 `index.js` + 14.5 `styles.css`) dist https://unpkg.com/@zendeskgarden/react-buttons/. So not a massive size savings – but with a substantially improved theming experience.

austin94:
> So not a massive size savings – but with a substantially improved theming experience

The size savings will become more visible for areas that are re-exporting our components. There are some areas where this will avoid duplicated CSS sheets.

Also, once we get ESM bundles added the tree-shaking aspect will help some.

utils/test/svg-mock.js
```diff
@@ -5,8 +5,10 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React from 'react';
+import styled from 'styled-components';
 
-const MockSVG = props => <span {...props}>SVG</span>;
+const MockSVG = styled.svg.attrs(() => ({
```
austin94: Why is `styled-components` necessary for this? Could we just use a standard component. Like:

```jsx
const MockSVG = props => <svg xmlns="http://www.w3.org/2000/svg" {...props} />;
```

packages/typography/src/views/lists/OrderedListItem.js
```diff
@@ -31,7 +31,7 @@ const StyledOrderedListItemContent = styled(MD)`
   ${props => listItemContentCSS(props)};
 `;
 
-const OrderedListItem = ({ children, ...props }) => {
+const OrderedListItem = ({ children }) => {
```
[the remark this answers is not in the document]
jzempel: I'm addressing this in the upcoming `typography` PR: https://github.com/zendeskgarden/react-components/compare/next...jzempel/styled-typography#diff-df84cee4e13dc0397a975151f9b34145R17

utils/test/svg-mock.js
jzempel: Yeah, had it that way originally and then decided to be fancy. Reverting.

packages/typography/src/views/lists/UnorderedListItem.js
```diff
@@ -26,7 +26,7 @@ const StyledUnorderedListItemContent = styled(MD)`
   ${props => listItemContentCSS(props)};
 `;
 
-const UnorderedListItem = ({ children, ...props }) => {
+const UnorderedListItem = ({ children }) => {
```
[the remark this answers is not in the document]
jzempel: I'm addressing this in the upcoming `typography` PR: https://github.com/zendeskgarden/react-components/compare/next...jzempel/styled-typography#diff-b6aaaaa36f05ab8393bcd03def2f60b7R17

PR #415, opened 2019-08-16
fix(datepicker): correct minValue and maxValue displays and allow same date selection within Range

austin94:
## Description

While making some additional datepicker examples I found a few issues that I have resolved in this PR.

- `minValue` and `maxValue` displays were off by a single day due to us comparing date values without finding a common day baseline
- `DatepickerRange` usages were unable to have a start and end date that were the same value

## Checklist

      designer as a reviewer)
      component


jzempel:
It would be good if the range example could be tweaked under this PR. The current grid layout isn't working very nicely.

Screen Shot 2019-08-22 at 9 45 37 AM

Instead, we'd want those fields to display at a fixed `280px` width directly above the corresponding begin/end months.

Issue #419, opened 2019-08-17
Documentation about @zendeskgarden/react-theming 


vhiairrassary:
[The installation snippet](https://github.com/zendeskgarden/react-components/blob/122e1f02c09a7cdc0a43085e97bd5b5ba6b1fb31/packages/theming/README.md#installation) for `@zendeskgarden/react-theming` is:
```sh
npm install @zendeskgarden/react-theming

# Peer Dependencies - Also Required
npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
```

But I am not sure `@zendeskgarden/react-theming` should be listed as a `peer dependencies` (and [it is not in the associated package.json](https://github.com/zendeskgarden/react-components/blob/122e1f02c09a7cdc0a43085e97bd5b5ba6b1fb31/packages/theming/package.json#L21-L26)). I hope this is a valid report that could improve the documentation.

ryanseddon:
Thanks @vhiairrassary you are indeed correct that react-theming shouldn't be mentioned in the peer dependency install instructions, care to create a PR to fix that?

vhiairrassary:
Sure, I just created the PR: https://github.com/zendeskgarden/react-components/pull/421

PR #421, opened 2019-08-20
docs: correct react-theming peer dependencies

vhiairrassary:
Fixes https://github.com/zendeskgarden/react-components/issues/419


ryanseddon:
Thanks @vhiairrassary this should be deployed shortly

PR #424, opened 2019-08-21
feat(tags): CSS-in-JS refactor

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

 - BREAKING CHANGES
    - no more `styles.css` dist
    - removed `Avatar` and `Close` exports -> `Tag.Avatar` and `Tag.Close`
    - renamed tag `type` -> `hue` prop
- Restructured source
  - `src/components` – high-level package components
  - `src/styled` – styled-components variants (formerly `views`)
- All tag CSS contained within components and themable! (no more CSS module dependencies)

## Checklist

      designer as a reviewer)
      component~


packages/tags/src/views/Tag.js
```diff
@@ -5,110 +5,175 @@
...
+
+  return css`
+    background-color: ${backgroundColor};
+    color: ${foregroundColor};
+
+    &:hover {
+      color: ${foregroundColor}; /* <a> element reset */
+    }
+
+    &.focus-visible {
```
ryanseddon: The useFocusVisible hook prefixes the class with `garden` should we do that here too?

packages/tags/README.md
```diff
@@ -15,24 +15,19 @@ npm install react react-dom prop-types styled-components @zendeskgarden/react-th
...
   <Tag pill>
-    <Avatar>
-      <img src="images/amir.png" alt="Example Avatar" />
-    </Avatar>
+    <Tag.Avatar>
+      <img src="images/user.png" alt="Example User" />
+    </Tag.Avatar>
     Default tag with avatar
-    <Close onClick={() => alert('remove tag')} />
+    <Tag.Close onClick={() => alert('remove tag')} />
```
ryanseddon: Can we add an `aria-label` here to show best practice a11y, I know it's defaulted but just to make people aware that if they want this translated it should be specified by the consumer.

packages/tags/examples/advanced.md
```diff
@@ -0,0 +1,101 @@
...
+            )}
+            <Ellipsis>{tag}</Ellipsis>
+            {state.close && <Tag.Close onClick={() => alert(`Delete "${state.text}" tag`)} />}
+          </Tag>
+        ))}
+      </FauxInput>
+      <Range
+        onChange={event => setState({ width: event.target.value })}
+        style={{ marginTop: 8 }}
+        value={state.width}
```
ryanseddon: Super nit-picky but the flutter of avatars changing as you move the range slider is a bit jarring.

packages/tags/src/views/Tag.js
jzempel: Possibly. For now I want to leave these `.focus-visible` classes in place so that we can grep the codebase for locations (also in buttons) that should be addressed when `@zendeskgarden/container-focusvisible` is introduced. I'm also not exactly sure we want to retain the original `focused` assumptions that @austin94 made on this tags component – another topic to discuss as we add the `useFocusVisible` hook.

packages/tags/README.md
jzempel: Good call. Done.

packages/tags/examples/advanced.md
jzempel: 💅 done.

PR #432, opened 2019-08-26
feat(theming): upgrade Theming to use TypeScript

austin94:
## Description

This PR introduces TypeScript into our `react-theming` package. It is split into 2 commits.

1. The first commit adds a global `DefaultTheme` to styled-components and migrates the Theming utilities to TS
1. The second commit migrates the existing TS usages of `react-theming` to pass the Typescript compiler

## Detail

Everything seemed to go fairly smoothly with this, but there are some areas that I would like to comb through with @jzempel if possible.

The `getColor` utility has several different ways to pass a hue/shade that make it difficult to type correctly. I would like to go through all the use-cases to ensure the overrides are correct.

Also, I would like to type our DefaultTheme creation in a more generic way to help with flexibility.

## Checklist

      designer as a reviewer)
      component


austin94:
@jzempel thanks for walking me through some of the color compute logic. I have pushed the recent changes.

PR #439, opened 2019-09-03
feat(dropdowns): introduce Multiselect component

austin94:
## Description

This PR introduces a `Multiselect` component within the `react-dropdowns` package. This is the final item necessary to finalize the `react-autocomplete` deprecation.

## Detail

This component has a similar API to the previous element. It also includes the complications of that design so this isn't the most elegant solution. This component is starting to stretch the "global" `<Dropdown>` API that is shared with Menu, Select, and Autocomplete.

Downshift is working on a [collection of hooks](https://github.com/silviuavram/downshift-hooks) that would replicate what we have achieved in `react-dropdowns`. Hopefully we can transition to those once they are fleshed out some more.

## Checklist

      designer as a reviewer)
      component


austin94:
Coveralls now has a corrected coverage https://coveralls.io/builds/25638668

packages/dropdowns/src/styled/field/StyledMultiselect.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isSmall: boolean;
+  isVisible: boolean;
+}>`
+  && {
+    flex-basis: 60px;
+    flex-grow: 1;
+    margin: 2px;
+    width: inherit;
+    min-width: 60px;
+    line-height: ${props => (props.isSmall ? 20 / 14 : 32 / 14)};
```
ryanseddon: Where are these values from? How did you arrive at them?

packages/dropdowns/src/styled/field/StyledMultiselect.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  ${props =>
+    !props.isVisible &&
+    `
+  &&& {
```
ryanseddon: Ooof how come we need to triple class here?

packages/dropdowns/examples/multiselect.md
```diff
@@ -0,0 +1,138 @@
...
+          setInputValue(changes.inputValue);
+        }
+      }}
+    >
+      <Field>
+        <Label>Multiselect with debounce</Label>
+        <Hint>This example include basic debounce logic</Hint>
+        <Multiselect
+          small
+          renderItem={({ value, removeValue }) => (
```
ryanseddon: Do you see this as a potential issue down the track if consumers wanted more control over how to render the multiselect children?

packages/dropdowns/src/styled/field/StyledMultiselect.tsx
austin94: These values are originally from the `react-autocomplete` package. All of this styling is a direct copy. These are to match some non-identified constants in `css-forms` which we won't have to worry about post CSS-in-JS

packages/dropdowns/src/styled/field/StyledMultiselect.tsx
austin94: There is very high specificity with the `css-forms` usage. These overrides are already in use in `react-autocomplete`

packages/dropdowns/examples/multiselect.md
austin94: Very much so 😞I attempted to match the context model that is used throughout the other components in `react-dropdowns` i.e.

```jsx
<Dropdown>
  <Field>
    <Multiselect>
      <Tag>Hi</Tag><Tag>Hi2</Tag>{/** etc */}
    </Multiselect>
  <Field>
  <Menu>{ /** etc */}</Menu>
</Dropdown>
```

Unfortunately, there were some issues with ensuring our `useSelection` hook is called correctly. Since `getItemProps()` isn't idempotent I was unable to ensure that the children were rendered at the correct time.

Also, with the move to hooks there isn't a guarantee on the ordering of `useEffect` hooks, which even thought it's a good thing, makes ordering difficult here. This `renderItem` API allows us to enforce the ordering and number of items displayed without too much magic.

packages/dropdowns/examples/multiselect.md
```diff
@@ -0,0 +1,138 @@
...
+      downshiftProps={{ defaultHighlightedIndex: 0 }}
+      onStateChange={changes => {
+        if (Object.prototype.hasOwnProperty.call(changes, 'inputValue')) {
+          setInputValue(changes.inputValue);
+        }
+      }}
+    >
+      <Field>
+        <Label>Multiselect with debounce</Label>
+        <Hint>This example include basic debounce logic</Hint>
```
jzempel: *includes

packages/dropdowns/examples/multiselect.md
```diff
@@ -0,0 +1,138 @@
+The `Multiselect` component renders a customizable "tag" for each selected item.
+
+This can be any element, but the surrounding field is designed to work with
+our standard `Tag` component.
+
+```jsx static
+<Dropdown>
+  <Field>
+    <Label>Example Multiselect</Label>
+    <Multiselect
```
jzempel: This fails to dismiss the expanded menu on chevron click. Compare to https://garden.zendesk.com/react-components/autocomplete/#multiselect where clicking the chevron shows/hides the menu.

Additionally, clicking the cursor text area shows/hides the menu and it shouldn't. It's like the click action on these two controls needs to be reversed.

There's also an unfortunate rendered flash of `+ X more` onMouseDown that would be nice to have cleaned up.

packages/dropdowns/examples/multiselect.md
```diff
@@ -0,0 +1,138 @@
...
+        if (Object.prototype.hasOwnProperty.call(changes, 'inputValue')) {
+          setInputValue(changes.inputValue);
+        }
+      }}
+    >
+      <Field>
+        <Label>Multiselect with debounce</Label>
+        <Hint>This example include basic debounce logic</Hint>
+        <Multiselect
+          small
```
jzempel: You're using a `small` multi-select with a default menu size. Please make the two sizes match.

packages/dropdowns/src/styled/field/StyledMultiselect.tsx
```diff
@@ -0,0 +1,66 @@
...
+    opacity: 0;
+    height: 0;
+    min-height: 0;
+    width: 0;
+    min-width: 0;
+  }
+`}
+`;
+
+export const StyledMoreAnchor = styled.div`
```
jzempel: Sadness that we can't modify our `StyledAnchor` here. 🙁

PR #442, opened 2019-09-04
feat(button): migrate react-buttons to TypeScript

austin94:
## Description

This PR migrates `react-buttons` to TypeScript. Overall, pretty straight-forward, but I have pointed out a few changes below that will affect other packages.

## Detail

The `react-buttons` package leverages `React.forwardRef` heavily. The default type definition for `forwardRef` does not include the `WeakValidationMap<T>` value for `propTypes`. Since we want to support this value for our non-TS consumers I have pony-filled the `ForwardRefExoticComponent` definition with a custom `react.d.ts` in our utils folder.

```ts
import { ForwardRefExoticComponent, NamedExoticComponent } from 'react';

declare module 'react' {
  /* eslint-disable-next-line */
  export interface ForwardRefExoticComponent<P> extends NamedExoticComponent<P> {
    defaultProps?: Partial<P>;
    propTypes?: WeakValidationMap<P>;
  }
}
```

This allows us to use `React.fowardRef` without type casting. This may be something we want to up-stream to DefinitelyTypes.

## Checklist

      designer as a reviewer)
      component


packages/buttons/src/index.ts
```diff
@@ -11,4 +11,3 @@ export { default as ButtonGroup } from './components/ButtonGroup';
 export { default as ChevronButton } from './components/ChevronButton';
 export { default as IconButton } from './components/IconButton';
 export { default as SplitButton } from './components/SplitButton';
-export { default as ButtonGroupContainer } from './containers/ButtonGroupContainer';
```
jzempel: :+1:

packages/buttons/src/styled/StyledButtonGroup.spec.tsx
```diff
@@ -37,10 +36,11 @@ describe('StyledButtonGroup', () => {
       </StyledButtonGroup>
     );
 
+    /* stylelint-disable */
     expect(getByTestId('group-button')).toHaveStyleRule('position', 'relative', {
-      modifier: css`
+      modifier: (`
```
jzempel: is the `css` removal intended?

austin94: Yes, there were stylelint issues when this got moved to TypeScript that I couldn't remove. Also, the tests still pass so I don't think they are necessary in this context.

PR #448, opened 2019-09-10
feat(tabs): deprecate TabsContainer component

austin94:
## Description

This PR deprecates the `TabsContainer` within the `react-tabs` package. It also migrates us to the `container-tabs` package.

## Detail

There should be no change in the API or consumption model for this component. There was some additional focus styling needed as we now use the "roving tab index" model of selection, but overall not too bad.

## Checklist

      designer as a reviewer)
      component


packages/tabs/src/containers/TabsContainer.example.md
```diff
@@ -1,82 +1,7 @@
...
-          {tab} content
-        </TabPanel>
-      ))}
-    </TabsView>
-  )}
-</TabsContainer>;
-```
+This component has been deprecated in favor of the API provided in the
+[@zendeskgarden/container-tabs](https://www.npmjs.com/package/@zendeskgarden/container-tabs)
+package.
```
ryanseddon: Should you keep the examples here until it's removed?

packages/tabs/src/views/Tab.js
```diff
@@ -28,6 +28,16 @@ const Tab = styled.div.attrs(props => ({
     [TabStyles['is-selected']]: props.selected
   })
 }))`
+  &&& {
+    ${props =>
+      !props.focused &&
+      `
+      :before {
+        box-shadow: none;
```
ryanseddon: Is this fixing a bug? Why is this included?

packages/tabs/src/elements/Tabs.js
```diff
@@ -8,16 +8,17 @@
...
-import TabsContainer from '../containers/TabsContainer';
 import TabsView from '../views/TabsView';
 import TabList from '../views/TabList';
 import Tab from '../views/Tab';
 
 /**
  * High-level abstraction for basic Tabs implementations.
  */
-export default class Tabs extends ControlledComponent {
+class Tabs extends ControlledComponent {
```
ryanseddon: Why not convert to a functional component here and use `useTabs` instead? It helps reduce code size and is easier to read. See #431

packages/tabs/src/containers/TabsContainer.example.md
austin94: Sure, I will add that back.

packages/tabs/src/elements/Tabs.js
austin94: This would be a breaking change with removing `onStateChange` and how people could have overrode it.

We can include that API change when we change the Tabs API in CSS-in-JS.

austin94:
> People will have a lot of warnings in tests and dev if we push this into a minor release

I originally had the warning at the root of the file, which would have shown the warning at the root import. I've now moved this to within the `TabsContainer` component (and the previous deprecation in `BreadcrumbsContainer`) to reduce the visibility to those that only use it directly.

packages/tabs/src/views/Tab.js
jzempel: Feels like this hack should be commented in code.

austin94: I've added a comment for this.

PR #449, opened 2019-09-10
chore(build): add bundleSize check to all packages

austin94:
## Description

This PR enables bundle size thresholds using the [bundlesize](https://www.npmjs.com/package/bundlesize) library.

Each package controls it's threshold with the `zendeskgarden:max_size` attribute in it's `package.json`.

This can also be hooked up with Github comments at a later time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,13 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH="$HOME/.yarn/bin:$PATH"
 
-install: yarn --frozen-lockfile --ignore-scripts
+install: yarn --frozen-lockfile
 
 script:
-  - yarn lerna bootstrap --ignore-scripts
   - yarn lint
   - yarn tsc
   - yarn format
+  - yarn bundlesize --config bundlesize.config.js
   - yarn test:all --coverage --runInBand
 
 after_success: yarn coveralls < demo/coverage/lcov.info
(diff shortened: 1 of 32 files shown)


.travis.yml
```diff
@@ -13,13 +13,13 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH="$HOME/.yarn/bin:$PATH"
 
-install: yarn --frozen-lockfile --ignore-scripts
+install: yarn --frozen-lockfile
```
ryanseddon: This will slow down the build time, means it'll bootstrap and build the demos for all PRs

austin94:
> Is the plan to have a github action to run this and comment on the PR too as a next step?

That could be a follow on for sure. Right now we will get a separate PR check which should keep us going until then.

.travis.yml
austin94: This shouldn't be building the demos. That is under a separate npm script.

For us to check the bundle size we have to build the packages. Also, with TypeScript we will need to produce the type-definitions from the build to test and lint correctly.

This change is already merged in the `next` branch as well.

> slow down

This will slow it down, but only for the webpack output. Looks like it's only added a few minutes right now. Once we remove the deprecated packages in the next major that time should be cut in half.

jzempel:
A little late, but `zendeskgarden:max_size` should've been added to the `.template` package.

austin94:
> A little late, but zendeskgarden:max_size should've been added to the .template package.

I've added this line under #448 since it's a one-liner.

PR #460, opened 2019-09-19
feat(buttons): deprecate ButtonGroupContainer

austin94:
## Description

This PR deprecates the `ButtonGroupContainer` component within `react-buttons`.

## Detail

In this deprecating I opted to mimic the `onStateChange` callback rather than use the `ControlledComponent` from `react-selection`. Since there was only one value being provided this ended up being simpler.

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/buttons/src/elements/ButtonGroup.js
+++ b/packages/buttons/src/elements/ButtonGroup.js
@@ -59,7 +59,6 @@ const ButtonGroup = ({ id, selectedKey, onStateChange, children, ...otherProps }
           selected: key === selectedItem,
           focused: key === focusedItem,
           item: key,
-          ref: focusRef,
           focusRef,
           ...child.props
         })
--- a/packages/buttons/src/elements/ButtonGroup.js
+++ b/packages/buttons/src/elements/ButtonGroup.js
@@ -5,110 +5,85 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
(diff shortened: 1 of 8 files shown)


packages/buttons/src/elements/ButtonGroup.js
```diff
@@ -5,110 +5,85 @@
...
+      const focusRef = React.createRef();
+
+      return cloneElement(
+        child,
+        getButtonProps({
+          key,
+          selected: key === selectedItem,
+          focused: key === focusedItem,
+          item: key,
+          ref: focusRef,
```
hzhu: Could you clarify the purpose of line 62? The ref value is also passed on line 63 and it seems like the value is already [defined](https://github.com/zendeskgarden/react-containers/blob/master/packages/selection/src/useSelection.js#L237) in `useSelection`. What are the consequences of removing line 62?

austin94: Good catch, this isn't needed anymore. There was a recent change to `useSelection` that doesn't make this necessary, but this would still work.

We should make a follow-on task to update our READMEs in `react-containers` to use the correct syntax. The example in https://www.npmjs.com/package/@zendeskgarden/container-buttongroup is showing my original syntax.

hzhu: Oh nice, thanks for the clarification!

PR #462, opened 2019-09-20
feat(loaders): deprecate ScheduleContainer BREAKING

### LINKED ISSUE #147: Rethink loading icon speed API

austin94:
- [x] **BREAKING CHANGE**

## Description

This PR deprecates the `ScheduleContainer` found in `react-loaders` and migrates to the `useSchedule` hook in `container-schedule`. This deprecation REQUIRES a **BREAKING CHANGE** on the `Dots` and `Spinner` component where the `velocity` prop is now a standard `duration` (in MS).

Closes #147 

## Detail

With this change it became easier to bring in the work from the `next` branch rather than start from scratch. This PR now includes all `next` changes for `react-loaders` (without the theming improvements).

## Timing

This PR should stay queued until we are ready to merge v7

## Checklist

      designer as a reviewer)
      component


austin94:
It looks like the `ScheduleContainer` was never actually exported publicly. I have just removed it rather than following a deprecation path.

PR #463, opened 2019-09-23
feat(forms): deprecate react-ranges and migrate components to react-forms

austin94:
## Description

This PR deprecates the `react-ranges` package and migrates the `Range` and `MultiThumbRange` components to `react-forms`.

## Detail

Due to the structure of the form components:

```jsx
<Field>
  <Label>...</Label>
  <Range />
</Field>
```

Labels aren't able to dynamically change their appearance based on the context they are nested in (i.e. checkboxes and radios). These components are currently using the input `Label`, `Hint`, and `Message` which have the equivalent styling.

## Checklist

      designer as a reviewer)
      component


packages/forms/examples/range.md
```diff
@@ -0,0 +1,39 @@
+```jsx
+<Field>
+  <Label>Default Range</Label>
+  <Hint>Supports all native props</Hint>
+  <Range />
```
jzempel: This needs `margin-top: 8px` to compensate for not having access to the range-specific hint.

packages/forms/examples/range.md
```diff
@@ -0,0 +1,39 @@
...
+    : 'Youre good!';
+
+<Field>
+  <Label>Dynamic Validation</Label>
+  <Hint>
+    Move range to view changes. [value="
+    {state.value}
+    "]
+  </Hint>
+  <Range
```
jzempel: Needs `margin: 8px 0`.

packages/forms/examples/multi-thumb-range.md
```diff
@@ -0,0 +1,62 @@
...
+  minValue: 0,
+  maxValue: 100
+};
+
+<Field>
+  <Label>Default MultiThumb Range</Label>
+  <Hint>
+    Min Value: {state.minValue} | Max Value: {state.maxValue}
+  </Hint>
+  <MultiThumbRange
```
jzempel: Add `margin-top: 8px`.

packages/forms/examples/multi-thumb-range.md
```diff
@@ -0,0 +1,62 @@
...
+  minValue: 0,
+  maxValue: 100
+};
+
+<Field>
+  <Label>MultiThumb Range - Step Size (5)</Label>
+  <Hint>
+    Min Value: {state.minValue} | Max Value: {state.maxValue}
+  </Hint>
+  <MultiThumbRange
```
jzempel: Add `margin-top: 8px`.

packages/forms/examples/multi-thumb-range.md
```diff
@@ -0,0 +1,62 @@
...
+  minValue: 25,
+  maxValue: 35
+};
+
+<Field>
+  <Label>Disabled MultiThumb Range</Label>
+  <Hint>
+    Min Value: {state.minValue} | Max Value: {state.maxValue}
+  </Hint>
+  <MultiThumbRange
```
jzempel: Add `margin-top: 8px`.

packages/forms/src/fields/MultiThumbRange.js
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    if (keyIntercepted) {
+      e.preventDefault();
+      e.stopPropagation();
+    }
+  };
+
+  // This may be increased to enforce separation between thumbs
+  const MIN_DISTANCE_PX = 0;
```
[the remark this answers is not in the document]
austin94: This only affects the visual distance between elements, not the actual values returned. i.e. this could be used to only allow the edges of the thumbs to touch.

When we went over the original design for this we decided to allow overlapping since both values could technically be the same.

packages/forms/src/fields/MultiThumbRange.spec.js
```diff
@@ -0,0 +1,462 @@
...
+      });
+
+      it('increments minValue on RIGHT key', () => {
+        const { getAllByTestId } = render(
+          <MultiThumbRange minValue={15} maxValue={75} step={5} onChange={onChangeSpy} />
+        );
+
+        fireEvent.keyDown(getAllByTestId('thumb')[0], { keyCode: KEY_CODES.RIGHT });
+        expect(onChangeSpy).toHaveBeenCalledWith({ minValue: 20, maxValue: 75 });
+      });
```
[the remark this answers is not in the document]
austin94: > priority querying

We use the role and text queries more often in `react-containers`. Since many of our consumers need to test via `data-test-id` attributes we use the same pattern internally to catch areas that we forget to spread props to their children.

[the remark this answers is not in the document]
austin94: > render against changed output

This is how we (try) handle the testing in our more modern components. The tests within the `react-ranges` packages have gone through our `enzyme -> @testing-library/react` migration so the tests don't follow the best practices for integration testing. Since this is a hook-conversion I didn't want to mess with the tests as much as possible to avoid regressions.

packages/forms/src/fields/Range.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    it('updates backgroundSize after onChange event', () => {
+      const { getByTestId } = render(
+        <Field>
+          <Range min={0} max={-25} value={25} data-test-id="range" />
+        </Field>
+      );
+
+      const range = getByTestId('range');
+
+      fireEvent.change(range, { target: { value: 85 } });
```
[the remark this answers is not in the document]
austin94: For some background, as I mentioned above most of these tests have gone through an `enzyme -> @testing-library/react` migration so these patterns are a little off.

We have found diminishing returns with doing assertions for a "before" visual state. This becomes more apparent as we move to CSS-in-JS more heavily.

But, we can always use more pre-assertions for some of the logical changes. 😄 I've added in several more assertions for the non-style changes.

packages/forms/src/fields/MultiThumbRange.spec.js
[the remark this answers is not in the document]
hzhu: Thanks for the explanation @austin94!

packages/forms/src/fields/Range.spec.js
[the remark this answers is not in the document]
hzhu: > `enzyme -> @testing-library/react` migration so these patterns are a little off.

Oh this makes sense. Thanks for the explanation.

packages/forms/src/styled/index.js
```diff
@@ -43,3 +43,11 @@ export { default as StyledToggleLabel } from './toggle/StyledToggleLabel';
...
 export { default as StyledToggleInput } from './toggle/StyledToggleInput';
 export { default as StyledToggleMessage } from './toggle/StyledToggleMessage';
+
+/**
+ * Range styles
+ */
+export { default as StyledRangeSingleThumb } from './range/StyledRangeSingleThumb';
+export { default as StyledRangeLabel } from './range/StyledRangeLabel';
+export { default as StyledRangeHint } from './range/StyledRangeHint';
+export { default as StyledRangeMessage } from './range/StyledRangeMessage';
```
jzempel: I think the problem with spacing is that these styled Label/Hint/Message components are not being used. Once these are configured, with corresponding context, in `Label.js`, `Hint.js`, and `Message.js` I think we'll see the spacing mentioned under my previous comments.

packages/forms/examples/range.md
austin94: Corrected.

packages/forms/examples/multi-thumb-range.md
austin94: Corrected.

packages/forms/examples/multi-thumb-range.md
austin94: Corrected.

packages/forms/examples/multi-thumb-range.md
austin94: Corrected.

PR #468, opened 2019-09-25
feat(loaders): introduce Progress component

sunesimonsen:
- [ ] **BREAKING CHANGE?** 

## Description

Introduces a new progress component.

## Detail

Screenshot 2019-10-04 at 09 00 00

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/loaders/src/Progress.js
+++ b/packages/loaders/src/Progress.js
@@ -1,0 +1,76 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import { zdColorGreen600, zdColorGrey200, zdSpacingXs } from '@zendeskgarden/css-variables';
+
+const COMPONENT_ID = 'loaders.progress';
+
+const HEIGHTS = {
+  small: 2,
+  medium: 6,
+  large: 12
+};
+
+const sizeToBorderRadius = size => HEIGHTS[size] / 2;
+
+const ProgressBackground = styled.div.attrs(props => ({
+  borderRadius: sizeToBorderRadius(props.size)
+}))`
+  margin: ${zdSpacingXs} 0;
+  border-radius: ${({ borderRadius }) => borderRadius}px;
+  background-color: ${zdColorGrey200};
+  color: ${({ color }) => color || zdColorGreen600};
+`;
+
+const ProgressIndicator = styled.div.attrs(props => ({
+  height: HEIGHTS[props.size],
+  borderRadius: sizeToBorderRadius(props.size)
+}))`
+  transition: width 0.1s ease-in-out;
+  border-radius: ${({ borderRadius }) => borderRadius}px;
+  background: currentColor;
+  width: ${({ value }) => value}%;
+  height: ${({ height }) => height}px;
+`;
+
+const Progress = forwardRef(({ value, size, ...other }, ref) => {
+  const percentage = Math.max(0, Math.min(100, value));
+
+  return (
+    <ProgressBackground
+      data-garden-id={COMPONENT_ID}
+      data-garden-version={PACKAGE_VERSION}
+      aria-valuemax="100"
+      aria-valuemin="0"
+      aria-valuenow={percentage}
+      ref={ref}
+      role="progressbar"
+      size={size}
+      {...other}
+    >
+      <ProgressIndicator value={percentage} size={size} />
+    </ProgressBackground>
+  );
+});
+
+Progress.propTypes = {
+  color: PropTypes.string,
+  /** The progress as a value between 0 and 100 */
+  value: PropTypes.number.isRequired,
+  size: PropTypes.oneOf(['small', 'medium', 'large'])
+};
+
+Progress.defaultProps = {
+  value: 0,
+  size: 'medium'
+};
+
+export default Progress;
--- a/packages/loaders/src/Progress.spec.js
+++ b/packages/loaders/src/Progress.spec.js
@@ -1,0 +1,93 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render } from 'garden-test-utils';
+import Progress from './Progress';
+import { zdColorGreen600 } from '@zendeskgarden/css-variables';
+
+describe('Progress', () => {
+  describe('without a value', () => {
+    it('renders a progress bar with 0% progress', () => {
+      const { getByRole } = render(<Progress />);
+
+      expect(getByRole('progressbar')).toHaveAttribute('aria-valuenow', '0');
+    });
+  });
+
+  it('renders a progress bar with correct aria attributes', () => {
+    const { getByRole } = render(<Progress value={40} />);
+
+    const progress = getByRole('progressbar');
+
+    expect(progress).toHaveAttribute('aria-valuemin', '0');
+    expect(progress).toHaveAttribute('aria-valuenow', '40');
+    expect(progress).toHaveAttribute('aria-valuemax', '100');
+  });
+
+  it('renders a progress indicator', () => {
+    const { getByRole } = render(<Progress value={40} />);
+
+    expect(getByRole('progressbar').firstChild).toHaveStyleRule('width', '40%');
+  });
+
+  describe('when given a value below zero', () => {
+    it('renders a progress bar with 0% progress', () => {
+      const { getByRole } = render(<Progress value={-42} />);
+
+      expect(getByRole('progressbar')).toHaveAttribute('aria-valuenow', '0');
+    });
+  });
+
+  describe('when given a value above 100', () => {
+    it('renders a progress bar with 100% progress', () => {
+      const { getByRole } = render(<Progress value={666} />);
+
+      expect(getByRole('progressbar')).toHaveAttribute('aria-valuenow', '100');
+    });
+  });
+
+  describe('with a small size', () => {
+    it('renders a small progress bar', () => {
+      const { getByRole } = render(<Progress value={42} size="small" />);
+
+      expect(getByRole('progressbar').firstChild).toHaveStyleRule('height', '2px');
+    });
+  });
+
+  describe('with the default size', () => {
+    it('renders a default progress bar', () => {
+      const { getByRole } = render(<Progress value={42} />);
+
+      expect(getByRole('progressbar').firstChild).toHaveStyleRule('height', '6px');
+    });
+  });
+
+  describe('with a large size', () => {
+    it('renders a large progress bar', () => {
+      const { getByRole } = render(<Progress value={42} size="large" />);
+
+      expect(getByRole('progressbar').firstChild).toHaveStyleRule('height', '12px');
+    });
+  });
+
+  describe('without a color set', () => {
+    it('renders a blue progress bar by default', () => {
+      const { getByRole } = render(<Progress value={42} />);
+
+      expect(getByRole('progressbar')).toHaveStyleRule('color', zdColorGreen600);
+    });
+  });
+
+  describe('with a color set', () => {
+    it('renders a colored progress bar', () => {
+      const { getByRole } = render(<Progress value={42} color="red" />);
+
+      expect(getByRole('progressbar')).toHaveStyleRule('color', 'red');
+    });
+  });
+});
--- a/packages/loaders/README.md
+++ b/packages/loaders/README.md
@@ -28,6 +28,20 @@ import { Dots } from '@zendeskgarden/react-loaders';
 </ThemeProvider>;
 ```
 
+### Progress
+
+```jsx static
+import { ThemeProvider } from '@zendeskgarden/react-theming';
+import { Progress } from '@zendeskgarden/react-loaders';
+
+/**
+ * Place a `ThemeProvider` at the root of your React application
+ */
+<ThemeProvider>
+  <Progress value={40} />
+</ThemeProvider>;
+```
+
 ### Spinner
 
 ```jsx static
--- a/packages/loaders/src/Progress.example.md
+++ b/packages/loaders/src/Progress.example.md
@@ -1,0 +1,232 @@
+```jsx
+<Grid>
+  <Row>
+    <Col>
+      Cat.png
+      <Progress value={10} />
+    </Col>
+  </Row>
+  <Row>
+    <Col>
+      Dog.png
+      <Progress value={40} />
+    </Col>
+  </Row>
+</Grid>
+```
+
+### Sizes
+
+```jsx
+<Grid>
+  <Row>
+    <Col>
+      Small
+      <Progress value={10} size="small" />
+    </Col>
+  </Row>
+  <Row>
+    <Col>
+      Default
+      <Progress value={40} />
+    </Col>
+  </Row>
+  <Row>
+    <Col>
+      Large
+      <Progress value={80} size="large" />
+    </Col>
+  </Row>
+</Grid>
+```
+
+### Colors
+
+```jsx
+const { zdColorGrey600, zdColorBlue600, zdColorRed600 } = require('@zendeskgarden/css-variables');
+
+<Grid>
+  <Row>
+    <Col>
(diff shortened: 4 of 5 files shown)


packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,63 @@
...
+const heights = {
+  small: 2,
+  large: 12,
+  default: 6
+};
+
+const sizeToHeight = size => heights[size || 'default'];
+const sizeToBorderRadius = size => sizeToHeight(size) / 2;
+
+const ProgressBarContainer = styled.div.attrs(props => ({
```
ryanseddon: Can you use the `<progress>` element? Or is it tricky to do custom styles on?

sunesimonsen: We discussed that, but it is very hard to style as it maps to a native control in different OS'es.

sunesimonsen:
Sorry for the wait - I'll pick this up soon.

sunesimonsen:
@austin94 should you be able to change the color of a progress bar?

jzempel:
@sunesimonsen we will want to be able to change the color fairly easily – similar to the existing loaders. I suspect whatever approach you take will be fine as this is an area that we will smooth out with our CSS-in-JS release currently in-progress on the `next` branch.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,72 @@
...
+  border-radius: ${props => props.borderRadius}px;
+  background: ${zdColorGrey200};
+`;
+
+const ProgressBarIndicator = styled.div.attrs(props => ({
+  height: sizeToHeight(props.size),
+  borderRadius: sizeToBorderRadius(props.size)
+}))`
+  border-radius: ${props => props.borderRadius}px;
+  background: ${zdColorBlue600};
```
jzempel: In order to more easily handle color changes, you might set `color: ${zdColorBlue600}` on the `ProgressBarContainer` and then set `background-color: currentColor` here on the `ProgressBarIndicator`.

sunesimonsen:
@jzempel should we add a transition on the width to make it more smooth when the progress changes?

jzempel:
@sunesimonsen most likely. These micro color transitions tend to be `.1s ease-in-out` throughout Garden. But we probably need guidance from a designer.

/cc @allisonacs @rossmoody @ginnywood

allisonacs:
@sunesimonsen is this pre-published somewhere so I can review? Would love to see it in action (as well as get added as an official reviewer).

sunesimonsen:
@allisonacs it is not really interactive, so the screenshot is pretty much it right now. But I'll at the colors as the other loaders, then we can try to publish it.

allisonacs:
@sunesimonsen ok cool. If I might request some kind ability to show what it looks like as it animates in the published demo, that'd be really cool for folks—same way as we demo the skeleton loaders.

sunesimonsen:
@allisonacs demo https://garden.zendesk.com/react-components/loaders/#progress

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,76 @@
...
+import styled, { keyframes } from 'styled-components';
+
+const COMPONENT_ID = 'loaders.progress';
+
+const { zdColorBlue600, zdColorGrey200, zdSpacingXs } = require('@zendeskgarden/css-variables');
+
+const heights = {
+  small: 2,
+  large: 12,
+  default: 6
```
jzempel: We've been moving our APIs to use `medium` and then setting that size in `defaultProps`.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,76 @@
...
+const Progress = ({ value, size, ...other }) => {
+  const percentage = Math.max(0, Math.min(100, value));
+
+  return (
+    <ProgressBarContainer
+      data-garden-id={COMPONENT_ID}
+      data-garden-version={PACKAGE_VERSION}
+      aria-valuemax="100"
+      aria-valuemin="0"
+      aria-valuenow={percentage}
```
jzempel: Should there be a default `aria-valuetext`? Either that, or throw an error if it's missing like https://github.com/zendeskgarden/react-containers/blob/5b06c16f04ee9792f002073e8e4e5a041a710f01/packages/accordion/src/useAccordion.js#L67

packages/loaders/src/Progress.js
sunesimonsen: That is also a much better pattern, I copied this from another component. I'll change that.

packages/loaders/src/Progress.js
sunesimonsen: We can just make it a required prop, I didn't know if you wanted to enforce that.

jzempel: Yeah, tbh I'm not exactly clear. From reading this – https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuetext_attribute – it may be that we should keep it as optional and just rely on `aria-valuenow` as the default.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,76 @@
...
+
+const heights = {
+  small: 2,
+  medium: 6,
+  large: 12
+};
+
+const sizeToBorderRadius = size => heights[size] / 2;
+
+const ProgressBarContainer = styled.div.attrs(props => ({
```
hzhu: What does everyone else think about the "container" in the name of this component? One concern might be that consumers of Garden may get this component confused with similarly named `react-containers` components, which are light weight render prop components that wrap a custom React hook. 

```js
useBreadcrumb() -> <BreadcrumbContainer>
useSelection() -> <SelectionContainer>
useAccordion() -> <AccordionContainer>
useTabs() -> <TabsContainer>

??? -> <ProgressBarContainer>
```

Or does "container" mean wrapping something (anything), and in this case, `<ProgressBarContainer>` wraps `<ProgressBarIndicator>` with CSS?

sunesimonsen: Yes, you are right that might be a bit confusing. It is just the wrapper element. What about `ProgressBackground`?

sunesimonsen: I changed the naming to `ProgressBackground` and `ProgressIndicator`.

hzhu: Awesome @sunesimonsen, that sounds great!

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,76 @@
...
+Progress.propTypes = {
+  color: PropTypes.string.isRequired,
+  /** The progress as a value between 0 and 100 */
+  value: PropTypes.number.isRequired,
+  size: PropTypes.oneOf(['small', 'medium', 'large'])
+};
+
+Progress.defaultProps = {
+  value: 0,
+  color: zdColorBlue600,
```
hzhu: The default `color` seems bit strange on the prop sheet because it includes the `require` statement.

Screen Shot 2019-10-02 at 10 54 35 AM

Also, do you think the default color should be `inherit`? Similar to the default color of the `<Spinner>` component?

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { css } from 'styled-components';
```
hzhu: `css` is imported into the file, but doesn't seem to be used 👀 .

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,93 @@
...
+import { render, renderRtl } from 'garden-test-utils';
```
hzhu: Seems like `renderRtl` is also not used in this test file.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { keyframes } from 'styled-components';
```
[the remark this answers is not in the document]
sunesimonsen: Mhh I'm used to linting catching these things - thanks for pointing it out.

packages/loaders/src/Progress.js
sunesimonsen: I talked with @jzempel about it, we agreed to use blue as the default color. Should I maybe inline the color code, or default it inside of the code instead?

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,93 @@
...
+describe('Progress', () => {
+  describe('without a value', () => {
+    it('renders a progress bar with 0% progress', () => {
+      const { container } = render(<Progress />);
+
+      expect(container.firstChild).toHaveAttribute('aria-valuenow', '0');
+    });
+  });
+
+  it('renders a progress bar with correct aria attributes', () => {
```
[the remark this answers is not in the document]
sunesimonsen: I was just copying patterns from other tests files, that seems to do it this way. I'm a huge fan of test id's and was basically the person introducing the concept in Zendesk, long before there was anything called react-testing-library - as far as I know. That is the main way we query for element in Guide tests. I love them so much that I added first class support for them in [my test library](https://unexpected.js.org/unexpected-dom/assertions/DOMElement/queried-for-test-id/). So yes, let's definitely do that.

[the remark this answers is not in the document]
sunesimonsen: I chose to use the `getByRole` as that makes the most sense to be. I don't think consumers will ever need to get a hold on the progress indicator element as the aria attributes already startes the values, so I don't want to add a test id there - which means I'm still using firstChild to get to that element.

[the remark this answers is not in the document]
hzhu: Oh wow, thanks for sharing the history with me. I'm definitely going to check out your library!

packages/loaders/src/Progress.js
hzhu: Awesome sounds good! I think inlining the color code or defaulting it in the code would be nice if it makes the prop sheet more reader friendly (e.g. without the `require`) statement.

packages/loaders/src/Progress.spec.js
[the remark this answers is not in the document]
sunesimonsen: Funnily enough I crossed paths with https://github.com/kentcdodds, before he went on to make his hugely popular testing library: https://github.com/sunesimonsen/react-dom-testing/issues/3#issuecomment-373952510 :-)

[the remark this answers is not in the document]
hzhu: @sunesimonsen that's funny and awesome! 

```jsx
function sel(id) {
  return `[data-test="${id}"]`
}
```

That's code we used on my previous team and it's so cool that it originally came from your work!

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,94 @@
...
+      expect(getByRole('progressbar')).toHaveAttribute('aria-valuenow', '0');
+    });
+  });
+
+  it('renders a progress bar with correct aria attributes', () => {
+    const { getByRole } = render(<Progress value={40} />);
+
+    const progress = getByRole('progressbar');
+
+    expect(progress).toHaveAttribute('role', 'progressbar');
```
hzhu: If you would like, you can remove this assertion because the one on lines 25 will implicitly verify that a role of `progressbar` exists. Not a big deal or a blocker though.

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+      expect(getByRole('progressbar')).toHaveAttribute('aria-valuenow', '100');
+    });
+  });
+
+  describe('with a small size', () => {
+    it('renders a small progress bar', () => {
+      const { getByRole } = render(<Progress value={42} size="small" />);
+
+      expect(getByRole('progressbar').firstChild).toHaveStyleRule('height', '2px');
```
hzhu: Not requesting a change, but would like to put my thoughts out there for discussion.

I feel like looking up a target element via `.firstChild` _can_ be bit brittle. This is probably more true for application code than library code, but it's always stuck with me. For example, by adding a non-content styled `div`, these tests would break:

```jsx
const Progress = ({ value, size, ...other }) => {
  const percentage = Math.max(0, Math.min(100, value));

  return (
    <ProgressBackground
      data-garden-id={COMPONENT_ID}
      data-garden-version={PACKAGE_VERSION}
      aria-valuemax="100"
      aria-valuemin="0"
      aria-valuenow={percentage}
      role="progressbar"
      size={size}
      {...other}
    >
      <div id="tracking-pixel"></div>
      <ProgressIndicator value={percentage} size={size} />
    </ProgressBackground>
  );
};
```

I know this is a pattern Garden has had (and I'm new!) so I'm **not** suggesting we change this. Also, this is probably not that important in this scenario because the code example above is contrived and would probably never happen in Garden. But just wanted to put my thoughts out on targeting elements by using `.firstChild`.

packages/loaders/src/Progress.spec.js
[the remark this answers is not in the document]
sunesimonsen: Good ideas often is invented many times, so I don't claim to be the inventor of test ids, only in my own code ;-)

allisonacs:
@sunesimonsen I was thinking that we should stay away from blue for the default color. Blue-600 denotes interactivity, and a user can't interact with the progress bar directly. I was thinking we default it to green-600 by default (optimistically assuming success).

sunesimonsen:
@allisonacs I changed the default color to green.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,75 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const COMPONENT_ID = 'loaders.progress';
+
+const { zdColorGreen600, zdColorGrey200, zdSpacingXs } = require('@zendeskgarden/css-variables');
```
austin94: This should be a standard `import {} from '..'` structure rather than the `require` syntax.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,75 @@
...
+
+const heights = {
```
austin94: [nit] we normally ALL_CAPS constants.

packages/loaders/src/Progress.js
```diff
@@ -0,0 +1,75 @@
...
+  borderRadius: sizeToBorderRadius(props.size)
+}))`
+  transition: width 0.1s ease-in-out;
+  border-radius: ${({ borderRadius }) => borderRadius}px;
+  background: currentColor;
+  width: ${({ value }) => value}%;
+  height: ${({ height }) => height}px;
+`;
+
+const Progress = ({ value, size, ...other }) => {
```
austin94: Should we be using `React.forwardRef` here? This would allow our Tooltip and other positioning packages to position PopperJS properly with `ref`.

packages/loaders/src/Progress.spec.js
```diff
@@ -0,0 +1,93 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render } from 'garden-test-utils';
+import Progress from './Progress';
+const { zdColorGreen600 } = require('@zendeskgarden/css-variables');
```
austin94: Same as above. Could this use the standard `import from` syntax?

packages/loaders/src/Progress.spec.js
austin94: We _could_ target this with a `data-garden-test-id`, which is removed during the `production` builds, but not sure it would be worth it in this use-case. Here is the logic for that https://github.com/zendeskgarden/react-components/blob/master/babel.config.js#L24-L28

sunesimonsen: I don't think it would be worth it for this case.

packages/loaders/src/Progress.spec.js
sunesimonsen: Mhhh I probably copied this from somewhere, sorry about that.

sunesimonsen: Ahh I copied it from an example, sorry.

sunesimonsen:
Thanks for all of the reviews!

PR #477, opened 2019-09-26
feat(avatars): migrate to TypeScript

jzempel:
- [ ] **BREAKING CHANGE?** 

## Description

Migrates `@zendeskgarden/react-avatars` to TypeScript on the `next` branch.

## Checklist

      designer as a reviewer)~
      component~


packages/avatars/src/components/Avatar.tsx
```diff
@@ -5,27 +5,27 @@
...
+  /** Set the avatar background color */
+  backgroundColor?: string;
+  /** Set the color for child SVG or `Avatar.Text` components */
+  foregroundColor?: string;
+  /** Set the color of the surface behind the avatar – used to manipulate the inner status rings */
+  surfaceColor?: string;
+  /** Applies system styling */
+  isSystem?: boolean;
+  size?: 'extrasmall' | 'small' | 'medium' | 'large';
+  status?: 'available' | 'away';
```
hzhu: I noticed on [line 196](https://github.com/zendeskgarden/react-components/pull/477/files#diff-a6953e6648bf616b20ee1d05b13b0c76R196) in the `IStyledAvatarProps` that `status` includes an optional `active` status, which isn't included in the `status` for `IAvatarProps`. 

Should `active` be added here too?

```ts
status?: 'available' | 'active' | 'away';
```

packages/avatars/src/components/Avatar.tsx
```diff
@@ -62,26 +62,15 @@ const Avatar = React.forwardRef(
   }
 );
 
-Avatar.propTypes = {
```
hzhu: @austin94, I recalled you had mentioned that `propType` checks are important in our components because not all of our consumers use TypeScript. I might have heard completely wrong, but could you elaborate on the heuristics of when `propType` checks should be removed (or included) when authoring a React component in TypeScript?

For example, in [this](https://github.com/zendeskgarden/react-containers/pull/108/files#diff-792be63542e4df272f763d73537a4f23R24) React component, the props are typed annotated, but also uses React's `propType` checking. @jzempel if you have an answer I'd love to hear your opinion too!

packages/avatars/src/components/Avatar.tsx
jzempel: Good question, but no. The styling is directly impacted by the 3-part status. In the higher-abstraction `Avatar` component, we only deal with the two statuses – and any occurrence of `badge` overrides to ensure that the component is styled as `active`.

hzhu: Thanks for the explanation!

packages/avatars/README.md
```diff
@@ -9,7 +9,7 @@ This package includes components relating to avatars in the
 npm install @zendeskgarden/react-avatars
 
 # Peer Dependencies - Also Required
-npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
```
austin94: I believe this should still be a peer-dependency since we are including `propTypes` along with the type defintiions.

packages/avatars/src/components/Avatar.tsx
jzempel: I've restored this.

PR #481, opened 2019-10-11
fix(datepicker): ensure mouse events are propogated correctly

austin94:
## Description

There is currently a bug where the `DatepickerRange` doesn't persist values correctly if certain inputs are not `blurred` before a mouse interaction.

## Detail

This PR corrects the issue above by updating the blurred state if either the start or end input receives focus.

I have also updated a visual bug where the selected day was receiving the `bold` font weight instead of `semi-bold`

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/datepickers/src/DatepickerRange/utils/datepicker-range-reducer.ts
+++ b/packages/datepickers/src/DatepickerRange/utils/datepicker-range-reducer.ts
@@ -125,7 +125,7 @@ export const datepickerRangeReducer = ({
         }
       }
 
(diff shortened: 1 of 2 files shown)


hzhu:
Should we write a unit test to accompany this bug fix?

austin94:
> Should we write a unit test to accompany this bug fix?

Looking at the branch coverage it looks like this is actually already covered. I think this has something to do with event ordering which seems to be fine in the JSDOM environment.

PR #487, opened 2019-10-16
chore(internal): update react-popper dependencies

austin94:
## Description

This PR upgrades all `react-popper` dependencies across the packages.

## Detail

Because `Tooltip` is always rendering it's content in the DOM, and hiding it with `aria-hidden`, there was a noticeable flash of positioning whenever the tooltip became visible.

To help mitigate this flash I have added a quick `10ms` fade transition to the tooltip. This is fast enough to not be noticeable, but removes the initial flash of positioning. Please let me know if you know of a better strategy here.

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/tooltips/src/elements/Tooltip.js
+++ b/packages/tooltips/src/elements/Tooltip.js
@@ -5,14 +5,14 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { cloneElement } from 'react';
+import React, { cloneElement, useRef, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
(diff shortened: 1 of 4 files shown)


packages/tooltips/src/elements/Tooltip.js
```diff
@@ -23,13 +23,21 @@ import LightTooltip from '../views/LightTooltip';
  * due to constraints in our arrow css. We must ensure that the container
  * of the tooltip can retain it's relative positioning. Without this
  * container Popper would apply absolute positioning.
+ *
+ * This wrapper also includes an opacity transition. It allows Popper to
+ * re-position the tooltip without having a visible shift. The transition
+ * is fast enough that it should not be perceptable.
```
jzempel: *perceptible

PR #503, opened 2019-10-24
fix(tooltips): fix inconsistent default size prop

hzhu:
## Description

This pull request addresses the inconsistent default `size` prop for tooltips. 

## Details
Currently, the light tooltip has a default prop size of `LARGE` and the dark tooltip has a default prop size of `SMALL`. You can see this inconsistency in the "before" screenshots below. **This PR fixes the inconsistency by updating the light tooltip to have the same default prop as the dark tooltip, which is `small`.**

## Screenshots

**Before:** _Notice that the light tooltip has a default size prop of `large`, and dark tooltip has a default size of `small`. This is inconsistent._

Before

**After:** _Notice that the light tooltip has a default size prop of `small`, and dark tooltip has a default size of `small`. This is now consistent._

AFTER2

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/tooltips/src/views/LightTooltip.js
+++ b/packages/tooltips/src/views/LightTooltip.js
@@ -70,7 +70,7 @@ LightTooltip.propTypes = {
 
 LightTooltip.defaultProps = {
   arrow: true,
-  size: SIZE.LARGE
+  size: SIZE.SMALL
 };
 
 /** @component */
--- a/packages/tooltips/README.md
+++ b/packages/tooltips/README.md
@@ -24,7 +24,7 @@ Standard tooltip usages.
  */
 import '@zendeskgarden/react-tooltips/dist/styles.css';
 
-import { ThemingProvider } from '@zendeskgarden/react-theming';
+import { ThemeProvider } from '@zendeskgarden/react-theming';
 import { Tooltip } from '@zendeskgarden/react-tooltips';
 
 /**


packages/tooltips/src/views/LightTooltip.js
```diff
@@ -70,7 +70,7 @@ LightTooltip.propTypes = {
 
 LightTooltip.defaultProps = {
   arrow: true,
-  size: SIZE.LARGE
```
hzhu: Alternatively, the default `size` prop of both light and dark tooltip could be updated to be `SIZE.MEDIUM`.

allisonacs:
@hzhu I want to provide some background for you. We when implemented it this way, this was actually a "feature not a bug." At small sizes, Garden recommends using the dark tooltip as it is more noticeable, so dark tooltips are small size by default. Although technically possible, we don't recommend using small/light tooltips. On the other hand, when a toolitp has enough content to require a bigger font and left-aligned text, we recommend those tooltips are sized medium and we recommend those tooltips have a light background. Hence, the light default size is medium, not small. We were trying to create simple guardrails, making it easier to the right thing, and harder to the wrong thing. 

I can see where that would be unpredictable from an eng pov, so I'll leave it to y'all how to you want to think about this in code.

hzhu:
@allisonacs thanks for the clarification! Sorry I jumped the gun on this one.

PR #504, opened 2019-10-24
feat(datepicker): add theming values to react-datepicker

austin94:
## Description

This PR converts our design token usages in `react-datepickers` to use the theming values provided in the `next` branch.

## Checklist

      designer as a reviewer)
      component~


packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -52,36 +54,40 @@ export const StyledHeaderPaddle = styled.div<{
...
+  color: ${props => getColor('neutralHue', 600, props.theme)};
   cursor: pointer;
   border-radius: 50%;
 
   ${props => props.isHidden && `visibility: hidden;`};
 
   :hover {
-    color: ${PALETTE.grey[800]};
-    background-color: ${rgba(PALETTE.blue[600], 0.08)};
+    color: ${props => getColor('foreground', 800, props.theme)};
```
austin94: @jzempel let's have some sit-down time to figure out `foreground` vs `neutralHue` usages.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -52,36 +54,40 @@ export const StyledHeaderPaddle = styled.div<{
...
+    background-color: ${props => getColor('primaryHue', 600, props.theme, 0.08)};
   }
 
   :active {
-    background-color: ${rgba(PALETTE.blue[600], 0.2)};
-    color: ${PALETTE.grey[800]};
+    background-color: ${props => getColor('primaryHue', 600, props.theme, 0.2)};
+    color: ${props => getColor('foreground', 800, props.theme)};
```
jzempel: Never use `getColor` for `foreground` – only `props.theme.colors.foreground`. But in this case, I think it's a variation on `neutralHue`. But we can go over together.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -108,11 +122,15 @@ export const StyledRangeCalendar = styled.div`
     padding: 0;
   }
 
-  ${StyledDatepicker}:${props => (isRtl(props) ? 'last-of-type' : 'first-of-type')} {
-    margin-right: ${DEFAULT_THEME.space.md};
+  ${StyledDatepicker}:${props => (props.theme.rtl ? 'last-of-type' : 'first-of-type')} {
+    margin-right: ${props => props.theme.space.md};
```
jzempel: Use `space.base`.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -146,30 +172,31 @@ const retrieveStyledDayColor = ({
...
+    return getColor('background', 600, theme);
   }
 
   if (isToday) {
     return 'inherit';
   }
 
   if (isPreviousMonth) {
-    return PALETTE.grey[600];
+    return getColor('forground', 600, theme);
```
jzempel: *`forground` misspelled.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -146,30 +172,31 @@ const retrieveStyledDayColor = ({
   isDisabled,
   isSelected,
   isToday,
-  isPreviousMonth
+  isPreviousMonth,
+  theme
 }: IStyledDayProps) => {
   if (isDisabled) {
-    return PALETTE.grey[400];
+    return getColor('foreground', 400, theme);
```
jzempel: These should all be `neutralHue`.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -146,30 +172,31 @@ const retrieveStyledDayColor = ({
...
   }
 
   if (isSelected && !isDisabled) {
-    return PALETTE.white;
+    return getColor('background', 600, theme);
```
jzempel: Should be `props.theme.colors.background`.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -184,10 +211,10 @@ export const StyledDay = styled.div<IStyledDayProps>`
   justify-content: center;
   align-items: center;
   border-radius: 50%;
-  font-size: ${props => (props.small ? DEFAULT_THEME.fontSizes.sm : DEFAULT_THEME.fontSizes.md)};
+  font-size: ${props => (props.small ? props.theme.fontSizes.sm : props.theme.fontSizes.md)};
```
jzempel: Shouldn't this be `props.isCompact` (formerly known as `isSmall`)?

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -5,44 +5,46 @@
...
-  isRtl,
-  retrieveComponentStyles
-} from '@zendeskgarden/react-theming';
-import rgba from 'polished/lib/color/rgba';
+import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
+import { retrieveComponentStyles, getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import math from 'polished/lib/math/math';
 
-const retrieveSpacing = ({ isSmall }: { isSmall?: boolean }) => {
+const retrieveSpacing = ({ isSmall, theme }: { isSmall?: boolean } & ThemeProps<DefaultTheme>) => {
```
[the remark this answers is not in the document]
austin94: This aligns with the public prop `small` as defined by design when we created the initial API. I don't think making this API differ from the public prop or have a breaking change for this since there isn't a noticeable gain.

packages/datepickers/src/styled/styled-datepicker.tsx
```diff
@@ -5,44 +5,46 @@
...
   }
 
-  return DEFAULT_THEME.space.xl;
+  return theme.space.xl;
 };
 
 export const StyledDatepicker = styled.div<{
   isSmall: boolean;
 }>`
   /* stylelint-disable */
```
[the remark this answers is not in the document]
austin94: This was needed with a previous version of our stylelint config. Was able to remove.

packages/datepickers/src/styled/styled-datepicker.tsx
austin94: This was left as `small` to align with the public prop. I have renamed it to `isSmall` to help differentiate. I will rename it to `isCompact` based on our discussion above.

packages/datepickers/src/styled/styled-datepicker.tsx
austin94: Was hoping to have this discussion above with my opening question. We will need to have a 100-900 scale with the `foreground` color to accurately theme icon buttons and other content which uses the "text" colors.

Will move to neutral for now.

packages/datepickers/src/Datepicker/Datepicker.tsx
```diff
@@ -330,7 +330,8 @@ Datepicker.defaultProps = {
   animate: true,
   eventsEnabled: true,
   zIndex: 1000,
-  locale: 'en-US'
+  locale: 'en-US',
+  theme: DEFAULT_THEME
```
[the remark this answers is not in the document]
austin94: We would also need access to the default `document` in many of the other components where `withTheme` is used.

Is there a specific reason that we wouldn't treat these elements/usages the same as the `styled` components with their default theme?

jzempel:
Crossing out the "designer approved" checklist item is no good. Let's figure out how to get these changes in front of a designer. In addition, it's possible that a basic example should be enhanced with knobs to control all of the styling variations.

packages/datepickers/src/Datepicker/Datepicker.tsx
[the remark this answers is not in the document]
austin94: As discussed offline let's use the `defaultProps` approach to align with our other usages in the `styled` elements

packages/datepickers/src/styled/StyledDay.ts
```diff
@@ -0,0 +1,96 @@
...
+  :hover {
+    background-color: ${getColor('primaryHue', 600, props.theme, 0.08)};
+    color: ${getColor('primaryHue', 800, props.theme)};
+  }
+
+  :active {
+    background-color: ${getColor('primaryHue', 600, props.theme, 0.2)};
+    color: ${getColor('primaryHue', 800, props.theme)};
+  }
+  `}
```
jzempel: The scattered collection of color values throughout this file is difficult to rationalize. After having taken a variety of approaches with CSS ordering (and failed), I want us to consistently the following approach:

1. Base/core CSS properties (display, position, transition, direction, resets, etc.)
1. `${sizeStyles(props)}` – a function that collects all size-related properties (height, width, font-size, line-height, padding, etc) into one location.
    1. associated size pseudo-classes (in LVHFA order) – however, these should be very rare.
1. Psuedo-classes related to base/core properties (LVHFA order)
1. `${colorStyles(props)}` – a function with ALL color styles (border-color, background-color, color, box-shadow)
    1. with psuedo-class properties related to color (LVHFA order)
    1. and color related to `:disabled`/`[aria-disabled]`
1. Base/core `:disabled`/`[aria-disabled]` propreties.
1. Associated styled-component styles (if needed) – repeat the ordering shown above
1. `retrieveComponentStyles`

Again, after having done this for a long time, I believe this gives us the advantage of consolidating the styles most-impacted over time (size & color) into predictable locations. Also, the macro ordering is indicative of what needs to take CSS precedence.

Take a closer look at existing patterns under:
- https://github.com/zendeskgarden/react-components/tree/next/packages/buttons/src/styled
- https://github.com/zendeskgarden/react-components/tree/next/packages/tags/src/styled
- https://github.com/zendeskgarden/react-components/tree/next/packages/typography/src/styled
- https://github.com/zendeskgarden/react-components/tree/jzempel/styled-forms/packages/forms/src/styled (in progress)

Study and repeat what you see there. Let's talk if you feel like those patterns are not working here.

packages/datepickers/src/styled/StyledDay.ts
```diff
@@ -0,0 +1,96 @@
...
+const StyledDay = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID
+})<IStyledDayProps>`
+  display: flex;
+  position: absolute;
+  align-items: center;
+  justify-content: center;
+  border-radius: 50%;
+  background-color: ${retrieveBackgroundColor};
+  cursor: ${props => (props.isDisabled ? 'inherit' : 'pointer')};
```
jzempel: There shouldn't be an `isDisabled` prop here (or throughout `datepickers`). Use `[aria-disabled]` if applying `:disabled` styling to html element that does not accept the `disabled` attribute.

packages/datepickers/src/styled/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  return `${theme.space.base * 10}px`;
+};
+
+export const boldedStyling = css<{ isSmall: boolean }>`
+  line-height: ${props => props.theme.lineHeights.md};
+  font-size: ${props => (props.isSmall ? props.theme.fontSizes.sm : props.theme.fontSizes.md)};
+  font-weight: ${props => props.theme.fontWeights.semibold};
+`;
```
jzempel: This is an anti-pattern – only used 2 places and combining size-related (`line-height` + `font-size`) together with base/core (`font-weight`) properties. Providing these inline is more readable and maintainable over time.

packages/datepickers/src/styled/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+export const retrieveSpacing = ({
+  isSmall,
+  theme
+}: { isSmall?: boolean } & ThemeProps<DefaultTheme>) => {
+  if (isSmall) {
+    return `${theme.space.base * 8}px`;
+  }
+
+  return `${theme.space.base * 10}px`;
+};
```
jzempel: The fact that this value is subsequently being `math`-multiplied feels hacky and difficult to make sense of when reading through the CSS blocks where it is being called. The value of this utility is questionable – I almost wonder if you should provide a base `Styled` components that provide consistent height/width and can be extended.

packages/datepickers/styleguide.config.js
```diff
@@ -61,8 +61,8 @@ module.exports = {
     {
       name: 'Components',
       components: [
-        '../../packages/datepickers/src/Datepicker/[A-Z]*.{ts,tsx}',
-        '../../packages/datepickers/src/DatepickerRange/[A-Z]*.{ts,tsx}'
+        '../../packages/datepickers/src/elements/Datepicker/[A-Z]*.{ts,tsx}',
+        '../../packages/datepickers/src/elements/DatepickerRange/[A-Z]*.{ts,tsx}'
```
jzempel: TY!

packages/datepickers/src/styled/index.ts
```diff
@@ -5,5 +5,14 @@
...
+export { default as StyledDatepicker } from './StyledDatepicker';
+export { default as StyledRangeCalendar } from './StyledRangeCalendar';
+export { default as StyledHeader } from './StyledHeader';
+export { default as StyledHeaderPaddle } from './StyledHeaderPaddle';
+export { default as StyledHeaderLabel } from './StyledHeaderLabel';
+export { default as StyledCalendar } from './StyledCalendar';
+export { default as StyledCalendarItem } from './StyledCalendarItem';
+export { default as StyledDayLabel } from './StyledDayLabel';
+export { default as StyledHighlight } from './StyledHighlight';
+export { default as StyledDay } from './StyledDay';
```
jzempel: Default exports are a departure from what we've done with other CSS-in-JS refactors.

packages/datepickers/src/styled/StyledHighlight.ts
```diff
@@ -0,0 +1,48 @@
...
+    props.isStart &&
+    `
+    border-radius: ${props.theme.rtl ? '0 50% 50% 0' : '50% 0 0 50%'};
+  `}
+
+  ${props =>
+    props.isEnd &&
+    `
+    border-radius: ${props.theme.rtl ? '50% 0 0 50%' : '0 50% 50% 0'};
+  `}
```
jzempel: Pull out `border-radius: ${props => ...}` and simplify the logic for which of the two `border-radius` values is used.

packages/datepickers/src/styled/StyledCalendarItem.ts
```diff
@@ -0,0 +1,33 @@
...
+const StyledCalendarItem = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID
+})<{
+  isSmall?: boolean;
+}>`
+  display: inline-block;
+  position: relative;
+  width: ${retrieveSpacing};
+  height: ${retrieveSpacing};
+  vertical-align: top;
```
jzempel: Just to be clear, when properties are easy to take in all in one glance – as they are here – it does not make sense to worry about breaking out size/color-related styles. So nice setup here. Although, it would even be better if the `retrieveSpacing` were simply included inline where it can be reasoned about in the context of the styled element.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -0,0 +1,56 @@
...
+  justify-content: center;
+  transform: ${props => props.theme.rtl && 'rotate(180deg)'};
+  border-radius: 50%;
+  cursor: pointer;
+  width: ${retrieveSpacing};
+  height: ${retrieveSpacing};
+
+  color: ${props => getColor('neutralHue', 600, props.theme)};
+
+  ${props => props.isHidden && `visibility: hidden;`};
```
jzempel: Pull `visibility` out so that properties can be read together as `property: value`. In this case `visibility: ${props => props.isHidden && 'hidden'}`. Don't remove it from the ordering of the base/core properties above.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { retrieveSpacing } from './utils';
+
+const COMPONENT_ID = 'datepickers.header_paddle';
+
+const StyledHeaderPaddle = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID
+})<{
+  isSmall: boolean;
+  isHidden?: boolean;
```
jzempel: Consider the value of connecting with the native global `[hidden]` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden. YMMV. But the fact that you are specifying the `display` below makes this a viable option.

.stylelintrc
```diff
@@ -6,9 +6,10 @@
   ],
   "rules": {
     "declaration-empty-line-before": null,
+    "declaration-colon-newline-after": null,
     "keyframes-name-pattern": null,
     "value-keyword-case": ["lower", { ignoreKeywords: ['dummyValue'] }],
     "block-no-empty": null,
-    "number-leading-zero": null
+    "number-leading-zero": null,
```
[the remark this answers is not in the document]
austin94: Weird, looks like VS code included this formatting. Removed.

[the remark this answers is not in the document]
jzempel: 10b. Pshaw!

packages/datepickers/src/styled/StyledDatepicker.ts
```diff
@@ -0,0 +1,32 @@
...
+  'data-garden-id': COMPONENT_ID
+})<{
+  isSmall: boolean;
+}>`
+  background-color: ${props => props.theme.colors.background};
+  padding: ${props =>
+    `${props.isSmall ? props.theme.space.base * 4 : props.theme.space.base * 5}px`};
+  color: ${props => props.theme.colors.foreground};
+
+  direction: ${props => props.theme.rtl && 'rtl'};
```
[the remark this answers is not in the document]
austin94: I've removed the space, but stylelint is saying that this ordering is correct.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
austin94: I think in a previous version of `styled-components` this would have actually rendered `visibility: false`. It looks like this has been cleaned up. Updated.

.stylelintrc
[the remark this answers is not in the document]
hzhu: Pulled the latest branch down and it looks like the comma is still there.

packages/datepickers/src/styled/StyledDatepicker.ts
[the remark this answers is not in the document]
hzhu: The value for the `direction` attribute can be either `false` or `"rtl"`, which are two different types. It probably doesn't affect the UI, but should this line be re-written as:

```tsx
  direction: ${props => props.theme.rtl ? 'rtl': 'ltr'};
```

I'm curious why TypeScript didn't complain here, I'm guessing it's because the value for `direction` is a function that gets evaluated? Or maybe I'm misunderstanding how this works. TypeScript does complain about using the `&&` in CSS over [here](https://github.com/zendeskgarden/react-containers/pull/111#discussion_r332165935) in the react-containers repo.

packages/datepickers/src/styled/StyledDay.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import styled, { DefaultTheme, ThemeProps } from 'styled-components';
+import { retrieveComponentStyles, getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+interface IStyledDayProps extends ThemeProps<DefaultTheme> {
+  isPreviousMonth?: boolean;
+  isToday?: boolean;
+  isDisabled?: boolean;
+  isSelected?: boolean;
+  isSmall: boolean;
```
hzhu: I noticed @jzempel commented on some sizing props elsewhere in this PR about using `isCompact` rather than `isSmall`. Would that apply to this line of code as well?

packages/datepickers/src/styled/StyledDatepicker.ts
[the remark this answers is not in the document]
austin94: `styled-components` has some 🌈 "magic" which helps decide when to apply your style rule.

In this case `false` (or `undefined` as well) are not valid values for the `direction` style rule and are not applied. This stops you from accidentally having `color: undefined`'s everywhere.

packages/datepickers/src/styled/StyledDay.ts
austin94: From our offline conversation, the combination of the states and `isDisabled` logic make this difficult to read with an attribute.

I have refactored this area to a helper function to help with readability.

packages/datepickers/src/styled/StyledDatepicker.ts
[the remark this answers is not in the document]
hzhu: Great thanks for the clarification 👍 .

packages/datepickers/src/styled/StyledMenu.tsx
```diff
@@ -64,11 +64,11 @@ interface IStyledMenuViewProps extends HTMLProps<HTMLUListElement> {
  * Accepts all `<ul>` props
  */
 const StyledMenuView = styled.ul.attrs<IStyledMenuViewProps>(props => ({
-  'data-garden-id': COMPONENT_ID,
+  'data-garden-id': (props as any['data-garden-id']) || COMPONENT_ID,
```
jzempel: Did you determine this "or" treatment is needed after all?

packages/datepickers/src/styled/StyledRangeCalendar.ts
```diff
@@ -0,0 +1,36 @@
...
+  ${StyledDatepicker} {
+    padding: 0;
+  }
+
+  ${props =>
+    props.theme.rtl
+      ? `${/* sc-selector */ StyledDatepicker}:last-of-type {margin-right: ${props.theme.space
+          .base * 5}px}`
+      : `${/* sc-selector */ StyledDatepicker}:first-of-type {margin-right: ${props.theme.space
+          .base * 5}px}`}
```
jzempel: Can this move under the `${StyledDatepicker}` block above as `&:last-of-type` and `&:first-of-type`?

packages/datepickers/src/styled/index.ts
jzempel: Why was this resolved?

packages/datepickers/src/styled/StyledRangeCalendar.ts
jzempel: I've used a construct like this in other packages:

```js
&:{props => props.theme.rtl ? 'first' : 'last'}-of-type: { margin-right: ...; }
```

jzempel: Oops, please note that Abstract specifies 16px space between calendars in compact mode.

packages/datepickers/src/styled/index.ts
austin94: I've corrected this to use named exports. Are you seeing the most up-to-date file?

packages/datepickers/src/styled/StyledMenu.tsx
austin94: This isn't needed.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -0,0 +1,73 @@
...
+}>`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transform: ${props => props.theme.rtl && 'rotate(180deg)'};
+  visibility: ${props => props.isHidden && 'hidden'};
+  border-radius: 50%;
+  cursor: pointer;
+
+  ${/* sc-block */ retrieveSizing}
```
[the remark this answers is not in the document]
austin94: We have used this in other styled-component usages for some time. This helps stylelint infer what type of rules to use.

Take a look at https://www.styled-components.com/docs/tooling#tags

packages/datepickers/src/elements/Datepicker/Datepicker.tsx
```diff
@@ -330,7 +330,8 @@ Datepicker.defaultProps = {
   animate: true,
```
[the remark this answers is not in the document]
austin94: `isAnimate` isn't grammatically correct.  The `animate`, `arrow` and other props are shared with the `Menu` and `Tooltip` components. If we want to change those I believe a separate PR would help address it consistently.

.stylelintrc
```diff
@@ -6,6 +6,7 @@
   ],
   "rules": {
     "declaration-empty-line-before": null,
+    "declaration-colon-newline-after": null,
```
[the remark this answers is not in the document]
austin94: This rule conflicts with prettier for our interpolation tagging. This aligns it with the `stylelint-config-styled-components` configuration.

packages/datepickers/src/styled/StyledRangeCalendar.ts
austin94: This value was purposefully increased during development with the designer. Let's revisit this as a separate task if this is something that we want to update.

austin94: > I've used a construct like this in other packages

Prettier starts to format values into invalid CSS with different combinations. This was the most readable option with formatting for me.

I have raised it into the parent `${StyledDatepicker}` block though.

packages/datepickers/src/styled/StyledDay.ts
```diff
@@ -0,0 +1,86 @@
...
+}))<IStyledDayProps>`
+  display: flex;
+  position: absolute;
+  align-items: center;
+  justify-content: center;
+  border-radius: 50%;
+  cursor: ${props => (props.isDisabled ? 'inherit' : 'pointer')};
+  width: 100%;
+  height: 100%;
+  line-height: ${props => props.theme.lineHeights.md};
```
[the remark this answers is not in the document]
austin94: Removed as this isn't needed.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
jzempel: (per offline convo)

packages/datepickers/src/elements/Datepicker/Datepicker.tsx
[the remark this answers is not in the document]
jzempel: Yeah, let's revisit.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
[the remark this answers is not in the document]
jzempel: TIL

PR #505, opened 2019-10-30
feat(loaders): migrate theming and TS to react-loaders

austin94:
## Description

This PR migrates `react-loaders` to be "next ready". This includes:
- new package structure to align with other packages
- migration to TypeScript
- adding theming support

## Detail

Now that we have a few of these migrations under our belt I think there are some opportunities to help make this process more reproducible in the future for the implementers and reviewers.

To help simplify the diffs for future upgrades I believe that splitting the upgrades into 2 PRs could be beneficial.

1. Migrate to TS and new package structure
  * Moving to TS helped me _immensely_ with this upgrade.
  * Accomplishing this as early as possible will help with correct theming and making sure our API surfaces aren't changing unintentionally.
  * Since we're already renaming files in this step it worked well to introduce the `examples/elements/styled/utils` file structure here
2. Introduce CSS-in-JS and theming
  * With TS support it becomes much easier to migrate each `styled` component to our theming values.

@jzempel @hzhu let me know what you think of ☝️ 

## Checklist

      designer as a reviewer)
      component~


austin94:
Updated demo site available at https://austin-next.netlify.com/loaders/

jzempel:
@austin94 I think the approach you've described is sound. The only thing I would add (for clarification) under the second "Introduce CSS-in-JS and theming" is:

- a knobs-style example demo'ing most/all of the component props.

We _must_ have this to support thorough visual review from ourselves (as we go about moving the CSS into JS) and from others (both developers and designers). It is also critical to help frame component context for our documentation refactor. This step cannot be skipped.

Please reference the following `next` branch examples for expectations:
- https://competent-shirley-cc2f82.netlify.com/avatars/
- https://competent-shirley-cc2f82.netlify.com/forms/
- https://competent-shirley-cc2f82.netlify.com/loaders/ (not bad, although several of these should just be consolidated into a *Basic* demo and the "Basic Usage" examples are not super valuable – adding visual clutter to the documentation).
- https://competent-shirley-cc2f82.netlify.com/tags/
- https://competent-shirley-cc2f82.netlify.com/typography/

packages/loaders/src/styled/StyledInlineTypingIndicator.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)}
+`;
+
+StyledInlineTypingSVG.defaultProps = {
+  theme: DEFAULT_THEME
+};
+
+export const StyledInlineTypingIndicator: React.FC<IStyledTypingSvgProps> = props => {
```
jzempel: Why is this under `styled`? Feels like the structure should be represented under `elements/Inline`.

jzempel:
Oh @austin94, which PR (one or two) are the prop alignment changes intended to go under?

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -0,0 +1,121 @@
...
+  theme,
+  dark
+}: IStyledSkeletonProps & ThemeProps<DefaultTheme>) => {
+  // Disabling stylelint due to conflicts with prettier and linear-gradient formatting
+  return css`
+    /* stylelint-disable */
+    background-image: linear-gradient(
+      ${theme.rtl ? '-45deg' : '45deg'},
+      transparent,
+      ${dark ? getColor('kale', 700, theme, 0.4) : rgba(theme.colors.background, 0.6)},
```
jzempel: This should be using `getColor('chromeHue', ...)`.

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -0,0 +1,121 @@
...
+};
+
+export const StyledSkeleton = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSkeletonProps>`
+  display: inline-block;
+  position: relative;
+  animation: ${fadeInAnimation} 750ms linear;
+  border-radius: ${props => props.theme.space.base}px;
```
jzempel: Use `props.theme.borderRadii.md`.

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -0,0 +1,121 @@
...
+  width: ${props => props.customWidth};
+  height: ${props => props.customHeight};
+  overflow: hidden;
+  line-height: 0.6;
```
jzempel: Line height is typically a function of intended pixel height divided by font-size. How may that be represented here?

packages/loaders/src/styled/StyledProgress.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+type SIZE = 'small' | 'medium' | 'large';
+
+const HEIGHTS: Record<string, number> = {
+  small: 2,
+  medium: 6,
+  large: 12
```
jzempel: How do these values relate to the DEFAULT_THEME? Are they based on `space` units? If so (similar to grid), these should be pulled (or calculated on `base`) accordingly.

packages/loaders/src/styled/StyledInlineTypingIndicator.tsx
austin94: Moved to the `element` area.

packages/loaders/src/styled/StyledSkeleton.ts
austin94: Oh neat. Didn't realize we had a `chromeHue`!

packages/loaders/src/styled/StyledSkeleton.ts
austin94: The `Skeleton` component is meant to fill any element it is put in. Since both of those values will always be inherited is there a way to calculate this?

packages/loaders/src/styled/StyledProgress.ts
austin94: We shouldn't be using DEFAULT_THEME to calculate these since that wouldn't make the values updated if `base` is themed, right? I've moved this to a utility method which uses the current theme to calculate the values from `theme.space.base`

austin94:
> which PR (one or two) are the prop alignment changes intended to go under?

@jzempel I lean toward the first one. For many of the components we are wanting to introduce non-theming breaking changes to some of the API (using context instead of manually iterating children).

The less we have in the CSS-in-JS PR the better since much of the styling is changing from the CSS world.

packages/loaders/src/styled/StyledSkeleton.ts
jzempel: There may not be. Although, usually our designers identify values with some basis in mind and it is nice when we can capture that intent in code. For example, we frequently see 14px fonts occupying 20px of space. Therefore the (general) intent is a 1.42857 line-height, regardless of font size. (Side note: these unitless values tend to lose their effectiveness at extreme sizes.)

packages/loaders/src/elements/Skeleton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyledSkeleton } from '../styled';
+
+interface ISkeletonProps extends React.HTMLAttributes<HTMLDivElement> {
+  width?: string;
+  height?: string;
+  /** Apply styling for use with dark backgrounds */
+  dark?: boolean;
```
jzempel: Oops, this should be renamed to `isDark` per our boolean prop guidelines.

packages/loaders/src/styled/StyledSkeleton.ts
austin94: I now calculate this based on the `fontSize` and `lineHeights` provided by the theme.

PR #507, opened 2019-10-30
feat(forms): CSS-in-JS refactor

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

- BREAKING CHANGES
  - several COMPONENT_ID renames for consistency
  - see https://github.com/zendeskgarden/react-components/blob/jzempel/styled-forms/BREAKING_CHANGES.md for details
- Restructured source
  - `src/elements` – high-abstraction package components
  - `src/styled` – styled-components variants
- Features
  - Add Checkbox `indeterminate` prop
- All tag CSS contained within components and themable! (no more CSS module dependencies)

## Detail

- TypeScript to be handled as a follow-on PR.
- Demo pre-published for review: https://competent-shirley-cc2f82.netlify.com/forms/

## Checklist

      designer as a reviewer)
      component~


.stylelintrc
```diff
@@ -5,10 +5,12 @@
...
   ],
   "rules": {
+    "declaration-colon-newline-after": null,
     "declaration-empty-line-before": null,
     "keyframes-name-pattern": null,
     "value-keyword-case": ["lower", { ignoreKeywords: ['dummyValue'] }],
     "block-no-empty": null,
-    "number-leading-zero": null
+    "number-leading-zero": null,
+    "selector-type-no-unknown": null
```
austin94: Does this remove the need for us to use the `/* sc-selector */` comments!? That's awesome

BREAKING_CHANGES.md
```diff
@@ -24,6 +24,16 @@
 
 - removed; see `@zendeskgarden/react-forms`
 
+## @zendeskgarden/react-forms
+
+- modified theming COMPONENT_IDs for alignment with `data-garden-id`
```
austin94: Do we want to call these out under breaking changes? With us changing the structure and styling of the components these would probably break for people regardless.

packages/forms/src/elements/Checkbox.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Must be rendered within a `<Field>` element; accepts all
+ * `<input type="checkbox">` attributes and events.
+ */
+const Checkbox = React.forwardRef(({ indeterminate, children, ...props }, ref) => {
+  const { getInputProps } = useFieldContext();
+  const inputRef = inputElement => {
+    inputElement && (inputElement.indeterminate = indeterminate);
+  };
+  const combinedRef = inputElement => {
```
austin94: Just leaving a note for us to revisit why the `useCombinedRefs` utility didn't work for this use-case.

packages/forms/src/elements/MultiThumbRange.js
```diff
@@ -7,101 +7,37 @@
 
 import React, { useState, useEffect, useCallback, useRef } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
-import styled from 'styled-components';
 import debounce from 'lodash.debounce';
 import { KEY_CODES } from '@zendeskgarden/container-utilities';
+import { withTheme, isRtl, getDocument } from '@zendeskgarden/react-theming';
```
austin94: Could we reference the theme directly here instead of using the `isRtl` helper? Same with `getDocument`

packages/forms/src/index.js
```diff
@@ -5,25 +5,21 @@
...
+export { default as Radio } from './elements/Radio';
+export { default as Range } from './elements/Range';
+export { default as Textarea } from './elements/Textarea';
+export { default as Toggle } from './elements/Toggle';
 
-export const styled = styledElements;
+/** Other */
+export { default as FauxInput } from './elements/FauxInput';
+export { default as MediaInput } from './elements/MediaInput';
+export { default as MultiThumbRange } from './elements/MultiThumbRange';
```
austin94: Even though this isn't backed by a native `<input />` I still classify the `MultiThumbRange` as a field. Should this be moved up to the `Fields` block?

packages/forms/src/styled/checkbox/StyledCheckInput.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const dashSvg = props => {
+  const color = props.theme.colors.background;
+
+  return `
+    <svg xmlns="http://www.w3.org/2000/svg" width="12" height="12">
+      <path
+        fill="none"
+        stroke="${color}"
```
austin94: Just to confirm, the `currentColor` value wouldn't work here since we are apply this as a background-image?

packages/forms/src/styled/common/StyledField.js
```diff
@@ -0,0 +1,33 @@
...
+
+const COMPONENT_ID = 'forms.field';
+
+export const StyledField = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  direction: ${props => props.theme.rtl && 'rtl'};
+  /* as <fieldset> resets */
+  margin: 0;
```
austin94: Since this is locked as a `div` do we need these resets from the CSS side?

packages/forms/src/styled/radio/StyledRadioInput.spec.js
```diff
@@ -6,13 +6,40 @@
...
     const { container } = render(<StyledRadioInput />);
 
+    expect(container.firstChild.nodeName).toBe('INPUT');
     expect(container.firstChild).toHaveAttribute('type', 'radio');
   });
+
+  it('renders expected checked styling', () => {
+    const { container } = render(<StyledRadioInput defaultChecked />);
+
+    expect(container.firstChild).toHaveStyleRule('background-color', PALETTE.blue[600], {
```
austin94: Open question, should we be asserting that our elements have applied the _correct color_ or the _correct hue_?

If we are saying that the element should be applying `primary600` it may make more sense to assert:

```js
expect(container.firstChild).toHaveSytleRule('background-color', getColor('primaryHue', 600, DEFAULT_THEME));
```

BREAKING_CHANGES.md
jzempel: Probably fair to add a "blanket statement" to the effect that all referenced COMPONENT_IDs should be re-checked for naming accuracy.

packages/forms/src/styled/checkbox/StyledCheckInput.js
jzempel: Correct, it would not work.

packages/forms/src/styled/radio/StyledRadioInput.spec.js
jzempel: I think you're right. For example if `primaryHue` changes some day, we don't want all these tests to break. Just me ... being lazy. I'll get these spruced up.

.stylelintrc
jzempel: I think so – probably why I didn't know about the https://www.styled-components.com/docs/tooling#interpolation-tagging 😮

packages/forms/src/elements/Checkbox.js
jzempel: Yeah, meant to mention. In `useCombinedRefs` the `useEffect` messes with the timing of applying `indeterminate` to the underlying DOM node. The approach taken here can be seen in https://github.com/smooth-code/react-merge-refs/blob/master/src/index.js and might be a case for adding a `mergeRefs` utility.

packages/forms/src/styled/common/StyledField.js
jzempel: It is valid (and perhaps desirable – see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#example – in limited cases) to use `<Field as="fieldset">`. It's trivial for us to keep these and style accordingly.

packages/forms/src/elements/MultiThumbRange.js
jzempel: Removed `isRtl`; left `getDocument` as it guarantees a valid document (which does not exist in the default theme).

packages/forms/src/elements/FauxInput.js
```diff
@@ -0,0 +1,58 @@
...
+  );
+});
+
+FauxInput.propTypes = {
+  /** Apply compact styling */
+  isCompact: PropTypes.bool,
+  /** Remove borders and padding */
+  isBare: PropTypes.bool,
+  /** Apply inset `box-shadow` styling on focus */
+  focusInset: PropTypes.bool,
```
hzhu: This prop doesn't appear to be a native HTML attribute. Is there a particular reason why this prop is not using the `isBoolean` naming format (e.g. `isFocusInset`)?

jzempel: This prop is used throughout: `forms`, `dropdowns`, `buttons`. I'm fine with an improved name (`hasFocusInset`?), but we'll need to agree that this is the right thing to do. Let's discuss. /cc @austin94

jzempel: `hasInsetFocus`?

packages/forms/src/elements/Checkbox.js
```diff
@@ -0,0 +1,57 @@
...
+          targetRef(inputElement);
+        } else {
+          targetRef.current = inputElement;
+        }
+      }
+    });
+  };
+
+  return (
+    <CheckboxContext.Provider value={{}}>
```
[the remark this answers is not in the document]
jzempel: The context object is used within the common `Hint`, `Label` and `Message` components to select the correct styled component that corresponds with the given field. The `value` is superfluous. I'm not totally sure this is the ideal pattern, but it's what we have in the original: https://github.com/zendeskgarden/react-components/blob/master/packages/forms/src/fields/common/Label.js.

packages/forms/src/elements/Range.js
```diff
@@ -34,33 +33,25 @@ const Range = React.forwardRef(({ min, max, step, ...otherProps }, ref) => {
 
       setBackgroundSize(`${percentage}%`);
     },
-    [max, min]
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
+    [max, min, step]
```
hzhu: Could you clarify the purpose of adding `step` into the dependency array?

I pulled down the branch locally and remove `step` from the dependency array and did not notice a difference with `<Range>`. Does the `react-hooks/exhaustive-deps` lint rule not apply in this particular situation? I think that `step` is not needed in the dependency array since `step` isn't used within the `useCallback` code — but maybe there is something I'm missing?

jzempel: with `step` removed from the array:
- use the "knobs" to set the step value to 1
- slide thumb to a value
- change knob to 50
result: background size fails to update.

hzhu: Awesome, this makes sense. The `step` value is an implicit dependency.

packages/forms/src/elements/Checkbox.js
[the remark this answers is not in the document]
hzhu: > I'm not totally sure this is the ideal pattern

Awesome thank you for clarifying! I'm still wrapping my head around this pattern and don't have a suggestion. One thing we could do would be to replace the Provider's value of `{}` with `true`. That way that there wouldn't be [unintentional renders](https://reactjs.org/docs/context.html#caveats).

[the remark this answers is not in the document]
jzempel: I wonder if we should replace `useCheckboxContext`, `useRadioContext`, `useToggleContext` with a single `useInputContext` (or `useCheckedInputContext`) that receives a string indicating the style of input that the context is providing ("checkbox", "radio", "toggle").

@austin94 do you see any advantage to this?

packages/forms/examples/advanced.md
```diff
@@ -0,0 +1,162 @@
...
+      onChange={event =>
+        setState({
+          checkboxChecked: event.target.checked,
+          checkboxIndeterminate: event.nativeEvent.shiftKey
+        })
+      }
+    >
+      <Label>Controlled checkbox</Label>
+      <Hint>
+        Must be checked (hold shift key for <Code>indeterminate</Code>)
```
hzhu: The indeterminate checkbox works for me on latest Chrome, Safari, and IE11, but it doesn't seem to work on Firefox (68.0.2).

ginnywood:
@jzempel cool, all of that sounds good!

> hmm, nice catch. When did that change? We've had them spaced at 4px in the CSS since day one (https://garden.zendesk.com/css-components/forms/text/).

I’m looking back through our Abstract commits, and it looks like we’ve had them spaced at 8px since day one. I never noticed until I saw it side-by-side with validation messages on other components and realized they were different!

jzempel:
@ginnywood 

> I’m looking back through our Abstract commits, and it looks like we’ve had them spaced at 8px since day one. I never noticed until I saw it side-by-side with validation messages on other components and realized they were different!

I would say definitely off on the code side, as these fail to match-up with the corresponding checkbox/radio/toggle messages. I'll get it fixed. Again, great catch.

packages/forms/examples/advanced.md
jzempel: You found a cool issue with FF: https://bugzilla.mozilla.org/show_bug.cgi?id=559506. I'm not going to sub out the key, as there are various issues with the other options (alt, cmd, ctrl) – so let's leave this as a minor quirk limited to the example.

hzhu: @jzempel thanks for looking into it, much appreciated!

PR #512, opened 2019-11-05
fix(dropdowns): allow inline style to override Menu width

austin94:
## Description

By default the `Menu` component matches the width of the `Select` element which triggers it. In some use-cases the width of the `Select` can be dramatically smaller than a default width of a `Menu`.

This PR allows inline width overrides to take precedence over the width of the paired `Select`.

```jsx
<Menu style={{ width: 500 }}>
  // Renders at 500px width
</Menu>
```

## Checklist

      designer as a reviewer)~
      component~


jzempel:
How does this play well with `max-width` or `min-width` if those are desired?

austin94:
> How does this play well with max-width or min-width if those are desired?

This should work as expected in the non-Select use-cases. You could define `width: inherit` or another valid value

PR #519, opened 2019-11-08
fix(buttons): apply `isBoolean` prop updates and knob-style examples

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

...also:
- remove `min-width` CSS
- moved component exports under `elements`.

## Detail

Pre-published for review: https://competent-shirley-cc2f82.netlify.com/buttons/

## Checklist

      designer as a reviewer)
      component~


packages/buttons/src/elements/Button.tsx
```diff
@@ -38,19 +38,19 @@ const Button: React.FunctionComponent<
 > = React.forwardRef<HTMLButtonElement, IButtonProps>((props, ref) => {
   const focusInset = props.focusInset || useContext(ButtonGroupContext);
 
-  return <StyledButton ref={ref} focusInset={focusInset} {...props} />;
+  return <StyledButton ref={ref} {...props} focusInset={focusInset} />;
```
jzempel: Reordered to ensure `focusInset` takes precedence with group-styled buttons.

packages/buttons/src/styled/StyledExternalIcon.ts
```diff
@@ -21,6 +21,7 @@ export const StyledExternalIcon = styled(NewWindowIcon).attrs({
   transform: ${props => props.theme.rtl && 'scaleX(-1)'};
   margin-bottom: -0.085em;
   padding-left: 0.25em;
+  box-sizing: content-box;
```
jzempel: Discovered this issue and fix with new knobs-style examples.

ginnywood:
They’re so beautiful.
itssobeautiful

I think I found some visual bugs.
**Chevron button**
Focus inset on a primary button is inset by 1px, leaving a “border” outside the focus ring.
Screen Shot 2019-11-08 at 4 59 51 PM
**Split button**
With `default` applied, the border between the main button and the chevron button looks too thick. With `primary` applied, the text doesn’t look vertically centered (it looks low)
**Button group**
With `primary` applied, the text doesn’t look vertically centered (it looks low) and the borders between buttons look thicker than 1px
Screen Recording 2019-11-08 at 05 11 PM

jzempel:
hey @ginnywood good eye :wink:

There's a fair bit of border styling CSS connected with buttons that participate in a group (chevron, split button, group buttons). Trade-offs have been made to prevent expensive styling bloat for unlikely scenarios.

### re: Chevron Button

I updated the example to force apply `isBasic` prop styling if the solo chevron has `isPrimary` styling toggled. The default is for the chevron to keep it's border so that separation styling can be applied when it is added as part of a SplitButton – which is the typical case.

### re: group border thickness

The actual spacing in every case is 1px. However, the browser rendering can be a bit flaky as the treatment is is stack the 1px borders on top of each other and handle layering via `z-index`. The results are different from one browser/screen to the next. I'm not convinced there's a whole lot we can do here.

### re: text centering

I'm not seeing the vertical shift on my screen. The horizontal shift is expected as there are borders being knocked out. Let's look at this together. There's not much room for tweaking the button `line-height` CSS properties without messing things up.

PR #520, opened 2019-11-09
fix(avatars): img scaling and move `components` to `elements`

jzempel:
- [ ] **BREAKING CHANGE?** 

## Description

from zendeskgarden/css-components#249

## Checklist

      designer as a reviewer)
      component~


hzhu:
I noticed that `object-fit` [isn't supported](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) in IE and it was not discussed in zendeskgarden/css-components#249. Not sure if it's worth it, but there seems to be quite a few [fallback techniques](https://stackoverflow.com/questions/37792720/ie-and-edge-fix-for-object-fit-cover).

jzempel:
@hzhu this is a (rare) spot where Garden won't provide perfect fidelity for IE11 and that's ok. It's not worth imposing the fallbacks. Our answer would be "use a square image".

PR #524, opened 2019-11-11
feat(tabs): migrate react-tabs to CSS-in-JS

austin94:
## Description

This PR migrates `react-tabs` to `CSS-in-JS`. I've attempted to maintain the logic used in [css-components](https://github.com/zendeskgarden/css-components/tree/master/packages/tabs) since the logic for managing the `::before` focus-state is pretty complex.

If you can think of anything to simplify the focus logic please discuss below!

## Checklist

      designer as a reviewer)
      component~


austin94:
https://austin-next.netlify.com/tabs/ published for review

ginnywood:
I noticed an existing bug—unrelated to the CSS-in-JS work: the selected tab text should be semibold.

austin94:
@jzempel applying a `font-weight` of semiBold is causing a pixel-shift when selection changes. Was there a reason that tabs originally didn't include this styling?

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
-      }
-    `}
+  display: inline-block;
+  position: relative;
+  transition: color 0.25s ease-in-out;
+  border-bottom: ${props => props.theme.borderStyles.solid} transparent;
+  border-width: ${props => props.theme.borderWidths.md};
+  cursor: pointer;
+  padding: ${props => props.theme.space.base * 2.5}px ${props => props.theme.space.base * 7}px
+    ${props => props.theme.space.base * 1.5}px;
```
jzempel: This `padding-bottom` is a function of top padding less border widths and should be expressed as such. See https://github.com/zendeskgarden/css-components/blob/c2a6c1f7c10b9e5f86a89b0f64b4b2d677e51c0b/packages/tabs/src/_base.css#L11. Along with recovering anchor styling, this is where having a separate `sizeStyles` starts to make sense as our recommended approach.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
   isSelected?: boolean;
 }
 
+/**
+ * A high specificity is needed to apply the border-color in vertical orientations
+ */
+const colorStyles = ({ theme, isSelected }: IStyledTabProps & ThemeProps<DefaultTheme>) => css`
+  color: ${isSelected ? getColor('primaryHue', 600, theme) : 'inherit'};
+
+  &&& {
```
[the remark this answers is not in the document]
jzempel: or better, just calculate `border-color` the same way that you do with `color` above:

```css
border-color: ${isSelected ? getColor(...) : 'transparent'};
```

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
+  text-decoration: none;
+  text-overflow: ellipsis;
+
+  ${colorStyles}
+  ${focusStyles}
+
+  &[aria-disabled='true'] {
+    border-color: transparent;
+    cursor: default;
+    color: ${props => getColor('neutralHue', 400, props.theme)};
```
jzempel: Move this color-related value up to the `colorStyles` function (along with `border-color` above). Leave `cursor` here.

packages/tabs/src/styled/StyledTabs.ts
```diff
@@ -20,26 +19,74 @@ interface IStyledTabsProps {
...
+        border-left: 0;
+        border-right-style: ${theme.borderStyles.solid};
+        border-right-color: transparent;
+        text-align: right;
+      `}
+    }
+
+    ${StyledTabPanel} {
+      margin-left: ${theme.space.base * 8}px;
+      width: 100%;
```
jzempel: I'm curious where this `width` property came from and what it's supposed to be doing?

jzempel:
@ginnywood I suspect the semi-bold you're seeing is a bug with the Sketch symbol. We would have to think super hard about introducing that kind of treatment here as it is very much subject to the jitter that Austin mentioned. We've never had that as part of tabs CSS styling.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
+    color: ${getColor('primaryHue', 600, theme)};
+  }
+
+  &[data-garden-focus-visible]::before {
+    position: absolute;
+    top: ${theme.space.base * 2.5}px;
+    right: ${theme.space.base * 6}px;
+    left: ${theme.space.base * 6}px;
+    border-radius: ${theme.borderRadii.md};
+    box-shadow: inset ${theme.shadows.sm(getColor('primaryHue', 600, theme, 0.35)!)};
```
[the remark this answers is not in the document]
austin94: This would require us to have two areas that style the `::before` element on focus-visible. I found it harder to reason when focus styling is spaced around in multiple helper functions.

With this being one of the few components that focus a pseudo-element I'd prefer to keep this logic separate from the colors.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
-        box-shadow: none;
-      }
-    `}
+  display: inline-block;
+  position: relative;
+  transition: color 0.25s ease-in-out;
+  border-bottom: ${props => props.theme.borderStyles.solid} transparent;
+  border-width: ${props => props.theme.borderWidths.md};
+  cursor: pointer;
+  padding: ${props => props.theme.space.base * 2.5}px ${props => props.theme.space.base * 7}px
```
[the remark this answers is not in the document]
austin94: Why does this need to be retained? This functionality was never possible in the previous React components.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,102 @@
...
+  }
+
+  &:active::before {
+    box-shadow: none;
+  }
+`;
+
+const focusStyles = ({ theme }: ThemeProps<DefaultTheme>) => css`
+  &[data-garden-focus-visible] {
+    color: ${getColor('primaryHue', 600, theme)};
```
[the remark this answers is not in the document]
austin94: Since this component has to style the `::before` element I broke this out into a separate helper function. I found it easier to co-locate all of this special logic in a single helper.

packages/tabs/src/styled/StyledTab.ts
austin94: I've moved this to a `sizeStyles` helper to perform this logic.

packages/tabs/src/styled/StyledTabs.ts
austin94: Great question. I think I was playing around with the table display styling and thought it was necessary. Removed.

packages/tabs/src/styled/StyledTab.ts
austin94: I've split this style definition.

packages/tabs/src/styled/StyledTab.ts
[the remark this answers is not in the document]
austin94: This still requires a specificity hack to beat the styling. Since we have to override the default border-color of `transparent` provided in `StyledTabs`.

This hack allows us to avoid using context to provide an `isVertical` prop which we spread to all the components so I think it's still worth it to include.

ginnywood:
@jzempel cool—sounds like semibold selected text isn’t worth it. I’ll update our Sketch libraries so we’re in sync. Thanks!

hzhu:
Should there be a `#1f73b7` border left style for the active tab in the vertical `<Tabs>` example? I noticed that it is missing the left border color styling for the active tab:

tabs-example-after

Here's the production vertical `<Tabs>` example (with the left border color styling):

tabs-example-before

packages/tabs/examples/default-tabs.md
```diff
@@ -70,7 +77,7 @@ initialState = {
       >
```
hzhu: Probably not related to this PR, but I noticed a React warning in the Advanced Usage example:

rendering-pure

I wasn't able to pinpoint the impure part of the render yet, but will keep looking into it.

packages/tabs/src/elements/Tab.tsx
```diff
@@ -34,7 +37,6 @@ const Tab: React.FC<ITabProps> = ({ disabled, item, ...otherProps }) => {
         focusRef,
         index: tabsPropGetters.tabIndexRef.current++,
         isSelected: item === tabsPropGetters.selectedItem,
-        isFocused: item === tabsPropGetters.focusedItem,
```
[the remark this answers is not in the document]
austin94: Now that our focus-visible styling is provided by the global `useFocusVisible()` hook we no longer need to hook into the value returned from the hook. We just use the data-attribute to style in the styled component.

packages/tabs/examples/default-tabs.md
austin94: This is visible often in our `stylguidist` environment. Not related to this PR

austin94:
@hzhu I'm not seeing the vertical styling issue you have. Could you pull again and test?

packages/tabs/src/elements/Tab.tsx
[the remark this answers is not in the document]
hzhu: Thanks for the clarification!

packages/tabs/src/styled/StyledTab.ts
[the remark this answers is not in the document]
austin94: From our offline conversation, this isn't something that was supported in the previous components. Closing for now.

hzhu:
> @hzhu I'm not seeing the vertical styling issue you have. Could you pull again and test?

Pulled and tested again, not seeing it anymore 👍 .

packages/tabs/examples/default-tabs.md
```diff
@@ -43,23 +25,34 @@ tabs = ['Tab 1', 'Tab 2', 'Tab 3'];
 
 initialState = {
   isVertical: true,
```
jzempel: Default the layout to horizontal.

packages/tabs/src/elements/TabPanel.tsx
```diff
@@ -11,6 +11,9 @@ import { StyledTabPanel } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 interface ITabPanelProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * A value to uniquely identity a TabPanel. Provided to the `onChange` event.
```
jzempel: This lacks clarity. The `item` is what connects the tab panel to its associated tab.

packages/tabs/src/styled/StyledTab.ts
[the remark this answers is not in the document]
austin94: These are now split.

packages/tabs/src/styled/StyledTab.ts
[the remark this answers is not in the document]
austin94: This helper has now been removed.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,120 @@
...
 interface IStyledTabProps {
-  isHovered?: boolean;
-  isFocused?: boolean;
-  isActive?: boolean;
   disabled?: boolean;
   isSelected?: boolean;
 }
 
+/**
+ * A high specificity is needed to apply the border-color in vertical orientations
```
jzempel: This comment is way too easy to miss and it doesn't apply to the function. It belongs inline with the line below. We've got to have the ability to comment CSS inline. Styled components docs suggest that it can be stripped on compile: https://www.styled-components.com/docs/tooling#minification. Even if it can't for now (which is a problem that should be fixed) we need to be able to put comments inline and it's worth the added characters to properly document our code.

To that end, I'd like you to recover the comments from https://github.com/zendeskgarden/css-components/tree/master/packages/tabs/src and re-apply them to this migration.

One possible approach that mitigates the added bloat would be to spell out explanations in a header comment, something like this:

```js
/**
 * Consolidated Tab color styling
 *
 * 1. Ensure border-color is applied in vertical orientations.
 * 2. etc.
 * 3. ...
 */
```

And then inline in the CSS-in-JS:

```css
border-color: ${isSelected && 'currentColor'} !important; /* [1] */
```

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,120 @@
...
+ */
+const colorStyles = ({ theme, isSelected }: IStyledTabProps & ThemeProps<DefaultTheme>) => {
+  const selectedColor = getColor('primaryHue', 600, theme);
+
+  return css`
+    color: ${isSelected ? selectedColor : 'inherit'};
+
+    &&& {
+      border-color: ${isSelected && 'currentColor'};
+    }
```
jzempel: Ok so when we run into situations like this, where all other options have been exhausted, it's time to use the dreaded `!important`. Adding `&` until we "win" is brittle.

One other option would be to add the following line under the `${StyledTab}` portion of the `StyledTabs` vertical styling:

```css
&[aria-selected='true'] {
  border-color: currentColor;
}
```

The latter is probably what I would opt for, but YMMV.

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,120 @@
...
+      color: ${selectedColor};
+    }
+
+    &:active {
+      border-color: currentColor;
+      color: ${selectedColor};
+    }
+
+    &[data-garden-focus-visible] {
+      color: ${getColor('primaryHue', 600, theme)};
```
jzempel: Use the `selectedColor` variable.

packages/tabs/src/styled/StyledTabs.ts
```diff
@@ -20,26 +19,64 @@ interface IStyledTabsProps {
...
+    }
+
+    ${StyledTab} {
+      display: block;
+      margin-bottom: ${theme.space.base * 5}px;
+      margin-left: ${theme.rtl && '0'};
+      border-left: ${theme.rtl && '0'};
+      border-right-style: ${theme.rtl && theme.borderStyles.solid};
+      border-bottom-style: none;
+      border-left-style: ${!theme.rtl && theme.borderStyles.solid};
```
jzempel: My preference would be to use constructs similar to what you have below throughout this block. i.e.

```css
border-${theme.rtl ? 'right' : 'left'}-style: ${theme.borderStyles.solid};
```

It improves scanability for trying to figure out how the logic is going to trigger one property over another. And it starts to set a pattern precedent in our code for "this is an RTL-manipulated CSS property".

packages/tabs/src/elements/TabPanel.tsx
austin94: I've updated this wording.

packages/tabs/src/styled/StyledTab.ts
austin94: I believe we are minifying correctly at this time. When we initially added styled-component I believe we had the babel plugin misconfigured which is why we were seeing comments original.

I have added in the comments.

packages/tabs/src/styled/StyledTab.ts
austin94: I will use the `aria-selected` selector. But, from our offline conversation, this selector is a side-effect of the `useSelection()` hook. I'm still unsure if this is something that we should bake into the styles.

austin94: JK, that doesn't work with horizontal rules and adds additional specificity issues. Going to use `!important`

packages/tabs/src/styled/StyledTab.ts
```diff
@@ -5,55 +5,118 @@
...
+  position: relative;
+  transition: color 0.25s ease-in-out;
+  border-bottom: ${props => props.theme.borderStyles.solid} transparent;
+  border-width: ${props => props.theme.borderWidths.md};
+  cursor: pointer;
+  overflow: hidden; /* [1] */
+  vertical-align: top; /* [2] */
+  user-select: none;
+  text-align: center;
+  text-decoration: none; /* [3] */
```
jzempel: This is looking better, but there is no reference point for `[3]` above.

austin94: This was an artifact from the `anchor` styling. Removed this line and double-checked the styling.

jzempel: That's no good because now `<Tab as="a" href="...">` is broken:

Screen Shot 2019-11-13 at 2 54 58 PM

As we discussed (offline) that's a scenario we need to retain view styling support for.

PR #526, opened 2019-11-13
feat(tooltips): migrate react-tooltips to TS

austin94:
- [x] **BREAKING CHANGE**

## Description

This PR migrates `react-tooltips` to TypeScript. It also includes refactoring the structure of the package to align with other `next` packages.

## Breaking changes

- Removed exports
  - LightTooltip
  - TooltipView
  - GARDEN_PLACEMENTS
  - POPPER_PLACEMENTS
  - getPopperPlacement
  - getRtlPopperPlacement
- rename `arrow` prop to `showArrow`
- `appendToBody` prop is now `appendToNode`
  - You must now pass the HTML element which you would like the tooltip to append to
- Tooltip `trigger` prop is removed
  - Tooltip `children` now accepts a single element which acts as the triggering element
- Tooltip `content` prop now accepts Tooltip content (previous `children`)
- The Tooltip trigger no longer has a wrapping `<div>` element
  - An optional `refKey` prop has been added for retrieving the ref from the triggering element

## Checklist

      designer as a reviewer)~
      component


utils/build/declarations.d.ts
```diff
@@ -25,3 +25,4 @@ declare module '@zendeskgarden/css-menus';
 declare module '@zendeskgarden/css-forms';
 declare module '@zendeskgarden/css-arrows';
 declare module '@zendeskgarden/css-tabs';
```
jzempel: Side note: it looks like this got missed with the `react-tabs` CSS-in-JS.

packages/tooltips/README.md
```diff
@@ -31,7 +31,9 @@ import { Tooltip } from '@zendeskgarden/react-tooltips';
  * Place a `ThemeProvider` at the root of your React application
  */
 <ThemeProvider>
-  <Tooltip trigger={<button>Trigger top placement</button>}>This is a small tooltip</Tooltip>
+  <Tooltip content="This is a small tooltip">
```
jzempel: Rename main section heading to `Usage` (vs ~Usages~)

packages/tooltips/examples/default.md
```diff
@@ -0,0 +1,100 @@
+The `Tooltip` component acts as a high-abstraction API over the
+[useTooltip()](https://www.npmjs.com/package/@zendeskgarden/container-tooltip)
+hook.
+
+All props passed to the root element are proxied onto the tooltip element.
```
jzempel: Can we clarify with "All ~props~ events and attributes ~passed~ applied to the root element ..."

packages/tooltips/examples/default.md
```diff
@@ -0,0 +1,100 @@
...
+          </Field>
+        </StyledSpacer>
+      </Well>
+    </Col>
+    <Col md={4} style={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
+      <Tooltip
+        content="This is a small tooltip"
+        showArrow={state.showArrow}
+        placement={state.placement}
+        initialIsVisible
```
jzempel: This example is so nice. Is there a way we can improve by preventing `closeTooltip` on blur? Either that, or apply a very long (10 minutes) delayMilliseconds here.

packages/tooltips/examples/boundary-detection.md
```diff
@@ -0,0 +1,23 @@
+```jsx
```
jzempel: Preface with some explanatory text about the user's need to scroll. Otherwise, it just looks like a broken (empty) example box.

packages/tooltips/examples/default.md
```diff
@@ -0,0 +1,100 @@
...
+          </Dropdown>
+        </StyledSpacer>
+        <StyledSpacer>
+          <Field>
+            <Toggle
+              checked={state.showArrow}
+              onChange={e => setState({ showArrow: e.target.checked })}
+            >
+              <Label>Arrow</Label>
+            </Toggle>
```
hzhu: Is there a way to keep the focus on the underlying input element for `<Toggle>` after it is triggered? 

Currently, space keypress causes the toggle to tigger, then puts the focus on the `<body>` tag. A keyboard user can select the toggle once via space key, but the second space press will cause the browser to scroll to the bottom of the page since focus is now on the `<body>`. 

This can be reproduced by selecting the `<Toggle>` twice with a space key:

boom2

packages/tooltips/examples/default.md
```diff
@@ -0,0 +1,100 @@
...
+const StyledSpacer = styled.div`
+  margin-bottom: ${props => props.theme.space.sm};
+`;
+
+<Grid>
+  <Row>
+    <Col md={4}>
+      <Well recessed>
+        <StyledSpacer>
+          <Dropdown selectedItem={state.placement} onSelect={placement => setState({ placement })}>
```
hzhu: I noticed that the dropdown loses focus after selecting an option via keyboard. After selecting an option via keyboard, the `activeElement` becomes `<body>`. Is there a way to retain focus on the dropdown after selecting an option?

focus-lost-1

packages/tooltips/examples/boundary-detection.md
hzhu: I was thinking something similar before seeing this comment. My thought was to include some explanatory text within the example, but explanatory outside the example would be cool too.

```
// Scroll down to view tooltip with collision detection.










<Tooltip>






```

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -40,43 +43,54 @@ const TooltipWrapper = styled.div`
...
+  showArrow,
   size,
   type,
+  appendToNode,
   zIndex,
-  initialIsVisible,
   ...otherProps
 }) => {
-  const scheduleUpdateRef = useRef();
+  const scheduleUpdateRef = useRef<any>();
```
hzhu: Do you think there is any benefit of typing this as `() => void` rather than `any`? 

TypeScript compiler passes with:

```ts
const scheduleUpdateRef = useRef<() => void>();
```

BREAKING_CHANGES.md
```diff
@@ -111,6 +111,24 @@ should be re-checked for ID naming accuracy.
...
+## @zendeskgarden/react-tooltips
+
+- Removed exports
+  - LightTooltip
+  - TooltipView
+  - GARDEN_PLACEMENTS
+  - POPPER_PLACEMENTS
+  - getPopperPlacement
+  - getRtlPopperPlacement
+- rename `arrow` prop to `showArrow`
```
jzempel: We should stick with our boolean naming for these changes, so `hasArrow`.

utils/build/declarations.d.ts
austin94: Removed.

packages/tooltips/README.md
austin94: Renamed.

packages/tooltips/examples/default.md
austin94: Since we are limiting the examples to a single knobs area these usages need to mimic the expected interaction. I would prefer to leave this as is so users can see our default timing.

packages/tooltips/examples/default.md
austin94: This is due to `stylguidist` updating it's global example state. We have a global `State` helper component which I've used to stop the example from re-rendering.

packages/tooltips/src/elements/Tooltip.tsx
austin94: This works! When I originally moved this into TS `scheduleUpdate` was typed as an `any` within `react-popper`. This now matches their value.

BREAKING_CHANGES.md
austin94: Updated.

packages/tooltips/examples/default.md
austin94: Resolved with answer above.

packages/tooltips/examples/boundary-detection.md
austin94: Now that we have defaulted the `boundariesElement` to be against the `viewport` this examples doesn't really show much. I'm gong to remove it.

packages/tooltips/examples/default.md
hzhu: > We have a global State helper component which I've used to stop the example from re-rendering.

Thank you for the explanation, that's very helpful to know about the `State` helper 🙏.  Checked again and the example is working great.

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -40,43 +43,54 @@ const TooltipWrapper = styled.div`
...
+  eventsEnabled?: boolean;
+  id?: string;
+  content: React.ReactNode;
+  /**
+   * These placements differ from the default naming of Popper.JS placements to help
+   * assist with RTL layouts.
+   **/
+  placement?: GARDEN_PLACEMENT;
+  /** Passes options to [Popper.JS Instance](https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#new-popperreference-popper-options) */
+  popperModifiers?: any;
```
hzhu: Do you think the [`Modifiers` interface](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/index.d.ts#L56) from PopperJS could be used here as the type?

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -141,20 +177,12 @@ const Tooltip = ({
 };
 
 Tooltip.propTypes = {
-  /** Appends the tooltip to the body element */
-  appendToBody: PropTypes.bool,
-  arrow: PropTypes.bool,
-  children: PropTypes.node,
-  /** Milliseconds of delay before open/close of tooltip is initiated */
+  appendToNode: PropTypes.any,
```
hzhu: Was there a particular reason this prop type is `PropTypes.any` rather than `PropTypes.node`?

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -141,20 +177,12 @@ const Tooltip = ({
...
   delayMilliseconds: PropTypes.number,
-  /** Whether Popper.js should update based on DOM resize events */
   eventsEnabled: PropTypes.bool,
   id: PropTypes.string,
-  trigger: PropTypes.node,
-  /**
-   * These placements differ from the default naming of Popper.JS placements to help
-   * assist with RTL layouts.
-   **/
+  content: PropTypes.any.isRequired,
```
hzhu: What do you think about using`PropTypes.node.isRequired` for the `content` prop?

packages/tooltips/examples/default.md
jzempel: Please work with @allisonacs to determine whether your assertion is helpful or confusing. 

One thought is that you could add yet another `<Range>` knob connected to `delayMilliseconds`. Set it high, by default. But let the user control the value so they clearly understand when/why tooltips are disappearing in this example. In this case, you should also add a boolean toggle for `initialIsVisible`.

packages/tooltips/src/elements/Tooltip.tsx
austin94: I originally saw this wasn't exported from the `react-popper` package so I didn't use it. We should be ok since PopperJS is a `dependency`. Added.

packages/tooltips/src/elements/Tooltip.tsx
austin94: The `createPortal()` definition takes an `Element` which isn't a subset of `PropTypes.node`. `any` is required here.

packages/tooltips/examples/default.md
austin94: I've added a range for the `delayMilliseconds` value!

packages/tooltips/src/elements/Tooltip.tsx
hzhu: Awesome, thanks for clarifying!

PR #528, opened 2019-11-16
feat(grid): migrate to TS and update API

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

- Prop renames:
  - Grid `debug` -> `isDebug`
  - Grid `fluid` -> `isFluid`
  - Row `gutters` -> `isCollapsed`
    - Note: boolean logic is reversed for API clarity, but default layout without any prop specified remains the same.
    - Previously `<Grid gutters={false} />`; currently `<Grid isCollapsed />`

## Detail

Knob-style example updates will happen in conjunction with the follow-on CSS-in-JS theming PR.

## Checklist

      designer as a reviewer)~
      component~


packages/grid/src/styled/StyledGrid.ts
```diff
@@ -5,46 +5,32 @@
...
-    [GridStyles.container]: !props.fluid,
+    [GridStyles['container-fluid']]: props.isFluid,
+    [GridStyles.container]: !props.isFluid,
 
     // Debug styling
-    [GridStyles[`is-debug`]]: props.debug,
+    [GridStyles[`is-debug`]]: props.isDebug,
 
     // RTL styling
     [GridStyles['is-rtl']]: isRtl(props)
```
jzempel: will address this `isRtl` usage in the follow-on CSS-in-JS theming PR

packages/grid/src/elements/Col.tsx
```diff
@@ -0,0 +1,60 @@
...
+  /** Use flexbox alignment utilities to horizontally align */
+  alignSelf?: 'start' | 'center' | 'end';
+  justifyContent?: 'start' | 'center' | 'end' | 'around' | 'between';
+  order?: any;
+}
+
+/**
+ * Accepts all `<div>` attributes and events
+ */
+export const Col = React.forwardRef<HTMLDivElement, IColProps>((props, ref) => (
```
austin94: I assume we are doing both named and default exports to help allow `styleguidist` to understand the `forwardRef` syntax?

Could we _only_ use the named export when possible? This would align our `elements` usage with our `styled` usages.

packages/grid/src/styled/StyledCol.ts
```diff
@@ -55,44 +58,31 @@ const retrieveColClassNames = ({
...
+  lg?: number | string | boolean;
+  xl?: number | string | boolean;
+  offsetXs?: number | string | boolean;
+  offsetSm?: number | string | boolean;
+  offsetMd?: number | string | boolean;
+  offsetLg?: number | string | boolean;
+  offsetXl?: number | string | boolean;
+  alignSelf?: 'start' | 'center' | 'end';
+  justifyContent?: 'start' | 'center' | 'end' | 'around' | 'between';
+  order?: any;
```
hzhu: I noticed that `StyledCol` component takes order as a string [here](https://github.com/zendeskgarden/react-components/pull/528/files#diff-e9d9dd98622fd439d2d246cb9d144f60R143). Would this be better suited to be typed as `string`? Or does it also accept other types?

packages/grid/src/elements/Col.tsx
jzempel: Oops, this wasn't intentional.

It appears that I can use named exports all the way through and styleguidist still picks them up. Are we sure now is the right time to start pivoting to using a named export pattern? The packages will be in a 50%/50% state.

packages/grid/src/styled/StyledCol.ts
jzempel: I'm cleaning this up in the CSS-in-JS PR. Stay tuned ;)

packages/grid/src/elements/Col.tsx
austin94: Since we are fully on styleguidist for at least the short-term I think making the move is ok to allow us to keep documenting as we have in the past.

> packages will be in 50%/50% state

With us being in TypeScript land I don't think this hurts us as much since the code-completion and build will help make sure we are using the exports correctly.

jzempel: Nvmd. I lied. The only way the current syntax works with styleguidist is with named exports. Will update...

Screen Shot 2019-11-18 at 10 02 13 AM

PR #529, opened 2019-11-18
feat(tooltips): migrate react-tooltips to CSS-in-JS

austin94:
## Description

This PR migrates the `react-tooltips` package to CSS-in-JS. The bulk of the styling in this PR relates to arrow styling and placement.

## Detail

I tried several different implementations for the arrow styling. PopperJS (and `react-popper`) have direct support for arrow elements which are a separate element from the tooltip. [react-popper example](https://github.com/FezVrasta/react-popper#usage). Our designs include a colored border and box-shadow around the tooltip and it's associated arrow which requires us to use styled `::before/::after` elements.

I have moved the arrow style helper to its own file, `arrow-styles.ts`. This helper applies arrow styling which adapts based on a `data-placement="current-placement"` attribute. This placement is provided by the current "PopperPlacement" of a positioned element.

I originally implemented a version of this helper which uses the `placement` prop to dynamically generate the required arrow styling, but in the end it was more readable for me to use the `data-placement` approach.  This also aligns with some other libraries which use `popper.js` and CSS-in-JS.

## Checklist

      designer as a reviewer)
      component~


austin94:
These changes have been pre-published for review at https://austin-next.netlify.com/tooltips/

ginnywood:
The tooltips are visible on page load, so at first I was a little confused about how to see the delay. I think I would expect to focus on the buttons to see the tooltips.

I think I found a couple visual bugs:
The default tooltip should have 2px corner radius.
The light tooltip body text should be grey-700.

austin94:
> The tooltips are visible on page load, so at first I was a little confused

This was added since we are now moving all examples to the "knob" examples.

> The light tooltip body text should be grey-700

It looks like I made the title and body color the same for the light tooltip. This is now corrected to `grey-700` for the body text.

> The default tooltip should have 2px corner radius

Locally I am seeing a 2px borderRadii for the default (dark) tooltip. Could you help confirm there is an issue here?

ginnywood:
> Locally I am seeing a 2px borderRadii for the default (dark) tooltip. Could you help confirm there is an issue here?
Sorry—my mistake, this is only Sketch rendering it differently.

jzempel:
Piling on along with @allisonacs comment. The knobs example feels weird to me. Either the `delay` should be set very large by default in the example so that the tooltips don't prematurely dismiss. Even better would be the addition of a `debug` (`isDebug`?) prop that removes the display-on-hover/dismiss-on-blur logic. Not only would this be beneficial for our example, but I think it could be helpful for use during development (similar to how we have a debug setting on `grid`).

packages/tooltips/src/styled/StyledLightTooltip.ts
```diff
@@ -5,27 +5,62 @@
...
+    &::before,
+    &::after {
+      font-size: ${arrowFontSize};
+    }
+  `;
+};
+
+const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
+  return css`
+    border: ${theme.borderWidths.sm} ${theme.borderStyles.solid}
```
jzempel: Use `borders.sm` from the theme since this is a shorthand CSS property.

packages/tooltips/src/styled/StyledLightTooltip.ts
```diff
@@ -5,27 +5,62 @@
...
+      color: ${theme.colors.foreground};
+    }
+  `;
+};
+
 /**
  * Accepts all `<div>` props
  */
-export const StyledLightTooltip = styled(StyledTooltip).attrs(props => ({
+export const StyledLightTooltip = styled(StyledTooltip).attrs({
```
jzempel: From a styling POV, it would be better if all of this were incorporated into `StyledTooltip` – leveraging the `type` and removing `large`/`small` defaultProps – effectively removing several styling overrides. Keep the `styled` components agnostic with respect to size/color. Let the hi-abstraction component dictate which is expected, by default. (i.e. "light" tooltips shouldn't render at the "small" size and vice versa).

packages/tooltips/src/styled/StyledLightTooltip.ts
```diff
@@ -5,27 +5,62 @@
...
+      font-size: ${arrowFontSize};
+    }
+  `;
+};
+
+const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
+  return css`
+    border: ${theme.borderWidths.sm} ${theme.borderStyles.solid}
+      ${getColor('neutralHue', 300, theme)};
+    box-shadow: 0 ${theme.space.base * 3}px ${theme.space.base * 5}px 0
```
jzempel: Use the `shadows.lg` theming function here. Improve code readability by assigning a `boxShadow` variable outside of the `css` block.

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -5,78 +5,128 @@
...
+    paragraphMarginTop = `${theme.space.base * 2}px`;
+  } else if (size === 'medium') {
+    padding = 'padding: 1em';
+    maxWidth = '140px';
+    lineHeight = (theme.space.base * 4) / stripUnit(theme.fontSizes.sm);
+  }
+
+  if (size === 'extra-large' || size === 'large') {
+    fontSize = theme.fontSizes.md;
+    titleDisplay = 'block';
```
jzempel: See comment on `StyledTitle` re: `display`.

packages/tooltips/src/styled/StyledLightTooltip.ts
```diff
@@ -5,27 +5,62 @@
...
+    arrowFontSize = `${theme.space.base * 4 + 2}px`;
+  }
+
+  return css`
+    &::before,
+    &::after {
+      font-size: ${arrowFontSize};
+    }
+  `;
+};
```
jzempel: Incorporating this into the `StyledTooltip` code should make this slightly less obtuse. It's always just whatever the `small`/`medium`/`large` sizes are plus `2px` (basically accounting for two `borderWidths.sm` strokes).

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -5,78 +5,128 @@
...
+
+  &[aria-hidden='true'] {
+    display: none;
+  }
+
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
 
 StyledTooltip.defaultProps = {
   hasArrow: true,
```
jzempel: Shouldn't have to default with this opinion – let the higher-abstraction component decide.

packages/tooltips/src/styled/arrow-styles.ts
```diff
@@ -0,0 +1,193 @@
...
+  const rightClipPath = 'polygon(100% 0, 100% 100%, calc(100% - 1px) 100%, 0 1px, 0 0)';
+  const topClipPath = 'polygon(100% 0, 100% 1px, 1px 100%, 0 100%, 0 0)';
+
+  let arrowFontSize = `${theme.space.base * 1.75}px`;
+
+  if (size === 'extra-large') {
+    arrowFontSize = `${theme.space.base * 4}px`;
+  } else if (size === 'large') {
+    arrowFontSize = theme.fontSizes.sm;
+  }
```
jzempel: Move this into `StyledTooltip` as it is directly associated with that component as opposed to this more generic arrow styling.

I guess the big question is how we can pull this arrow styling out into something that can be re-used across components? Feels like we would be better served if this moved to `theming` or (blerg) `utilities`.

packages/tooltips/src/styled/StyledParagraph.ts
```diff
@@ -16,8 +15,13 @@ const COMPONENT_ID = 'tooltip.paragraph';
  */
 export const StyledParagraph = styled.div.attrs({
```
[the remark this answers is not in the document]
austin94: Consumers often put several paragraphs of content within the this boundary. Especially with it being ready by an `aria-describedby` usage I don't think this would be necessary.

packages/tooltips/src/styled/StyledTitle.ts
```diff
@@ -16,8 +15,15 @@ const COMPONENT_ID = 'tooltip.title';
  */
 export const StyledTitle = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: TooltipStyles['c-tooltip__title']
+  'data-garden-version': PACKAGE_VERSION
 })`
+  display: none;
```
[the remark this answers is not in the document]
austin94: Wouldn't this be the same logic, just swapping the display values in each section? We already override the display based on if it is `large` or `extra-large`.

packages/tooltips/src/styled/StyledLightTooltip.ts
austin94: I have moved this to the `StyledTooltip` component

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -5,78 +5,128 @@
...
+  }
+
+  if (size === 'extra-large') {
+    padding = `${theme.space.base * 10}px`;
+    maxWidth = '460px';
+    lineHeight = (theme.space.base * 5) / stripUnit(theme.fontSizes.md);
+    paragraphMarginTop = `${theme.space.base * 2.5}px`;
+  } else if (size === 'large') {
+    padding = `${theme.space.base * 5}px`;
+    maxWidth = '270px';
```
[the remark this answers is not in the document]
austin94: I am now calculating this as `theme.space.base * 67.5`

packages/tooltips/src/styled/StyledLightTooltip.ts
austin94: I originally left the original CSS structure intact to reduce the number of changes in a single PR. I have now restructured this to be a single file that handles all permutations.

packages/tooltips/src/styled/StyledTooltip.ts
austin94: Moved to the element level.

packages/tooltips/src/styled/arrow-styles.ts
austin94: I have moved the size-specific logic to the `StyledTooltip` component.

> how we can pull this arrow styling out into something that can be re-used across components?

From our offline conversation, with the differences between arrow (specifically animation) don't think we should try and solve this reuse issue until we tackle that package.

austin94:
@jzempel @allisonacs I have updated the example to include the following changes:
- `size` dropdown which defaults to an undefined, default size
- `visibility` toggle
  - Rather than introduce an `isDebug` prop I have add a controlled `isVisible` prop for consumers that need to override our default state handling.
  - The `useTooltip()` hook is one of the few components that only provides uncontrolled state so there is some additional `getControlledValue()` usage to allow us to control the value.

@jzempel I have also removed the `StyledLightTooltip` component and moved all size-specific arrow styles to the `StyledTooltip` element.

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -5,78 +5,178 @@
...
-  return 'margin: 6px';
+  return css`
+    margin: ${margin};
+    border-radius: ${borderRadius};
+    padding: ${padding};
+    max-width: ${maxWidth};
+    line-height: ${lineHeight};
+    word-wrap: ${wordWrap};
+    white-space: ${whiteSpace};
+    /*  Manipulate arrow sizing (which is based on ems). */
```
jzempel: If I'm reading it right, I think this comment might be stale. Arrow sizing is manipulated via the `arrowFontSize` applied to the `::before/::after` below.

packages/tooltips/examples/basic.md
```diff
@@ -69,7 +86,7 @@ const StyledSpacer = styled.div`
                   min={100}
                   max={10000}
```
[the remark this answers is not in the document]
austin94: I've updated the example with these changes.

packages/tooltips/src/styled/StyledTooltip.ts
austin94: I brought this comment over from https://github.com/zendeskgarden/css-components/blob/master/packages/tooltips/src/_base.css#L48-L53

I think it was originally implying that the font-size is impacting the `::before,::after` styling in `c-arrow`. This is still based on `em`'s. We are using

```ts
const arrowSize = `${stripUnit(theme.fontSizes.sm) / stripUnit(theme.fontSizes.md)}em`;
```

in `arrow-styles` right now.

Should I just shorten this to `/* Manipulate arrow sizing */`?

packages/tooltips/src/styled/StyledTooltip.ts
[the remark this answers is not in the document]
allisonacs: Looking at `austin-next`, the rendered components look like they have the correct spacing.

jzempel:
@allisonacs we've looked at this before with https://garden.zendesk.com/css-components/tooltips/ and there's nothing we can do when the arrow falls into the fat part of the box shadow.

austin94:
@allisonacs It looks like this styling issue isn't new to this PR. I've confirmed that it exists in our css-components and react-components packages for all arrows in a "bottom" position.

### `css-arrows` 

image

I think our screenshots are making the colors more noticeable than a normal usage.

... Looks like @jzempel beat me to the punch with his comment 😄

jzempel:
fwiw, I'm not opposed to revisiting our styling approach to arrows. While the designs hold-up nicely, the styling piece is utilitarian but less-than-perfect. Best deferred for future investigation.

allisonacs:
I'd be happy to add a ticket to the backlog re: tooltip styling. We're
seeing more and more of them as we move toward more contextual controls, so
making sure we're happy with both the design and the code approach is
probably worthwhile.

>

austin94:
@allisonacs I have updated the string length, visibility toggle, and size warning in the example.

packages/tooltips/examples/basic.md
```diff
@@ -0,0 +1,189 @@
...
+                  checked={state.isVisible}
+                  onChange={e => setState({ isVisible: e.target.checked })}
+                >
+                  <Label>Force visibility</Label>
+                </Toggle>
+              </Field>
+            </StyledSpacer>
+            <StyledSpacer>
+              <Field>
+                <Label>Delay ({state.delayMS} ms)</Label>
```
hzhu: I noticed an a11y with this element while running Axe / Lighthouse.

Screen Shot 2019-11-21 at 9 52 55 AM

austin94: This is an axe false-positive relating to the `useField()` hook. This doesn't cause any errors for audits or screenreaders.

jzempel: As a side note, I do think our `useField` hook is a bit brute force and could use some finesse in the future. There are other semantic configurations that aren't handled very elegantly (i.e. Label/Hint over multiple radio inputs).

PR #544, opened 2019-11-22
fix(tabs): use state helper to avoid example re-rendering

hzhu:
## Description

React Styleguidest updates to global state is causing the toggles to lose focus. This causes the `activeElement` to lose focus on state changes. To solve this problem, I used the global `<State>` to encapsulate the example state so the toggles do not lose focus.

## Screenshots

**Before:** User focuses on the toggle component and presses the `space` key twice. The first press triggers the toggle. The second key press scrolls the user to the bottom of the page:

**After:** User focuses on the toggle component and presses the `space` key multiple times. The key presses trigger the toggle back and forth, without scrolling the user to the bottom of the page:

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/packages/tabs/examples/default-tabs.md
+++ b/packages/tabs/examples/default-tabs.md
@@ -21,9 +21,9 @@ const { Well } = require('@zendeskgarden/react-notifications/src');
 const { Code } = require('@zendeskgarden/react-typography/src');
 const { Field, Label, Toggle } = require('@zendeskgarden/react-forms/src');
 
-tabs = ['Tab 1', 'Tab 2', 'Tab 3'];
+const tabs = ['Tab 1', 'Tab 2', 'Tab 3'];
 
-initialState = {
+const initialState = {
   isVertical: false,
   isDisabled: false,
   selectedItem: tabs[0]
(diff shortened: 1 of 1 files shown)


packages/tabs/examples/default-tabs.md
```diff
@@ -33,48 +33,52 @@ const StyledSpacedField = styled(Field)`
...
 
-<Grid>
-  <Row>
-    <Col md={4}>
-      <Well recessed>
-        <StyledSpacedField>
-          <Toggle
-            checked={state.isVertical}
-            onChange={e => setState({ isVertical: e.target.checked })}
+<State initialState={initialState}>
```
austin94: The `initialState` global actually does some state "magic" with styleguidist. It may be cleaner to just move that object inline to this prop.

hzhu: Oh no, magic. What do you think about declaring these variables using `const` instead? That way styleguist can't re-declare another value for `initialState`. 

// before
```jsx
tabs = ['Tab 1', 'Tab 2', 'Tab 3'];

initialState = {
  isVertical: false,
  isDisabled: false,
  selectedItem: tabs[0]
};
```

// before
```jsx
const tabs = ['Tab 1', 'Tab 2', 'Tab 3'];

const initialState = {
  isVertical: false,
  isDisabled: false,
  selectedItem: tabs[0]
};
```

austin94: 👍 Looks good to me!

PR #545, opened 2019-11-23
feat(grid): CSS-in-JS refactor

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

Breaking changes detailed under https://github.com/zendeskgarden/react-components/blob/0543a947ba033724ed2684ee9d431aed4baa221d/BREAKING_CHANGES.md#zendeskgardenreact-grid.

Feature enhancements:
- Connected to customizable `theming` breakpoints
- Added responsive variants for most `Row`/`Col` props (these were always available in the CSS, but missing for React)
- Grid gutter width is controlled by a prop that is connected to customizable `theming` space variables (defaults to `md` which equals 20px)
- Grid total column count controlled by a prop (defaults to 12)
- Grid `debug` styling no longer impacts padding/spacing
- Fixed RTL layout (previously broken for `offset` columns)

## Detail

Demo pre-published for review: https://competent-shirley-cc2f82.netlify.com/grid/

## Checklist

      designer as a reviewer)
      component~


jzempel:
Thanks @allisonacs. I attempted an update and re-published a demo. Plz let me know if this works a bit better.

packages/grid/src/styled/StyledRow.spec.tsx
```diff
@@ -6,71 +6,140 @@
...
+      });
+
+      describe('Responsively', () => {
+        Object.keys(DEFAULT_THEME.breakpoints).forEach(breakpoint => {
+          const key = `alignItems${breakpoint[0].toUpperCase()}${breakpoint.substring(1)}`;
+
+          it(`renders ${key}=${alignItems} flex alignment`, () => {
+            const props = { [key]: alignItems };
+            const { container } = render(<StyledRow {...props} />);
+            const minWidth = (DEFAULT_THEME.breakpoints as any)[breakpoint];
```
jzempel: @austin94 I'm wondering if there's a better TS hack than the `as any` used here.

packages/grid/src/utils/types.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export type TYPE_ALIGN_ITEMS = 'start' | 'end' | 'center' | 'baseline' | 'stretch';
```
austin94: For naming in other packages, we've used ALL_CAPS to denote types. It may be more readable to drop the `TYPE_` prefix.

If you use the `F2` option in VSCode it will rename all usages for you 😄

packages/grid/src/utils/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export type TYPE_JUSTIFY_CONTENT = 'start' | 'end' | 'center' | 'between' | 'around';
+export const ARRAY_JUSTIFY_CONTENT: Array<TYPE_JUSTIFY_CONTENT> = [
+  'start',
+  'end',
+  'center',
+  'between',
+  'around'
+];
+
+export type TYPE_NUMBER = number | string;
```
austin94: This naming seems ambiguous right now. `foo = foo | bar`

Is there a more specific name for this type that we could use? i.e. `COLUMN_SIZE`

packages/grid/src/elements/Col.tsx
```diff
@@ -7,38 +7,91 @@
...
-  alignSelf?: 'start' | 'center' | 'end';
-  justifyContent?: 'start' | 'center' | 'end' | 'around' | 'between';
-  order?: any;
+  /**
+   * Determine the size, relative to the total number of `columns` in the grid,
+   * for all screen sizes
+   */
+  size?: TYPE_NUMBER;
+  /** Determine the size for extra-small screen sizes */
+  xs?: TYPE_NUMBER | boolean;
```
austin94: If the sizing props have a consistent type of `string | number | boolean` would it make sense to create another type for this?

packages/grid/src/styled/StyledCol.spec.tsx
```diff
@@ -6,149 +6,206 @@
  */
 
 import React from 'react';
-import { render } from 'garden-test-utils';
+import math from 'polished/lib/math/math';
+import { render, renderRtl } from 'garden-test-utils';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming/src';
```
austin94: This would break during the build. Make sure you don't have any `src` imports throughout all the files. Sometimes TypeScript auto imports these paths by accident.

packages/grid/src/utils/types.ts
jzempel: But as you can see here, it's meant to differentiate from the `ARRAY` constant.

packages/grid/src/utils/types.ts
jzempel: I had it particularly named originally. But as you go through the diff, you'll see that it applies anywhere that a prop accepts a numeric value – not just columns. The generalized naming works better.

packages/grid/src/utils/types.ts
jzempel: Are you suggesting to rename all the `type` constants and leave the array naming as-is?

packages/grid/src/styled/StyledCol.spec.tsx
jzempel: oops, fixed.

packages/grid/src/elements/Col.tsx
jzempel: I thought about adding a `TYPE_SIZE` for these. But then it's weird that `size` would not be `TYPE_SIZE`. So then I thought about adding a `TYPE_NUMBER_OR_BOOLEAN`, but that reads worse than just saying `TYPE_NUMBER | boolean`.

packages/grid/src/utils/types.ts
austin94: > Are you suggesting to rename all the type constants and leave the array naming as-is

That could make sense. We just don't normally define arrays for the `propType` definitions in other components. If this feels right to you feel free to leave it.

packages/grid/src/utils/types.ts
austin94: > it applies anywhere that a prop accepts a numeric value – not just columns

Then let's name this `GRID_NUMBER` or something generic. Normally with something as basic as `number | string` you would just add that to the other type definitions.

packages/grid/src/elements/Col.tsx
austin94: Could we define 2 additional types here for the `xs|sm|etc.` and `offset*` props? This would allow:

```tsx
xs: TYPE_COLUMN,
offsetXs: TYPE_COLUMN_OFFSET
```

Even thought they allow roughly the same basic types `string | number | (sometimes) boolean` these values aren't necessarily related.

jzempel: Added `BREAKPOINT` constant to cover `number | string | boolean` types.

packages/grid/src/styled/StyledRow.spec.tsx
austin94: There are probably some more specific semantics for this, but it makes sense to me 😄 Since this is a test file I don't think we need to change this. LGTM

PR #546, opened 2019-11-23
feat(breadcrumbs): CSS-in-JS Refactor

hzhu:
## Description

This pull request refactors CSS to JS.

## Details

Breadcrumbs has been updated to use 12x12 chevron stroke icons from svg-icons. For a **visual check**, a review link for the breadcrumb package Storybook is published [**here**](https://gallant-blackwell-63466f.netlify.com/breadcrumbs/).

## Checklist

      designer as a reviewer)
      component~

### DIFF
--- a/packages/breadcrumbs/package.json
+++ b/packages/breadcrumbs/package.json
@@ -21,7 +21,7 @@
   "types": "dist/typings/index.d.ts",
   "dependencies": {
     "@zendeskgarden/container-breadcrumb": "^0.3.2",
-    "classnames": "^2.2.5"
+    "polished": "^3.4.2"
   },
   "peerDependencies": {
     "@zendeskgarden/react-theming": "^6.0.0 || ^7.0.0",
@@ -31,9 +31,7 @@
     "styled-components": "^4.2.0"
   },
   "devDependencies": {
-    "@zendeskgarden/css-breadcrumbs": "0.2.18",
-    "@zendeskgarden/react-theming": "^7.0.0",
-    "@zendeskgarden/react-utilities": "^7.0.0"
+    "@zendeskgarden/react-theming": "^7.0.0"
   },
   "keywords": [
     "components",
--- a/packages/breadcrumbs/src/elements/Breadcrumb.spec.tsx
+++ b/packages/breadcrumbs/src/elements/Breadcrumb.spec.tsx
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import { render } from 'garden-test-utils';
+import { getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
 import { Breadcrumb } from './Breadcrumb';
 
@@ -72,9 +73,12 @@ describe('Breadcrumb', () => {
 
       items.forEach((item, i) => {
         if (i === lastItemIndex) {
-          expect(item.parentElement).toHaveClass('is-current');
+          expect(item.parentElement).toHaveStyleRule(
+            'color',
+            getColor(DEFAULT_THEME.colors.neutralHue, 600)
+          );
         } else {
-          expect(item.parentElement).not.toHaveClass('is-current');
+          expect(item.parentElement).toHaveStyleRule('color', 'inherit');
         }
       });
     });
--- a/packages/breadcrumbs/src/elements/Breadcrumb.tsx
+++ b/packages/breadcrumbs/src/elements/Breadcrumb.tsx
@@ -7,8 +7,12 @@
 
 import React, { Children, cloneElement, HTMLAttributes } from 'react';
 import { useBreadcrumb } from '@zendeskgarden/container-breadcrumb';
-
-import { StyledBreadcrumb, StyledBreadcrumbItem } from '../styled';
+import {
+  StyledBreadcrumb,
+  StyledBreadcrumbItem,
+  StyledCenteredBreadcrumbItem,
+  StyledChevronIcon
+} from '../styled';
 
 export const Breadcrumb = React.forwardRef<HTMLElement, HTMLAttributes<HTMLElement>>(
   (props, ref) => {
@@ -27,7 +31,14 @@ export const Breadcrumb = React.forwardRef<HTMLElement, HTMLAttributes<HTMLEleme
         );
       }
 
-      return <StyledBreadcrumbItem>{child}</StyledBreadcrumbItem>;
+      return (
+        <>
+          <StyledBreadcrumbItem>{child}</StyledBreadcrumbItem>
+          <StyledCenteredBreadcrumbItem>
+            <StyledChevronIcon />
+          </StyledCenteredBreadcrumbItem>
+        </>
+      );
     });
 
     return (
--- a/packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
+++ b/packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
@@ -6,19 +6,20 @@
  */
 
 import styled from 'styled-components';
-import classNames from 'classnames';
-import BreadcrumbStyles from '@zendeskgarden/css-breadcrumbs';
-import { retrieveComponentStyles, isRtl } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'breadcrumbs.list';
 
-export const StyledBreadcrumb = styled.ol.attrs(props => ({
+/**
+ * 1. <ol> reset.
+ */
+export const StyledBreadcrumb = styled.ol.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(BreadcrumbStyles['c-breadcrumb'], {
-    // RTL
-    [BreadcrumbStyles['is-rtl']]: isRtl(props)
-  })
-}))`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  margin: 0; /* [1] */
+  padding: 0; /* [1] */
+  list-style: none; /* [1] */
+  font-size: ${props => props.theme.fontSizes.md};
+  direction: ${props => props.theme.rtl && 'rtl'};
 `;
--- a/packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
+++ b/packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
@@ -6,19 +6,27 @@
  */
 
 import React from 'react';
+import stripUnit from 'polished/lib/helpers/stripUnit';
 import { render } from 'garden-test-utils';
+import { getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { StyledBreadcrumbItem } from './StyledBreadcrumbItem';
 
 describe('StyledBreadcrumbItem', () => {
   it('renders default styling', () => {
     const { container } = render(<StyledBreadcrumbItem />);
+    const lineHeight = (DEFAULT_THEME.space.base * 5) / stripUnit(DEFAULT_THEME.fontSizes.md);
 
-    expect(container.firstChild).toHaveClass('c-breadcrumb__item');
+    expect(container.firstChild).toHaveStyleRule('line-height', lineHeight.toString());
+    expect(container.firstChild).toHaveStyleRule('white-space', 'nowrap');
+    expect(container.firstChild).toHaveStyleRule('font-size', 'inherit');
   });
 
   it('renders current styling', () => {
     const { container } = render(<StyledBreadcrumbItem isCurrent />);
 
-    expect(container.firstChild).toHaveClass('is-current');
+    expect(container.firstChild).toHaveStyleRule(
+      'color',
+      getColor(DEFAULT_THEME.colors.neutralHue, 600)
+    );
   });
 });
--- a/packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
+++ b/packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
@@ -5,10 +5,9 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled from 'styled-components';
-import classNames from 'classnames';
-import { retrieveComponentStyles } from '@zendeskgarden/react-theming';
-import BreadcrumbStyles from '@zendeskgarden/css-breadcrumbs';
+import styled, { css } from 'styled-components';
+import stripUnit from 'polished/lib/helpers/stripUnit';
+import { getColor } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'breadcrumbs.item';
 
@@ -16,13 +15,36 @@ export interface IStyledBreadcrumbItemProps {
   isCurrent?: boolean;
 }
 
-export const StyledBreadcrumbItem = styled.li.attrs<IStyledBreadcrumbItemProps>(props => ({
+/**
+ * These CSS pseudo-classes are used to match the equivalent of :any-link, which
+ * is not used here because it is not currently supported in Microsoft Edge.
+ */
+const linkStyles = ({ isCurrent }: IStyledBreadcrumbItemProps) => css`
+  & > :link,
+  & > :visited {
+    white-space: inherit;
+  }
+
+  ${isCurrent &&
+    `
+      & > :link,
+      & > :visited,
+      & > :link:hover,
+      & > :visited:hover,
+      & > :link:focus,
+      & > :visited:focus {
+        color: inherit;
+      }
+    `}
+`;
+
+export const StyledBreadcrumbItem = styled.li.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(BreadcrumbStyles['c-breadcrumb__item'], {
-    // State
-    [BreadcrumbStyles['is-current']]: props.isCurrent
-  })
-}))<IStyledBreadcrumbItemProps>`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledBreadcrumbItemProps>`
+  line-height: ${props => (props.theme.space.base * 5) / stripUnit(props.theme.fontSizes.md)};
+  white-space: nowrap;
+  color: ${props => (props.isCurrent ? getColor(props.theme.colors.neutralHue, 600) : 'inherit')};
+  font-size: inherit;
+  ${linkStyles};
 `;
--- a/packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
+++ b/packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
@@ -1,0 +1,39 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { HTMLAttributes } from 'react';
+import styled, { ThemeProps, DefaultTheme } from 'styled-components';
+import em from 'polished/lib/helpers/em';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import ChevronRightStrokeIcon from '@zendeskgarden/svg-icons/src/12/chevron-right-stroke.svg';
+
+/**
+ * styled-components injects a theme prop that is rendered as an attribute
+ * of the SVG icon in the browser. This function removes the `theme` prop injected
+ * by styled-components.
+ */
+const ValidChevronIcon: React.FC<HTMLAttributes<SVGElement> & ThemeProps<DefaultTheme>> = props => {
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+  const { theme, ...validProps } = props;
+
+  return <ChevronRightStrokeIcon {...validProps} />;
+};
+
+/**
+ * Accepts all `<svg>` props
+ */
+export const StyledChevronIcon = styled(ValidChevronIcon).attrs({
+  role: 'presentation',
+  'aria-hidden': 'true'
+})`
+  transform: ${props => props.theme.rtl && `rotate(180deg);`};
+  margin: 0 ${props => em(props.theme.space.base, props.theme.fontSizes.md)};
+`;
+
+StyledChevronIcon.defaultProps = {
+  theme: DEFAULT_THEME
+};
--- a/packages/breadcrumbs/src/styled/StyledBreadcrumb.spec.tsx
+++ b/packages/breadcrumbs/src/styled/StyledBreadcrumb.spec.tsx
@@ -7,18 +7,23 @@
 
 import React from 'react';
 import { render, renderRtl } from 'garden-test-utils';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { StyledBreadcrumb } from './StyledBreadcrumb';
 
 describe('StyledBreadcrumb', () => {
   it('renders default styling correctly', () => {
     const { container } = render(<StyledBreadcrumb />);
 
-    expect(container.firstChild).toHaveClass('c-breadcrumb');
+    expect(container.firstChild).toHaveStyleRule('display', 'flex');
+    expect(container.firstChild).toHaveStyleRule('margin', '0');
+    expect(container.firstChild).toHaveStyleRule('padding', '0');
+    expect(container.firstChild).toHaveStyleRule('list-style', 'none');
+    expect(container.firstChild).toHaveStyleRule('font-size', DEFAULT_THEME.fontSizes.md);
   });
 
   it('renders RTL styling correctly', () => {
     const { container } = renderRtl(<StyledBreadcrumb />);
 
-    expect(container.firstChild).toHaveClass('is-rtl');
+    expect(container.firstChild).toHaveStyleRule('direction', 'rtl');
   });
 });
--- a/packages/breadcrumbs/src/styled/StyledCenteredBreadcrumbItem.tsx
+++ b/packages/breadcrumbs/src/styled/StyledCenteredBreadcrumbItem.tsx
@@ -1,0 +1,16 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { StyledBreadcrumbItem } from './index';
+
+export const StyledCenteredBreadcrumbItem = styled(StyledBreadcrumbItem).attrs({
+  'aria-hidden': true
+})`
+  display: flex;
+  align-items: center;
+`;
--- a/packages/breadcrumbs/src/styled/index.ts
+++ b/packages/breadcrumbs/src/styled/index.ts
@@ -6,4 +6,6 @@
  */
 
 export { StyledBreadcrumb } from './StyledBreadcrumb';
+export { StyledChevronIcon } from './StyledChevronIcon';
 export { StyledBreadcrumbItem } from './StyledBreadcrumbItem';
+export { StyledCenteredBreadcrumbItem } from './StyledCenteredBreadcrumbItem';


packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,21 @@
...
+    'data-garden-version': PACKAGE_VERSION
+  };
+})`
+  ${props => `
+    display: flex;
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    font-size: ${props.theme.fontSizes.md};
+    direction: ${props.theme.rtl ? 'rtl' : 'ltr'};
```
hzhu: I recall CSS properties should be in a specific order. I used [this](https://github.com/brandon-rhodes/Concentric-CSS/blob/master/style.css) as a guide when ordering. Is there docs for ordering that I should refer to? Please let me know if the ordering here is wrong.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
```diff
@@ -6,19 +6,41 @@
...
+      modifier: '::after'
+    });
+  });
+
+  it('renders the correct style for the ::after chevron pseudo element for RTL', () => {
+    const { container } = renderRtl(<StyledBreadcrumbItem />);
+
+    expect(container.firstChild).toHaveStyleRule('transform', 'rotate(180deg)', {
+      modifier: '::after'
+    });
```
hzhu: The two tests above are newly added. They assert that users should see the correct chevron direction in the breadcrumb depending on their writing system (LTR vs RTL).

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,36 @@
...
-  })
+  'data-garden-version': PACKAGE_VERSION
 }))<IStyledBreadcrumbItemProps>`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${props => `
+    line-height: ${props.theme.lineHeights.md};
+    color: ${props.isCurrent ? getColor(props.theme.colors.neutralHue, 600) : 'inherit'}
+    ${
+      props.isCurrent === undefined || props.isCurrent === false
+        ? `::after {
```
hzhu: Previously the `::before` pseudo element was used, but now `::after` is used.

I had a difficult time wrapping my head around using `::before` to position the chevrons. I opted for `::after` because all breadcrumbs need a chevron _after_ itself besides the current breadcrumb (or last breadcrumb).

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,36 @@
...
+        transform:  ${props.theme.rtl ? 'rotate(180deg)' : 'none'};
+        display: inline-block;
+        margin: 0 ${props.theme.space.xs};
+        background: no-repeat 50% / 1em url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16' color='%2368737D'%3E%3Cpath fill='none' stroke='currentColor' stroke-linecap='round' d='M4.5 15l6.6-6.6c.2-.2.2-.5 0-.7L4.5 1'/%3E%3C/svg%3E");
```
hzhu: I copied over the chevron we have now in production and bumped up the size to 16 x 16. 

This was the most straight forward approach, but I suspect there are much better ways to do this. I also know that chevron icons exists in the [svg-icons](https://garden.zendesk.com/svg-icons/16px/) project, which might be useful here.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
jzempel: Order is [concentric](https://rhodesmill.org/brandon/2011/concentric-css/) as defined by https://github.com/zendeskgarden/stylelint-config/blob/master/plugins/stylelint-order.js. Use `yarn lint:css` to check while you're getting used to the ordering.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
jzempel: Use `::before` since `::after` has the affect of polluting the user space. See [internal-only link](https://github.com/zendeskgarden/css-forms/pull/22) for details. We tend to use `::after` only when needed:
- swapping `::before`/`::after` `background-image` on hover
- z-index styling in [css-arrows](https://github.com/zendeskgarden/css-components/tree/master/packages/arrows)

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
jzempel: You should be using the styling seen [here](https://garden.zendesk.com/css-components/breadcrumbs/). We got this in recently so that you would have the right designs to work off of.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
jzempel: Having said ☝️, it's going to make a lot more sense for you to move the chevron out into it's own DOM element. Compare tag [CSS](https://github.com/zendeskgarden/css-components/blob/master/packages/tags/src/_remove.css) with [CSS-in-JS](https://github.com/zendeskgarden/react-components/blob/next/packages/tags/src/elements/Close.tsx) for reference.

Unless it's going to be [tricky](https://github.com/zendeskgarden/react-components/blob/next/packages/forms/src/styled/radio/StyledRadioInput.js) we need to take this opportunity to inline SVGs. Unlike CSS, we control the DOM here and this instance should be fairly straightforward.

jzempel: Actually, this pattern might work better for you: https://github.com/zendeskgarden/react-components/blob/next/packages/buttons/src/styled/StyledExternalIcon.ts

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
```diff
@@ -6,19 +6,41 @@
...
+import { render, renderRtl } from 'garden-test-utils';
+import { getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { StyledBreadcrumbItem } from './StyledBreadcrumbItem';
 
 describe('StyledBreadcrumbItem', () => {
   it('renders default styling', () => {
     const { container } = render(<StyledBreadcrumbItem />);
 
-    expect(container.firstChild).toHaveClass('c-breadcrumb__item');
+    expect(container.firstChild).toHaveStyleRule('line-height', DEFAULT_THEME.lineHeights.md);
```
hzhu: This should retain a unit less value rather than using a pixel value found in the default theme.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,21 @@
...
+export const StyledBreadcrumb = styled.ol.attrs(() => {
+  return {
+    'data-garden-id': COMPONENT_ID,
+    'data-garden-version': PACKAGE_VERSION
+  };
+})`
+  ${props => `
+    display: flex;
+    margin: 0;
+    padding: 0;
```
hzhu: These are CSS resets that have useful comments on why they exists in their respective CSS files. Those comments should be brought over during the CSS-in-JS migration as they will be helpful for understanding why these exists.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
hzhu: Regarding using `<svg>` for the separator:

> To prevent screen reader announcement of the visual separators between links, they are added via CSS:

> The separators are part of the visual presentation that signifies the breadcrumb trail, which is already semantically represented by the nav element with its label of Breadcrumb. So, using a display technique that is not represented in the accessibility tree used by screen readers prevents redundant and potentially distracting verbosity.

> Each link has a border on one side that is skewed with the CSS’ transform property so it resembles a slash.

Source: https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

An `aria-hidden` can be applied to the separator to prevent it from being picked up on screen readers for the `<svg>` approach. Another challenge with the `<svg>` approach is centering the chevron — managed to center it with `position:relative;top:1px;` which is not great but seems to work.

packages/breadcrumbs/src/elements/Breadcrumb.spec.tsx
```diff
@@ -72,9 +73,12 @@ describe('Breadcrumb', () => {
 
       items.forEach((item, i) => {
         if (i === lastItemIndex) {
-          expect(item.parentElement).toHaveClass('is-current');
+          expect(item.parentElement).toHaveStyleRule(
```
austin94: I don't think this has to be changed, just want to surface another testing strategy I had to use with CSS-in-JS.

We have a babel transform which removes all `data-test-*` attributes in production builds. If you didn't want to assert a color style here you could also use a `data-test-current="true|false"` with `toHaveAttribute()` if it makes more sense.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,24 @@
...
+ * 1. <ol> reset.
+ */
+export const StyledBreadcrumb = styled.ol.attrs(() => {
+  return {
+    'data-garden-id': COMPONENT_ID,
+    'data-garden-version': PACKAGE_VERSION
+  };
+})`
+  ${props => `
+    display: flex; ${/* [1] */ ''}
```
austin94: Comments can be inline (no template strings). They are removed in production builds with the babel plugin.

```
display: flex; /* [1] */
```

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,24 @@
...
-    // RTL
-    [BreadcrumbStyles['is-rtl']]: isRtl(props)
-  })
-}))`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+/**
+ * 1. <ol> reset.
+ */
+export const StyledBreadcrumb = styled.ol.attrs(() => {
+  return {
```
austin94: This can be an immediate object return since we no longer use `props` here.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
```diff
@@ -6,19 +6,27 @@
  */
 
 import React from 'react';
+import { stripUnit } from 'polished';
```
austin94: Use the `lib` imports to help reduce bundle size for consumers. Probably something like `polished/lib/utilities/stripUnit`

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,34 @@
...
+import { getColor } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'breadcrumbs.item';
 
 export interface IStyledBreadcrumbItemProps {
   isCurrent?: boolean;
 }
 
-export const StyledBreadcrumbItem = styled.li.attrs<IStyledBreadcrumbItemProps>(props => ({
+export const StyledBreadcrumbItem = styled.li.attrs(() => ({
```
austin94: Direct object return should be used

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,34 @@
...
-  })
+  'data-garden-version': PACKAGE_VERSION
 }))<IStyledBreadcrumbItemProps>`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${props => `
+    line-height: ${(props.theme.space.base * 5) / stripUnit(props.theme.fontSizes.md)}}
+    color: ${props.isCurrent ? getColor(props.theme.colors.neutralHue, 600) : 'inherit'}
+    & a {
+      &:any-link {
+        color: ${props.isCurrent ? 'inherit' : ''};
```
austin94: There isn't a need to do an `else` return of an empty string. Styled-components will remove this rule if undefined/false is returned. Change to:

``
${props.isCurrent && 'inherit'}
```

File: packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts [hunk 13]
```diff
@@ -6,23 +6,34 @@
...
-    [BreadcrumbStyles['is-current']]: props.isCurrent
-  })
+  'data-garden-version': PACKAGE_VERSION
 }))<IStyledBreadcrumbItemProps>`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${props => `
+    line-height: ${(props.theme.space.base * 5) / stripUnit(props.theme.fontSizes.md)}}
+    color: ${props.isCurrent ? getColor(props.theme.colors.neutralHue, 600) : 'inherit'}
+    & a {
+      &:any-link {
```
austin94: I'm assuming this came from the CSS side of things, but doesn't this break in IE11 and Edge https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link ?

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const LTR_COMPONENT_ID = 'breadcrumbs.chevron_right_stroke_icon';
+const RTL_COMPONENT_ID = 'breadcrumbs.chevron_left_stroke_icon';
+
+const getGardenId = (rtl: boolean) => (rtl ? RTL_COMPONENT_ID : LTR_COMPONENT_ID);
+
+const ChevronStrokeIcon = (props: SVGAttributes<SVGElement>) => {
+  const { rtl } = useContext(ThemeContext);
+
+  return rtl ? <ChevronLeftStrokeIcon {...props} /> : <ChevronRightStrokeIcon {...props} />;
```
austin94: Rather than using context to conditionally render 2 separate icons (twice the bundle hit) could we use CSS to `rotate(180deg)`?

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
hzhu: Awesome, reducing the bundle size sounds good. I'll make that change.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,24 @@
...
+    'data-garden-version': PACKAGE_VERSION
+  };
+})`
+  ${props => `
+    display: flex; ${/* [1] */ ''}
+    margin: 0; ${/* [1] */ ''}
+    padding: 0; ${/* [1] */ ''}
+    list-style: none; ${/* [1] */ ''}
+    font-size: ${props.theme.fontSizes.md};
+    direction: ${props.theme.rtl ? 'rtl' : 'ltr'};
```
jzempel: Usually we try to leave CSS defaults out of the render. This can just be `direction: ${props.theme.rtl && 'rtl'};` – styled-components will take care of the rest.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.spec.tsx
jzempel: `import stripUnit from 'polished/lib/helpers/stripUnit';`

(although spec files don't hit the bundle ;)

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
jzempel: It would be good if you break this out to an `anyLinkStyles` function – similar to what we did with postCSS [input range pseudo](https://github.com/jonathantneal/postcss-input-range) here: https://github.com/zendeskgarden/react-components/blob/1067eae528c03996c34c364046bdbff8c756e4c0/packages/forms/src/styled/range/StyledRangeInput.js#L34-L48

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const StyledChevronIcon = styled(ChevronStrokeIcon).attrs(props => ({
+  'data-garden-id': `${getGardenId(props.theme.rtl)}`,
+  'data-garden-version': PACKAGE_VERSION,
+  'aria-hidden': 'true'
+}))`
+  display: inline-block;
+  position: relative;
+  top: 1px;
+  margin: 0 calc(8 / 14 * 1em);
```
jzempel: Looks like you've got yourself backed into a corner with this one. Whenever you find yourself "pixel-nudging" it's a good indication that it's time to step back and look at the problem differently. The breadcrumb is already flexbox, so it seems like this should just be a flex item with appropriate `align-self` applied.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,34 @@
...
+      &:any-link:hover {
+        color: ${props.isCurrent ? 'inherit' : ''};
+      }
+      &:any-link:focus {
+        color: ${props.isCurrent ? 'inherit' : ''};
+      }
+    }
+  `}
+  white-space: nowrap;
+  font-size: inherit;
```
jzempel: @hzhu once this is cleaned up based on existing comments, let's get together and review CSS property ordering for styled components. Thanks :pray:

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Accepts all `<svg>` props
+ */
+export const StyledChevronIcon = styled(ChevronStrokeIcon).attrs(props => ({
+  'data-garden-id': `${getGardenId(props.theme.rtl)}`,
+  'data-garden-version': PACKAGE_VERSION,
+  'aria-hidden': 'true'
+}))`
+  display: inline-block;
+  position: relative;
+  top: 1px;
```
hzhu: This pixel pushing isn't great for maintainability so I'm going to look into using flexbox to center the breadcrumb separator.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
hzhu: Thanks for the suggestion 👍. I updated it so that CSS default is left out of the render.

packages/breadcrumbs/package.json
```diff
@@ -31,7 +31,6 @@
     "styled-components": "^4.2.0"
   },
   "devDependencies": {
-    "@zendeskgarden/css-breadcrumbs": "0.2.18",
     "@zendeskgarden/react-theming": "^7.0.0",
     "@zendeskgarden/react-utilities": "^7.0.0"
```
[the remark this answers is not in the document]
hzhu: I confirmed the utility methods aren't used and removed the dependency from the `package.json`.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
hzhu: Looks like styled-components isn't removing the style.

Screen Shot 2019-11-28 at 10 52 33 AM

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
hzhu: @jzempel thanks for the suggestion! I added an `anyLinkStyles` function that takes a string template of CSS and applies it to the `any-link` pseudo-class selectors. What do you think about the following?

```css
/**
 * 1. CSS targets Microsoft Edge
 */
const anyLinkStyles = (styles: string) => `
  & a {
    &:any-link { ${styles} }
    &:any-link:hover { ${styles} }
    &:any-link:focus { ${styles} }
    &:link { ${styles} } /* [1] */
    &:visited { ${styles} } /* [1] */
    &:focus:link { ${styles} } /* [1] */
    &:hover:link { ${styles} } /* [1] */
    &:focus:visited { ${styles} } /* [1] */
    &:hover:visited { ${styles} } /* [1] */
  }
`;
```

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
hzhu: @jzempel that sounds great! I've re-ran the CSS lint and re-ordered CSS according to [stylelint-order](https://github.com/zendeskgarden/stylelint-config/blob/master/plugins/stylelint-order.js).

Screen Shot 2019-11-30 at 5 45 15 PM

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
hzhu: Thanks for bringing this up, I will be more aware of avoiding pixel pushing in the future. The `<ol>` is flexbox but applying various `align-self` values did not center the chevron. However, I applied `display:flex;` and `align-items:center;` to the list items and that got the chevrons centered.

packages/breadcrumbs/src/elements/Breadcrumb.spec.tsx
hzhu: Nice, thanks for surfacing another testing strategy!

The test above this one, (`applies current aria values correctly`) also verifies that the last item is current by asserting that it has `aria-current=page`. The `toHaveStyleRule` is used to verify the color style because this test description, (`receives current styling if last`), involves the current styling.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,22 @@
...
+  'data-garden-version': PACKAGE_VERSION
+})`
+  ${props => `
+    display: flex; /* [1] */
+    margin: 0; /* [1] */
+    padding: 0; /* [1] */
+    list-style: none; /* [1] */
+    font-size: ${props.theme.fontSizes.md};
+  `};
+  ${props => props.theme.rtl && 'direction: rtl'}
```
austin94: This syntax you're using is a little different than what we do in the other components. I would use:

```tsx
styled.ol`
  display: flex;
  margin: 0;
  /* etc */
  font-size: ${props => props.theme.fontSizes.md};
  direction: ${props => props.theme.rtl && 'rtl'}
`;
```

if you have a block which uses _a lot_ of themed values you can use an alternative signature that is similar to

```tsx
styled.ol(props => `
  display: flex;
  font-size: ${props.theme.fontSizes.md};
  /* etc */
`);
```

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,44 @@
...
+})<IStyledBreadcrumbItemProps>`
+  ${props => `
+    display: flex;
+    align-items: center;
+    line-height: ${(props.theme.space.base * 5) / stripUnit(props.theme.fontSizes.md)};
+    white-space: nowrap;
+  `}
+  ${props =>
+    `color: ${props.isCurrent ? getColor(props.theme.colors.neutralHue, 600) : 'inherit'};`}
+  ${props => props.isCurrent && anyLinkStyles(`color: inherit;`)}
```
austin94: If this is only used in 1 place, is there a reason to pass the color style as an attribute? This could be more performant if you include this value inline above.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+const LTR_COMPONENT_ID = 'breadcrumbs.chevron_right_stroke_icon';
+const RTL_COMPONENT_ID = 'breadcrumbs.chevron_left_stroke_icon';
+
+const getGardenId = (rtl: boolean) => (rtl ? RTL_COMPONENT_ID : LTR_COMPONENT_ID);
+
+/**
+ * Accepts all `<svg>` props
+ */
+export const StyledChevronIcon = styled(ChevronRightStrokeIcon).attrs(props => ({
+  'data-garden-id': `${getGardenId(props.theme.rtl)}`,
```
austin94: The `getGardenId` utility returns a string, you shouldn't wrap this in a template.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+  'data-garden-version': PACKAGE_VERSION,
+  'aria-hidden': 'true'
+}))`
+  ${props => props.theme.rtl && `transform:rotate(180deg);`}
```
austin94: To match the other components use

```tsx
transform: ${props => props.theme.rtl && 'rotate(180deg)'};
```

packages/breadcrumbs/src/elements/Breadcrumb.tsx
```diff
@@ -27,7 +26,12 @@ export const Breadcrumb = React.forwardRef<HTMLElement, HTMLAttributes<HTMLEleme
         );
       }
 
-      return <StyledBreadcrumbItem>{child}</StyledBreadcrumbItem>;
+      return (
+        <StyledBreadcrumbItem>
+          {child}
+          <StyledChevronIcon />
```
jzempel: I'm not convinced this is an effective component layout. By making this icon a child of the item, you're failing to leverage the outer flexbox container which makes the `StyledBreadcrumbItem` need to function as both a flexbox container and flexbox item.

I'd recommend refactoring so that the `StyledChevronIcon` is a specific type of `StyledBreadcrumbItem` and is positioned as a sibling (aka direct child of `StyledBreadcrumb`).

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
jzempel: :+1: to @austin94's comments here. Please remain consistent with other refactored CSS-in-JS components.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,44 @@
...
 export interface IStyledBreadcrumbItemProps {
   isCurrent?: boolean;
 }
 
-export const StyledBreadcrumbItem = styled.li.attrs<IStyledBreadcrumbItemProps>(props => ({
+/**
+ * 1. CSS targets Microsoft Edge
+ */
+const anyLinkStyles = (styles: string) => `
+  & a {
```
jzempel: This is not the current styling seen in the CSS https://unpkg.com/browse/@zendeskgarden/css-breadcrumbs@0.2.20/dist/index.css. If you are styling for a child `a` tag, then none of the `:any-link` pseudo is necessary. The point of `:any-link` is to apply styling to any "link-ish" type element. See https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link for details.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
jzempel: You failed to capture the `white-space: inherit;` CSS property for `:any-link` children. Since that doesn't have `:hover`/`:focus` implications, you'll need to re-think the `anyLinkStyles` function.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -6,23 +6,44 @@
...
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(BreadcrumbStyles['c-breadcrumb__item'], {
-    // State
-    [BreadcrumbStyles['is-current']]: props.isCurrent
-  })
-}))<IStyledBreadcrumbItemProps>`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledBreadcrumbItemProps>`
+  ${props => `
```
jzempel: This `props` block is not a pattern we use throughout the `CSS-in-JS` refactoring.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import ChevronRightStrokeIcon from '@zendeskgarden/svg-icons/src/12/chevron-right-stroke.svg';
+
+const LTR_COMPONENT_ID = 'breadcrumbs.chevron_right_stroke_icon';
+const RTL_COMPONENT_ID = 'breadcrumbs.chevron_left_stroke_icon';
```
jzempel: Do not vary the component ID based on RTL. Please see the explanation of "Components" under https://competent-shirley-cc2f82.netlify.com/theming/#default_theme. The consumer has access to props in order to make styling tweaks. This flavor of ID is unnecessary (and unexpected based on all of our other styled components).

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Accepts all `<svg>` props
+ */
+export const StyledChevronIcon = styled(ChevronRightStrokeIcon).attrs(props => ({
+  'data-garden-id': `${getGardenId(props.theme.rtl)}`,
+  'data-garden-version': PACKAGE_VERSION,
+  'aria-hidden': 'true'
+}))`
+  ${props => props.theme.rtl && `transform:rotate(180deg);`}
+  margin: 0 calc(8 / 14 * 1em);
```
jzempel: This is the right calculation from CSS, but it should be reasoned as a function based on theming values. Based on what we see in the designs, it is safe to assume the intent is a `4px` horizontal margin – so simply base it on `space.base`:

`em(props.theme.space.base, props.theme.fontSizes.md)` (`em` is a function from polished)

Also, there should be (almost) no `calc` in our CSS.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
hzhu: Thanks for the suggestion @austin94 — I'll make sure to be consistent with the existing code.

packages/breadcrumbs/src/elements/Breadcrumb.tsx
hzhu: > I'd recommend refactoring so that the StyledChevronIcon is a specific type of StyledBreadcrumbItem

Thanks for pointing this out 👍. This was one of the alternatives I considered as a result of exploring breadcrumb components from other open source projects. It did not cross my mind to leverage the outer flexbox container, but that makes a lot of sense to do so. 

I will restructure the code so that the chevron is a direct sibling of the`StyledBreadcrumb`. That way, the UI can leverage the outer flexbox container.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
hzhu: Whoops — thanks for pointing this out.  I will make sure to stay consistent with other styled-components CSS in the future.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
hzhu: Thanks for the suggestion 👍. I have removed component ID related code in this styled component because it was unnecessary and unexpected.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
hzhu: I removed this code entirely because of JZ's comments pointed out that this was unnecessary (which sounds fair to me).  I will go ahead and mark this as resolved.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
hzhu: > Also, there should be (almost) no calc in our CSS.

Thanks for pointing this unreasoned line of CSS out. I had a feeling I was cutting corners. My apologies and will make sure to be more thorough about hard coding magical values. I have updated this particular line of CSS to:

```jsx
margin: 0 ${props => em(props.theme.space.base * 2, props.theme.fontSizes.md)};
```

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -5,24 +5,41 @@
...
 
 const COMPONENT_ID = 'breadcrumbs.item';
 
 export interface IStyledBreadcrumbItemProps {
   isCurrent?: boolean;
 }
 
-export const StyledBreadcrumbItem = styled.li.attrs<IStyledBreadcrumbItemProps>(props => ({
+/**
+ * These CSS indivual pseudo-classes are used to match the equivilant of :any-link.
```
rossmoody: Mega nit: equivalent

hzhu: Mega thanks 🙏@rossmoody.

packages/breadcrumbs/src/elements/Breadcrumb.tsx
```diff
@@ -27,7 +26,14 @@ export const Breadcrumb = React.forwardRef<HTMLElement, HTMLAttributes<HTMLEleme
         );
       }
 
-      return <StyledBreadcrumbItem>{child}</StyledBreadcrumbItem>;
+      return (
+        <>
+          <StyledBreadcrumbItem>{child}</StyledBreadcrumbItem>
+          <StyledBreadcrumbItem style={{ display: 'flex' }} aria-hidden>
```
rossmoody: Why does this one get an inline `css` designation?

hzhu: Good question! **I applied a `style` attribute to `StyledBreadcrumbItem` (which renders an `<li>`) so that the contents of the `<li>` can be centered**. In this particular case, the content is a chevron SVG icon. The HTML for this UI looks like:

```html
<ol> // flex container
  <li>one</li>
  <li> // make this a flex container to center chevron
    <svg>chevron</svg>
  </li>
  <li>two</li>
  <li>
    <svg>chevron</svg>
  </li>
  <li>three</li>
</ol>
```

Alternatively, I could have relied on the `<ol>` flex container, and I did not have to apply flex to each `<li>` that had a chevron. That would look like the following:

```html
<ol> // flex container
  <li>one</li>
  <svg>chevron</svg> // centered because it's a direct child of `<ol>` flex container
  </li>
  <li>two</li>
  <svg>chevron</svg> // centered because it's a direct child of `<ol>` flex container
  <li>three</li>
</ol>
```

I avoided the second alternative because the [HTML spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element) recommends only `<li>` as the contents of an `<ol>`.

austin94: If this requires custom styling, we should be applying this through styled-components. I think it could still live in this file though.

```tsx
const StyledCenteredBreadcrumbItem = styled(StyledBreadcrumbItem)`
  display: flex;
`;
```

hzhu: @rossmoody @austin94  I've cleaned this up, thank you.

packages/breadcrumbs/src/styled/StyledBreadcrumb.ts
```diff
@@ -6,19 +6,21 @@
...
-  className: classNames(BreadcrumbStyles['c-breadcrumb'], {
-    // RTL
-    [BreadcrumbStyles['is-rtl']]: isRtl(props)
-  })
-}))`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  align-items: center;
```
jzempel: Shouldn't need this `align-items`.

packages/breadcrumbs/src/styled/StyledBreadcrumbChevronItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { StyledBreadcrumbItem, StyledChevronIcon } from './index';
+
+const StyledCenteredBreadcrumbItem = styled(StyledBreadcrumbItem)`
+  display: flex;
```
jzempel: Add `align-items: center` here.

packages/breadcrumbs/src/styled/StyledBreadcrumbChevronItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+`;
+
+export const StyledBreadcrumbChevronItem = () => (
+  <StyledCenteredBreadcrumbItem aria-hidden>
+    <StyledChevronIcon />
+  </StyledCenteredBreadcrumbItem>
+);
```
jzempel: This is functionality we should see under `elements` - not under `styled`.

packages/breadcrumbs/src/styled/StyledBreadcrumbItem.ts
```diff
@@ -5,24 +5,41 @@
...
+ */
+const linkStyles = ({ isCurrent }: IStyledBreadcrumbItemProps) => css`
+  & > :link,
+  & > :visited {
+    white-space: inherit;
+  }
+
+  ${isCurrent &&
+    `
+      & > :link, & > :visited, & > :link:hover, & > :visited:hover, & > :link:focus, & > :visited:focus {
```
jzempel: Please format this line with line breaks after the commas for readabililty.

packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Accepts all `<svg>` props
+ */
+export const StyledChevronIcon = styled(ValidChevronIcon).attrs({
+  role: 'presentation',
+  'aria-hidden': 'true'
+})`
+  transform: ${props => props.theme.rtl && `rotate(180deg);`};
+  margin: 0 ${props => em(props.theme.space.base * 2, props.theme.fontSizes.md)};
```
jzempel: @rossmoody please confirm – there should only be a 4px margin around the chevron, not 8px?

hzhu: Confirmed with @rossmoody that this should be `4px`. I'll go ahead and get this updated. Thanks for pointing it out!

PR #547, opened 2019-11-25
Allow the portal container to be passed in as a prop

luis-almeida:
## Description

Being able to define the portal container would allow us to render the modal into a shadow root element.

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/modals/src/elements/Modal.js
+++ b/packages/modals/src/elements/Modal.js
@@ -46,7 +46,7 @@ const isOverflowing = element => {
  * High-level abstraction for basic Modal implementations. Accepts all `<div>` props.
  */
 const Modal = React.forwardRef(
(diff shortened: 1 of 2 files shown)


packages/modals/src/elements/Modal.js
```diff
@@ -132,6 +132,10 @@ Modal.propTypes = {
    * @param {Object} event - DOM event that triggered the close action
    */
   onClose: PropTypes.func,
+  /**
+   * The DOM element where the modal will be rendered to
+   **/
+  container: PropTypes.instanceOf(Element),
```
austin94: Could we rename this prop to `appendToNode`? We use this style in other components where it is optional, might make sense to continue it here.

luis-almeida:
Thanks @austin94 for the review.
I address your comments in my last two commits. Can you take another look when you have time?

PR #555, opened 2019-12-06
feat(tables): migrate react-tables to TS

austin94:
## Description

This PR migrates `react-tables` to TS. It includes some heavy-rewrites of the examples to support the default `<table>` styling that is now used.

## Breaking changes

```
## @zendeskgarden/react-tables

- no longer packages a `styles.css` dist; CSS is self-contained
- All components now use standard `<table>` styling
- Removed all `focused`, `active`, and `hovered` states from components
- Component props renames:
  - `Cell`
    - `minimum` -> `isMinimum`
    - `truncate` -> `isTruncated`
    - `menu` -> `hasOverflow`
  - `Row`
    - `striped` -> `isStriped`
    - `selected` -> `isSelected`
```

## Checklist

      designer as a reviewer)
      component


packages/tables/examples/selectable-table.md
```diff
@@ -66,7 +66,7 @@ const isSelectAllChecked = (selectedRows, rows) => {
   <Body>
     {state.rows.map(row => (
       <Row key={row.id} selected={state.selectedRows[row.id]}>
```
hzhu: I noticed that the selected row styling was missing. I think renaming the `selected` prop to `isSelected` would fix it.

Notice: styling is not applied:
TruncatedTitle

packages/tables/examples/paginated-table.md
```diff
@@ -36,34 +36,30 @@ const getPagedData = (data, currentPage, pageSize) => {
 
 <div>
   <Table size="small">
-    <Caption>
-      <XL tag={Caption} style={{ marginBottom: DEFAULT_THEME.space.sm }}>
-        Paginated Tickets
-      </XL>
+    <Caption style={{ marginBottom: DEFAULT_THEME.space.sm }}>
```
hzhu: I noticed that in production, the margin bottom is applied to the `<XL>` immediately below this line. What do you think about removing the margin here, and applying it back to the `<XL>`?

Before (production): 
Screen Shot 2019-12-06 at 2 36 42 PM

Current (feature branch), the spacing between the `<Caption>` and `<XL>` is very close:
Screen Shot 2019-12-06 at 2 40 34 PM

jzempel:
@austin94 please include @allisonacs for example review coverage and please publish the demo to ease the review for all of us. Thanks.

packages/tables/styleguide.config.js
```diff
@@ -5,61 +5,69 @@
...
+  resolver: reactDocgen.resolver.findAllComponentDefinitions,
   sections: [
     {
       name: '',
       content: '../../packages/tables/README.md'
     },
     {
-      name: 'Tables',
-      content: '../../packages/tables/src/examples/tables.md',
+      name: 'Examples',
```
hzhu: This is an open ended question. I noticed that early on, CSS-in-JS example sections are organized like this:

```
Examples
├── Basic
└── Advanced
Elements
├── Element 1
├── Element 2
└── Element 3
```

Should the examples for `react-tables` be organized under "Basic" and "Advanced" too? Or is `react-tables` complex enough to require a different organization strategy? When would one know _not_ to follow the above organization (e.g. organizing under a "Basic", "Advanced", etc)?

austin94: I've included the basic styling/format props as knobs in the first basic example. All of the other examples include a lot of custom logic and styling for their use-case.

austin94: Additionally, having the separated example naming helps with readability. If I remember correctly the "Virtual scrolling" example is the most shared deep-link across all the documentation.

austin94:
Pre-published at https://austin-next.netlify.com/tables/

packages/tables/styleguide.config.js
hzhu: Thanks for the clarification!

packages/tables/src/elements/OverflowButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+    >
+      &nbsp;
+    </StyledOverflowButton>
+  );
+});
+
+OverflowButton.propTypes = {
+  isHovered: PropTypes.bool,
+  isActive: PropTypes.bool,
+  isFocused: PropTypes.bool
```
jzempel: Do we see this getting cleaned-up with CSS-in-JS?

packages/tables/src/elements/Row.tsx
```diff
@@ -0,0 +1,43 @@
...
+      ref={ref}
+      {...otherProps}
+    />
+  );
+});
+
+Row.propTypes = {
+  isStriped: PropTypes.bool,
+  isFocused: PropTypes.bool,
+  isHovered: PropTypes.bool,
```
jzempel: Why do we want to expose these props at the element level? We don't want folks manually tweaking hover/focus states.

packages/tables/src/elements/SortableCell.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <StyledHeaderCell aria-sort={ariaSortValue} width={width} {...cellProps}>
+      <StyledSortableButton sort={sort} ref={ref} {...otherProps} />
+    </StyledHeaderCell>
+  );
+});
+
+SortableCell.propTypes = {
+  sort: PropTypes.oneOf(['asc', 'desc']),
+  isFocused: PropTypes.bool,
+  isActive: PropTypes.bool,
```
jzempel: Same comment here re: not wanting users to manually control focus/active states.

packages/tables/src/elements/Table.tsx
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Accepts all `<table>` attributes and events
+ */
+export const Table = React.forwardRef<
+  HTMLTableElement,
+  IStyledTableProps & HTMLAttributes<HTMLTableElement>
+>((props, ref) => <StyledTable ref={ref} {...props} />);
+
+Table.propTypes = {
+  size: PropTypes.oneOf(['small', 'large'])
```
jzempel: Following on other components, we should have a `medium` (default).

packages/tables/src/styled/StyledHeaderCell.ts
```diff
@@ -0,0 +1,19 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles } from '@zendeskgarden/react-theming';
+import { StyledCell } from './StyledCell';
+
+const COMPONENT_ID = 'tables.header_cell';
+
+export const StyledHeaderCell = styled(StyledCell.withComponent('th')).attrs({
```
jzempel: Prefer `as` attr vs the `withComponent` deprecation candidate.

packages/tables/examples/basic.md
```diff
@@ -0,0 +1,162 @@
...
+            <ToggleLabel>Grouped rows</ToggleLabel>
+          </Toggle>
+        </FormsField>
+      </Layout.Col>
+    </Layout.Row>
+  </Well>
+  <StyledSpacer />
+  <Layout.Row>
+    <Layout.Col>
+      <Table size={state.rowSize === 'default' ? undefined : state.rowSize}>
```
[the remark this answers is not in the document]
austin94: With the examples listed about I ended up with a large if statement which rendered 5 unique examples. I did several iterations and ended up combining some of the simpler examples, but some of these needed to be split out to make them usable as base-line examples.

packages/tables/src/elements/Row.tsx
austin94: Overall, I was planning on addressing these API removals in the CSS-in-JS PR.

But, `Row` may need to be our exception here. There are several use-cases where a consumer may need to manually apply this styling based on a table state. The `drag-and-drop` example is a good use-case for making the `isHovered` prop available.

packages/tables/src/elements/SortableCell.tsx
austin94: I was planning to address this in the CSS-in-JS PR to ensure I bring over all the necessary styling.

packages/tables/src/elements/OverflowButton.tsx
austin94: I was planning to address this in the CSS-in-JS PR to ensure I bring over all the necessary styling.

packages/tables/src/styled/StyledHeaderCell.ts
austin94: I've moved to the `as` prop in `attrs`.

Looking through the docs, [forwardedAs](https://www.styled-components.com/docs/api#forwardedas-prop) also looks really useful. We will have to watch our `peerDependency` range for some of these utilities as we are on the low-end for some of these props.

packages/tables/src/elements/Table.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import React, { HTMLAttributes } from 'react';
+import PropTypes from 'prop-types';
+
+import { StyledTable, IStyledTableProps } from '../styled';
+
+/**
+ * Accepts all `<table>` attributes and events
+ */
+export const Table = React.forwardRef<
```
hzhu: I noticed that there are [existing](https://github.com/zendeskgarden/react-components/blob/next/packages/chrome/src/elements/Chrome.spec.tsx#L26) [unit](https://github.com/zendeskgarden/react-components/blob/next/packages/grid/src/elements/Grid.spec.tsx#L19) [tests](https://github.com/zendeskgarden/react-components/blob/next/packages/tabs/src/elements/Tab.spec.tsx#L18) which assert that components exported to consumers should forward refs.

Should this PR include unit test for `Table` ( and `Body`, `Caption`, `Cell`, `GroupRow`, `Head`, `HeaderCell`, `HeaderRow`, `OverflowButton`, `Row`) now that these components have been updated to support ref forwarding?

packages/tables/examples/basic.md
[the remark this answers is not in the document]
jzempel: I still feel like there's a both/and option here. Do the combo thing, but break out the individual examples for code readability. This is the direction I would be trying to take this example, but YMMV.

packages/tables/src/elements/Table.tsx
austin94: I've now added these `ref` checks. I also found a invalid DOM nesting warnings from `react-dom` (thanks react-testing-library 🎉 ). I've updated the `Cell/Row` examples to include wrapping `Table/Body` elements

austin94:
> I would expect the blue color to change back to grey if I move my mouse off the control but it stays blue bc that's where the browser focus is. Can we fix that?

We can address this in the CSS-in-JS PR if we feel it is necessary.

> Maybe a toggle to turn the caption on/off, and something that inserts/removes the <MD> as well?

I've added a toggle to the example and modified the styling to use our theming values.

allisonacs:
> > I would expect the blue color to change back to grey if I move my mouse off the control but it stays blue bc that's where the browser focus is. Can we fix that?
> 
> We can address this in the CSS-in-JS PR if we feel it is necessary.

Yea, let's make sure we do this. It feels off right now.

allisonacs:
@austin94 the first part of my comment was that it seemed like there were more individual examples than I would have anticipated. Is there a way to cleanly cut down on the number of examples?

austin94:
> Is there a way to cleanly cut down on the number of examples?

I don't see a way to combine the remaining example while still allowing developers to use them as an implementation reference. Each of them require custom rendering and styling logic that is complex enough to not be shared.

If we want to spend more time combining these examples I believe it would be best to estimate it in a separate PR.

PR #556, opened 2019-12-06
feat(notifications): migrate react-notifications to TS 

hzhu:
## Description

This pull request migrates the `react-notifications` package to TypeScript in preparation for the CSS-in-JS work.

## Detail

This PR comprises of the following work for the `react-notifications` package:

1) Configure and convert files to TypeScript.
2) Organize files into the `src/elements` and `src/styled` directories.
3) Update Storybook example with Knobs. 

## Demo

A demo pre-published for review can be found at https://henry-next.netlify.com/notifications. 

## Checklist

      designer as a reviewer)
      component~


packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,212 @@
...
+discernible text.
+
+```jsx
+const { Field, Label, Radio } = require('@zendeskgarden/react-forms/src');
+
+const initialState = { type: 'success' };
+
+<Grid>
+  <Row>
+    <Col size={12}>
```
austin94: [nit] not related to the `notifications` part of the PR, but I think the `Col` defaults to `[size="12"]`. You might be able to drop it.

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,212 @@
...
+      <div role="group" aria-label="alert types radio">
+        <Field>
```
austin94: This should use a `react-typography` element here. This is now a `label` without an associated `input`

packages/notifications/package.json
```diff
@@ -44,5 +44,5 @@
   },
   "zendeskgarden:library": "GardenNotifications",
   "zendeskgarden:max_size": "3 kB",
-  "zendeskgarden:src": "src/index.js"
+  "zendeskgarden:src": "src/index.ts"
```
austin94: This should also include the `types` field

packages/notifications/src/elements/Alert.ts
```diff
@@ -0,0 +1,24 @@
...
+import { VALIDATION } from '../utils/validation';
+
+export interface IAlertProps {
+  type?: VALIDATION;
+}
+
+/**
+ * Supports all `<div>` props
+ */
+export const Alert = StyledAlert as React.FC<IAlertProps>;
```
austin94: This is downcasting from `forwardRefExoticComponent` to `functionComponent` and removes the ability to use the `ref` prop.

You can use `forwardRef` and use the `IStyledAlertProps` interface as the props.

packages/notifications/src/elements/Notification.ts
```diff
@@ -0,0 +1,24 @@
...
+import { VALIDATION } from '../utils/validation';
+
+export interface INotificationProps {
+  type?: VALIDATION;
+}
+
+/**
+ * Supports all `<div>` props
+ */
+export const Notification = StyledNotification as React.FC<INotificationProps>;
```
austin94: Same comment as `Alert`

packages/notifications/src/elements/content/Paragraph.ts
```diff
@@ -0,0 +1,14 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { StyledParagraph } from '../../styled';
+
+/**
+ * Used for multi-line Notification content. Supports all `<div>` props
+ */
+export const Paragraph = StyledParagraph as React.FC;
```
austin94: This removes the ability to apply any props to the `Paragraph` tag.

Use the `forwardRef` API and include `HTMLAttributes<HTMLElement>` as the props

packages/notifications/src/elements/content/Title.ts
```diff
@@ -0,0 +1,14 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { StyledTitle } from '../../styled';
+
+/**
+ * Used for Notification titles. Supports all `<div>` props
+ */
+export const Title = StyledTitle as React.FC;
```
austin94: Same comments as `Paragraph`

packages/notifications/src/utils/validation.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+export enum VALIDATION {
+  SUCCESS = 'success',
+  WARNING = 'warning',
+  ERROR = 'error',
+  INFO = 'info'
+}
+
+export type ValidationTypes = 'success' | 'warning' | 'error' | 'info';
```
austin94: In many areas you use both `VALIDATION` and `ValidationTypes`. Also, I'm not sure if you have tried to build this yet, but named `ENUMS` probably won't work with our current babel target.

I suggest following the structure of the other packages and having a `type VALIDATION = 'success' | 'warning' | 'error' | 'info'`

hzhu: > I'm not sure if you have tried to build this yet, but named ENUMS probably won't work with our current babel target.

This seems to pass the build, but I had issues with using the named `enum` in development which is why I needed the extra `ValidationTypes`. I will go back and try to consolidate everything to a `type VALIDATION`. Thanks for the suggestion!

packages/notifications/src/elements/Alert.ts
hzhu: Should we also update `react-dropdowns` to not downcast to `React.FC`? I was following how [the Dropdown components](https://github.com/zendeskgarden/react-components/blob/next/packages/dropdowns/src/Fields/Message.ts#L19) were migrated to TypeScript.

Seems like a lot of our existing components don't support ref forwarding. Moving forward, I'll make sure to configure components to support ref forwarding if the component is exported to consumers.

Thanks for the suggestion, I'll go ahead and update these components and re-request a review.

packages/notifications/package.json
hzhu: Sounds good, I can add the `types` field. Should `"typings": ./index.d.ts`? That was [already in the file](https://github.com/zendeskgarden/react-components/blob/next/packages/notifications/package.json#L21).

packages/notifications/examples/basic.md
hzhu: Sounds good, I'll go ahead and use that.

packages/notifications/src/elements/Alert.ts
austin94: > Seems like a lot of our existing components don't support ref forwarding.

Yep, this wasn't very consistent before the CSS-in-JS migration. When it's time to move `react-dropdowns` over I will clean up those usages.

packages/notifications/src/utils/validation.ts
jzempel: @hzhu not totally sure it addresses what you're trying to achieve, but consider the types and array constants seen here: https://github.com/zendeskgarden/react-components/blob/next/packages/grid/src/utils/types.ts

jzempel:
When this is ready, please pre-publish the demo to your own netlify and add design reviewers.

packages/notifications/src/utils/validation.ts
hzhu: @austin94 

A named `enum` is [used](https://github.com/zendeskgarden/react-components/blob/next/packages/dropdowns/src/styled/field/StyledMessage.ts#L15) in the `react-dropdowns`. Could you confirm if named `enums` are okay to use? I'm using a named `enum` in this PR, and it seems to work. 

Also, I didn't find any packages in the `next` branch having `type VALIDATION = 'success' | 'warning' | 'error' | 'info'`. Which packages can I find that?

@jzempel 

That file is beneficial, thank you for sharing! I went ahead and used it as a guideline. Instead of using creating an `Array` of types (like [this](https://github.com/zendeskgarden/react-components/blob/next/packages/grid/src/utils/types.ts#L9)), I created an `enum` of types because `react-notification` uses object property dot notation [here](https://github.com/zendeskgarden/react-components/blob/next/packages/notifications/src/Alert.js#L34) and [here](https://github.com/zendeskgarden/react-components/blob/next/packages/notifications/src/Notification.js#L34) — a behavior that a typed array would not support.

I ended up with the following, which worked well for me:

```ts
/*
./react-components/packages/notifications/src/utils/types.ts
*/

export type VALIDATION_TYPES = 'success' | 'warning' | 'error' | 'info';

export enum VALIDATION {
  SUCCESS = 'success',
  WARNING = 'warning',
  ERROR = 'error',
  INFO = 'info'
}
```

packages/notifications/package.json
austin94: That `index.d.ts` can be removed. This was an attempt to manually include type definitions for some of our TS consumers.

You should be able to just have `"types": "dist/typings/index.d.ts",` like the other packages

packages/notifications/src/elements/Alert.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyledAlert, IStyledAlertProps } from '../styled';
+import { VALIDATION } from '../utils/types';
+
+/**
+ * Supports all `<div>` props
+ */
+export const Alert = React.forwardRef<HTMLDivElement, IStyledAlertProps>((props, ref) => (
```
austin94: These components are able to consume our custom props from `IStyledAlertProps`, but they're also able to consume all valid HTMLAttributes for a DIV element.

A more accurate type definition would we `IStyledAlertProps & HTMLAttributes<HTMLDivElement>`

austin94: This should be replicated across all `forwardRef` usages

packages/notifications/src/utils/validation.ts
austin94: I would avoid using that value, even if it's in use with `react-dropdowns`. That was the first TS package so there are some things in there we need to align with our other packages.

> created an enum of types because react-notification uses object property dot notation here and here

This dot-notation is a side-effect of using a CONSTANTS object. With TS this isn't needed. We can rely on the type-system to enforce this. You can just use `validation === 'success'` instead of `validation === VALIDATION.SUCCESS`

BREAKING_CHANGES.md
```diff
@@ -4,6 +4,14 @@ NOTE: many style-related COMPONENT_IDs were renamed for consistency based
 on the CSS-in-JS restructure. All component-level theming customizations
 should be re-checked for ID naming accuracy.
 
+## @zendeskgarden/react-notifications
+
+- Boolean props for this package is updated to align with `isBoolean` naming. The following are
```
rossmoody: Nit: "are updated..."

packages/notifications/package.json
hzhu: Great, I removed the file and updated the `package.json` 👍 .

packages/notifications/src/utils/validation.ts
hzhu: That sounds good. The code has been updated to not use property lookup on an object. That way it now relies on the type-system alone to enforce safety.

packages/notifications/src/elements/Alert.tsx
hzhu: Nice catch, I've updated the type definition for the generic to accept `HTMLAttributes<HTMLDivElement>`.

BREAKING_CHANGES.md
hzhu: The markdown linter asked to drop this line at 100 characters. The sentence continues below this line, but let me know if the sentence needs to be changed for different reason.

hzhu: Reviewing this line, I think you might have meant adding the word "updated" 👀.

hzhu: @rossmoody, did you think adding the word provides clarity (or maybe some other benefit)? I went ahead and added the word. 

I updated

> - Boolean props for this package is updated to align with `isBoolean` naming. The following are prop name changes for the `<Well>` component:
>
>   - `floating` -> `isFloating`
>   - `recessed` -> `isRecessed`

to

> - Boolean props for this package is updated to align with `isBoolean` naming. The following are updated prop name changes for the `<Well>` component:
>
>   - `floating` -> `isFloating`
>   - `recessed` -> `isRecessed`

hzhu: Third times a charm. Updated to:

> - Boolean props for this package are updated to align with `isBoolean` naming. The following are  prop name changes for the `<Well>` component:

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,240 @@
...
+      <Well isRecessed>
+        <Field>
+          <Toggle
+            checked={state.isMultiLine}
+            onChange={event => setState({ isMultiLine: event.target.checked })}
+          >
+            <Label>Multi-line</Label>
+          </Toggle>
+        </Field>
+      </Well>
```
jzempel: Why is this in its own well rather than being combined with the controls above?

ginnywood:
Looking really good! I think the way you’ve structured the knobs is logical and concise.

I think I found a preexisting visual bug. Alerts and Notifications should have 40px left padding. Looks like right now, it’s 50px.

jzempel:
@hzhu @allisonacs per conversations with both of you, let's drop the "multi-line" knob. It's super confusing with respect to the API which is "use a `<Paragraph>` if your text spans multiple lines, otherwise just drop in plain text". I think we might agree that the API is also confusing and Garden could do a better job of setting users up for success. But that's a job for later.

@hzhu plz work with @allisonacs to either:
1. apply both a single and multi-line example to each example, or
1. drop whatever is determined to be the least helpful layout from each example

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,240 @@
...
+const capitalize = string => string.charAt(0).toUpperCase() + string.slice(1);
+
+const initialState = { type: 'success' };
+
+<Grid>
+  <Row>
+    <Col>
+      <Well isRecessed>
+        <div role="group" aria-label="alert types radio">
+          <MD style={{ color: '#2f3941', fontWeight: 600 }}>Alert Types</MD>
```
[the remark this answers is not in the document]
austin94: I'm not sure the `<Field><Label>Foo</Label></Field>` you used there is valid. It is currently showing axe warnings.

The label that is rendered has a `[for=""]` attribute that is pointing to a non-existent input.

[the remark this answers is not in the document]
jzempel: :sob: I think I stole that from another old Styleguidist example. We don't have a good story around the layout for a radio group. I've added an issue to our backlog.

allisonacs:
@hzhu let's just remove the multiline examples. The "just us a `<p>`" will render paragraphs when needed, but the primary use cases for all these items is shorter content.

packages/notifications/examples/basic.md
hzhu: It does look strange, I think since this wasn't associated to a prop, I left it on a separate `<Well>`. I moved it all into a single `<Well>`.

packages/notifications/examples/basic.md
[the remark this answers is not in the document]
hzhu: I've removed the inline style replaced this with a `<StyledMd>` in the meanwhile 👍.

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,269 @@
...
+            checked={state.isMultiLine}
+            onChange={event => setState({ isMultiLine: event.target.checked })}
+          >
+            <Label>Show Multi-line</Label>
+          </Toggle>
+        </Field>
+      </Well>
+    </Col>
+    <Col md={8}>
+      <Notification type={state.type}>
```
hzhu: > By default in the sketch libraries, these are 300px wide. We don't spec a width for these in code, but at least showing them similarly to the alert example with the controls side-by-side will be helpful.

Currently, the `<Notification>` example renders to fit its parent container, similar to the current production example. It feels like an exact size would visually limit the different sizes that the `<Notification>` can be seen. Are there pain points around how `<Notification>` is currently presented in the production examples?

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,269 @@
...
+`;
+
+const alertTitles = {
+  success: 'Success Alert',
+  error: 'Error Alert',
+  warning: 'Warning Alert',
+  info: 'Info Alert'
+};
+
+const initialState = { type: 'success', showClose: true };
```
hzhu: @jzempel @allisonacs 

> ...let's drop the "multi-line" knob. It's super confusing with respect to the API which is "use a <Paragraph> if your text spans multiple lines, otherwise just drop in plain text". 

Does the same apply to the `<Alert>` and its adjacent child component `<Close>` too? There isn't a close related prop for `<Alert>`. Similar to how `<Notification>` doesn't have a multi-line prop. `close` and `multi-line` aren't on prop sheets.

Maybe I need to reconsider what a knob means for Garden's examples — knobs don't necessarily map to a prop on the prop sheet. Instead, they are form elements that dynamically show possible renders, which include different usages in combination with various child components (e.g., rendering a `<Paragraph>` within a `<Notification>` or `<Close>`  within a `<Alert> `.

It felt strange to not have a knob for multi-line for `<Notification>`, but have a show close knob for `<Alert>`. I added a knob for both, but renamed their labels to "Show Close" and "Show Multi-line":

Examples

packages/notifications/examples/basic.md
hzhu: Follow up: the first example from this PR had a `<Notification>` that took the full width of the screen. But now that it lives in the same row as the radio group, the widths of the `<Notification>` are good to go. Confirmed with @allisonacs in person.

packages/notifications/examples/basic.md
hzhu: @allisonacs and I walked through the UI that was recorded above. We both think it's sufficient. @jzempel, I know we discussed removing the multi-line toggle, but I added the multi-line toggle back so that there's consistency with `<Alert>`'s "Show Close" toggle. It felt strange having a "Show Close", but not having a "Show Multi-line".

packages/notifications/examples/basic.md
```diff
@@ -0,0 +1,240 @@
...
+      <Notification type={state.type}>
+        <Title>
+          Notification: {capitalize(state.type)} {state.isMultiLine ? '(Multi-line)' : ''}
+        </Title>
+        {state.isMultiLine ? (
+          <Paragraph>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
+            incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
+            exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure
+            dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
```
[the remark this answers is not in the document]
hzhu: I updated this to use veggie flavor ipsum. Moving forward I'll update the existing ipsum to the preferred flavor of ipsum 👍.

PR #560, opened 2019-12-17
feat(tables): migrate to CSS-in-JS

austin94:
- [x] **BREAKING CHANGE**

## Description

This PR moves `react-tables` to CSS-in-JS. It is pre-published to https://austin-next.netlify.com/tables/

## Details

I have left several of the semantic props (focus/active/hovered) available for styling due to usages with drag-drop and virtualization libraries.

## Checklist

      designer as a reviewer)
      component~


packages/tables/src/styled/StyledCell.ts
```diff
@@ -21,18 +20,68 @@ export interface IStyledCellProps {
...
+  } else if (props.size === 'small') {
+    verticalPadding = '6px';
+  }
+
+  if (props.isMinimum) {
+    if (props.theme.rtl) {
+      rightPadding = `${props.theme.space.base * 3}px`;
+    }
+
+    rightPadding = '0';
```
jzempel: [nit] calculating `horizontalPadding` would feel more intuitive throughout (vs. `rightPadding`) and then apply below based on RTL/not-RTL.

packages/tables/src/styled/StyledHeaderCell.ts
```diff
@@ -5,16 +5,73 @@
...
+const truncatedStyling = css`
+  ${StyledOverflowButton} {
+    max-width: 100%;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+`;
+
+const sizeStyles = (props: IStyledCellProps & ThemeProps<DefaultTheme>) => {
+  let paddingTop = undefined;
```
jzempel: Fix all padding calculations to be `(rowHeight - lineHeight) / 2`. They are not font-sized based.

packages/tables/src/styled/StyledTable.tsx
```diff
@@ -6,29 +6,37 @@
  */
```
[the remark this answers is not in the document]
austin94: Renamed.

packages/tables/src/styled/StyledTable.tsx
```diff
@@ -6,29 +6,37 @@
...
+})<IStyledTableProps>`
+  display: table;
+  border: none; /* [1] */
+  width: 100%; /* [1] */
+  table-layout: fixed; /* [1] */
+  border-collapse: collapse; /* [1] */
+  border-spacing: 0; /* [1] */
+  line-height: ${props => props.theme.space.base * 5}px;
+  font-size: ${props => props.theme.fontSizes.md};
+  direction: ${props => props.theme.rtl && 'rtl'};
```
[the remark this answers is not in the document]
austin94: I've themed this to match the `foreground` color.

packages/tables/src/styled/StyledRow.tsx
```diff
@@ -18,19 +19,86 @@ export interface IStyledRowProps {
...
 }
 
-export const StyledRow = styled.tr.attrs<IStyledRowProps>(props => ({
+export const StyledBaseRow = styled.tr<IStyledRowProps>`
+  display: table-row;
+  transition: background-color 0.1s ease-in-out;
+  border-bottom: ${props =>
+    `${props.theme.borders.sm} ${getColor('neutralHue', 200, props.theme)}`};
+  background-color: ${props => props.isStriped && getColor('neutralHue', 100, props.theme)};
+  height: ${props => props.theme.space.base * 10}px;
```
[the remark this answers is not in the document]
austin94: Correct. This was being overridden by the other usages.

packages/tables/src/styled/StyledSortableButton.tsx
```diff
@@ -6,40 +6,140 @@
...
+})<IStyledSortableButtonProps>`
+  position: relative;
+  transition: color 0.25s ease-in-out, opacity 0.25s ease-in-out;
+  border: none; /* [1] */
+  background-color: transparent; /* [1] */
+  cursor: pointer;
+  padding: 0; /* [1] */
+  /* stylelint-disable-next-line property-no-unknown */
+  padding-${props => (props.theme.rtl ? 'left' : 'right')}: ${props => props.theme.iconSizes.md};
+  width: ${props => props.width};
```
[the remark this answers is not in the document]
austin94: This allows backwards compatibility with the older flexbox API where it was a required prop. Even though it's not required now, it is common for consumers to apply a fixed width to a column and enable truncation.

packages/tables/src/styled/StyledSortableButton.tsx
```diff
@@ -6,40 +6,140 @@
...
+  type: 'button'
+})<IStyledSortableButtonProps>`
+  position: relative;
+  transition: color 0.25s ease-in-out, opacity 0.25s ease-in-out;
+  border: none; /* [1] */
+  background-color: transparent; /* [1] */
+  cursor: pointer;
+  padding: 0; /* [1] */
+  /* stylelint-disable-next-line property-no-unknown */
+  padding-${props => (props.theme.rtl ? 'left' : 'right')}: ${props => props.theme.iconSizes.md};
```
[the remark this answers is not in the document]
austin94: Updated.

packages/tables/src/styled/StyledSortableButton.tsx
```diff
@@ -6,40 +6,140 @@
...
+  right: ${props => (props.theme.rtl ? 'auto' : '0')};
+  left: ${props => props.theme.rtl && '0'};
+  align-items: center;
+  justify-content: center;
+  opacity: 0;
+  width: ${props => props.theme.fontSizes.md};
+  height: 100%;
+  color: inherit;
+  fill: inherit;
+`;
```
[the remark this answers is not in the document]
austin94: Now that inline SVGs are being used that timing of this animation is handled differently. I've included this animation again, but we still need to tweak some of the timing.

packages/tables/src/styled/StyledCell.ts
```diff
@@ -21,18 +20,68 @@ export interface IStyledCellProps {
...
 
-export const StyledCell = styled.td.attrs<IStyledCellProps>(props => ({
+const truncatedStyling = css`
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
+
+const sizeStyling = (props: IStyledCellProps & ThemeProps<DefaultTheme>) => {
+  let padding = `${props.theme.fontSizes.xs} ${props.theme.fontSizes.sm}`;
```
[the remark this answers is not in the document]
austin94: Is this expressed anywhere in the CSS components? I'm only seeing https://github.com/zendeskgarden/css-components/blob/master/packages/tables/src/_base.css#L20-L22

The right padding value is already expressed as `${props.theme.space.base * 3}px`

packages/tables/src/styled/StyledCell.ts
austin94: The `padding-right` logic is determined by a combination of `rtl` and `isMinimum` and should only override a single left-right value with the parent `padding` declaration. I'm not sure combining these values would be clearer.

packages/tables/src/styled/StyledHeaderCell.ts
austin94: Updated.

packages/tables/src/styled/StyledCell.ts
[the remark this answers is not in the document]
austin94: I've updated the `verticalPadding` value to be dynamic.

packages/tables/src/styled/StyledGroupRow.tsx
```diff
@@ -6,16 +6,28 @@
...
-  className: classNames(TableStyles['c-table__row'], TableStyles['c-table__row--group'])
+  'data-garden-version': PACKAGE_VERSION
 })`
+  background-color: ${props => getColor('neutralHue', 100, props.theme)};
+  height: ${props => props.theme.space.base * 8}px;
+  line-height: ${props => props.theme.space.base * 4}px;
+  font-size: ${props => props.theme.fontSizes.sm};
+
+  ${StyledCell} {
+    padding: 6px 10px;
```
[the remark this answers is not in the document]
austin94: I've abstracted this section to a `sizeStyles` utility and themed these values.

packages/tables/src/styled/StyledSortableButton.tsx
```diff
@@ -6,40 +6,140 @@
...
 const COMPONENT_ID = 'tables.sortable';
 
 type SORT = 'asc' | 'desc';
 
+const StyledBaseIconWrapper = styled.div`
+  display: flex;
+  position: absolute;
+  top: 0;
+  right: ${props => (props.theme.rtl ? 'auto' : '0')};
+  left: ${props => props.theme.rtl && '0'};
```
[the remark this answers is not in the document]
austin94: The `right` value is different in its `rtl` state than the `left` value. Would this work?

austin94:
> 1. Check your demo, the `<Well>` should be `isRecessed`.

Looks like the recent `notifications` TypeScript PR didn't include a fix for this example. I've updated this prop.

> 3. Also looks like there are a number of `.tsx` files in the `/styled` folder that can be renamed `.ts`.

Updated.

> 2. Plz add an "overflow button" column to the basic demo so that we can easily see the impact of small/medium/large sizing on this component.

Could I add a "size" dropdown to the `Overflow menu` example instead? Overflow menus include custom components, custom popper boundaries, popper offsets, and accessibility values which aren't always necessary.

packages/tables/src/styled/StyledGroupRow.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const COMPONENT_ID = 'tables.group_row';
+
+const sizeStyles = (props: IStyledRowProps & ThemeProps<DefaultTheme>) => {
+  const height = `${props.theme.space.base * 8}px`;
+  const lineHeight = getLineHeight(props);
+
+  return css`
+    height: ${height};
+    line-height: ${lineHeight};
```
hzhu: Out of curiosity, would we be able to use the newly added [`getLineHeight`](https://github.com/zendeskgarden/react-components/blob/next/packages/theming/src/utils/getLineHeight.ts#L18) utility here?

packages/tables/src/styled/StyledOverflowButton.ts
```diff
@@ -0,0 +1,102 @@
...
+export const StyledOverflowButton = styled.button.attrs<IStyledOverflowButtonProps>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  type: 'button'
+})<IStyledOverflowButtonProps>`
+  display: block;
+  transition: opacity 0.25s ease-in-out, background-color 0.1s ease-in-out;
+  opacity: ${props => (props.isHovered || props.isFocused || props.isActive ? '1' : '0')};
+  z-index: ${props => (props.isActive ? '1' : '0')};
+  margin-top: calc(${props => stripUnit(getRowHeight(props)) / 2} - 1em);
```
hzhu: Looks like `margin-top` is overwritten by the `margin-top` located in [`StyledHeaderRow`](https://github.com/zendeskgarden/react-components/pull/560/files#diff-9bd59fa289b5b864cfd28e9eb386056eR35). Could `margin-top` be removed here?

packages/tables/src/styled/StyledHeaderRow.ts
```diff
@@ -5,17 +5,42 @@
...
 })`
+  border-bottom-color: ${props => getColor('neutralHue', 300, props.theme)};
+  height: ${getHeaderRowHeight};
+  vertical-align: bottom;
+  font-weight: ${props => props.theme.fontWeights.semibold};
+
+  ${StyledOverflowButton} {
+    opacity: 1;
+    margin-top: 0;
+    margin-bottom: calc(${props => stripUnit(getHeaderRowHeight(props)) / 2} - 1em);
```
hzhu: It looks like `calc` requires calculated values to have a unit, otherwise, the browser complains that the CSS value is invalid.

Screen Shot 2020-01-02 at 10 54 54 AM

packages/tables/src/styled/StyledGroupRow.ts
austin94: This value is used for multiple styles. This allows the value to only be calculated once, instead of for `line-height` and the `padding` style below.

packages/tables/src/styled/StyledOverflowButton.ts
austin94: That overridden value is only for when the `OverflowButton` is used within a `HeaderRow`. This value is the default for other usages within the body.

packages/tables/src/styled/StyledHeaderRow.ts
austin94: Good catch! I've moved to the `math()` polished helper to retain the unit

packages/tables/src/styled/StyledOverflowButton.ts
```diff
@@ -0,0 +1,102 @@
...
+export interface IStyledOverflowButtonProps {
+  isHovered?: boolean;
+  isActive?: boolean;
+  isFocused?: boolean;
+  size: SIZE;
+}
+
+const getOverflowButtonSize = () => {
+  return '2em';
+};
```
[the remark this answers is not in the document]
austin94: Renamed!

packages/tables/src/styled/StyledGroupRow.ts
hzhu: Excellent, thanks for the clarification! I think these are some benefits of keeping the values separate, and supplying them to padding and line-height, respectively. 

1) The adoption of a standard a way to add line-height, via JZ's `getLineHeight` util. A developer would not have to choose between using the util function or calculating their own `line-height` with `space.base`. Also, line-heights can always be unitless in Garden.

2) `getLineHeight` is named the same as `getLineHeight` from `react-theming`. Not a big deal, but there is a chance that someone could confuse this with the util function from `react-theming` that is named the same. Particularly if they didn't follow through the code to identify where the import is coming from.

3) Another scenario might be that a new contributor sees `getLineHeight` in this file first, and think that this is the idiomatic way to set line-heights in Garden. But then in code review, they are told to use the `getLineHeight` util from `react-theming`, rather than calculating their own line-height via `space.base`.

packages/tables/src/styled/StyledCell.ts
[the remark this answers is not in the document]
jzempel: It is not expressed in the CSS components. This is the right time to fix that. Padding above should definitely not be connected to font-size.

packages/tables/src/styled/StyledGroupRow.ts
jzempel: @hzhu this is a semi-unique styling where `line-height` needs to be expressed as pixels in order to workaround Chrome rounding bugs. We wouldn't want to use the generalized (unitless) `getLineHeight` utility here.

packages/tables/src/styled/StyledTable.ts
```diff
@@ -0,0 +1,47 @@
...
+  size?: SIZE;
+}
+
+export const getLineHeight = (props: ThemeProps<DefaultTheme>) => {
+  return `${props.theme.space.base * 5}px`;
+};
+
+/**
+ * 1. <table> reset
+ * 2. <th> reset
```
jzempel: I don't see the cross-reference for `[2]` below.

packages/tables/src/styled/StyledSortableButton.tsx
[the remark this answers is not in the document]
jzempel: I think it should work because `auto` is a CSS default. Did you try it?

packages/tables/src/styled/StyledSortableButton.tsx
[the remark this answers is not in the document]
jzempel: When will you plan to implement these "tweaks"?

jzempel:
Per https://github.com/zendeskgarden/react-components/pull/560#issuecomment-570266758... Yes, adding a `size` dropdown to the overflow example would be helpful.

packages/tables/src/styled/StyledGroupRow.ts
hzhu: Thanks, @jzempel, for the extra clarification. 

I ran this branch locally with the `getLineHeight` utility and applied a unitless `line-height` to `StyledGroupRow` to explore the Chrome rounding bugs. There appear to be no rounding issues, and the computed `line-height` seems to remain at `20px` for both the element and its descendants. Also, there seem to be no visual issues. I could be missing something, though, would you be able to confirm?

packages/tables/src/styled/StyledOverflowButton.ts
hzhu: Upon a second review, I noticed that `margin-top` also generates an invalid CSS property. Could the `math()` polished helper be used to retain the unit for the first value of the `calc` function?

Screen Shot 2020-01-07 at 1 17 54 PM

packages/tables/src/styled/StyledCell.ts
```diff
@@ -21,18 +22,64 @@ export interface IStyledCellProps {
...
+    rightPadding = '0';
+    width = '1em';
+  }
+
+  return css`
+    padding: ${padding};
+    /* stylelint-disable declaration-block-no-redundant-longhand-properties */
+    padding-top: ${verticalPadding};
+    padding-right: ${rightPadding};
+    padding-bottom: ${verticalPadding};
```
hzhu: The design mocks indicate that the top and bottom spacing should each be `10px`. I notice that on the "Overflow menu" example, the bottom spacing on each row appears to be larger than the top, and more than `10px`:

Screen Shot 2020-01-07 at 3 16 07 PM

packages/tables/src/styled/StyledOverflowButton.ts
austin94: I had a few of these `calc()` fixes stashed and forgot to add back in. I've added a `math()` call to address this.

packages/tables/src/styled/StyledCell.ts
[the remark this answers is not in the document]
austin94: I've updated `padding` to not use fontSize values

packages/tables/src/styled/StyledSortableButton.tsx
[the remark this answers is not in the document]
austin94: I think I had an invalid `calc` call which was messing up some unrelated styling here. I've updated this to your suggested syntax.

packages/tables/src/styled/StyledTable.ts
austin94: Removed.

packages/tables/src/styled/StyledCell.ts
austin94: When refactoring the `verticalPadding` and `padding` values I didn't handle the `hasOverflow` prop correctly. This is now corrected.

packages/tables/src/styled/StyledCell.ts
jzempel: Note that the padding is currently incorrect for RTL. There should be a left pad of 10px for non-RTL and a right pad of 10px for RTL. Once you correct the logic, it will make more sense to name this variable `horizontalPadding`. See https://austin-next.netlify.com/tables/?#selectable for the issue.

austin94: Updated to `horizontalPadding`

packages/tables/src/styled/StyledSortableButton.tsx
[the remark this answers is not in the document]
austin94: From our offline conversation, this is now removed.

jzempel:
@austin94 did you procure a designer review on this one?

PR #563, opened 2019-12-18
feat(notifications): Migrate to CSS-in-JS

hzhu:
##  Description

This pull request migrates `react-notifications` to CSS-in-JS and removes the dependencies on stylesheets backed by [@zendeskgarden/css-callouts](https://www.npmjs.com/package/@zendeskgarden/css-callouts).

## Detail

Styling has moved to a CSS-in-JS approach, and as a result, the components should look and behave as before. The API remains unchanged for consumers. 

**The demo for this pull request can be found [here](https://henry-next.netlify.com/notifications/) (Up to date as of 01/15/2020)**. 

## Checklist

      designer as a reviewer)
      component
- [x] :memo: tested in Chrome, Firefox, Safari, and Edge.


packages/notifications/src/styled/StyledAlert.spec.tsx
```diff
@@ -6,39 +6,293 @@
...
+      const { container } = render(<StyledAlert type="success" />);
+      const { colors, palette } = DEFAULT_THEME;
+      const { successHue } = colors;
+
+      // Close style
+      expect(container.firstChild).toHaveStyleRule('color', palette[successHue][500], {
+        modifier: css`
+          ${StyledClose}
+        ` as any
+      });
```
hzhu: I wrote a few tests which assert that the styling for `<Alert>` and `<Notification>` descendants are rendered correctly. I think these tests provide good value because the styling is heavily guarded by logic. These tests are likely more valuable when the project gains more contributors.

packages/notifications/src/styled/StyledIcon.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it('should render with the correct styling for LTR writing systems', () => {
+    const { container } = render(<StyledIcon type="info" />);
+    const { space } = DEFAULT_THEME;
+
+    expect(container.firstChild).toHaveStyleRule('left', `${space.base * 6}px`);
+    expect(container.firstChild).not.toHaveStyleRule('right');
+  });
+});
```
hzhu: Unfortunately, I wasn't able to come up with tests that verify that the component renders the correct SVG icon depending on the `type`. I will continue to think about a test for verifying the correct SVG icon.

packages/notifications/src/styled/StyledAlert.ts
```diff
@@ -5,37 +5,157 @@
...
+  const paddingLeft = rtl ? space.xl : `${space.base * 12.5}px`;
+
+  return `${paddingTop} ${paddingRight} ${paddingBottom} ${paddingLeft}`;
+};
+
 /**
  * Supports all `<div>` props
  */
-export const StyledAlert = styled.div.attrs<IStyledAlertProps>(props => ({
+export const StyledAlert = styled.div.attrs({
```
hzhu: > I had really hoped to see ideas around a generic StyledDiv fleshed out 

Please don’t lose hope yet 🙂. I’m all ears and would love to discuss and flesh out ideas around a generic `StyledDiv` now. We had briefly talked about this two days ago but I wasn’t 100% clear on what the generic `StyledDiv` meant and should have asked about an example.

I’ve continued to think about it. There seem to be five style properties, `position`, `border`, `border-radius`, `line-height`, and `font-size`, that are shared across the `<Alert>`, `<Notification>` and `<Well>` components which can be abstracted out into a [`<StyledDiv>`](https://github.com/zendeskgarden/react-components/blob/hzhu/notifications-css-in-js-ii/packages/notifications/src/styled/StyledDiv.ts). Then, the `<StyledDiv>` can be used to implement each of the notifications, like [this](https://github.com/zendeskgarden/react-components/blob/hzhu/notifications-css-in-js-ii/packages/notifications/src/styled/StyledAlert.ts#L142), [this](https://github.com/zendeskgarden/react-components/blob/hzhu/notifications-css-in-js-ii/packages/notifications/src/styled/StyledNotification.ts#L101), and finally [this](https://github.com/zendeskgarden/react-components/blob/hzhu/notifications-css-in-js-ii/packages/notifications/src/styled/StyledWell.ts#L33). Does this align with what you had in mind?

packages/notifications/src/styled/content/StyledClose.tsx
```diff
@@ -5,30 +5,51 @@
...
-    [CalloutStyles['is-hovered']]: props.isHovered
-  })
-}))<IStyledCloseProps>`
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledCloseProps>`
+  display: block;
+  position: absolute;
+  top: ${props => props.theme.space.xxs};
+  right: ${props => !props.theme.rtl && props.theme.space.xxs};
+  left: ${props => props.theme.rtl && props.theme.space.xxs};
```
hzhu: > For starters, please go through the entire PR and replace all `space.[dimension]` sizing with the corresponding `space.base` calculations. We want to pack a lot of leverage into a user's ability to modify the single `space.base` theme variable. The `xs`, `sm`, etc. sizing dimensions are generally available for convenience and used solely for `grid`.

It seems the `space` sizing is [internally computed](https://github.com/zendeskgarden/react-components/blob/next/packages/theming/src/elements/theme/index.ts#L133) using `space.base`, so these values may not need to be changed. There are two theme values that are used in this PR that isn’t computed internally with `space.base`: 

1) `border`
2) `fontSizes` 

I can replace these values with `space.base` calculations. Is there a place that documents which properties are computed using `space.base`? For example, as the end-user, if I modify the theme's `base` value, what parts of the design system will change its size?

packages/notifications/src/styled/StyledAlert.ts
```diff
@@ -5,37 +5,157 @@
...
+      break;
+    case 'warning':
+      color = palette[warningHue][700];
+      pseudoColor = palette[warningHue][800];
+      bgColor = getColor(warningHue, 700, theme, 0.15);
+      break;
+    case 'info':
+      color = palette[neutralHue][600];
+      pseudoColor = palette[neutralHue][800];
+      bgColor = getColor(neutralHue, 600, theme, 0.15);
```
hzhu: > Furthermore, I ran the notion of a flexible color hue application by design and got resoundingly positive feedback.

We spoke about the notion of a flexible color hue application a few days ago. I walked away not fully understanding, and that was my fault — I should have asked for an example. What’s an example of flexible color hue application? And what is an example of in-flexible color hue application? Could we highlight the in-flexible color hue application in this PR?

packages/notifications/src/styled/content/StyledClose.tsx
hzhu: Update: after speaking with @jzempel, I learned that the computed values happen before the user applies a custom theme. I'll update `theme.space` usages in this PR to be computed with `space.base` 👍 .

packages/notifications/src/styled/StyledAlert.ts
austin94: I'm not familiar with any non-validation background colors used in product. Why would we need to start supporting this now?

hzhu: Marking as resolved because this is closely related to the [thread ](https://github.com/zendeskgarden/react-components/pull/563#discussion_r359993043)about a flexible `StyledDiv` component.

packages/notifications/src/styled/StyledAlert.ts
hzhu: Spoke with @jzempel offline about what a `<StyledDiv>` might look like. I'm in the process of getting rid of `<StyledAlert>`, `<StyledNotification>`, and `<StyledWell`> in favor of a `<StyledDiv>`. I will push up a new commit for this once it's fleshed out. Stay tuned.

packages/notifications/src/styled/StyledWell.spec.tsx
```diff
@@ -6,31 +6,71 @@
...
+    expect(container.firstChild).toHaveStyleRule('border-radius', borderRadii.md);
+    expect(container.firstChild).toHaveStyleRule('background-color', `${colors.background}`);
+    expect(container.firstChild).toHaveStyleRule('padding', `${space.md} ${space.xl}`);
+    expect(container.firstChild).toHaveStyleRule(
+      'line-height',
+      ((space.base * 5) / stripUnit(fontSizes.md)).toFixed(5)
+    );
+    expect(container.firstChild).toHaveStyleRule('color', palette.grey[600]);
+    expect(container.firstChild).toHaveStyleRule('font-size', fontSizes.md);
+    expect(container.firstChild).not.toHaveStyleRule('box-shadow');
```
hzhu: Large blocks of style assertions are similar to snapshot testing. They are brittle because they change when the designs change. I'll go ahead and remove this, and only assert against syles that are guarded by logic.

packages/notifications/src/styled/StyledAlert.spec.tsx
jzempel: imo, it's smart to keep some basic testing in isolate around the styled (view) components.

packages/notifications/src/elements/Well.tsx
```diff
@@ -7,14 +7,14 @@
...
-import { StyledWell, IStyledWellProps } from '../styled';
+import { StyledWell, IStyledBaseProps } from '../styled';
 
 /**
  * Supports all `<div>` props
  */
 export const Well = React.forwardRef<
   HTMLDivElement,
-  IStyledWellProps & HTMLAttributes<HTMLDivElement>
+  IStyledBaseProps & HTMLAttributes<HTMLDivElement>
```
jzempel: `Well` should not be exposing a `hue` prop in the element API.

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -5,36 +5,84 @@
...
+
+  return css`
+    ${StyledClose} {
+      color: ${palette[neutralHue][600]};
+    }
+
+    ${StyledClose}:focus {
+      outline: none;
+      background-color: ${getColor(neutralHue, 600, theme, 0.15)};
+      color: ${palette[neutralHue][800]};
```
jzempel: These should not be applied to the `:focus` pseudo. Leverage `[data-garden-focusvisible]` instead.

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -5,36 +5,84 @@
...
+
+    ${StyledClose}:hover {
+      color: ${palette[neutralHue][800]};
+    }
+  `;
+};
+
+const colorStyles = (props: IStyledNotificationProps & ThemeProps<DefaultTheme>) => {
+  const { type, theme } = props;
+  const { colors, palette } = theme;
```
jzempel: Never reference the theme `palette` when you can use `getColor` instead.

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -5,36 +5,84 @@
...
 
 const COMPONENT_ID = 'notifications.notification';
 
 export interface IStyledNotificationProps {
   type?: VALIDATION_TYPE;
 }
 
+const closeStyles = (props: ThemeProps<DefaultTheme>) => {
+  const { theme } = props;
+  const { colors, palette } = theme;
```
jzempel: Never reference the theme `palette` when you can use `getColor` instead.

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -5,36 +5,84 @@
...
+import { StyledClose } from './content/StyledClose';
+import { StyledBase } from './StyledBase';
 
 const COMPONENT_ID = 'notifications.notification';
 
 export interface IStyledNotificationProps {
   type?: VALIDATION_TYPE;
 }
 
+const closeStyles = (props: ThemeProps<DefaultTheme>) => {
```
jzempel: Feels like these styles should be applied directly on `StyledClose` – leveraging the `hue` prop for all designed combos. If there are only opacity variations, let's get a designer involved to see if there's an easy win for normalizing close button color styling across the board.

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -5,36 +5,84 @@
...
+      color = palette[successHue][600];
+      break;
+    case 'error':
+      color = palette[dangerHue][600];
+      break;
+    case 'warning':
+      color = palette[warningHue][700];
+      break;
+    case 'info':
+      color = palette[neutralHue][800];
```
jzempel: Should be `props.theme.colors.foreground`.

packages/notifications/src/styled/StyledIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { getColor } from '@zendeskgarden/react-theming';
+
+export const StyledIcon = styled(({ children, ...props }) =>
+  React.cloneElement(Children.only(children), props)
+)`
+  position: absolute;
+  right: ${props => props.theme.rtl && `${props.theme.space.base * 6}px`};
+  left: ${props => !props.theme.rtl && `${props.theme.space.base * 6}px`};
+  margin-top: ${props => props.theme.space.base / 2}px;
+  color: ${props => props.hue && getColor(props.hue, 600, props.theme)};
```
jzempel: What purpose is this `color` property serving?

packages/notifications/src/styled/StyledAlert.ts
```diff
@@ -5,37 +5,79 @@
...
+  }
+
+  return css`
+    ${StyledClose} {
+      color: ${color};
+    }
+
+    ${StyledClose}:focus {
+      outline: none;
+      background-color: ${bgColor};
```
jzempel: `background-color` should be styled using the focusvisible technique.

packages/notifications/src/styled/StyledAlert.ts
```diff
@@ -5,37 +5,79 @@
...
+    }
+
+    ${StyledClose}:focus, ${StyledClose}:hover {
```
jzempel: [nit] please follow 1) `:hover`, 2) `:focus`, 3) `:active` psuedo ordering.

packages/notifications/src/styled/content/StyledClose.tsx
```diff
@@ -5,30 +5,51 @@
...
-    [CalloutStyles['is-hovered']]: props.isHovered
-  })
-}))<IStyledCloseProps>`
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledCloseProps>`
+  display: block;
+  position: absolute;
+  top: ${props => props.theme.space.base}px;
+  right: ${props => !props.theme.rtl && `${props.theme.space.base}px`};
+  left: ${props => props.theme.rtl && `${props.theme.space.base}px`};
```
jzempel: Replace with 

```${props => props.theme.rtl ? 'left' : 'right'}: ${props => `${props.theme.space.base}px`}```

for improved readability.

File: packages/notifications/src/styled/content/StyledClose.tsx [hunk 17]
```diff
@@ -5,30 +5,51 @@
...
+  transition: background-color 0.1s ease-in-out;
+  border: none; /* [1] */
+  border-radius: 50%;
+  background-color: transparent; /* [1] */
+  cursor: pointer;
+  padding: 0;
```
jzempel: Why is this `padding` reset needed?

packages/notifications/src/styled/content/StyledClose.tsx
```diff
@@ -5,30 +5,51 @@
...
+  width: ${props => props.theme.space.base * 7}px;
+  height: ${props => props.theme.space.base * 7}px;
+  overflow: hidden;
+  font-size: 0; /* [1] */
```
jzempel: Shouldn't be needed as this `font-size` is a reset for `<a>` tags.

packages/notifications/src/elements/Notification.tsx
```diff
@@ -8,15 +8,46 @@
...
+  warning: 'warningHue',
+  info: 'neutralHue'
+};
+
+const validationIcons = {
+  success: CheckCircleStrokeIcon,
+  error: AlertErrorStrokeIcon,
+  warning: AlertWarningStrokeIcon,
+  info: InfoStrokeIcon
+};
```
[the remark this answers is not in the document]
hzhu: That sounds good. The icons are now located in a separate file called `util/icons.ts` and the validation hues are located in `util/types.ts`.

packages/notifications/src/styled/content/StyledClose.tsx
hzhu: Looks like there are user agent stylesheet that adds padding:

Screen Shot 2020-01-08 at 2 45 36 PM

Should the padding be removed? Or should a `/* [1] */` be added to indicate padding reset?

packages/notifications/src/elements/content/Close.tsx
```diff
@@ -6,18 +6,11 @@
...
-  IStyledCloseProps & HTMLAttributes<HTMLButtonElement>
->((props, ref) => <StyledClose ref={ref} {...props} />);
-
-Close.propTypes = {
-  isFocused: PropTypes.bool,
-  isHovered: PropTypes.bool
-};
+export const Close = React.forwardRef<HTMLButtonElement, HTMLAttributes<HTMLButtonElement>>(
+  (props, ref) => <StyledClose ref={ref} {...props} />
+);
```
hzhu: > We should be using context to get the right type information to the Close component rather than expecting a consumer to match a Close type with its containing notification element.

@jzempel, I'm trying to understand the problems around `Close` so I can make the appropriate changes. How are consumers matching a `Close` type with its containing notification element? Does "consumers" refer to end consumers that use the `react-notifications` package like this? When you refer to "type information", is that one of "success", "error", "warning", or "info"?

```jsx
<Alert type="success">
  <Title>Success Alert</Title>
  Veggies es bonus vobis, proinde vos postulo essum magis kohlrabi welsh onion daikon.
  <Close onClick={() => alert('closing alert')} aria-label="Close Alert" />
</Alert>
```

packages/notifications/src/styled/StyledIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useContext } from 'react';
+import styled, { ThemeContext } from 'styled-components';
+import InfoStrokeIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
+import AlertErrorStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
+import CheckCircleStrokeIcon from '@zendeskgarden/svg-icons/src/16/check-circle-stroke.svg';
+import AlertWarningStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-warning-stroke.svg';
```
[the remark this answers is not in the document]
hzhu: Investigation indicates that these icons are not the culprit. Removal of importing these icons shave about `0.2kB` off the bundle size, from `6.14kB` down to `5.92kB`.

It appears that the culprit is importing `rgba` from the `polished` library in `src/styled/StyledBase.ts`. Removal of this import shaves about `2.95kB` off the bundle size, from `6.14kB` down to `3.18kB`. 

I incorrectly added `rgba` in `StyledBase.ts`. I removed the `rgba` usage, and the bundle size has dropped to `3.18kB`. Also, I have updated the bundle size limit of the notifications package to `3.5kB`.

packages/notifications/src/styled/StyledIcon.tsx
hzhu: This `color` property is styling the icon for `<Notification>`. A hue is calculated using the type, e.g. `success`, `error`, etc from `<Notification>`, and is then passed down as a prop to `<StyledIcon>`.

There isn't a `hue` prop passed to `<StyledIcon>` from `<Alert>`, so the Icon in `<Alert>` gets its color inherited from its parent (which determines the font color based on the type).

packages/notifications/src/styled/StyledNotification.ts
hzhu: Could you help paint a picture of how that might look like? Since `Close` is a child provided by the end consumer, the only way to receive a hue in `StyledClose` would be through context. Is that right? Also, in addition to hue, `StyledClose` would need to know which component (`Alert` or `Notification`) it was rendered in so that it can apply appropriate styles. How would `StyledClose` know what component it is a descendant of?

packages/notifications/src/styled/content/StyledClose.tsx
jzempel: My bad, the removal of this affects the alignment of the X within the button. It needs to be restored.

packages/notifications/src/elements/content/Close.tsx
jzempel: I'm not exactly sure this helps, but this is the way I think about the problem. The table shows that I can gather all the context needed at the Alert/Notification/Well element level to render whatever the correct treatment should be for each of the Base/Title/Close children. This means the `Styled` components don't need to establish a bunch of CSS cross-references. They simply render whatever hue they are given by the Title/Close wrapping elements.

| Element | type | StyledBase hue | StyledTitle hue | StyledClose hue |
| -------- | ----- | --------------- | --------------- | ---------------- |
| Alert | Success | green | green | green |
| Alert | Error | red | red | red |
| Alert | Warning | yellow | yellow | yellow |
| Alert | Info | grey | grey | grey |
| Notification | Success | `undefined` | green | grey |
| Notification | Error | `undefined` | red | grey |
| Notification | Warning | `undefined` | yellow | grey |
| Notification | Info | `undefined` | grey | grey |
| Well | `isRecessed === false` | `undefined` | grey | grey |
| Well | `isRecessed === true` | grey | grey | grey |

In other words, you shouldn't need to have `${StyledClose} { ... }` CSS scattered all over the place. It can all just live within the `StyledClose` component.

jzempel: Important that we would want to store the hue names (i.e. `successHue`, `errorHue`, `warningHue`, `neutralHue`) in context rather than the colors.

packages/notifications/src/styled/content/StyledParagraph.tsx
```diff
@@ -17,8 +15,8 @@ const COMPONENT_ID = 'notifications.paragraph';
  */
 export const StyledParagraph = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(CalloutStyles['c-callout__paragraph'])
+  'data-garden-version': PACKAGE_VERSION
 })`
+  margin: ${props => props.theme.space.base * 2.5}px 0 0;
```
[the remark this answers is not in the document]
allisonacs: @jzempel @hzhu yea, this is my bad. The space between the header and the paragraph should be 8px, not 10. That's my bad. I'm opening a bug branch now.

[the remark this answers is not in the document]
hzhu: Great, thanks @allisonacs I'll update the code to reflect that 👍.

packages/notifications/examples/basic.md
```diff
@@ -129,8 +129,8 @@ const initialState = {
   <Row>
     <Col md={4}>
       <Well isRecessed>
-        <div role="group" aria-label="alert types radio">
-          <StyledMd>Alert Types</StyledMd>
+        <div role="group" aria-label="notification types radio">
+          <StyledMd>Notification Types</StyledMd>
```
jzempel: Update this example to with a `Default` radio button (selected by default) which sets the notification type to `undefined`.

packages/notifications/src/elements/Alert.tsx
```diff
@@ -7,16 +7,37 @@
 
 import React, { HTMLAttributes } from 'react';
 import PropTypes from 'prop-types';
-import { StyledAlert, IStyledAlertProps } from '../styled';
-import { ARRAY_VALIDATION_TYPE } from '../utils/types';
+import InfoStrokeIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
+import { StyledAlert } from '../styled';
+import { StyledIcon } from '../styled/StyledIcon';
```
austin94: Should this import be available through `styled/index.ts`?

packages/notifications/src/styled/StyledWell.ts
```diff
@@ -5,32 +5,30 @@
...
-export const StyledWell = styled.div.attrs<IStyledWellProps>(props => ({
+export const StyledWell = styled(StyledBase).attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(CalloutStyles['c-callout'], {
-    // RTL
-    [CalloutStyles['is-rtl']]: isRtl(props),
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledWellProps>(props => {
+  const { theme } = props;
```
austin94: We don't desctructure theme/space like this in other components. Could we align this with our other usages?

packages/notifications/src/elements/Notification.tsx
```diff
@@ -7,16 +7,31 @@
...
->((props, ref) => <StyledNotification ref={ref} {...props} />);
+>((props, ref) => {
+  const Icon = props.type ? validationIcons[props.type] : InfoStrokeIcon;
+  const hue = props.type && validationHues[props.type];
+
+  return (
+    <StyledNotification ref={ref} isFloating {...props}>
+      <StyledIcon hue={hue}>
+        <Icon />
+      </StyledIcon>
```
jzempel: Refer to designs; an icon is not a mandatory element for a Notification. When `type === undefined`, there is no icon.

packages/notifications/src/styled/content/StyledClose.tsx
```diff
@@ -6,29 +6,61 @@
...
+  background-color: transparent; /* [1] */
+  cursor: pointer;
+  padding: 0;
+  width: ${props => props.theme.space.base * 7}px;
+  height: ${props => props.theme.space.base * 7}px;
+  overflow: hidden;
+  color: ${props =>
+    props.hue
+      ? getColor(props.hue, props.hue === 'warningHue' ? 700 : 500, props.theme)
+      : getColor('neutralHue', 600, props.theme)};
```
jzempel: Color for the close icon is the `600` shade (unless `'warningHue'` which is 700). I think the CSS got it wrong compared to the current designs.

packages/notifications/src/styled/content/StyledParagraph.tsx
```diff
@@ -6,19 +6,17 @@
...
 import styled from 'styled-components';
-import classNames from 'classnames';
-import CalloutStyles from '@zendeskgarden/css-callouts';
 import { retrieveComponentStyles } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'notifications.paragraph';
 
 /**
- * Used for multi-line Notification content. Supports all `<div>` props
+ * Used for multi-line Notification content. Supports all `<p>` props
```
jzempel: Please update the `Paragraph` element JSDoc accordingly (div -> p)

packages/notifications/src/styled/StyledWell.ts
jzempel: After that ☝️ this can all just become a straight template string without any `css` return.

jzempel:
When the next review is ready, please be sure that the demo page allows us to check `:focus` states. Thanks.

packages/notifications/src/utils/icons.ts
```diff
@@ -0,0 +1,26 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import AlertErrorStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
+import CheckCircleStrokeIcon from '@zendeskgarden/svg-icons/src/16/check-circle-stroke.svg';
+import AlertWarningStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-warning-stroke.svg';
+import InfoStrokeIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
+import { VALIDATION_TYPE } from './types';
+
+export const validationIcons = {
```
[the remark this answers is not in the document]
hzhu: Yes, I think that would be valuable. Thanks for the suggestion!

packages/notifications/src/elements/Alert.tsx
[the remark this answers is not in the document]
hzhu: No particular reason and might have been an oversight on my part. Let me to export the component from the `index.ts`.

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -0,0 +1,68 @@
...
+  } else {
+    backgroundColor = props.theme.colors.background;
+    borderColor = getColor('neutralHue', 300, props.theme);
+    foregroundColor = getColor('neutralHue', 700, props.theme);
+  }
+
+  return css`
+    border-color: ${borderColor};
+    background-color: ${backgroundColor};
+    color: ${foregroundColor};
```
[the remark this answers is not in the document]
hzhu: Ah yes, the `color` isn't reflective of the designs. Factoring `type` works great.

packages/notifications/src/styled/StyledWell.ts
hzhu: Yeah aligning with other usages sound good. Fixed 👍 .

packages/notifications/src/styled/content/StyledClose.tsx
hzhu: Yeah that's correct. My mistake! This should be fixed with the latest commit.

packages/notifications/src/styled/content/StyledParagraph.tsx
hzhu: Updated, thanks 👍!

packages/notifications/src/elements/Alert.tsx
```diff
@@ -7,17 +7,36 @@
...
 import PropTypes from 'prop-types';
-import { StyledAlert, IStyledAlertProps } from '../styled';
-import { ARRAY_VALIDATION_TYPE } from '../utils/types';
+import { StyledAlert, StyledIcon } from '../styled';
+import { validationIcons, validationHues } from '../utils/icons';
+import { ARRAY_VALIDATION_TYPE, VALIDATION_TYPE, VALIDATION_HUE } from '../utils/types';
+import { NotificationsContext } from '../utils/useNotificationsContext';
+
+export interface IAlertProps {
+  type: VALIDATION_TYPE;
```
jzempel: The API lists `type` as `VALIDATION_TYPE`. The type strings need to be listed instead: https://henry-next.netlify.com/notifications/#alert. No one will know what `VALIDATION_TYPE` means.

packages/notifications/src/elements/Notification.tsx
```diff
@@ -7,16 +7,34 @@
 
 import React, { HTMLAttributes } from 'react';
 import PropTypes from 'prop-types';
+import InfoStrokeIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
 import { StyledNotification, IStyledNotificationProps } from '../styled';
+import { StyledIcon } from '../styled/StyledIcon';
```
jzempel: Combine with previous as `import { StyledIcon } from '../styled';`

packages/notifications/src/elements/Notification.tsx
```diff
@@ -7,16 +7,34 @@
...
+  const Icon = props.type ? validationIcons[props.type] : InfoStrokeIcon;
+  const hue = props.type && validationHues[props.type];
+
+  return (
+    <StyledNotification ref={ref} type={props.type} isFloating {...props}>
+      {props.type ? (
+        <StyledIcon hue={hue}>
+          <Icon />
+        </StyledIcon>
+      ) : null}
```
jzempel: cleaner as `{props.type && <StyledIcon ...`

packages/notifications/src/styled/content/StyledTitle.tsx
```diff
@@ -6,19 +6,24 @@
...
 import { retrieveComponentStyles } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'notifications.title';
 
 /**
- * Used for Notification titles. Supports all `<div>` props
+ * Used for Notification titles. Supports all `<div>` props.
+ * For improved semantics, pass an "h1" or "strong" to the `as` prop.
+ * Additional styles can be applied to the `as` element to override
+ * user agent stylesheets.
```
jzempel: Move this documentation to `/elements/Title` so that it shows in the public API prop sheet.

packages/notifications/src/elements/Alert.tsx
```diff
@@ -7,17 +7,36 @@
...
-import { StyledAlert, IStyledAlertProps } from '../styled';
-import { ARRAY_VALIDATION_TYPE } from '../utils/types';
+import { StyledAlert, StyledIcon } from '../styled';
+import { validationIcons, validationHues } from '../utils/icons';
+import { ARRAY_VALIDATION_TYPE, VALIDATION_HUE } from '../utils/types';
+import { NotificationsContext } from '../utils/useNotificationsContext';
+
+export interface IAlertProps {
+  type: 'success' | 'warning' | 'error' | 'info';
+}
```
hzhu: Props should have a helpful description comment, particularly because it gets generated into the documentation prop sheet.

PR #569, opened 2019-12-19
chore(build): allow preMajors from next branch

austin94:
## Description

This PR enables the ability to release `preMajor` versions in the `next` branch. It will force all packages to be deployed under the `next` tag in NPM.


austin94:
Also, this removes the CHANGELOG logic while we are in the `next` branch. No reason to muddy the flow there is "nightly" type messages.

PR #573, opened 2020-01-06
chore: use `getLineHeight` utility

jzempel:
- [ ] **BREAKING CHANGE?** 

## Description

- swap out various `line-height` calculations with `getLineHeight` from theming
- fix button heights at various zoom levels

## Detail

@ginnywood the components that got touched by this PR include:
- breadcrumb
- portions of chrome
- loaders
- tags
which are all 1:1 swaps. But the most important for visual review would be the fixes for buttons, which can be seen here: https://silly-haibt-18bd35.netlify.com/buttons/

## Checklist

      designer as a reviewer)
      component~

### DIFF
--- a/packages/buttons/src/styled/StyledButton.ts
+++ b/packages/buttons/src/styled/StyledButton.ts
@@ -32,14 +32,14 @@ const getBorderRadius = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) =
   return props.theme.borderRadii.md;
 };
 
-export const getLineHeight = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
+export const getHeight = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
   if (props.size === SIZE.SMALL) {
-    return props.theme.space.base * 8;
+    return `${props.theme.space.base * 8}px`;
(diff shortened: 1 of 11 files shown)


packages/buttons/src/styled/StyledButton.ts
```diff
@@ -224,7 +222,8 @@ const sizeStyles = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
 
     retVal = css`
       padding: 0 ${em(math(`${padding} - ${props.theme.borderWidths.sm}`), fontSize)};
-      line-height: ${math(`${lineHeight} - (${props.theme.borderWidths.sm} * 2)`)};
+      height: ${height};
+      line-height: ${lineHeight};
```
hzhu: Out of curiosity, since button text doesn't wrap in Garden (or do they?), what are the consequences of allowing the browser to handle line-height for buttons? Alternatively, what are the consequences of setting the `line-height` value to `1`?

I tried this out in the `basic.md` example, and it (visually) looks the same.

jzempel: This could be legacy. As I remember, the use of `line-height` ensures a consistent result across elements (i.e. `a` or `input type="button"`), browsers, and fonts (we used to use custom).

PR #575, opened 2020-01-09
chore(lint): create styled-component default theme eslint rule

austin94:
## Description

This PR introduces a local ESLINT rule that requires all `styled` usages to have a `DEFAULT_THEME` provided via `defaultProps`.

## Detail

It turns out adding a local ESLINT rule is actually pretty crazy. I had to:

- Create a local, private package within our `utils` package
- Add it to our `node_modules` path via a relative path in the `package.json`
- Enable it like a standard rule under the `garden-local` rule naming

The `garden-local/require-default-theme` rule ensures that all styled `TaggedTemplateExpression`s include a `defaultProp` of `theme`. This is able to work across both TS and JS files.

I tried to get the "suggested fix" syntax working within VS Code, but it doesn't seem to be working locally for me.

## Example Errors

Screen Shot 2020-01-09 at 9 28 50 AM

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -49,7 +49,7 @@
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/interface-name-prefix": ["error", "always"],
     "@typescript-eslint/no-non-null-assertion": "off",
-    "garden-local/require-default-theme": "warn"
+    "garden-local/require-default-theme": "error"
   },
   "overrides": [
     {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,7 @@
     "PACKAGE_VERSION": true,
     "COMPONENT_IDS": true
   },
-  "plugins": ["prettier", "react", "jest", "jsx-a11y", "react-hooks", "@typescript-eslint"],
+  "plugins": ["prettier", "react", "jest", "jsx-a11y", "react-hooks", "@typescript-eslint", "garden-local"],
   "env": {
     "es6": true,
     "browser": true,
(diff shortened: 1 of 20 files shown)


utils/eslint/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "private": true,
+  "name": "eslint-plugin-garden-local",
+  "description": "Local eslint ruless",
```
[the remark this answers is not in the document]
austin94: Fixed!

austin94:
@hzhu haha definitely needs some more comments. I've included this reasoning in the ESLint rule

```
/**
 * The `require-default-theme` rule requires that all styled-component usages include
 * a default `theme` value.
 *
 * This ensures that all components are able to render without a parent `ThemeProvider`.
 */
```

Without this rule simple component usages, like `<Button>`, would error if no parent `ThemeProvider` was provided.

hzhu:
Ah yes, that makes sense. Thank you @austin94!

.eslintrc.json
```diff
@@ -48,13 +48,15 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/interface-name-prefix": ["error", "always"],
-    "@typescript-eslint/no-non-null-assertion": "off"
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "garden-local/require-default-theme": "warn"
```
jzempel: Why not `"error"`?

austin94: Because I'm dumb :) Moved to error

PR #594, opened 2020-01-17
feat(forms): migrate react-forms to TS

austin94:
## Description

This PR migrates `react-forms` to TS. Overall, it's mostly just migrating directly from JS to TS without any prop or API changes.

Take a look at https://austin-next.netlify.com/forms/ to view the updated prop-sheets

I've also brought in the Tooltip performance fix from #593 since it's a one-liner

## Checklist

      designer as a reviewer)~
      component~


packages/forms/src/styled/text/StyledTextInput.ts
```diff
@@ -185,11 +184,18 @@ const sizeStyles = props => {
   `;
 };
 
-export const StyledTextInput = styled.input.attrs(props => ({
+export interface IStyledTextInputProps {
```
jzempel: Should `IStyledTextInputProps` extend `ThemeProps<DefaultTheme>`?

austin94: `ThemeProps` is inferred by the styled-components usage so we don't need to manually include it here.

jzempel: Right, but it needs to be repeatedly and-ed in for the function `props` usage.

austin94: Correct, but that because those helpers are being used within a `styled-component`. The interfaces are only for the API exposed by the components. This lets us reuse these interfaces across the high-abstaction elements without muddying the prop sheets.

Searching for `& ThemeProps` shows us using this pattern across most of the packages. If we want to readdress how we handle these CSS helpers we might want to do that in a separate PR.

packages/forms/src/elements/common/Field.tsx
```diff
@@ -5,18 +5,23 @@
...
+}
+
 /**
  * Provides layout and ID context for child components; accepts all `<div>`
  * attributes and events.
  */
-const Field = React.forwardRef((props, ref) => {
-  const getMessageProps = messageProps => ({ role: 'alert', ...messageProps });
+export const Field = React.forwardRef<HTMLDivElement, IFieldProps>((props, ref) => {
+  const getMessageProps = (messageProps: any) => ({ role: 'alert', ...messageProps });
```
hzhu: Should `messageProps` be typed?

austin94: Since this is defined inline it's much more difficult to type it with generics. With this only being used internally I didn't want to rewrite the logic to make it fit our prop-getter structure more easily. If it's something you think we need I can dig into it some more.

hzhu: Cool, thanks for the clarification! I was curious, but it's not something I think we need.

PR #596, opened 2020-01-17
fix(demo): remove global.Row `alignItems` attrs override.

jzempel:
- [ ] **BREAKING CHANGE?** 

## Description

With updated `styled-components` the `global.Row` attrs override is hostile as `alignItems` cannot be modified via props.

## Detail

While updating the tooltips demo, I discovered that the `text-align` styling on a parent can have unintended consequences for tooltip styling. A CSS property was added to correct that issue.

## Checklist

      designer as a reviewer)~
      component~

### DIFF
--- a/utils/styleguide/setup.js
+++ b/utils/styleguide/setup.js
@@ -24,11 +24,7 @@ global.DEFAULT_THEME = DEFAULT_THEME;
 global.PALETTE = PALETTE;
 global.ThemeProvider = ThemeProvider;
 global.Grid = Grid;
-global.Row = styled(Row).attrs({
-  alignItems: 'center'
-})`
-  /* stylelint-disable-line block-no-empty */
-`;
+global.Row = Row;
 global.Col = styled(Col)`
   margin-top: 4px;
   margin-bottom: 4px;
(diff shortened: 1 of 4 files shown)


utils/styleguide/setup.js
```diff
@@ -24,11 +24,7 @@ global.DEFAULT_THEME = DEFAULT_THEME;
 global.PALETTE = PALETTE;
 global.ThemeProvider = ThemeProvider;
 global.Grid = Grid;
-global.Row = styled(Row).attrs({
-  alignItems: 'center'
-})`
-  /* stylelint-disable-line block-no-empty */
-`;
+global.Row = Row;
```
austin94: Would it be better to allow the prop override of

```jsx
global.Row = styled(Row).attrs(props => ({
  alignItems: props.alignItems || 'center'	
}))`	
  /* stylelint-disable-line block-no-empty */	
`;
```

For most of examples we have centered styling. This would allow the prop overrides to affect it correctly.

jzempel: I went through all the examples. Actually, we do not have centered styling for most.

PR #597, opened 2020-01-18
fix(modals): use focusvisible in Modal component

hzhu:
## Description

Focusable elements in modals package examples persist their focused states causing the UI to show that multiple elements are focused after a user tabs through elements.

## Detail

It looks like the cause of this problem is because the modals' children are rendered within a portal, rather than the root node. To fix this issue I updated the modal component to pass the `modalRef` into `useFocusVisible`.

## Demo

Here is a [demo](https://blissful-babbage-396374.netlify.com/modals/) of the fixed example.

## Screenshots

**BEFORE:**

**AFTER:**

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/modals/package.json
+++ b/packages/modals/package.json
@@ -19,6 +19,7 @@
     "start": "../../utils/scripts/start.sh"
   },
   "dependencies": {
+    "@zendeskgarden/container-focusvisible": "^0.3.0",
     "@zendeskgarden/container-modal": "^0.4.0",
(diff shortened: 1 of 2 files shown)


hzhu:
> I don’t think we’ll want to ask Garden consumers to go through these motions. It might be better just to use the focusvisible hook directly in the API.

Baking in the focus visible concerns is much better for consumers, thanks for the suggestion! I've updated the PR description and title to reflect that.

PR #599, opened 2020-01-22
feat(modals): migrate react-modals to TS 

hzhu:
## Description

This pull request migrates the `react-modals` package to TypeScript in preparation for the CSS-in-JS work.

## Detail

This PR comprises of the following work for the `react-modals` package:

1) Configure and convert files to TypeScript.
2) Organize files into the `src/elements` and `src/styled` directories.
3) Replace Lorem ipsum with veggie ipsum text.
4) Rename boolean props using`isBoolean` convention: `isAnimated`, `isCentered`, and `isLarge`. 

## Notes

In a separate (immediately after) PR, I would like to update the examples to return the keyboard focus to the trigger element that was used to open a modal. 

## Demo

A demo pre-published for review can be found at [https://henry-next.netlify.com/modals](https://henry-next.netlify.com/modals/). 

## Checklist

      designer as a reviewer)~
      component


packages/modals/src/styled/StyledClose.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import styled from 'styled-components';
+import classNames from 'classnames';
+import ModalStyles from '@zendeskgarden/css-modals';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.close';
+
+export interface IStyledCloseProps {
+  hovered?: boolean;
```
hzhu: The `hovered` prop isn't renamed to use `isBoolean` naming convention because it will be removed in the CSS-in-JS migration.

packages/modals/package.json
```diff
@@ -18,11 +18,11 @@
     "build:demo": "../../utils/scripts/build-demo.sh",
     "start": "../../utils/scripts/start.sh"
   },
+  "types": "dist/typings/index.d.ts",
   "dependencies": {
     "@zendeskgarden/container-focusvisible": "^0.3.0",
     "@zendeskgarden/container-modal": "^0.4.0",
     "@zendeskgarden/container-utilities": "^0.3.0",
```
jzempel: These container versions should probably be updated.

packages/modals/src/elements/Modal.tsx
```diff
@@ -42,11 +38,47 @@ const isOverflowing = element => {
...
+   */
+  backdropProps?: HTMLAttributes<HTMLDivElement>;
+  /**
+   * Callback when a close action has been completed.
+   * Can be triggered from the backdrop and Close icon.
+   * @param {Object} event - DOM event that triggered the close action
+   */
+  onClose?: (event: KeyboardEvent | MouseEvent) => void;
+  /**
+   * Center modal
```
jzempel: Improve this doc, i.e. "Center the modal on the backdrop"

BREAKING_CHANGES.md
```diff
@@ -117,6 +117,21 @@ New features/fixes include:
...
+  prop name changes for the `<Modal>` component:
+
+  - `animate` -> `isAnimated`
+  - `center` -> `isCentered`
+  - `large` -> `isLarge`
+
+- `<Close>` no longer accepts the `focused` and `hovered` props. The styling for these states are
+  included with the component, and focusvisible is used.
+
+- The `<Backdrop>` and `<ModalView>` components are deprecated. The abstraction for these
```
austin94: > deprecated

Implies that they can still consume and use these components until they are removed. We should be able to just drop them.

packages/modals/src/elements/Body.tsx
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { HTMLAttributes } from 'react';
+import { StyledBody } from '../styled';
+import { useModalsContext } from '../utils/useModalContext';
```
austin94: Could we name this `useModalContext` (singular) to match the file name?

packages/modals/src/elements/Modal.spec.tsx
```diff
@@ -9,27 +9,27 @@ import React from 'react';
...
+  let onCloseSpy: jest.Mock;
 
-  const BasicExample = props => (
+  const BasicExample = (props: IModalProps) => (
     <Modal
       {...props}
       id={MODAL_ID}
       data-test-id="modal"
-      backdropProps={{ 'data-test-id': 'backdrop' }}
+      backdropProps={{ 'data-test-id': 'backdrop' } as any}
```
austin94: I'm surprised `data-*` attributes aren't included with HTMLAttributes used in the props interface. If this is needed we may want to just type `backdropProps` as `any`.

packages/modals/src/elements/Modal.spec.tsx
```diff
@@ -9,27 +9,27 @@ import React from 'react';
...
     >
       <Header data-test-id="header">Example Header</Header>
       <Body data-test-id="body">Body content</Body>
       <Footer data-test-id="footer">
-        <button onClick={props.onClose}>Confirm</button>
+        <button onClick={props.onClose as any}>Confirm</button>
```
austin94: Rather than casting this as `any` you could call it from within an inline method. This may align better with how people would actually consume this in TS.

BREAKING_CHANGES.md
hzhu: You're right, I've updated this sentence to be more explicit and clear. Thank you!

packages/modals/src/elements/Body.tsx
hzhu: Oops, I forgot to update this. It should be singular now, rather than plural. Thanks!

packages/modals/src/elements/Modal.spec.tsx
hzhu: I was surprised too!

One concern for typing `backdropProps` as `any` is that it might result in the prop sheet communicating to consumers that `any` type can be passed for `backdropProps`, even strings, booleans, etc. By casting it here, consumers see the prop sheet type as:

```ts
HTMLAttributes<HTMLDivElement> | undefined
```

This way they are informed that only `div` attributes can be passed to `backdropProps`. Is this a valid concern? Let me know what you think.

packages/modals/src/elements/Modal.spec.tsx
hzhu: Oh cool, I did not know this. Thanks for the suggestion!

packages/modals/src/elements/Modal.spec.tsx
austin94: This should be fine! It's more specific to the actual use-case as well.

PR #600, opened 2020-01-24
feat(dropdowns): migrate menu styling to CSS-in-JS

austin94:
- [X] **BREAKING CHANGE**

## Description

This PR migrates the `css-menus` and `css-arrows` related styling to CSS-in-JS. The other `css-forms` related styling will be migrated in another PR.

The main styling changes in this PR are:
- New focus visual treatment to align with forms/buttons
- Remove `<Separator />` from non-tree examples
- Move from `::before,::after` icons to inline SVGs
  - This affects selection state, header icons, previous item, next item, and add item
  - This move affects the styling around MediaFigures so please double check my work here 🙏

Pre-published at https://austin-next.netlify.com/dropdowns/

## Known Issues

~Currently PopperJS is having trouble understanding the initial size of our Menus. This is creating an initial miss-placement when the menu is rendered in some positions. Please review the styling while I try and debug this issue.~

Was able to fix this by moving our `is-hidden` logic to a node that directly affects PopperJS.

## Breaking Changes

I have removed the `active`, `focused`, and `hovered` props from all publicly consumable menu-related elements.

## Checklist

      designer as a reviewer)
- ~[ ] :nail_care: view component styling is based on a Garden CSS
      component~


packages/dropdowns/src/styled/items/header/StyledHeaderIcon.ts
```diff
@@ -5,17 +5,47 @@
...
+interface IStyledHeaderIcon {
+  isCompact?: boolean;
+}
+
+const getIconSize = (props: IStyledHeaderIcon & ThemeProps<DefaultTheme>) => {
+  if (props.isCompact) {
+    return `${props.theme.space.base * 3}px`;
+  }
+
+  return `${props.theme.space.base * 4}px`;
```
jzempel: These should be `props.theme.iconSizes.sm` and `props.theme.iconSizes.md`.

packages/dropdowns/src/styled/items/StyledItemIcon.ts
```diff
@@ -0,0 +1,58 @@
...
+    height: calc(${props.theme.lineHeights.md} + ${math(`${getItemPaddingVertical(props)} * 2`)});
+  `;
+};
+
+const getIconSizeStyles = (props: IStyledItemIconProps & ThemeProps<DefaultTheme>) => {
+  let size = `${props.theme.space.base * 4}px`;
+
+  if (props.isCompact) {
+    size = `${props.theme.space.base * 3}px`;
+  }
```
jzempel: These should be `props.theme.iconSizes.sm` and `props.theme.iconSizes.md`.

packages/dropdowns/src/styled/items/StyledItemIcon.ts
```diff
@@ -0,0 +1,58 @@
...
+
+interface IStyledItemIconProps {
+  isCompact?: boolean;
+  isVisible?: boolean;
+}
+
+const getSizeStyles = (props: IStyledItemIconProps & ThemeProps<DefaultTheme>) => {
+  return css`
+    width: ${getItemPaddingHorizontal(props)};
+    height: calc(${props.theme.lineHeights.md} + ${math(`${getItemPaddingVertical(props)} * 2`)});
```
jzempel: Prefer `props.theme.space.base * 5` vs  `lineHeights`.

packages/dropdowns/src/elements/Menu/Items/Item.tsx
```diff
@@ -67,36 +68,41 @@ export const Item = React.forwardRef<HTMLLIElement, IItemProps>(
...
-      })
+    return (
+      <Component
+        {...getItemProps({
+          item: value,
+          isFocused,
+          ref,
+          isCompact,
+          'data-garden-is-focused': isFocused,
+          'data-garden-is-selected': isSelected,
```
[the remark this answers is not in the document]
austin94: These `data-garden-*` attributes are used throughout the tests to assert that an item is receiving a valid treatment based on key interaction. This is similar to the tests in `react-datepickers`

packages/dropdowns/src/elements/Menu/Menu.tsx
```diff
@@ -50,6 +50,7 @@ const Menu: React.FunctionComponent<IMenuProps & ThemeProps<DefaultTheme>> = pro
     maxHeight,
     style: menuStyle,
     zIndex,
+    isCompact,
```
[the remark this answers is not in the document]
austin94: These are now updated.

packages/dropdowns/examples/tree-menu.md
```diff
@@ -28,7 +28,6 @@ const renderItems = () => {
   return (
     <>
       <HeaderItem>General Settings</HeaderItem>
-      <Separator />
```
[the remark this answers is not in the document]
austin94: Updated.

packages/dropdowns/src/index.ts
```diff
@@ -22,12 +22,10 @@ export { HeaderItem } from './elements/Menu/Items/HeaderItem';
...
 export { ItemMeta } from './elements/Menu/Items/ItemMeta';
 export { MediaBody } from './elements/Menu/Items/MediaBody';
-export { default as MediaFigure } from './elements/Menu/Items/MediaFigure';
+export { MediaFigure } from './elements/Menu/Items/MediaFigure';
 export { MediaItem } from './elements/Menu/Items/MediaItem';
 export { NextItem } from './elements/Menu/Items/NextItem';
 export { PreviousItem } from './elements/Menu/Items/PreviousItem';
 export { GARDEN_PLACEMENT, POPPER_PLACEMENT } from './utils/garden-placements';
 
 export { resetIdCounter } from 'downshift';
```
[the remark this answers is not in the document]
austin94: I've added a quick explanation and a link out to the Downshift docs on how/when this should be used.

packages/dropdowns/src/styled/items/StyledNextIcon.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import styled from 'styled-components';
+import NextIconSvg from '@zendeskgarden/svg-icons/src/16/chevron-right-stroke.svg';
+import { retrieveComponentStyles, DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'dropdowns.next_item_icon';
+
+/**
+ * Accepts all `<li>` props
```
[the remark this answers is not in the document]
austin94: No. This was accidentally copied over. Removed.

packages/dropdowns/src/styled/items/StyledPreviousIcon.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import styled from 'styled-components';
+import PreviousIconSvg from '@zendeskgarden/svg-icons/src/16/chevron-left-stroke.svg';
+import { retrieveComponentStyles, DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'dropdowns.previous_item_icon';
+
+/**
+ * Accepts all `<li>` props
```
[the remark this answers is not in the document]
austin94: No. This was accidentally copied over. Removed.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -34,88 +76,124 @@ const retrieveMenuMargin = ({
...
+  className: props.isAnimated && 'is-animated'
 }))<IStyledMenuViewProps>`
-  && {
-    position: relative;
-  }
+  display: ${props => props.isHidden && 'inline-block'};
+  position: relative; /* [1] */
+  transition: ${props => props.isHidden && 'opacity .2s ease-in-out, .2s visibility 0s linear'};
+  visibility: ${props => props.isHidden && 'hidden'};
+  opacity: ${props => props.isHidden && '0'};
```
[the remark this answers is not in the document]
austin94: This was already broken in `master`, but I was able to correct all of this logic and also improve our `top|right|left` positioning.

In the original `css-components` the `is-hidden` class had to be applied to the menu element. Since we are using `::before/::after` arrow styling I've moved the `isHidden` logic to the element which wraps the menu (and actually controls PopperJS). This seems to have fixed a good chunk of the issues.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -34,88 +76,124 @@ const retrieveMenuMargin = ({
...
-    return `margin-bottom: ${marginAmount};`;
+const getArrowStyles = (props: IStyledMenuViewProps & ThemeProps<DefaultTheme>) => {
+  if (!props.hasArrow || !props.placement || props.isHidden) {
+    return undefined;
   }
 
-  if (placement === 'left' || placement === 'left-start' || placement === 'left-end') {
-    return `margin-right: ${marginAmount};`;
+  return arrowStyles(getArrowPosition(props.placement), {
+    size: '8.457px',
```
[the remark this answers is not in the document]
austin94: I noticed it being slightly different which is why I left it unthemed. If we can have that match it would be great 🙏

packages/dropdowns/src/styled/items/StyledItem.ts
```diff
@@ -5,46 +5,79 @@
...
-      background-color: inherit;
-    }
-  `};
+  display: block;
+  position: relative; /* [1] */
+  z-index: 0; /* [2] */
+  cursor: ${props => (props.disabled ? 'default' : 'pointer')};
+  padding: ${props => `${getItemPaddingVertical(props)} ${getItemPaddingHorizontal(props)}`};
+  text-decoration: none;
+  line-height: ${props => props.theme.lineHeights.md};
```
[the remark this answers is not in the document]
austin94: Would this be a good use-case for the `getLineHeight` utility? Since this affects text within the Item this feels like a good place for it.

packages/dropdowns/src/styled/items/StyledItemMeta.ts
```diff
@@ -6,19 +6,26 @@
...
  */
 export const StyledItemMeta = styled.span.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: MenuStyles['c-menu__item__meta']
-})`
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledItemMetaProps>`
+  display: ${props => (props.isCompact ? 'none' : 'block')};
+  line-height: ${props => props.theme.lineHeights.sm};
```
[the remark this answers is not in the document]
austin94: Same as https://github.com/zendeskgarden/react-components/pull/600#discussion_r371514754

Since this is meant to apply to text, why wouldn't we use these themed values?

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -6,15 +6,57 @@
...
+
+const ANIMATE_LEFT_KEYFRAME = keyframes`
+  0% {
+    margin-right: -${DEFAULT_THEME.space.base}px;
+  }
+
+  100% {
+    margin-right: 0;
+  }
+`;
```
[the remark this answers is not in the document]
austin94: I wasn't sure if this would have a negative performance impact. Looks like this doesn't affect it was long as the positions don't change rapidly. I've updated this to be instantiated inline.

austin94:
Looks like everyone got a deeper look than I was expecting while this was still in draft mode. I've resolved most of these comments raised so far:

> dropdown isCompact applied throughout all Styleguidist examples across packages

I've updated all non-dropdowns examples to use the new API.

> problems with API documentation
> improve missing JSDoc for function props
> replace VALIDATION constant with strings
> remove props that don't make sense (i.e. mediaLayout, tagLayout)
> still a lot of isFocused/isHovered being listed

These values all seem to relate to the `css-forms` related styling. These will be addressed in a separate PR.

> remove Menu isHidden – no reason to manipulate directly
> Menu is showing HTML style attribute

Removed.

> when will knobs-style examples be implemented for the field types? Should be very similar to https://competent-shirley-cc2f82.netlify.com/forms/#basic. This feels like a big problem with this PR as a lot of styling choices need to be taken on face value and don't have representation in the examples.

I will be introducing knobs-style examples for the forms related styling in the follow-on PR. I will be adding some basic knobs for the `Advanced usage` example to view different sizing and placement.

> we should be concerned about the full percentage point drop in test coverage – simply deleting all of the styled specs seems like a bad idea.
> it would be good to see ref spec testing on the affected element components

These tests have been added and I'm continuing to raise the coverage back up to ~99%.

> how is the current basic menu implementation holding together with https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html#rps2_label?

It seems to be holding together well, but I would prefer to handle this is a separate PR since it is a complex change by itself.

packages/dropdowns/src/styled/items/StyledItem.ts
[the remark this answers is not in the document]
austin94: Moved from our offline conversation.

packages/dropdowns/src/styled/items/StyledItemMeta.ts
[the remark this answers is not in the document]
austin94: Moved from our offline conversation.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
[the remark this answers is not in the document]
rossmoody: Def match imo.

packages/dropdowns/src/styled/items/media/StyledMediaBody.ts
```diff
@@ -7,18 +7,24 @@
...
  */
 export const StyledMediaBody = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: MenuStyles['c-menu__item--media__body']
-})`
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledMediaBodyProps>`
+  display: block;
+  margin-top: ${props => props.isCompact && '2px'};
```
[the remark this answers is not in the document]
austin94: I've moved this to a calculation based on the `MediaFigure` sizing.

packages/dropdowns/src/elements/Menu/Items/AddItem.spec.tsx
```diff
@@ -31,4 +31,23 @@ describe('AddItem', () => {
...
     expect(onSelectSpy.mock.calls[0][0]).toBe('add-item');
   });
+
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLLIElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: I noticed that this `data-test-id` isn't used in the test and can probably be removed.

packages/dropdowns/src/styled/items/StyledItem.ts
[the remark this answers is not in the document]
jzempel: This shouldn't be a `getLineHeight` because we need explicit pixel values here. So stick with the calculation.

packages/dropdowns/src/elements/Menu/Items/HeaderIcon.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+));
+
+describe('HeaderIcon', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: I noticed that this `data-test-id` isn't used in a couple of tests and can probably be removed.

packages/dropdowns/src/elements/Menu/Items/HeaderIcon.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+import React from 'react';
+import { render } from 'garden-test-utils';
+import { Dropdown, Trigger, Menu, HeaderItem, HeaderIcon } from '../../..';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const ExampleSvg = React.forwardRef<SVGSVGElement, {}>((props, ref) => (
+  <svg ref={ref} {...props} />
+));
```
hzhu: Not a blocker, but I was curious if there was a particular reason this component forwards a `ref`? The following appears to work and would yield the same test results:

```jsx
const ExampleSvg = () => <svg />;
```

packages/dropdowns/src/styled/items/media/StyledMediaItem.ts
```diff
@@ -5,23 +5,31 @@
...
-export const StyledMediaItem = styled(StyledItem).attrs(props => ({
+export const StyledMediaItem = styled(StyledItem).attrs<IStyledMediaItem>(props => ({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  className: classNames(props.className, MenuStyles['c-menu__item--media'])
-}))`
+  'data-compact': props.isCompact
+}))<IStyledMediaItem>`
+  ${StyledItemIcon} {
+    height: ${props => !props.isCompact && `calc(${getMediaFigureSize(props)} + 22px)`};
```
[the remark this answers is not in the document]
austin94: I've added a calculation here to make it use the `Item` padding and the vertical margin of the `MediaFigure`.

The current structure has the `MediaItem` aligning with the other item sizes in compact mode so we can just inherit the common styling from `StyledItemIcon`

packages/dropdowns/src/elements/Menu/Items/HeaderItem.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <HeaderItem containsIcon data-test-id="header-item">
+            Header Item
+          </HeaderItem>
+        </Menu>
+      </Dropdown>
+    );
+    const headerItem = getByTestId('header-item');
+
+    expect(headerItem).toHaveStyleRule('padding-left', undefined);
+    expect(headerItem).toHaveStyleRule('padding-right', undefined);
```
hzhu: An alternative would be to use the `.not` matcher so that the second argument can be omitted:

```jsx
expect(headerItem).not.toHaveStyleRule('padding-left');
expect(headerItem).not.toHaveStyleRule('padding-right');
```

packages/dropdowns/README.md
```diff
@@ -93,3 +93,9 @@ via the `downshiftProps` prop.
 Downshift provides several advanced customization features that can be very helpful when
 customizing this component. The [stateReducer](https://github.com/downshift-js/downshift#statereducer)
 pattern is a common customization strategy.
+
+### Server Side Rendering
+
+If you are using server side rendering you may need to [configure specific Downshift settings](https://github.com/downshift-js/downshift#resetidcounter).
+This package re-exports the Downshift `resetIdCunter` utility. It allows reseting the internal id
```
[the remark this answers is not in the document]
austin94: Typos corrected.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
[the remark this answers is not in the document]
austin94: I've moved this to match the `large` tooltip styling. Thanks for the help @rossmoody !

packages/dropdowns/src/elements/Menu/Items/AddItem.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/HeaderIcon.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/HeaderIcon.spec.tsx
austin94: I think I just got in the habit of using `forwardRef` 😄 Removed.

packages/dropdowns/src/elements/Menu/Items/HeaderItem.spec.tsx
austin94: Nice catch. Updated.

packages/dropdowns/src/elements/Menu/Items/MediaFigure.spec.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
+        </Trigger>
+        <Menu>
+          <MediaItem value="image">
+            <MediaFigure>
+              <img ref={ref} data-test-id="media-figure" src="test.png" alt="test" />
```
hzhu: I noticed that the `ref` is applied to the native `img` tag. Should this be on the `MediaFigure` component instead?

packages/dropdowns/src/elements/Menu/Items/MediaFigure.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import React, { HTMLAttributes } from 'react';
+import { StyledMediaFigure } from '../../../styled';
+import useMenuContext from '../../../utils/useMenuContext';
+
+/**
+ * Accepts all `<div>` props
+ */
+export const MediaFigure: React.FC<HTMLAttributes<any>> = props => {
```
hzhu: Should this component forward refs?

packages/dropdowns/src/styled/items/StyledItem.ts
```diff
@@ -5,46 +5,79 @@
...
-      background-color: inherit;
-    }
-  `};
+  display: block;
+  position: relative; /* [1] */
+  z-index: 0; /* [2] */
+  cursor: ${props => (props.disabled ? 'default' : 'pointer')};
+  padding: ${props => `${getItemPaddingVertical(props)} ${getItemPaddingHorizontal(props)}`};
+  text-decoration: none;
+  line-height: ${props => props.theme.space.base * 5}px;
```
hzhu: Could the [`getLineHeight`](https://github.com/zendeskgarden/react-components/blob/next/packages/theming/src/utils/getLineHeight.ts#L18) utility be used here?

packages/dropdowns/src/styled/items/StyledNextItem.ts
```diff
@@ -5,23 +5,26 @@
...
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION,
-  className: classNames(props.className, MenuStyles['c-menu__item--next'])
-}))`
+  'data-garden-version': PACKAGE_VERSION
+})`
+  ${StyledItemIcon} {
+    right: ${props => (props.theme.rtl ? 'auto' : '0')};
+    left: ${props => (props.theme.rtl ? '0' : 'auto')};
+  }
```
hzhu: Is there a particular reason why these styles were declared here, rather than directly within `StyledItemIcon`?

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -170,13 +206,41 @@ export interface IStyledMenuProps extends HTMLAttributes<HTMLUListElement> {
...
+        isAnimated={isAnimated}
+        zIndex={zIndex}
+      >
+        <StyledMenuView
+          hasArrow={hasArrow}
+          placement={placement}
+          isAnimated={isAnimated}
+          {...other}
+        >
           <StyledMaxHeightWrapper maxHeight={maxHeight}>{children}</StyledMaxHeightWrapper>
```
jzempel: This is broken HTML as it results in `ul` > `div` > `li`.

packages/dropdowns/src/styled/items/StyledItem.ts
jzempel: No, it needs pixel values in order to avoid cross-browser dimension shift.

packages/dropdowns/src/styled/items/StyledNextItem.ts
jzempel: The particular R/L icon positioning is only relevant for a `NextItem`.

packages/dropdowns/src/elements/Menu/Items/HeaderItem.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+describe('HeaderItem', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLLIElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: There are a few unused `data-test-id` in this file.

packages/dropdowns/examples/advanced-menu.md
```diff
@@ -5,52 +5,138 @@ All `Items` require a `value` prop that is provided with the `onSelect` callback
...
+        >
+          <HeaderItem containsIcon>
+            <HeaderIcon>
+              <ClipboardIcon />
+            </HeaderIcon>
+            Header Item
+          </HeaderItem>
+          <Item value="profile">Option 1</Item>
+          <Item value="settings">Option 2</Item>
+          <Item disabled>Disabled item</Item>
```
jzempel: Recommend removing this `disabled` item and adding a "Disable items" toggle to the controls. Two reasons:
1) it lets us vet that disabled styling works as intended across all of the item types
2) it removes a wonky visual anti-pattern where the first view of the menu plants a `separator` directly in line with the arrow

Screen Shot 2020-01-29 at 2 03 29 PM

packages/dropdowns/src/elements/Menu/Items/ItemMeta.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Dropdown, Trigger, Menu, MediaItem, MediaBody, ItemMeta } from '../../..';
+
+describe('ItemMeta', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: Unused `data-test-id` 👀 .

packages/dropdowns/src/elements/Menu/Items/MediaBody.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Dropdown, Trigger, Menu, MediaItem, MediaBody } from '../../..';
+
+describe('MediaBody', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: Unused `data-test-id` 👀 .

packages/dropdowns/src/elements/Menu/Items/MediaFigure.spec.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Dropdown, Trigger, Menu, MediaItem, MediaBody, MediaFigure } from '../../..';
+
+describe('MediaFigure', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLImageElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: Unused `data-test-id` 👀 .

packages/dropdowns/src/elements/Menu/Items/Separator.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Dropdown, Trigger, Menu, Item, Separator } from '../../..';
+
+describe('Separator', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLLIElement>();
+
+    const { getByTestId } = render(
+      <Dropdown>
+        <Trigger>
+          <button data-test-id="trigger">Test</button>
```
hzhu: Unused `data-test-id` 👀 .

packages/dropdowns/src/elements/Menu/Items/MediaFigure.spec.tsx
austin94: The `MediaFigure` component using `React.cloneElement` to apply styling directly onto an element. This test is ensuring that all props are spread to the correct element (itself).

packages/dropdowns/src/elements/Menu/Items/MediaFigure.tsx
austin94: Answered in https://github.com/zendeskgarden/react-components/pull/600#discussion_r372666448

packages/dropdowns/src/styled/menu/StyledMenu.tsx
austin94: With the old PopperJS positioning logic we had to define the overflow container. I'm surprised AXE never warned about this? It's possible the new `role` values made it semantically correct?

Either way, I've moved the `maxHeight` styling logic to `StyledMenu` and removed the `<div>`

packages/dropdowns/src/elements/Menu/Items/HeaderItem.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/ItemMeta.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/MediaBody.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/MediaFigure.spec.tsx
austin94: Removed.

packages/dropdowns/src/elements/Menu/Items/Separator.spec.tsx
austin94: Removed.

packages/dropdowns/examples/advanced-menu.md
austin94: Good points. I've added a disabled option to the knobs.

austin94:
> The updated demo shows a problem with menu positioning. In both start and top placements, there is an extra pixel of margin between the menu object and its trigger – both with or without arrow.

This is a side-effect of the new arrow size. I believe this extra space is also occurring in Tooltips

image

I've updated the margin calculation to include the additional size of the borders

packages/dropdowns/src/styled/menu/StyledMenu.tsx
austin94: This actually opened up a bigger can of worms than I was expecting. Modifying the DOM structure here is changing how Popper is able to calculate an accurate position while our animations are happening. Some additional complexity is that our arrow styling requires that any `overflow` values be applied to a nested element, not the Menu or it's parent element.

As a temporary fix I have made all nodes `<div>`. Even though semantically this isn't the best, it is a valid DOM structure that isn't throwing errors in AXE since Downshift is apply `role`s to all elements.

This change is complex enough that I think we should solve it with our other a11y questions in the follow-on PR.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -35,14 +40,16 @@ const retrieveMenuMargin = ({
...
-
   if (!placement) {
     return '';
   }
 
+  let marginAmount = shouldShowArrow({ hasArrow, placement })
+    ? getArrowSize({ theme })
+    : `${theme.space.base}px`;
+
+  marginAmount = math(`${marginAmount} - ${math(`${theme.borderWidths.sm} * 2`)}`);
```
jzempel: This is the wrong update, resulting in more problematic positioning than seen previously. We want the full 4px (or 8px for arrows) margin. It should not be reduced by border width.

The following pics try to highlight the 1px discrepancy seen before this code change. Usually these have something to do with unanticipated space being added to the DOM and my require display, alignment, or font-size tweaks in order to address...

Screen Shot 2020-01-30 at 9 42 40 AM

Screen Shot 2020-01-30 at 9 42 16 AM

These, on the other hand, are displaying as desired:

Screen Shot 2020-01-30 at 9 43 49 AM

Screen Shot 2020-01-30 at 9 44 23 AM

jzempel: Actually, upon closer inspection that 3rd "end" placement above could use a tweak, too. It impedes over the button border by 1px.

austin94: The 1px overlap for `start` and `end` positions seems to be a Popper calculation issue with the trigger. If I use a standard `<div>` it positions correctly.

While working to remove the `line-height` value I found that it is required for Popper to handle our `margin-*` animations correctly. Without a value the initial position of the popper is off-center.

I'm going to keep looking into this, but the positions seem to be valid now.

austin94: It turns out any non-zero font-size also corrected this issue. I've now added a `font-size: 0.01px` rule with a styling comment.

The positioning should be accurate now!

austin94:
> Apply disabled to either NextItem or PreviousItem to observe some bad styling issues. These need to be addressed (even though design doesn't approve of "disabled" next/prev).

There was some logic issues with the new `StyledItemIcon` usage in those components. Updated.

austin94:
> One final issue from me: `disabled` should be filtered from display on the `HeaderItem` prop sheet.

I was originally using the `IStyledItemProps` interface as the basis for the `HeaderItem`. With the amount of refactoring I've done this no longer makes sense. I've created a local props interface for this file.

packages/dropdowns/src/styled/menu/StyledMenu.tsx
```diff
@@ -167,13 +167,19 @@ interface IStyledMenuWrapperProps {
   isAnimated?: boolean;
 }
 
+/**
+ * 1. PopperJS requires a non-zero font-size to perform
+ * its initial placement correctly. This font-size value
+ * affects the content of the arrow styling ::before/::after
+ * elements.
```
jzempel: This second part isn't correct. The `font-size` has nothing to do with the arrow. It is simply pixel-sized now. See https://github.com/zendeskgarden/react-components/blob/next/packages/theming/src/utils/arrowStyles.ts for details.

austin94: I've updated this comment.

PR #601, opened 2020-01-26
feat(pagination): CSS-in-JS refactor

jzempel:
- [x] **BREAKING CHANGE?** 

## Description

All CSS moved into themed JS, making `styles.css` import unnecessary. Bundle size reduction from apx 19kb (12kb JS + 7kb CSS) down to 7.75kb combined. :tada:

Fixes:
- Pagination `renderPages` logic corrected to properly handle pages, padding, and gaps.
- Updated with improved ARIA for accessibility.

Feature enhancements:
- New `pageGap` prop for specifying gap positioning

## Detail

Demo pre-published for review: https://competent-shirley-cc2f82.netlify.com/pagination/

## Checklist

      designer as a reviewer)
      component~


packages/pagination/src/elements/Pagination.spec.tsx
```diff
@@ -8,10 +8,16 @@
 import React from 'react';
 import { render, fireEvent } from 'garden-test-utils';
 import { KEY_CODES } from '@zendeskgarden/container-utilities';
-
-import Pagination, { IPaginationProps, PAGE_TYPE } from './Pagination';
+import { default as Pagination, IPaginationProps, PAGE_TYPE } from './Pagination';
```
austin94: Should we use the standard default export? i.e.

```ts
import Pagination, { IPaginationProps, PAGE_TYPE } from '...';
```

packages/pagination/src/elements/Pagination.tsx
```diff
@@ -35,18 +31,27 @@ export interface IPaginationProps extends Omit<HTMLAttributes<HTMLUListElement>,
...
   /**
-   * @param {Any} currentPage - The newly selected page
+   * Position for the leading and trailing gap indicator, if needed based on
+   * current and total pages
    */
-  onChange?: (updatedCurrentPage: number) => void;
+  pageGap?: number;
   /**
-   * Allows custom props to be applied to each page element. Useful for QA attributes and localization.
+   * @param {Any} currentPage - The currently selected page
```
austin94: These JSDoc values are already created by TSDoc in the prop-sheet. Do we need to duplicate these here?

packages/pagination/src/elements/Pagination.spec.tsx
jzempel: No, without `default as ...` we miss the ability to test the `ref` exported via the default `withTheme` export.

packages/pagination/src/elements/Pagination.tsx
jzempel: Without the `param`, the argument isn't documented like I want it to be.

Screen Shot 2020-01-27 at 3 20 20 PM

PR #602, opened 2020-01-27
feat(modals): migrate to CSS-in-JS

hzhu:
##  Description

This pull request migrates `react-modals` to CSS-in-JS and removes the dependencies on stylesheets backed by [@zendeskgarden/css-modals](https://www.npmjs.com/package/@zendeskgarden/css-modals).

## Detail

Styling has moved to a CSS-in-JS approach. 

The demo for this pull request can be found [here](https://henry-next.netlify.com/modals/) **(updated as of 1/31)**.

## Checklist

      designer as a reviewer)
      component
- [x] :memo: tested in Chrome, Firefox, Safari, IE11, and Edge.


packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,107 @@
...
+  }
+
+  /* stylelint-disable-next-line */
+  @media (max-width: 699px) {
+    right: ${props => props.theme.rtl && props.theme.space.base * 6}px;
+    left: ${props => (props.theme.rtl ? 'auto' : `${props.theme.space.base * 6}px`)};
+  }
+
+  /* stylelint-disable-next-line */
+  @media screen and (-ms-high-contrast: active), screen and (-ms-high-contrast: none) {
```
hzhu: I wasn't able to target IE11 to center the modal using `_:-ms-input-placeholder`. However, this media query works. More than happy to switch this over to an alternative if there are suggestions.

packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,107 @@
...
+  flex-direction: column;
+  margin: ${props => (props.isCentered ? '0' : `${props.theme.space.base * 12}px`)};
+  border: ${props => `${props.theme.borders.sm} ${getColor('neutralHue', 300)}`};
+  border-radius: ${props => props.theme.borderRadii.md};
+  border-color: transparent;
+  box-shadow: ${boxShadow};
+  background-color: ${props => props.theme.colors.background};
+  width: ${props => (props.isLarge ? '800px' : '542px')};
+  min-height: 60px;
+  max-height: calc(100vh - 96px);
```
hzhu: Computing `60px` with `space.base` wasn't used because a minor adjustment in `space.base` may result in a very large `min-height` value. I also left the `calc` in here, as I was unsure what the CSS-in-JS equivalent is.

If `max-height` needs to be changed, I would love some suggestions, and the steps to get this value right.

packages/modals/src/styled/StyledBackdrop.ts
```diff
@@ -18,21 +16,46 @@ export interface IStyledBackdropProps {
...
+
+    to {
+      opacity: 1;
+    }
+  `;
+
+  return css`
+    animation-name: ${animationName};
+    animation-duration: 0.15s;
+    animation-timing-function: ease-in;
```
hzhu: I separated the animation CSS properties to longhand form so that an assertion can be made against `animation-duration` and `animation-timing-function` for modals that are rendered with the `isAnimated` prop. I wasn't sure if asserting against `animation-name` was suitable since it is a generated string (but maybe it might be?). Request for changes and suggestions are welcomed.

packages/modals/src/elements/Header.tsx
```diff
@@ -6,8 +6,17 @@
...
 
 import React from 'react';
-import { StyledHeader, IStyledHeaderProps } from '../styled';
+import styled from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import AlertErrorStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
 import { useModalContext } from '../utils/useModalContext';
+import { StyledHeader, IStyledHeaderProps } from '../styled';
+
+const StyledIcon = styled(AlertErrorStrokeIcon)`
```
austin94: Should this styled element style the StrokeIcon directly or style a div which positions its contents (the icon) correctly? 

I'm leaning towards the styled div solution as it lets you move this files to the `styled` directory and use multiple icon variants if it's something we want in the future.

packages/modals/src/styled/StyledBackdrop.ts
```diff
@@ -18,21 +16,46 @@ export interface IStyledBackdropProps {
...
+const animationStyles = () => {
+  const animationName = keyframes`
+    0% {
+      opacity: 0;
+    }
+
+    to {
+      opacity: 1;
+    }
+  `;
```
austin94: If this is a consistent animation lets move it as a constant so it isn't being newed up during each render.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -17,16 +15,46 @@ export interface IStyledCloseProps {
...
+  border: none;
+  border-radius: 50%;
+  background-color: transparent;
+  cursor: pointer;
+  padding: 0;
+  width: ${props => props.theme.space.base * 10}px;
+  height: ${props => props.theme.space.base * 10}px;
+  overflow: hidden;
+  text-decoration: none;
+  color: ${getColor('neutralHue', 600)};
```
austin94: Does this need a `props.theme` argument?

packages/modals/src/styled/StyledFooterItem.ts
```diff
@@ -7,15 +7,25 @@
...
+  margin-left: ${props => props.theme.space.base * 5}px;
+  min-width: 0;
+
+  &:first-child {
+    margin: ${props => !props.theme.rtl && 0};
+  }
+
+  &:last-child {
+    margin: ${props => props.theme.rtl && 0};
+  }
```
austin94: For these values you could dynamically compute the correct child. Like this

```
&:${props => props.theme.rtl ? 'last' : 'first'}-child {
  ...
}
```

packages/modals/src/styled/StyledHeader.spec.tsx
```diff
@@ -13,12 +13,12 @@ describe('StyledHeader', () => {
   it('renders default styling', () => {
     const { container } = render(<StyledHeader />);
 
-    expect(container.firstChild).toHaveClass('c-dialog__header');
+    expect(container.firstChild).toHaveStyleRule('color', '#2f3941');
```
austin94: Should these values be calculated from the `DEFAULT_THEME` or `PALETTE` objects?

packages/modals/src/styled/StyledClose.ts
hzhu: It seems like the `getColor` call here returns the correct value: `#68737d`, so it does work in this PR. Do you know if there are cases where not passing in the third argument (`theme`) results in unintended values?

I'll go with the safer approach here, and use `${props => getColor('neutralHue', 600, props.theme)}`. Thanks for pointing this out!

packages/modals/src/styled/StyledBackdrop.ts
hzhu: I've removed the function call. Now, the animations should not be newed up each render. Thanks for the suggestion!

packages/modals/src/styled/StyledFooterItem.ts
hzhu: Thanks for the suggestion. I tried it, but the linter is giving me some problems 🤔.

Screen Shot 2020-01-28 at 4 25 30 PM

packages/modals/src/elements/Header.tsx
hzhu: Awesome, this component has been moved to the `/styled` directory. The styling is still applied to the StrokeIcon directly. This component can be adjusted accordingly when the requirements change.

packages/modals/src/styled/StyledHeader.spec.tsx
hzhu: In another PR, @jzempel had mentioned to me to avoid calculating CSS units using `DEFAULT_THEME`'s space.base unit in spec files. As a result, I hardcoded the color value here.

I could update this assertion to look directly at the `DEFAULT_THEME`:

```jsx
expect(container.firstChild).toHaveStyleRule('color', DEFAULT_THEME.palette.red[600]);
```

Or it could use the `getColor` function to derive the color from the `DEFAULT_THEME`:

```jsx
expect(container.firstChild).toHaveStyleRule(
  'color',
  getColor('dangerHue', 600, DEFAULT_THEME)
);
```

Do ya'll have a preference?

austin94: Oh, didn't realize that was the direction we were going now. If that's the case feel free to resolve!

My only comment is that hard-coding these values makes these start to feel like snapshot tests rather than unit/integration tests.

packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,105 @@
...
+  50% {
+    transform: scale(1.05);
+  }
+
+  to {
+    opacity: 1;
+  }
+`;
+
+const animationStyles = css`
```
jzempel: Missing `animation-delay: 0.01s;`

packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,105 @@
...
+  /* stylelint-disable-next-line */
+  @media (max-height: 399px) {
+    top: ${props => props.theme.space.base * 6}px;
+    bottom: auto;
+    margin-bottom: ${props => props.theme.space.base * 6}px;
+    max-height: none;
+  }
+
+  /* stylelint-disable-next-line */
+  @media (max-width: ${props => (props.isLarge ? '799px' : '699px')}) {
```
jzempel: I think you could update to rely on the `props.theme.breakpoints.md` value for small here. The 799px should be calculated as the 800px `width` property minus 1px. Moving all these related values out to a `sizeStyles` function would make all of this much more clear.

packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,105 @@
...
+    right: ${props => props.theme.rtl && props.theme.space.base * 6}px;
+    left: ${props => (props.theme.rtl ? 'auto' : `${props.theme.space.base * 6}px`)};
```
jzempel: Use `${props.theme.rtl ? 'right' : 'left}: [calculated-value]` instead of the two properties.

packages/modals/src/styled/StyledModal.ts
jzempel: Hmm, I'm still seeing it work on the `css-components` side. We might need to discuss.

packages/modals/src/styled/StyledFooterItem.ts
hzhu: Tried various forms of `prettier-ignore` and `stylelint-disable`, but the error still shows when running `yarn lint`. There could be an [issue](https://github.com/stylelint/stylelint/issues/4401) with stylelint.

hzhu: On second thought, I have removed these styles in favor of the one that JZ suggested in another thread.

```css
margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 5}px;
```

packages/modals/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,20 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import AlertErrorStrokeIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
+
+export const StyledIcon = styled(AlertErrorStrokeIcon)`
+  position: absolute;
+  top: ${props => props.theme.space.base * 5.5}px;
```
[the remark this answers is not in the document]
hzhu: I could make that change — what do you suggest? Since `padding-top` is not located in this file, how would the calculation be made? Or would the `top` style be moved out of this file?

packages/modals/src/styled/StyledModal.ts
hzhu: The current media query does target IE11. I could using `_:-ms-input-placeholder` incorrectly in CSS-in-JS. This is what I tried:

```css
_:-ms-input-placeholder {
  right: ${props => props.isCentered && '50%'}; /* [1] */
  bottom: ${props => props.isCentered && '50%'}; /* [1] */
  transform: ${props => props.isCentered && 'translate(50%, 50%)'}; /* [1] */
}
```

jzempel: In order to make the hack work you need to use `_:-ms-input-placeholder, & { ...`

packages/modals/src/styled/StyledBackdrop.ts
```diff
@@ -18,21 +16,42 @@ export interface IStyledBackdropProps {
...
 }
 
+const animationName = keyframes`
+  0% {
+    opacity: 0;
+  }
+
+  to {
+    opacity: 1;
+  }
```
austin94: Should we combine percentage and `from/to` syntax? Does `from` work?

packages/modals/src/styled/StyledModal.ts
```diff
@@ -6,35 +6,99 @@
...
+  }
+
+  50% {
+    transform: scale(1.05);
+  }
+
+  to {
+    opacity: 1;
+  }
+`;
```
austin94: Same comment as above, should this be 100% instead?

packages/modals/src/styled/StyledBackdrop.ts
hzhu: This was moved over from https://unpkg.com/browse/@zendeskgarden/css-modals@6.4.19/dist/index.css, but `from` would work too.

jzempel: It works. This is an effect of `postcss` compiling to the smallest representation. The original is `0%` - `100%`: https://github.com/zendeskgarden/css-components/blob/master/packages/modals/src/_keyframes.css

packages/modals/src/styled/StyledIcon.ts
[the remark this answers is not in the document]
jzempel: You would `export` the necessary dimensions from the various locations. Again, might be unneeded overhead. This is actually a valid case for pulling the particular property (i.e. `top`) into the other styled file using `& ${StyledIcon} { ...` so that you can take advantage of the local set of size dimensions.

packages/forms/package.json
```diff
@@ -46,6 +46,6 @@
     "access": "public"
   },
   "zendeskgarden:library": "GardenForms",
-  "zendeskgarden:max_size": "17 kB",
+  "zendeskgarden:max_size": "17.1 kB",
```
jzempel: Set this to `17.5` in order to match the `jzempel/table-colors` branch.

packages/modals/src/styled/StyledFooterItem.ts
```diff
@@ -7,15 +7,24 @@
...
 })`
+  display: flex;
+  /* stylelint-disable property-no-unknown */
+  margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 5}px;
+  /* stylelint-enable property-no-unknown */
+  min-width: 0;
+
+  &:first-child {
+    margin-right: ${props => props.theme.rtl && 0};
+    margin-left: 0;
```
jzempel: Structure this just as you did the one-line `margin-...` property above. In other words, `margin-${props.theme.rtl ? ... }: 0;` This will always be the preferred pattern when you encounter directionality margining/padding/etc.

packages/modals/src/styled/StyledFooterItem.ts
```diff
@@ -7,15 +7,25 @@
...
 })`
+  display: flex;
+  /* stylelint-disable property-no-unknown */
+  margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 5}px;
+  /* stylelint-enable property-no-unknown */
+  min-width: 0;
+
+  &:first-child {
+    /* stylelint-disable property-no-unknown */
+    margin-${props => (props.theme.rtl ? 'right' : 'left')}: 0;
```
jzempel: Did you verify this? I think it's backwards.

hzhu: I verified it now and it looks like the margin is removed correctly, with respect to RTL and LTR.

PR #607, opened 2020-02-03
fix(dropdowns): overflow in Multiselect items

sunesimonsen:
## Description

We currently wrap multiselect items in a div that doesn't have overflow hidden, that makes it impossible to do text truncation. 

## Detail

Before:

Screenshot 2020-02-03 at 11 16 51

After: 

Screenshot 2020-02-04 at 19 08 29

### DIFF
--- a/packages/dropdowns/examples/multiselect.md
+++ b/packages/dropdowns/examples/multiselect.md
@@ -10,7 +10,8 @@ our standard `Tag` component.
     <Multiselect
       renderItem={({ value, removeValue }) => (
         <Tag size="large">
-          {value} <Close onClick={() => removeValue()} />
+          <span>{value}</span>
+          <Close onClick={() => removeValue()} />
         </Tag>
       )}
     />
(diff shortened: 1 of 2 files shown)


sunesimonsen:
Mhh putting a span around the text in the tag solved the ellipsis problem with this fix being present:

Screenshot 2020-02-03 at 11 25 09

sunesimonsen:
I don't understand why I would fail the pagination tests 🤔

sunesimonsen:
It seems like master also having this problem (just testing locally).

austin94:
Could you also include the `<span>{value}</span>` fix for the ellipsis? I think that was just missed when we wrote this example originally.

sunesimonsen:
@austin94 I fixed the readme, so now everything shows correctly. Would you mind releasing this when the tests gets sorted out?

austin94:
Applying `overflow: hidden` to the item wrapper clips the focus ring when an item is selected.

Applying a `max-width: 100%` may result in the experience we're looking for.

sunesimonsen:
@jzempel I need to fix this somehow or another, we have customers that can't see their labels. So I'm going to look more into it.

sunesimonsen:
I changed the StyledItemWrapper to use padding instead, that seems to solve the problem, please retest.

Screenshot 2020-02-04 at 08 24 00

sunesimonsen:
Thanks @austin94  for the help on getting this right.

PR #609, opened 2020-02-04
perf(loaders): prevent rerender for dots

li-kai:
## Description

In Lotus, we are seeing repeated renders of `Dots` even though they do not have any props that have changed.

Screenshot 2020-02-04 at 4 53 45 pm

## Detail

This is because `useSchedule` is continuously rescheduling as `loop`'s default is `true`, and causing rerenders.

Since Dots only needs rerender when `noAnimatedSVGSupport` is true, we will set `loop` based on this value.

DotsRerender

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/loaders/src/Dots.js
+++ b/packages/loaders/src/Dots.js
@@ -28,8 +28,8 @@ export default function Dots({
   delayMS = 750,
   ...other
 }) {
-  const { delayComplete } = useSchedule({ duration, delayMS });
   const noAnimatedSVGSupport = useCSSSVGAnimation();
+  const { delayComplete } = useSchedule({ duration, delayMS, loop: noAnimatedSVGSupport });
   const dotOne = useRef(null);
   const dotTwo = useRef(null);
   const dotThree = useRef(null);
--- a/packages/loaders/src/Dots.spec.js
+++ b/packages/loaders/src/Dots.spec.js
@@ -9,8 +9,12 @@ import React from 'react';
 import { render, act } from 'garden-test-utils';
 import mockDate from 'mockdate';
 import Dots from './Dots';
+import { useCSSSVGAnimation } from './hooks/useCSSSVGAnimation';
 
 jest.useFakeTimers();
+jest.mock('./hooks/useCSSSVGAnimation', () => ({
+  useCSSSVGAnimation: jest.fn(() => false)
+}));
 
 const DEFAULT_DATE = new Date(2019, 1, 5, 1, 1, 1);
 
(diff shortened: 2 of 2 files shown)


packages/loaders/src/Dots.js
```diff
@@ -62,9 +62,9 @@ export default function Dots({
       {...other}
     >
       <g fill="currentColor">
-        <DotsOneCircle duration={duration} ref={dotOne} />
-        <DotsTwoCircle duration={duration} ref={dotTwo} />
-        <DotsThreeCircle duration={duration} ref={dotThree} />
+        <DotsOneCircle duration={duration} ref={noAnimatedSVGSupport ? dotOne : null} />
+        <DotsTwoCircle duration={duration} ref={noAnimatedSVGSupport ? dotTwo : null} />
+        <DotsThreeCircle duration={duration} ref={noAnimatedSVGSupport ? dotThree : null} />
```
austin94: Why are is conditionally apply the refs here necessary?

packages/loaders/src/Dots.js
```diff
@@ -28,13 +28,13 @@ export default function Dots({
   delayMS = 750,
   ...other
 }) {
-  const { delayComplete } = useSchedule({ duration, delayMS });
   const noAnimatedSVGSupport = useCSSSVGAnimation();
+  const { delayComplete } = useSchedule({ duration, delayMS, loop: noAnimatedSVGSupport });
```
austin94: Good catch here 🎉

packages/loaders/src/Dots.js
```diff
@@ -28,13 +28,13 @@ export default function Dots({
...
   const dotOne = useRef(null);
   const dotTwo = useRef(null);
   const dotThree = useRef(null);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
austin94: `useLayoutEffect` will [cause a warning to be shown when it's server-side rendered](https://reactjs.org/docs/hooks-reference.html#uselayouteffect). This hook is synchronous so the animation may look better, but the rest of the app could see some performance hits while this animation is running, especially with how often `delayComplete` is updated.

Did you attempt to use `React.memo` for the `Dots*Circle` components and `StyledSvg` element? Wrapping these elements in a `memo` seemed to have a larger impact in the profiler results.

packages/loaders/src/Dots.spec.js
```diff
@@ -46,6 +50,60 @@ describe('Dots', () => {
...
+        }
+
+        .c1 {
+          -webkit-animation: hiIgQS 1250ms linear infinite;
+          animation: hiIgQS 1250ms linear infinite;
+        }
+
+        .c2 {
+          -webkit-animation: jfCXbz 1250ms linear infinite;
+          animation: jfCXbz 1250ms linear infinite;
```
hzhu: > relies on css animation if animated svg is not supported

It looks like CSS animation is used on Chrome 79 (latest). Does this mean that the Chrome doesn't support animated SVGs? I would have thought it was the older browsers like IE11 that don't support animated SVGs.

Screen Shot 2020-02-04 at 9 46 49 AM

packages/loaders/src/Dots.js
ryanseddon: I don't think it's worth checking if the ref is needed or not, theres little cost in applying a ref even if it isn't used.

packages/loaders/src/Dots.spec.js
ryanseddon: This is what styled-components outputs I know edge uses -webkit prefixes for some CSS

li-kai: I made a mistake with the test name. It should be "relies on svg transition attribute if css svg animation is not supported".

packages/loaders/src/Dots.js
li-kai: My primary concern is with the noise caused by repeated re-renderings in react devtools. The actual component's rendering isn't slow at all (<0.1ms).

austin94: > My primary concern is with the noise caused by repeated re-renderings in react devtools

Wrapping the `Dots*` and `StyledSvg` component in a `React.memo(...)` would clean up the profiler flamegraph a lot for you. Since we're using the `ref`'s to apply the transform property I think this would have some good performance impacts as well.

PR #611, opened 2020-02-05
feat(modals): Update the modal close button style

hzhu:
## Description

Update the modal close button to match icon buttons.

## Detail

Icon button states have been updated to use opacities for hover and active states. The "X" close button on a modal needs to be styled to match regular icon buttons.

## Demo

https://henry-next.netlify.com/modals#basic (Updated as of 2/6/2020).

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/package.json
+++ b/packages/modals/package.json
@@ -23,8 +23,7 @@
     "@zendeskgarden/container-focusvisible": "^0.3.0",
     "@zendeskgarden/container-modal": "^0.6.0",
     "@zendeskgarden/container-utilities": "^0.4.0",
-    "dom-helpers": "^5.1.0",
-    "polished": "^3.4.1"
+    "dom-helpers": "^5.1.0"
   },
   "peerDependencies": {
     "@zendeskgarden/react-theming": "^6.0.0 || ^7.0.0",
@@ -46,6 +45,6 @@
     "access": "public"
   },
   "zendeskgarden:library": "GardenModals",
-  "zendeskgarden:max_size": "12 kB",
+  "zendeskgarden:max_size": "9 kB",
   "zendeskgarden:src": "src/index.ts"
 }
--- a/packages/modals/package.json
+++ b/packages/modals/package.json
@@ -23,7 +23,8 @@
     "@zendeskgarden/container-focusvisible": "^0.3.0",
     "@zendeskgarden/container-modal": "^0.6.0",
     "@zendeskgarden/container-utilities": "^0.4.0",
-    "dom-helpers": "^5.1.0"
+    "dom-helpers": "^5.1.0",
+    "polished": "^3.4.1"
   },
   "peerDependencies": {
     "@zendeskgarden/react-theming": "^6.0.0 || ^7.0.0",
@@ -45,6 +46,6 @@
     "access": "public"
   },
   "zendeskgarden:library": "GardenModals",
-  "zendeskgarden:max_size": "9 kB",
+  "zendeskgarden:max_size": "12 kB",
   "zendeskgarden:src": "src/index.ts"
 }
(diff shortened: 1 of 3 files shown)


packages/modals/src/styled/StyledClose.ts
```diff
@@ -5,22 +5,42 @@
...
+      color: ${getColor('neutralHue' as string, 700, props.theme)};
+    }
+
+    &[data-garden-focus-visible] {
+      box-shadow: ${props.theme.shadows.md(rgba(baseColor as string, 0.35))};
+    }
+
+    &:active {
+      background-color: ${rgba(baseColor as string, 0.2)};
+      color: ${getColor('neutralHue' as string, 800, props.theme)};
```
jzempel: Why are these `as string` specifiers needed on `getColor`? I've not seen this in other areas of the code.

packages/modals/package.json
```diff
@@ -23,7 +23,8 @@
     "@zendeskgarden/container-focusvisible": "^0.3.0",
     "@zendeskgarden/container-modal": "^0.6.0",
     "@zendeskgarden/container-utilities": "^0.4.0",
-    "dom-helpers": "^5.1.0"
+    "dom-helpers": "^5.1.0",
+    "polished": "^3.4.1"
```
jzempel: `polished` should not be needed on this package. I'll explain below.

packages/modals/package.json
```diff
@@ -45,6 +46,6 @@
     "access": "public"
   },
   "zendeskgarden:library": "GardenModals",
-  "zendeskgarden:max_size": "9 kB",
+  "zendeskgarden:max_size": "12 kB",
```
jzempel: Revert after removing `polished`.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -5,22 +5,42 @@
...
+    background-color: transparent;
+    color: ${getColor('neutralHue' as string, 600, props.theme)};
+
+    &:hover {
+      background-color: ${rgba(baseColor as string, 0.08)};
+      color: ${getColor('neutralHue' as string, 700, props.theme)};
+    }
+
+    &[data-garden-focus-visible] {
+      box-shadow: ${props.theme.shadows.md(rgba(baseColor as string, 0.35))};
```
jzempel: Do not use `rgba` directly. Instead, prefer to call `getColor('primaryHue', 600, props.theme, 0.35)`

packages/modals/src/styled/StyledClose.ts
```diff
@@ -51,10 +66,7 @@ export const StyledClose = styled.button.attrs<IStyledCloseProps>({
     outline: none;
   }
 
-  &:hover {
-    color: ${props => getColor('neutralHue', 800, props.theme)};
-  }
-
+  ${props => colorStyles(props)}
```
jzempel: Nice, makes color manipulation much more maintainable going forward. :pray:

packages/modals/src/styled/StyledClose.ts
hzhu: Whoops, this was leftover from some earlier work, it's not needed and is removed 👍 .

packages/modals/src/styled/StyledClose.ts
hzhu: I originally called `getColor` instead of `rgba`, but opted to follow how `StyledButton.ts` [created it's `box-shadow`](https://github.com/zendeskgarden/react-components/blob/next/packages/buttons/src/styled/StyledButton.ts#L71). What is the heuristic that can be used to determine whether to use `rgba` versus `getColor`?

I've updated this branch to use `getColor`, and removed `polished/rgba` 👍.  The demo has also been updated.

jzempel: The opacity enhancement was added to `getColor` after buttons was converted and I missed updating that spot. Use `getColor`.

PR #613, opened 2020-02-06
fix(dropdowns): limit PopperJS computations while dropdowns are closed

austin94:
## Description

Garden `react-dropdowns` components are sporadically force-closing some versions of Chrome browsers in Android. This isn't consistent by to Android version, device, or chrome version.

Running LogCat with ADB shows a segmentation fault where Chromium is attempting to modify GPU thresholds for specific devices.

The only GPU intensive operations with `react-dropdowns` is the `translate(...)` operations done by PopperJS when the menu is positioned. These operations are consistent across all `<Menu>` usages so I'm not sure why it's only affecting `Select`, `Autocomplete`, and `Multiselect` components.

## Detail

To fix the issue I have removed the Popper computation reference while the menu is closed. This disables `react-popper` and seems to fix the issue across the problem browsers.

I attempted to disable GPU acceleration https://popper.js.org/docs/v1/#modifiers..computeStyle.gpuAcceleration with a Popper modifier, but that still didn't do the trick.

## Checklist

      designer as a reviewer)
      component

### DIFF
--- a/packages/dropdowns/src/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/Dropdown/Dropdown.tsx
@@ -212,22 +212,22 @@ const Dropdown: React.FunctionComponent<IDropdownProps> = props => {
         }}
         {...downshiftProps}
       >
-        {downshift => (
-          <DropdownContext.Provider
-            value={{
-              itemIndexRef,
-              previousItemRef,
-              previousIndexRef,
-              nextItemsHashRef,
-              popperReferenceElementRef,
-              selectedItems,
-              downshift: transformDownshift(downshift),
-              containsMultiselectRef
-            }}
-          >
-            {children}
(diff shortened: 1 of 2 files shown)


packages/dropdowns/src/Dropdown/Dropdown.tsx
```diff
@@ -212,22 +212,22 @@ const Dropdown: React.FunctionComponent<IDropdownProps> = props => {
...
-              popperReferenceElementRef,
-              selectedItems,
-              downshift: transformDownshift(downshift),
-              containsMultiselectRef
-            }}
-          >
-            {children}
-          </DropdownContext.Provider>
-        )}
+        {downshift => {
```
austin94: I've moved the context `value`'s to not be inline in an attempt to reduce unnecessary renders/positions.

PR #616, opened 2020-02-07
fix(notifications): Add ARIA role to Notification components

hzhu:
## Description

This pull request adds `role="alert"` to the `Alert`component and the `role="status"` to the `Notification` component.

## Detail

Recently, there was a report that `Alert` component content was not read aloud by screen readers. It required consumers to manually add a `role="alert"` to the underlying HTML element. This PR aims to bake in `role="alert"` and `role="status"` as a reasonable defaults.

## Demo

The demo for this branch can be found here: https://happy-hugle-9a77f5.netlify.com/notifications

## Verification

**Before:**

1) Navigate to the [published demo](https://henry-next.netlify.com/notifications/#basic) for `next`.
2) Turn on VoiceOver. 
3) Toggle between the different types of `Alert` and `Notification` components. 
4) Notice that the screen reader does not read the notification content aloud.

**After:**

1) Navigate to the [published demo](https://happy-hugle-9a77f5.netlify.com/notifications/#basic) for this branch.
1) Turn on VoiceOver. 
2) Toggle between the different types of `Alert` and `Notification` components. 
3) Notice that the screen reader says aloud the notification content aloud. For the `Alert`, the live region implicitly uses `assertive` and for the `Notification`, the live region implicitly uses `polite`.

## Checklist

### DIFF
--- a/packages/notifications/src/styled/StyledAlert.ts
+++ b/packages/notifications/src/styled/StyledAlert.ts
@@ -25,10 +25,11 @@ const colorStyles = (props: IStyledAlertProps & ThemeProps<DefaultTheme>) => css
 /**
  * Supports all `<div>` props
  */
-export const StyledAlert = styled(StyledBase).attrs({
+export const StyledAlert = styled(StyledBase).attrs(props => ({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION
-})<IStyledAlertProps>`
+  'data-garden-version': PACKAGE_VERSION,
+  role: props.role === undefined ? 'alert' : props.role
+}))<IStyledAlertProps>`
   ${colorStyles}
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
--- a/packages/notifications/src/styled/StyledNotification.ts
+++ b/packages/notifications/src/styled/StyledNotification.ts
@@ -54,10 +54,11 @@ const colorStyles = (props: IStyledNotificationProps & ThemeProps<DefaultTheme>)
 /**
  * Supports all `<div>` props
  */
-export const StyledNotification = styled(StyledBase).attrs({
+export const StyledNotification = styled(StyledBase).attrs(props => ({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION
-})<IStyledNotificationProps>`
+  'data-garden-version': PACKAGE_VERSION,
+  role: props.role === undefined ? 'status' : props.role
+}))<IStyledNotificationProps>`
   ${colorStyles}
 
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
--- a/packages/notifications/examples/basic.md
+++ b/packages/notifications/examples/basic.md
@@ -1,5 +1,7 @@
 ### Alert
 
+`<Alert>` components have an `role="alert"` by default and are announced by assistive
+technologies when they appear. The `role` attribute may be modified or removed if desired.
 All usages of `<Close />` must contain an `aria-label` or other assistive technique to have
 discernible text.
 
@@ -96,6 +98,9 @@ const initialState = { type: 'success', showClose: true };
 ```
 
 ### Notification
+
+`<Notification>` components have an `role="status"` by default and are announced by assistive
+technologies when they appear. The `role` attribute may be modified or removed if desired.
 
 The `<Paragraph>` component should be used to wrap multi-line content within a `<Notification>`.
 Otherwise, no wrapper is necessary. Usage of `<Close />` must contain an `aria-label` or other
--- a/packages/notifications/src/elements/Alert.spec.tsx
+++ b/packages/notifications/src/elements/Alert.spec.tsx
@@ -16,4 +16,22 @@ describe('Alert', () => {
 
     expect(container.firstChild).toBe(ref.current);
   });
+
+  it('has a default role attribute', () => {
+    const { container } = render(<Alert type="error" />);
+
+    expect(container.firstChild).toHaveAttribute('role', 'alert');
+  });
+
+  it('can have its role attribute modified', () => {
+    const { container } = render(<Alert type="info" role="status" />);
+
+    expect(container.firstChild).toHaveAttribute('role', 'status');
+  });
+
+  it('can have its role attribute removed', () => {
+    const { container } = render(<Alert type="info" role={null as any} />);
+
+    expect(container.firstChild).not.toHaveAttribute('role');
+  });
 });
--- a/packages/notifications/src/elements/Notification.spec.tsx
+++ b/packages/notifications/src/elements/Notification.spec.tsx
@@ -16,4 +16,22 @@ describe('Notification', () => {
 
     expect(container.firstChild).toBe(ref.current);
   });
+
+  it('has a default role attribute', () => {
+    const { container } = render(<Notification type="success" />);
+
+    expect(container.firstChild).toHaveAttribute('role', 'status');
+  });
+
+  it('can have its role attribute modified', () => {
+    const { container } = render(<Notification type="error" role="alert" />);
+
+    expect(container.firstChild).toHaveAttribute('role', 'alert');
+  });
+
+  it('can have its role attribute removed', () => {
+    const { container } = render(<Notification role={null as any} />);
+
+    expect(container.firstChild).not.toHaveAttribute('role');
+  });
 });


hzhu:
> maybe we should let consumers decide how/why they should modify live regions since it depends on the content?

I think the `Alert` component can have `role="alert"` by default since the use case for `Alert` component matches the description from MDN (requires immediate attention). Is there a case where consumers use `Alert`, but will not want the `role="alert"` attribute?

Maybe the `role="alert"` might not be useful for all `Notification` components. Maybe for this component, the `aria-live="polite"` can be used as a reasonable default? Would consumers ever use a `Notification` and not want the `aria-live="polite"` attribute? 

@austin94 what do you think? Do you have some suggestions?

packages/notifications/src/styled/StyledNotification.ts
```diff
@@ -56,7 +56,8 @@ const colorStyles = (props: IStyledNotificationProps & ThemeProps<DefaultTheme>)
  */
 export const StyledNotification = styled(StyledBase).attrs({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION
+  'data-garden-version': PACKAGE_VERSION,
+  role: 'alert' as string
```
jzempel: We know we want `role="alert"` for `Alerts`. Not so sure about `Notification` – should probably not interrupt the user (aka `aria-live="polite"`). Consult design (@ginnywood) for direction here. We should apply sensible defaults.

hzhu: > We know we want role="alert"

Sounds good, I will keep the `role="alert"` on `Alert`.

> Consult design (@ginnywood) for direction here.

@ginnywood could the `aria-live="polite"` attribute be used as a reasonable default for the `Notification` component since it doesn't interrupt the user? Would consumers ever use a `Notification` and not want the `aria-live="polite"` attribute? 

Alternatively, this component could be left without default for `aria-live`, and the consumers can be responsible for adding these attributes if desired depending on their use case. In this case, I can leave `Notifications` as is, and not make any modifications.

jzempel: Important to note that it can go the other way around. We apply a sensible default and the consumer is able to knock it out using `ariaLive={null}` or `undefined`. But the `attrs` on a Styled component would need to be structured to allow for that.

hzhu: I think the sensible defaults should be:

`Alert` -> `role="alert"`, implicitly has an implicit `aria-live` value of `assertive`.
`Status` -> `role="status"`, implicitly has an implicit `aria-live` value of `polite`.

With these additions, I don't see a need to add props or allow users to knock the `role` attributes off because these `role` attributes map nicely the purposes of the component. An `Alert` will always be used to communicate important information, while a `Notification` communicates information, but is not as urgent as what is communicated in an `Alert`.

@ginnywood @jzempel @austin94, what are your thoughts?

Resources:
https://www.w3.org/TR/wai-aria-1.1/#alert
https://www.w3.org/TR/wai-aria-1.1/#status

hzhu: > Add the ability to knock-out these roles and add testing to prove it is possible. We can't imagine all cases. Providing reasonable defaults is great, but we need to allow users to take control (role={null}), if necessary.

@jzempel thanks for the comments. I have thought a little bit more about the `role` attribute.

I think the `Alert` and `Notification` does not have to foresee all use cases because these components have specific use cases. The `Alert` component use case is to communicate urgent and important information. The `Notification` component use case is similar to `Alert`, but the information is less critical.

If a consumer uses an `Alert` component and is seeking to knock out the `role="alert"` attribute, then an `Alert` component might not be the best design pattern for their implementation. I want to propose that Garden's `Alert` and `Notification` does not provide a  `prop` to change or remove the `role` attribute.

I think other design systems have opted not to allow consumers to change the `role` attribute for the reasons described above. Specifically, I'm looking at [Evergreen's Alert](https://evergreen.segment.com/components/alert), [Polaris's Banner](https://polaris.shopify.com/components/feedback-indicators/banner), and [Material UI's Alert Banner](https://material-ui.com/components/alert/). Polaris makes some opinions on how to treat different types of alerts (`success`, `info`, vs. `critical`) with `role`. Still, the component will always have a `role` attribute that the consumer cannot modify.

jzempel: 😕 add `role={null}` to Evergreen. The "role" attribute goes away.

hzhu: @jzempel gotcha, I think an earlier mention of `ariaLive` got me thinking of a custom React `prop`. 

I pushed a commit and made the proper adjustments so the `role` can be overwritten. I opted not to add additional text in the `Alert` and `Notification` examples about modifying the `role` attribute because there doesn't seem to be a common use case for doing so. The native `role` attribute can now be overwritten.

allisonacs:
@ginnywood consider speaking with content strategy on this. I know V is looking at our `notifications` presently.

Also folks, please consider that we have things like the `info` alert, which I don't think would need an interruption.

jzempel:
@allisonacs the API should be as simple as “Alerts do X; Notifications do Y” and then let the user override when needed. We don’t know that infos shouldn’t alert, for example. Might be super important info ;)

packages/notifications/src/styled/StyledAlert.ts
```diff
@@ -25,10 +25,12 @@ const colorStyles = (props: IStyledAlertProps & ThemeProps<DefaultTheme>) => css
...
  * Supports all `<div>` props
  */
-export const StyledAlert = styled(StyledBase).attrs({
+export const StyledAlert = styled(StyledBase).attrs(props => ({
   'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION
-})<IStyledAlertProps>`
+  'data-garden-version': PACKAGE_VERSION,
+  role: 'alert' as string,
+  ...props
```
jzempel: Pls don’t spread props as attrs. This is bad for current or future HTML. Logically apply the role, instead.

hzhu: I thought props got forwarded under the hood by styled-components, so I didn't think spreading props into attrs would hurt. I thought only HTML attributes get applied. 

Could you give an example of issues that would arise when props and spread into attrs?

austin94: > I thought only HTML attributes get applied

This is true by default, but there are instances where it doesn't work as expected https://github.com/styled-components/styled-components/issues/439

This could also influence type-inference with TypeScript depending on custom props.

It may just be safer to use `role: (props.role || 'alert') as string`

hzhu: I see, thanks for the clarification and suggestion @austin94. I'll go ahead and use `role: (props.role || 'alert') as string`.

austin94: I just realized my original suggestion might not work 100% of the time. Does this work with truthy/falsy? We may want to use `props.role === undefined ? 'alert' : props.role`

ginnywood:
I’ve been documenting `Alerts` and `Notifications`, and this is how I’m thinking about it:
- `Notifications` are typically confirmations, and should not interrupt by default
- Danger and warning `Alerts` are critical information, and should interrupt by default. However, info and success are unlikely to be. I’d be comfortable with either having `Alerts` interrupt by default, allowing overrides, or not setting a default and letting users choose.

packages/notifications/src/styled/StyledAlert.ts
jzempel: Yeah, @austin94 that's the treatment.

@hzhu in the current code, we see `type` bleeding through as an attribute to the HTML. Austin's code will fix that.

In addition please add documentation and testing per my comments above.

hzhu:
> I’ve been documenting `Alerts` and `Notifications`, and this is how I’m thinking about it:
> 
> * `Notifications` are typically confirmations, and should not interrupt by default
> * Danger and warning `Alerts` are critical information, and should interrupt by default. However, info and success are unlikely to be. I’d be comfortable with either having `Alerts` interrupt by default, allowing overrides, or not setting a default and letting users choose.

This sounds good! To recap, this is what the current PR proposes:

`Alert` has a default `role="alert"` which implicitly uses `aria-live="assertive"`. Assistive technologies will interrupt, communicate the contents of the `Alert`, before proceeding. The `role` can be removed or modified for certain use cases, such as an "info" `Alert` with content that is not urgent or critical. For example: `<Alert type="info" role="status">`.

`Notification` has a default `role="status"` which implicitly uses `aria-live="polite"`. Assistive technologies will communicate the current contents, before communicating the `Notification`'s contents. The `role` can be removed or modified for certain use cases.

PR #619, opened 2020-02-10
feat(buttons): remove react-utilities dependency and deprecate package

austin94:
## Description

This PR deprecates `@zendeskgarden/conatiner-utilities` and includes a new warning message for consumers that consume it. We should be able to remove this package during `v9` to allow people to upgrade.

## Detail

The only package where these utilities were in `react-buttons`. Specifically the `ButtonGroup` component.  This component was still using our legacy `Children.map()` approach to applying nested props.

To remove this dependency I migrated the components to use custom context hooks to consume these prop-getters. The `ButtonGroup` can now support nesting and `styled(Button)` usages.

## Checklist

      designer as a reviewer)
      component~

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -265,3 +265,8 @@ New features/fixes include:
   - previously `<Code type="red">`
   - currently `<Code hue="red">`
 - rename `monospace` -> `isMonospace` prop
+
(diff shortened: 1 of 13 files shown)


austin94:
> Recommend removing link from demo/index.html

I've removed all `react-utilities` links from the demo index page as well as the root README.

PR #620, opened 2020-02-10
feat(datepickers): remove css-menus dependency

austin94:
## Breaking changes

- Rename `animate` prop to `isAnimated` to align with other packages

## Description

This PR brings the styling logic from `react-dropdowns` to `react-datepickers` and removes the `css-forms` dependency. This is the last remaining css-component package 🎉 

Available for review at: https://austin-next.netlify.com/datepickers/

## Checklist

      designer as a reviewer)
      component~


packages/datepickers/src/styled/StyledMenu.tsx
```diff
@@ -93,53 +123,51 @@ StyledMenuView.defaultProps = {
...
 }
 
 const StyledMenuWrapper = styled.div<IStyledMenuWrapperProps>`
-  ${retrieveMenuMargin};
+  transition: ${props =>
+    props.isHidden && props.isAnimated && 'opacity .2s ease-in-out, .2s visibility 0s linear'};
+  visibility: ${props => props.isHidden && 'hidden'};
+  opacity: ${props => props.isHidden && '0'};
+  z-index: ${props => props.zIndex};
+  font-size: 0.01px; /* [1] */
```
hzhu: I noticed that `[1]` reference doesn't point to a description. Should there be a description for this `[1]`?

austin94: Looks like I lost some comments during the refactor. I've added this styling comment back in.

packages/datepickers/src/elements/Datepicker/Datepicker.tsx
```diff
@@ -192,45 +190,39 @@ const Datepicker: React.FunctionComponent<IDatepickerProps & ThemeProps<DefaultT
...
-            modifiers={popperModifiers}
-            // Disable position updating on scroll events while menu is closed
-            eventsEnabled={state.isOpen && eventsEnabled}
-          >
-            {({ ref, style, scheduleUpdate, placement: currentPlacement }) => {
-              scheduleUpdateRef.current = scheduleUpdate;
 
-              const popperStyle = { ...style, zIndex };
+        <Popper
+          placement={popperPlacement as any}
```
hzhu: I think the type cast may not be needed because `popperPlacement` is a result of functions that return `POPPER_PLACEMENT`. Would you be able to confirm if that's the case?

austin94: I've removed this cast.

packages/datepickers/src/styled/StyledMenu.tsx
```diff
@@ -6,80 +6,110 @@
...
+  display: inline-block;
+  position: relative; /* [1] */
+  margin: 0; /* [2] */
+  box-sizing: border-box;
+  border: ${props => `${props.theme.borders.sm} ${getColor('neutralHue', 300, props.theme)}`};
+  border-radius: ${props => props.theme.borderRadii.md};
+  box-shadow: ${props =>
+    props.theme.shadows.lg(
+      `${props.theme.space.base * 5}px`,
+      '30px',
```
jzempel: Ideally this would be `${props.theme.space.base * 7.5}px` (missed it on `dropdowns`), although not super important to change.

jzempel: On the other hand, maybe we want this to be fixed?

austin94: I vote for keeping this fixed as a base change would could dramatically change a shadow. We may want to look at consolidating these shadows under a themed value/utility?

PR #627, opened 2020-02-13
feat(build): introduce rollup and ES module builds for all packages

austin94:
## BREAKING CHANGE

UMD bundles are no longer exported. These can be added in the future if deemed necessary.

## Description

This PR replaces Webpack with Rollup for our build process. This is similar to https://github.com/zendeskgarden/react-containers/pull/158 with some additional configuration needed to for `react-components`

## Detail

Rollup helped identify several dependency issues that have crept up while we have been developing in the `next` branch. To help review I've broken it into 3 commits:

1. Migrate from Webpack to Rollup
  - The first commit only includes the addition of the rollup config and other utilities
  - We no longer build UMD bundles as there are no known consumers at this time
  - This config dynamically computes `external` packages based off of the `peerDependencies` and `dependencies` of a package.
  - It also includes `svgr` and `svgo` configurations for our SVG usages.
2. Update packages for Rollup configuration
  -  The second commit includes dependency and `package.json` updates for all packages
  -  This includes the new `.size-snapshot` files and `yarn.lock` files for reproducible builds (bundle size comparisons)
3. Fix package issues for ES modules
  - Rollup was able to find several circular dependencies in our `styled` components. I refactored these elements to tree-shake correctly
  - Many packages were missing `@zendeskgarden/svg-icons` dependencies. I've added these to allow the builds to pass.
  - Also, now that we support `es` imports we no longer need to use the `polished/lib/*` imports. I've refactored these use `polished` directly.

## Checklist

      designer as a reviewer)~
      component~


package.json
```diff
@@ -3,9 +3,9 @@
   "name": "@zendeskgarden/react-components",
   "version": "0.0.0",
   "scripts": {
-    "build": "lerna run build --stream",
+    "build": "lerna run build --stream --concurrency 3",
```
austin94: Rollup is overall faster than webpack, but the default concurrency of "all the packages" is really bogging it down. Building 3 packages at a time seems like the sweet spot of `~55s`

package.json
```diff
@@ -3,9 +3,9 @@
...
     "build:analyze": "ANALYZE_BUNDLE=true yarn build:single",
-    "build:demo": "yarn build && lerna run build:demo --stream",
+    "build:demo": "lerna run build:demo --stream",
```
austin94: We don't need to rebuild the packages as it's already accomplished during `yarn install`

package.json
```diff
@@ -23,7 +23,8 @@
     "start:demo": "yarn build:demo && live-server demo",
     "tag": "utils/scripts/version-next.sh",
     "test": "yarn test:all --watch",
-    "test:all": "jest --config=utils/test/jest.config.js"
+    "test:all": "jest --config=utils/test/jest.config.js",
+    "version": "yarn run build && git add -A"
```
austin94: Since we inject the version number into our bundles this makes our bundle size change _after_ we version the package. This lifecycle hook ensures that all update `.size-snapshot` files are included.

.huskyrc
```diff
@@ -1,5 +1,5 @@
 {
   "hooks": {
-    "pre-commit": "lint-staged"
+    "pre-commit": "lint-staged && yarn lerna run build --since next"
```
jzempel: `next` won't be stable here. Is there a solution here? https://stackoverflow.com/questions/3161204/how-to-find-the-nearest-parent-of-a-git-branch

packages/theming/package.json
```diff
@@ -18,11 +19,12 @@
...
     "start": "../../utils/scripts/start.sh"
   },
+  "sideEffects": false,
   "types": "dist/typings/index.d.ts",
   "dependencies": {
-    "@zendeskgarden/container-focusvisible": "^0.4.0",
-    "@zendeskgarden/container-utilities": "^0.5.0",
-    "polished": "^3.4.1"
+    "@zendeskgarden/container-focusvisible": "0.4.0",
+    "@zendeskgarden/container-utilities": "0.5.0",
```
jzempel: Should `^` be used here?

.travis.yml
```diff
@@ -20,7 +20,6 @@ script:
   - yarn lint
   - yarn tsc
   - yarn format
-  - yarn bundlesize --config bundlesize.config.js
```
hzhu: I didn't notice a bundlesize check for this PR. Would bundle size checks still be a part of the CI build process? Does it still use `bundlesize-store.now.sh`?

How does this change the developer workflow? Is the new workflow to manually compare new snapshot to previous snapshot on each commit rather than relying on `yarn bundlesize --config bundlesize.config.js`?

package.json
jzempel: I'm not exactly following your comment. But this looks like it is going to add CI ⌛️ to the publish build process. Can we get rid of `version` by increasing the snapshot threshold so that the difference between `PACKAGE_VERSION` and actual version isn't considered to be a problem?

packages/theming/package.json
austin94: Updated.

.huskyrc
austin94: I'm not sure we can inject this value since we're in an `rc` file?

We already hard-code this value to tell lerna the release branch. We would only have to update this once we move it to `master`.

austin94:
> I'm not exactly following your comment. But this looks like it is going to add CI ⌛️ to the publish build process. Can we get rid of version by increasing the snapshot threshold so that the difference between PACKAGE_VERSION and actual version isn't considered to be a problem?

@jzempel This is only ran locally when we call `yarn run tag`. It shouldn't be ran in CI since we're only calling `lerna publish`. 

Adding a threshold would allow the build to pass, but it still creates new `.size-snapshot` files which would dirty the next commit someone has to make.

package.json
austin94: This is only ran locally when we call yarn run tag. It shouldn't be ran in CI since we're only calling lerna publish.

Adding a threshold would allow the build to pass, but it still creates new .size-snapshot files which would dirty the next commit someone has to make.

.travis.yml
austin94: The bundle size check is handled within Rollup when the `CI` environment variable is set. This follows the same process as https://github.com/zendeskgarden/react-containers/pull/158

The new workflow is that whenever a package is updated, the new size is reflected in the updated `.size-snapshot` file. These files are committed along with the PR so we can understand the bundle implications for every change.

austin94:
While messing with the [rollup perf option](https://rollupjs.org/guide/en/#perf) I found that our rollup plugin was using an older version of TS. I've updated it to use our matching version and disabled type checks since we already run `tsc` during our linting stage. It's now noticeably faster in Travis.

.travis.yml
hzhu: Awesome, sounds great!

package.json
jzempel: Makes sense. Thanks for the clarification.

PR #645, opened 2020-02-20
chore(lint): update eslint rules

austin94:
## Description

This PR is an audit of our existing ESLint rules. The only change is that we now require `displayName`s for all React components. This was disabled previously, but should be best practice for us going forward.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/elements/Body.tsx
+++ b/packages/modals/src/elements/Body.tsx
@@ -19,3 +19,5 @@ export const Body = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivEleme
     return <StyledBody ref={ref} {...getContentProps(props)} />;
   }
 );
+
+Body.displayName = 'Body';
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,16 +12,15 @@
   ],
   "settings": {
     "react": {
-      "version": "16.4.2"
+      "version": "detect"
     }
   },
   "globals": {
     "jest": true,
     "BASE_PATH_NAME": true,
-    "PACKAGE_VERSION": true,
-    "COMPONENT_IDS": true
+    "PACKAGE_VERSION": true
   },
-  "plugins": ["prettier", "react", "jest", "jsx-a11y", "react-hooks", "@typescript-eslint", "garden-local"],
+  "plugins": ["react-hooks", "garden-local"],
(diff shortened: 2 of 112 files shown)


packages/modals/src/elements/Body.tsx
```diff
@@ -19,3 +19,5 @@ export const Body = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivEleme
     return <StyledBody ref={ref} {...getContentProps(props)} />;
   }
 );
+
+Body.displayName = 'Body';
```
hzhu: Could you clarify why some components, such as `Modal.tsx` do not require a `displayName`?

austin94: Most `HOC`s [apply their own display names](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) to improve debugging. This is the case for `withTheme`.

Eslint is also limited in how it can detect React components https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md#about-component-detection

austin94: Just realized that `Modal` no longer uses the `withTheme` HOC. I'll look into this more deeply.

austin94: `forwardRef` has some [unique display name rules](https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools).

To help with dev tool identification I'm going to add named functions to ensure components are shown correctly

image

austin94: It turns out `forwardRef` was being identified in most areas successfully. The only "missing" display names were in `Modal` and `Tooltip` where the component returns a `createPortal()` call. I've added display names to both of these components now.

austin94:
> Looks like there should be a new `require-display-name` custom rule added for `eslint` under /utils.

There is a lot of complexity in determining whether a `displayName` is valid. https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md only missed 2 components due to our unique `createPortal` usage. We might want to stick with the existing solution to avoid complications as React introduces new APIs as well.

> Please add `displayName` to the `.template` package.

Updated! 👍

PR #652, opened 2020-02-26
docs(modals): include a11y note on restoring focus

hzhu:
docs(modals): include a11y note on restoring focus

## Description

This pull request adds a sentence about Garden's restoring focus capability, and potentially the need for developers to manage keyboard focus in certain circumstances.

## Detail

Recently Garden's `modals` package was updated to returns focus back to the trigger element. Specifically, the `modals` package uses one of `useFocusJail`'s newest features to [return the keyboard focus](https://github.com/zendeskgarden/react-containers/blob/master/packages/focusjail/src/useFocusJail.ts#L130) back to the trigger element.

### DIFF
--- a/packages/modals/examples/basic.md
+++ b/packages/modals/examples/basic.md
@@ -108,7 +108,9 @@ const onModalClose = () => setState({ isModalVisible: false });
 
 The `Modal` component uses the [useFocusJail()](https://www.npmjs.com/package/@zendeskgarden/container-focusjail)
 hook internally to limit focus and keyboard navigation
-to the Modal content.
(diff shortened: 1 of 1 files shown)


austin94:
Does the `Modal` component accept a `restoreFocus` prop that allows consumers to disable this functionality? I think we still need to pass that prop to the `useModal` hook.

hzhu:
> Does the `Modal` component accept a `restoreFocus` prop that allows consumers to disable this functionality? I think we still need to pass that prop to the `useModal` hook.

I had planned to have that in a separate PR.

PR #653, opened 2020-02-26
chore(markdown): include package examples in lint:markdown

hzhu:
## Description

This pull request updates the `lint:markdown` glob patterns to include markdown files that were added in v8 under `packages/*/examples/*md`. 

* Removed the `packages/*/src/*.md` pattern because no markdown file exists under `src`.
* Updated `lists.md` which fixes markdown line length issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "yarn lint:css && yarn lint:js && yarn lint:markdown",
     "lint:css": "stylelint \"packages/*/src/**/*.{js,ts,tsx}\" \"!packages/*/src/**/*.spec.{js,ts,tsx}\" && stylelint \"utils/**/*.{js,ts,tsx}\"",
     "lint:js": "eslint packages/*/src/ utils/ --ext js,ts,tsx --max-warnings 0",
-    "lint:markdown": "markdownlint README.md docs/*.md packages/*/src/**/*.md packages/*/src/*.md packages/*/README.md",
+    "lint:markdown": "markdownlint README.md docs/*.md packages/*/examples/*.md packages/*/src/**/*.md packages/*/README.md",
     "new": "utils/scripts/create-package.js",
     "prepare": "yarn build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "yarn lint:css && yarn lint:js && yarn lint:markdown",
     "lint:css": "stylelint \"packages/*/src/**/*.{js,ts,tsx}\" \"!packages/*/src/**/*.spec.{js,ts,tsx}\" && stylelint \"utils/**/*.{js,ts,tsx}\"",
     "lint:js": "eslint packages/*/src/ utils/ --ext js,ts,tsx --max-warnings 0",
-    "lint:markdown": "markdownlint README.md docs/*.md packages/*/src/**/*.md packages/*/src/*.md packages/*/README.md",
+    "lint:markdown": "markdownlint README.md docs/*.md packages/*/examples/*.md packages/*/README.md",
```
jzempel: What about all of the `*.example.md` files under `src` folders? Don't we want to lint those? See https://github.com/zendeskgarden/react-components/tree/master/packages/theming/src/utils

hzhu: > `arrowStyles.example.md` and `withTheme.example.md` under the `react-theming` package is not currently linted (both in `master` and in this branch). Eventually, these example files may be moved into `/examples` directory to be consistent with other packages. If that is agreed upon, I could make that change in this PR.

Originally was going to put that above statement in the PR description. Should I go ahead and migrate the examples? Or instead just add a glob pattern for it?

jzempel: The intent is that these examples stay connected to their API sheets. While we're still in styleguidist land, they need to stay where they are.

Add glob to pick them up for linting.

hzhu: > Restore glob to pick them up for linting.

I'll add a new glob pattern because these markdown files are not picked up in `master` branch.

hzhu: Edit: My mistake it was removed, and I added it back 👍.

PR #654, opened 2020-02-27
feat(forms): introduce new Tiles component

austin94:
## Description

This PR introduces the new `Tiles` pattern. Semantically this is similar to a [Radio Group](https://www.w3.org/TR/wai-aria-practices/#radiobutton), but this implementation uses native `<input type="radio" />` elements rather than roving tabIndex.

### Usage

Consumers can treat this as a controlled or uncontrolled input. It also allows more traditional `<form>` usages if needed.

### API

Common usage:

```jsx
<Tiles name="form-name" onChange={...} value={...}>
  <Tiles.Tile value="item-1">
    <Tiles.Icon>...</Tiles.Icon>
    <Tiles.Label>label</Tiles.Label>
    <Tiles.Description>description</Tiles.Description>
  </Tiles.Tile>
</Tiles>
```

### Layout and Variants

`Tiles` are currently used in a variety of sizes and layouts with only some needing mobile-responsiveness. Rather than bake all of these layouts into the component consumers can use our `<Grid>` component or their own implementations to define the column layouts.

I have included an `isVertical` layout that helps with modal and block usages.

## Preview

image

https://austin-next.netlify.com/forms/#tiles

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/examples/tiles.md
+++ b/packages/forms/examples/tiles.md
@@ -1,6 +1,79 @@
 The `Tiles` component is implemented using the [W3C Radio Group pattern](https://www.w3.org/TR/wai-aria-practices/#radiobutton).
 Each `Tile` may be treated as an individual radio input.
 
+### Components
+
+- `Tiles.Tile` Accepts all `div` attributes and the following props:
+  - [`value`] The value of the `input`
+  - [`disabled`] Whether the `Tile` is disabled
+- `Tiles.Label` Accepts all `span` attributes
+- `Tiles.Description` Accepts all `span` attributes
+
+### Layout
+
+The `Tiles` component provides no layout or responsive styling. For a responsive layout
+we use our [@zendeskgarden/react-grid package](https://garden.zendesk.com/react-components/grid/)
+in the examples below.
+
+```jsx
+const DocumentIcon = require('@zendeskgarden/svg-icons/src/16/file-document-stroke.svg').default;
+const ImageIcon = require('@zendeskgarden/svg-icons/src/16/file-image-stroke.svg').default;
+const PresentationIcon = require('@zendeskgarden/svg-icons/src/16/file-presentation-stroke.svg')
+  .default;
+const SpreadsheetIcon = require('@zendeskgarden/svg-icons/src/16/file-spreadsheet-stroke.svg')
+  .default;
+const FolderOpenIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
+
+<Tiles name="basic-example" aria-label="Example radio group selection" isStacked={state.isStacked}>
+  <Grid>
+    <Row>
+      <Col lg={3} md={4} sm={6}>
+        <Tiles.Tile value="document">
+          <Tiles.Icon>
+            <DocumentIcon />
+          </Tiles.Icon>
+          <Tiles.Label>Document</Tiles.Label>
+        </Tiles.Tile>
+      </Col>
+      <Col lg={3} md={4} sm={6}>
+        <Tiles.Tile value="image">
+          <Tiles.Icon>
+            <ImageIcon />
+          </Tiles.Icon>
+          <Tiles.Label>Image</Tiles.Label>
+        </Tiles.Tile>
+      </Col>
+      <Col lg={3} md={4} sm={6}>
+        <Tiles.Tile value="presentation">
+          <Tiles.Icon>
+            <PresentationIcon />
+          </Tiles.Icon>
+          <Tiles.Label>Presentation</Tiles.Label>
+        </Tiles.Tile>
+      </Col>
+      <Col lg={3} md={4} sm={6}>
+        <Tiles.Tile value="spreadsheet">
+          <Tiles.Icon>
+            <SpreadsheetIcon />
+          </Tiles.Icon>
+          <Tiles.Label>Spreadsheet</Tiles.Label>
+        </Tiles.Tile>
+      </Col>
+      <Col lg={3} md={4} sm={6}>
+        <Tiles.Tile value="folder">
+          <Tiles.Icon>
+            <FolderOpenIcon />
+          </Tiles.Icon>
+          <Tiles.Label>Folder</Tiles.Label>
+        </Tiles.Tile>
+      </Col>
+    </Row>
+  </Grid>
+</Tiles>;
+```
+
+### Advanced Usage
+
 ```jsx
 const { Well } = require('@zendeskgarden/react-notifications/src');
 const ChevronIcon = require('@zendeskgarden/svg-icons/src/16/chevron-box-stroke.svg').default;
@@ -13,7 +86,7 @@ const NumberIcon = require('@zendeskgarden/svg-icons/src/16/number-stroke.svg').
 const DecimalIcon = require('@zendeskgarden/svg-icons/src/16/decimal-stroke.svg').default;
 const CalendarIcon = require('@zendeskgarden/svg-icons/src/16/calendar-stroke.svg').default;
 const CreditCardIcon = require('@zendeskgarden/svg-icons/src/16/credit-card-stroke.svg').default;
-const AsterickIcon = require('@zendeskgarden/svg-icons/src/16/asterisk-stroke.svg').default;
+const AsteriskIcon = require('@zendeskgarden/svg-icons/src/16/asterisk-stroke.svg').default;
 
 const items = [
   { value: 'dropdown', label: 'Dropdown', icon: <ChevronIcon /> },
@@ -25,7 +98,7 @@ const items = [
   { value: 'decimal', label: 'Decimal', icon: <DecimalIcon /> },
   { value: 'date', label: 'Date', icon: <CalendarIcon /> },
   { value: 'credit-card', label: 'Credit card', icon: <CreditCardIcon /> },
-  { value: 'wildcard', label: 'Wildcard', icon: <AsterickIcon /> }
+  { value: 'wildcard', label: 'Wildcard', icon: <AsteriskIcon /> }
 ];
 
 initialState = {
--- a/packages/forms/examples/tiles.md
+++ b/packages/forms/examples/tiles.md
@@ -6,6 +6,7 @@ Each `Tile` may be treated as an individual radio input.
 - `Tiles.Tile` Accepts all `div` attributes and the following props:
   - [`value`] The value of the `input`
   - [`disabled`] Whether the `Tile` is disabled
+- `Tiles.Icon` Accepts all `span` attributes
 - `Tiles.Label` Accepts all `span` attributes
 - `Tiles.Description` Accepts all `span` attributes
 
@@ -24,7 +25,7 @@ const SpreadsheetIcon = require('@zendeskgarden/svg-icons/src/16/file-spreadshee
   .default;
 const FolderOpenIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
 
-<Tiles name="basic-example" aria-label="Example radio group selection" isStacked={state.isStacked}>
+<Tiles name="basic-example" aria-label="Example radio group selection">
   <Grid>
     <Row>
       <Col lg={3} md={4} sm={6}>
@@ -102,7 +103,7 @@ const items = [
 ];
 
 initialState = {
-  isStacked: false,
+  isCentered: true,
   disabled: false
 };
 
@@ -113,10 +114,10 @@ initialState = {
         <Col>
           <Field>
             <Toggle
-              checked={!!state.isStacked}
-              onChange={event => setState({ isStacked: event.target.checked })}
+              checked={!!state.isCentered}
+              onChange={event => setState({ isCentered: event.target.checked })}
             >
-              <Label>Stacked</Label>
+              <Label>Centered</Label>
             </Toggle>
           </Field>
         </Col>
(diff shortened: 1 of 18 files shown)


ginnywood:
One other thing: in a vertical Tile, the space between the icon and the text should be 20px. Looks like it is 16 right now.

austin94:
@ginnywood I've addressed the following feedback:

> My specs were wrong on borders for hover and active states

I've updated these values to match radios/checkboxes

> Icons should remain grey, but get progressively darker on hover and active
> Text color should always match icon color

I've brought over the IconButton styling for this.

> Selected Tile has hover and active states too—should match a Radio/Checkbox

These now match Radio/Checkbox

> Disabled default state background should match disabled Text-input/Dropdown
> Disabled selected state background match disabled Checkbox/Radio

I've combined these two styles. Let me know what you think.

> Instead of “vertical”, I propose “stacked”

I think this makes sense considering all row/column layout is handled by the consumer.

I've published the new changes at https://austin-next.netlify.com/forms/#tiles

packages/forms/examples/tiles.md
```diff
@@ -0,0 +1,91 @@
...
+const CheckboxDoubleIcon = require('@zendeskgarden/svg-icons/src/16/check-box-double-stroke.svg')
+  .default;
+const TextIcon = require('@zendeskgarden/svg-icons/src/16/text-stroke.svg').default;
+const MultilineIcon = require('@zendeskgarden/svg-icons/src/16/multiline-stroke.svg').default;
+const CheckIcon = require('@zendeskgarden/svg-icons/src/16/check-lg-stroke.svg').default;
+const NumberIcon = require('@zendeskgarden/svg-icons/src/16/number-stroke.svg').default;
+const DecimalIcon = require('@zendeskgarden/svg-icons/src/16/decimal-stroke.svg').default;
+const CalendarIcon = require('@zendeskgarden/svg-icons/src/16/calendar-stroke.svg').default;
+const CreditCardIcon = require('@zendeskgarden/svg-icons/src/16/credit-card-stroke.svg').default;
+const AsterickIcon = require('@zendeskgarden/svg-icons/src/16/asterisk-stroke.svg').default;
```
jzempel: * `AsteriskIcon` :wink: #rickroll

packages/forms/src/styled/tiles/StyledTileLabel.ts
```diff
@@ -0,0 +1,39 @@
...
+  overflow: hidden;
+  text-align: ${props => !props.isStacked && 'center'};
+  text-overflow: ellipsis;
+  line-height: ${props => props.theme.space.base * 5}px;
+  white-space: nowrap;
+  font-size: ${props => props.theme.fontSizes.md};
+  font-weight: ${props => props.theme.fontWeights.semibold};
+  /* stylelint-disable-next-line property-no-unknown */
+  margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props =>
+  props.isStacked && props.theme.space.base * 13}px;
```
jzempel: Technically, I think the intent here is to represent 20px of space between the icon and the label. So the more accurate calculation would be `space.base * 5 + iconSize`, where the icon size is `iconSizes.md * 2`.

packages/forms/src/styled/tiles/StyledTileIcon.ts
```diff
@@ -0,0 +1,47 @@
...
+export const StyledTileIcon = styled.span.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledTileIconProps>`
+  display: block;
+  position: ${props => props.isStacked && 'absolute'};
+  top: ${props => props.theme.space.base * 6}px;
+  text-align: center;
+  line-height: 0;
+  ${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 5}px;
```
jzempel: Neither `right/left` or `top` is applicable for non-stacked tiles. Feels like the absolute positioning could all be moved to `sizeStyles` and simplified by using local variables.

packages/forms/src/styled/tiles/StyledTileDescription.ts
```diff
@@ -0,0 +1,35 @@
...
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledTileDescriptionProps>`
+  display: block;
+  margin-top: ${props => props.theme.space.base}px;
+  text-align: ${props => !props.isStacked && 'center'};
+  line-height: ${props => props.theme.space.base * 4}px;
+  font-size: ${props => props.theme.fontSizes.sm};
+  /* stylelint-disable-next-line property-no-unknown */
+  margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props =>
+  props.isStacked && props.theme.space.base * 13}px;
```
jzempel: Same comment here re: the `iconSize` + `margin` calculation.

packages/forms/src/styled/tiles/StyledTileDescription.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export const StyledTileDescription = styled.span.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledTileDescriptionProps>`
+  display: block;
+  margin-top: ${props => props.theme.space.base}px;
+  text-align: ${props => !props.isStacked && 'center'};
+  line-height: ${props => props.theme.space.base * 4}px;
```
jzempel: Should be using the `getLineHeight` utility here.

packages/forms/src/styled/tiles/StyledTileLabel.ts
```diff
@@ -0,0 +1,39 @@
...
+export const StyledTileLabel = styled.span.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledTileLabelProps>`
+  display: block;
+  margin-top: ${props => (props.isStacked ? 0 : props.theme.space.base * 2)}px;
+  overflow: hidden;
+  text-align: ${props => !props.isStacked && 'center'};
+  text-overflow: ellipsis;
+  line-height: ${props => props.theme.space.base * 5}px;
```
jzempel: Should be using the `getLineHeight` utility here.

packages/forms/src/styled/tiles/StyledTile.ts
```diff
@@ -0,0 +1,145 @@
...
+  padding: 0;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+  white-space: nowrap;
+`;
+
+StyledTileInput.defaultProps = {
+  theme: DEFAULT_THEME
+};
```
jzempel: Would prefer to see this in its own file with the rules for view components applied:
- `data-garden-id` & `data-garden-package` attrs
- theming override capabilities via `retrieveComponentStyles`

packages/forms/src/elements/tiles/components/Tile.tsx
```diff
@@ -0,0 +1,83 @@
...
+        {...inputProps}
+        disabled={disabled}
+        value={value}
+        onBlur={() => setIsFocused(false)}
+        onFocus={e => {
+          e.persist();
+
+          /**
+           * The focus-visible attribute provided by `react-theming` is
+           * unable to persit to the `<label>`. This manually applies focus
```
jzempel: *persist

packages/forms/examples/tiles.md
ginnywood: Yes, I’d initially designed this to be used for checkboxes too. We don’t currently have a use case, but from a design perspective we’re approved and prepared to introduce that concept whenever the engineering effort makes sense.

austin94: Updated.

packages/forms/src/styled/tiles/StyledTileDescription.ts
austin94: Updated.

packages/forms/src/styled/tiles/StyledTileLabel.ts
austin94: Updated.

packages/forms/src/styled/tiles/StyledTile.ts
austin94: This is a hidden input that is not visually show and only included for a11y/interaction. I'm not sure why this would need to be themed?

packages/forms/src/elements/tiles/components/Tile.tsx
austin94: Updated.

packages/forms/src/styled/tiles/StyledTileLabel.ts
austin94: Updated.

packages/forms/src/styled/tiles/StyledTileDescription.ts
austin94: Updated.

packages/forms/src/styled/tiles/StyledTileIcon.ts
austin94: I've moved this logic to the existing `sizeStyles` utility to help clarify.

packages/forms/src/elements/tiles/components/Tile.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import React, { HTMLAttributes, useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import { useTilesContext } from '../../../utils/useTilesContext';
+import { StyledTile, StyledTileInput } from '../../../styled';
+
+interface ITileProps {
+  /** The value of the input */
+  value?: string;
+  /** Whether the radio is disabled */
```
jzempel: Should "radio" be "tile"?

packages/forms/src/elements/tiles/Tiles.tsx
```diff
@@ -0,0 +1,91 @@
...
+  extends ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>> {
+  Tile: typeof Tile;
+  Description: typeof Description;
+  Icon: typeof Icon;
+  Label: typeof Label;
+}
+
+/* eslint-disable react/display-name */
+/**
+ * Accepts all `<div>` attributes
```
jzempel: Without improved documentation here, it is very hard to understand how to use the `Tiles` component just by viewing the API. Unless there is a way to make Styleguidist pick up the `Tiles.Tile` etc. components below, there should probably be a minimal usage example along with showing how it should be incorporated with the `Grid`.

packages/forms/src/elements/tiles/components/Tile.tsx
austin94: Updated.

packages/forms/src/elements/tiles/Tiles.tsx
austin94: I've added a new basic example and some documentation regarding layout and the static property components.

packages/forms/src/styled/tiles/StyledTile.ts
jzempel: I still think the clearer pattern is to separate styled components, and to not assert whether there is theming benefit or not. But YMMV – I'll go with your preference here.

packages/forms/examples/tiles.md
```diff
@@ -0,0 +1,164 @@
+The `Tiles` component is implemented using the [W3C Radio Group pattern](https://www.w3.org/TR/wai-aria-practices/#radiobutton).
+Each `Tile` may be treated as an individual radio input.
+
+### Components
+
+- `Tiles.Tile` Accepts all `div` attributes and the following props:
+  - [`value`] The value of the `input`
+  - [`disabled`] Whether the `Tile` is disabled
+- `Tiles.Label` Accepts all `span` attributes
+- `Tiles.Description` Accepts all `span` attributes
```
jzempel: What about `Tiles.Icon`?

austin94: Added!

packages/forms/src/styled/tiles/StyledTile.ts
austin94: I've moved this to a standalone file to separate them.

packages/forms/src/elements/tiles/Tiles.spec.tsx
```diff
@@ -0,0 +1,253 @@
...
+    });
+
+    it('checks input when tile is selected', () => {
+      const { container } = render(
+        <Tiles name="example" value="item-1">
+          <Tiles.Tile disabled value="item-1" />
+        </Tiles>
+      );
+
+      expect(container.querySelector('input')).toBeChecked();
```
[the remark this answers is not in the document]
austin94: I had just assumed this didn't work for inputs that didn't have a defined `aria-labelledby`! Updated.

packages/forms/src/elements/tiles/Tiles.spec.tsx
```diff
@@ -0,0 +1,253 @@
...
+        <Tiles name="example">
+          <Tiles.Tile data-test-id="tile" disabled />
+        </Tiles>
+      );
+
+      expect(getByTestId('tile')).toHaveAttribute('aria-disabled', 'true');
+      expect(container.querySelector('input')).toBeDisabled();
+    });
+
+    it('checks input when tile is selected', () => {
```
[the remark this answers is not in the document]
austin94: I've added an uncontrolled usage test.

packages/forms/src/elements/tiles/Tiles.spec.tsx
```diff
@@ -57,47 +57,69 @@ describe('Tiles', () => {
...
+    it('checks input when tile is uncontrolled', () => {
+      const { getByLabelText, getByText } = render(
+        <Tiles name="example">
+          <Tiles.Tile value="item-1">
+            <Tiles.Label>label</Tiles.Label>
+          </Tiles.Tile>
+        </Tiles>
+      );
+
+      fireEvent.click(getByText('label'));
```
hzhu: Not a blocker, but this should probably use `getByLabelText` also.

austin94: This would actually the backing `input`. This tests that the `label` is able to be clicked.

PR #658, opened 2020-03-02
feat(accordions): introduce new Stepper component

### LINKED ISSUE #408: Feature Request: Stepper component
## Expectations

I expect to have a `<Stepper />` component which allows for a controlled progression through defined steps. Similar to the [salesforce lightning progress indicator](https://react.lightningdesignsystem.com/components/progress-indicators/) component.

### Fine Print

- Component: This may want to live in `react-wells` depending on the amount of shared styling between different stepper variants

hzhu:
## Description

This pull request introduces the new accordions package to Garden which will house the `Stepper` component. It includes code for the `Stepper`, tests, and documentation.

## Details

* The recommended maximum steps for a `Stepper` is five. However, the Styleguist examples allow it to render various amounts of steps (3, 5, or 10) and optional long labels to stress test the component.

* I would like to propose the first package that uses function React components in the `Stepper` demo. It allows the demo to use a more modern approach when demonstrating the `Stepper`.

* ~The package uses one new dependency, `react-transition-group` which provides transition states that are helpful for transitions and animations. The addition of this package results in a small, 1% increase in bundle size~. 

* Accessibility: there is a small note in the Styleguist example that encourages API consumers to communicate to users on assistive devices which step the user is currently on.

## Preview

The preview of the `Stepper` component and its documentation can be found here: https://henry-master.netlify.com/accordions (up-to-date as of 3/10/20).

## Checklist

      designer as a reviewer)

closes #408


packages/steppers/src/styled/StyledStepContent.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return undefined;
+};
+
+export const StyledStepContent = styled.div.attrs<IStyledStepContent>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledStepContent>`
+  transition: max-height 0.5s ease-in-out;
```
hzhu: These transition values may need to be changed. I went with what looked natural, however, I don't have a good set of heuristics for determining these values. I looked at `Chrome`'s subnav transition, however, I thought that it transitioned a bit quickly and that a `Stepper` animates a bit differently (the previous step has an exit transition).

@ginnywood @jzempel feel free to provide new transitions value in this thread, its heuristics, and I can make adjustments to the transition 👍.

packages/steppers/src/styled/StyledStep.ts
```diff
@@ -0,0 +1,45 @@
...
+  position: ${props => props.isHorizontal && 'relative'};
+  flex: ${props => props.isHorizontal && '1'};
+
+  &:last-of-type ${StyledLine} {
+    display: ${props => props.theme.rtl && 'none'};
+  }
+
+  &:first-of-type ${StyledLine} {
+    display: ${props => !props.theme.rtl && 'none'};
+  }
```
hzhu: Originally had declared a single selector with string template in the selector, however due to a parse error, the `display` style was separated into using two selectors.

originally:

```js
&:${props => (props.theme.rtl ? 'last-of-type' : 'first-of-type')} ${StyledLine} {
  display: none;
}
```

parseError:

Screen Shot 2020-03-01 at 5 03 51 PM

packages/steppers/src/styled/StyledLabel.ts
```diff
@@ -0,0 +1,43 @@
...
+} from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'steppers.step_label';
+
+interface IStyledLabelProps {
+  isActive?: boolean;
+  isHorizontal?: boolean;
+}
+
+export const StyledLabel = styled.div.attrs<IStyledLabelProps>({
```
hzhu: @ginnywood I know we discussed applying horizontal margins to the labels, however, there could be some trade-offs (margins only noticed on Steppers with 10+ steps with long labels, effects of other parts of the layout, etc). Would you be down to sit together and see how we could make adjustments to the margins?

packages/steppers/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const Stepper = forwardRef<HTMLDivElement, IStepperProps>(
+  ({ isHorizontal, activeIndex, ...props }, ref) => {
+    const currentIndexRef = useRef(0);
+    const stepperContext = { isHorizontal, activeIndex, currentIndexRef };
+
+    React.useEffect(() => {
+      return () => {
+        currentIndexRef.current = 0;
+      };
+    }, [props.children]);
```
hzhu: This effect is only used to perform clean up — it resets the current index so that each `Step` can compute the correct step when a `Stepper` has dynamic steps (e.g. the `basic` example which can dynamically render 3, 5, or 10 steps). Without this cleanup, dynamically changing the number of steps causes `Step` to have incorrect calculations. 

This cleanup should not result in a performance hit. It runs once when the component mounts, unmounts, and each time the `activeIndex` changes (`props.children` change).

packages/steppers/src/styled/StyledLabel.ts
ginnywood: Yes, grab some time with me!

packages/steppers/src/styled/StyledStepContent.ts
ginnywood: It feels a little sluggish to me right now. I think it could speed up a bit.
The main issue is sequencing: when the new step is transitioning to active, it first opens, _then_ slides up into place. Instead, it should happen simultaneously, sliding up into the area previously occupied by the previous step, while opening at the same time.
Let’s talk about this.

packages/steppers/src/elements/components/StepLabel.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const styledIcon = (
+      <StyledIcon isActive={isActive} isHorizontal={isHorizontal}>
+        {isCompleted ? <CheckCircleStrokeIcon /> : stepIcon}
+      </StyledIcon>
+    );
+
+    return (
+      <StyledLabel ref={ref} isActive={isActive} isHorizontal={isHorizontal} {...props}>
+        {isHorizontal ? (
+          <StyledIconFlexContainer>{styledIcon}</StyledIconFlexContainer>
```
hzhu: Only applying flex to the `StyledLabel` worked across all browsers except IE11 — in IE11, this caused the long labels to render onto a single line. To address this issue for IE but also have the layout correct for other browsers, I restructured the HTML/CSS so that Step Icons are centered within their flex container in horizontal steppers.

jzempel:
I think we said the package should be named `accordions`, and be setup to contain both `Stepper` and `Accordion` components (at a minimum).

packages/steppers/src/styled/StyledStep.ts
jzempel: https://github.com/stylelint/stylelint/issues/4401

packages/accordions/src/elements/components/Step.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render } from 'garden-test-utils';
+import { Stepper } from '../Stepper';
+import { Step } from './Step';
+
+describe('Step', () => {
```
hzhu: @austin94, I noticed that in `Tiles`, a single spec file exists that tests `Tiles` and all of its static property components. 

Should I delete all the `.spec` files under the `/components` directory which test static property components, and move these tests over to a single `Stepper.spec.tsx`?

packages/accordions/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Installation
+
+```sh
+npm install @zendeskgarden/react-accordions
+
+# Peer Dependencies - Also Required
+npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
+```
+
+## Usage
```
hzhu: This `Usage` section could be structured like the following when the Accordion component is introduced:

```
## Accordion Usage
<Accordion />

## Stepper Usage
<Stepper />
```

packages/accordions/src/elements/components/StepContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { StyledStepContent } from '../../styled';
+import { useStepContext, useStepperContext } from '../../utils';
+
+export const StepContent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
+    const contentRef = useCombinedRefs<HTMLDivElement>(ref);
+    const { activeIndex, isHorizontal } = useStepperContext();
+    const { currentStepIndex } = useStepContext();
+    const isActive = currentStepIndex === activeIndex;
+    const isVertical = isHorizontal === false;
```
austin94: Is this necessary? Could the ternary below use `isHorizontal` directly?

packages/accordions/src/utils/useStepContext.ts
```diff
@@ -0,0 +1,24 @@
...
+  currentStepIndex: number;
+}
+
+export const StepContext = createContext<IStepContext | undefined>(undefined);
+
+export const useStepContext = () => {
+  const context = useContext(StepContext);
+
+  if (context === undefined) {
+    throw new Error('useStepContext must be used within a StepContext.Provider');
```
austin94: This error is meant to help consumers understand required nesting, but it's using wording/components that only affect our implementation.

Another wording may be "This component must be rendered within a `Stepper.Step` component"

packages/accordions/src/utils/useStepperContext.ts
```diff
@@ -0,0 +1,26 @@
...
+  currentIndexRef: React.MutableRefObject<number>;
+}
+
+export const StepperContext = createContext<IStepperContext | undefined>(undefined);
+
+export const useStepperContext = () => {
+  const context = useContext(StepperContext);
+
+  if (context === undefined) {
+    throw new Error('useStepperContext must be used within a StepperContext.Provider');
```
austin94: Same wording as `useStepContext`

packages/accordions/examples/basic.md
```diff
@@ -0,0 +1,249 @@
+### Steppers
+
+The `Stepper` component accepts an `activeIndex` prop and determines the current step and computes
+the completed steps. The `activeIndex` can be used to help communicate to assistive devices in
+which step the user is currently on using
+[ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
+
+The default layout of the `Stepper` is vertical, and renders step content using `StepContent`.
```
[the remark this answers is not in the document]
hzhu: That's a fair point. My goal was to ensure consumers know that horizontal Stepper requires content to live outside it. What do you think if the entire paragraph was reduced to the following sentence?

> For a horizontal `Stepper`, consumers should render step content outside of the `Stepper` component instead of using `Stepper.Content`.

packages/accordions/src/elements/components/StepContent.tsx
hzhu: I like reading the return statement as "if it is vertical return this JSX", particularly cause `StepContent` serves vertical Steppers. It was a stylistic and reading preference, but I can remove it.

packages/accordions/src/elements/components/Step.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByTestId } = render(
+      <Stepper activeIndex={0} isHorizontal>
```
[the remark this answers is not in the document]
hzhu: It is a required `Stepper` context value, and TypeScript is not satisfied with `Stepper`'s `isHorizontal` default prop 😭.

I think I can work around it with:

```jsx
const stepperContext = { isHorizontal: isHorizontal || false, activeIndex, currentIndexRef };
```

What are your thoughts?

packages/accordions/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+  Step: typeof Step;
+  Label: typeof StepLabel;
+  Content: typeof StepContent;
+}
+
+interface IStepperProps extends HTMLAttributes<HTMLDivElement> {
+  /** Used to show the current step and compute completed steps */
+  activeIndex: number;
+  /** Displays the Stepper in a horizontal layout */
+  isHorizontal: boolean;
```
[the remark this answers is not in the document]
hzhu: These are the required values for `StepperContext`. So the props can be made optional, but a TypeScript workaround may be needed to keep TS happy:

```jsx
const stepperContext = { 
  isHorizontal: isHorizontal || false, 
  activeIndex: activeIndex || 0, 
  currentIndexRef 
};
```

What are your thoughts?

packages/accordions/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const Stepper = forwardRef<HTMLDivElement, IStepperProps>(
+  ({ isHorizontal, activeIndex, ...props }, ref) => {
+    const currentIndexRef = useRef(0);
+    const stepperContext = { isHorizontal, activeIndex, currentIndexRef };
+
+    React.useEffect(() => {
+      return () => {
+        currentIndexRef.current = 0;
+      };
+    }, [props.children]);
```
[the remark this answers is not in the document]
hzhu: Seems like `props.children` isn't necessary if the goal is to run the side effect on every child render, so I removed the dep array. I did a little refactoring: use direct import, moved the reset to `useEffect` body, and removed the dependency array.

packages/steppers/src/styled/StyledLabel.ts
hzhu: @ginnywood the steps now have horizontal spacing (`12px` each side). This was achieved by applying horizontal padding and a `min-width` on each step. The step label should have horizontal spacing now 👍.

packages/steppers/src/styled/StyledStep.ts
hzhu: Tried various `stylelint` disable comments and adding arbitrary CSS selector and styling like the OP in the issue with no luck. Open to suggestions, but I think having a style for both LTR and RTL is okay.

packages/accordions/examples/basic.md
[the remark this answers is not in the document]
austin94: Looks good!

packages/accordions/src/elements/Stepper.tsx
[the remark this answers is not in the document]
austin94: If there are values provided by `defaultProps` you may be able to use:

```
{
    isHorizontal: isHorizontal!,
    activeIndex: activeIndex!
}
```

packages/accordions/examples/basic.md
```diff
@@ -0,0 +1,247 @@
...
+const GlobeStrokeIcon = require('@zendeskgarden/svg-icons/src/12/globe-stroke.svg').default;
+const CarStrokeIcon = require('@zendeskgarden/svg-icons/src/12/car-stroke.svg').default;
+const PencilStrokeIcon = require('@zendeskgarden/svg-icons/src/12/pencil-stroke.svg').default;
+const PhoneStrokeIcon = require('@zendeskgarden/svg-icons/src/12/phone-stroke.svg').default;
+const LightningBoltStrokeIcon = require('@zendeskgarden/svg-icons/src/12/lightning-bolt-stroke.svg')
+  .default;
+const SearchStrokeIcon = require('@zendeskgarden/svg-icons/src/12/search-stroke.svg').default;
+
+const StyledHorizontalContent = styled.div`
+  padding: 18px 0;
```
austin94: Are we able to use our `theme` here?

packages/accordions/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,75 @@
...
+ *  - `Stepper.StepContent`
+ */
+// eslint-disable-next-line react/display-name
+export const Stepper = forwardRef<HTMLDivElement, IStepperProps>(
+  ({ isHorizontal, activeIndex, ...props }, ref) => {
+    const currentIndexRef = useRef(0);
+    const stepperContext = {
+      isHorizontal: isHorizontal || false,
+      activeIndex: activeIndex || 0,
+      currentIndexRef
```
austin94: From my other comment, if we provide values through `defaultProps` we should respect those values.

If may be ok to use:

```
isHorizontal: isHorizontal!
```

packages/accordions/src/elements/components/Step.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const steps = [1, 2, 3];
+    const { queryAllByTestId } = render(
+      <Stepper isHorizontal>
+        {steps.map(step => (
+          <Step key={step} />
+        ))}
+      </Stepper>
+    );
+
+    expect(queryAllByTestId('step-line').length).toBe(steps.length);
```
austin94: Are you able to use the `toHaveLength()` utility here?

packages/accordions/src/elements/components/Step.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const steps = [1, 2, 3];
+    const { queryAllByTestId } = render(
+      <Stepper activeIndex={0}>
+        {steps.map(step => (
+          <Step key={step} />
+        ))}
+      </Stepper>
+    );
+
+    expect(queryAllByTestId('step-line').length).toBe(0);
```
austin94: Are you able to use the `toHaveLength()` utility here?

packages/accordions/src/styled/StyledContent.ts
```diff
@@ -0,0 +1,45 @@
...
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'steppers.step_content';
+
+export interface IStyledContent {
+  isActive?: boolean;
+}
+
+const inactiveStyling = css`
+  & > div {
```
austin94: Rather than have a naked `<div>` within the `StepContent` would it make sense to have this a styled component? We could pass an `isActive` prop to it to control this animation.

packages/accordions/src/styled/StyledContent.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const StyledContent = styled.div.attrs<IStyledContent>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledContent>`
+  margin: ${props =>
+    `${props.theme.space.base * 1.5}px ${
+      props.theme.rtl ? props.theme.space.base * 3 : '0'
+    }px ${props.theme.space.base * 1.5}px ${props.theme.rtl ? '0' : props.theme.space.base * 3}px`};
+  padding: ${props => props.theme.space.base * 4}px;
```
austin94: There are some repeated values here? Would it be cleaner to have this in a `sizeStyles` utility?

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,62 @@
...
+  !props.isHorizontal && `${props.theme.space.base * 3}px`};
+  margin-bottom: ${props => props.isHorizontal && props.theme.space.base}px;
+  border-radius: 100%;
+  background: ${props =>
+    props.isActive
+      ? getColor('neutralHue', 600, props.theme)
+      : getColor('neutralHue', 200, props.theme)};
+  width: ${props => props.theme.space.base * 6}px;
+  height: ${props => props.theme.space.base * 6}px;
+  line-height: ${props => getLineHeight(props.theme.lineHeights.lg, props.theme.fontSizes.sm)};
```
rossmoody: Maybe a min-width in here to stop this formatting? 
image

packages/accordions/src/styled/StyledLabel.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const StyledLabel = styled.div.attrs<IStyledLabelProps>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledLabelProps>`
+  display: ${props => !props.isHorizontal && 'flex'};
+  flex: ${props => props.isHorizontal && '1'};
+  flex-direction: ${props => (props.isHorizontal ? 'column' : 'row')};
+  align-items: center;
+  text-align: center;
```
rossmoody: Might need some logic here to left align Label in horizontal applications for when it wraps: 
image

packages/accordions/src/styled/StyledContent.ts
rossmoody: Unsure the best solve from my tinkering and I know it's not what I came up with (`padding: ${props => props.theme.space.base * 5 + 2}px;`) but the content will need pushed over to be justified with the label: https://share.getcloudapp.com/2NuBwWB2

hzhu: Thanks @rossmoody, content should now be justified to the label for both writing modes.

packages/accordions/src/styled/StyledLabel.ts
hzhu: Spoke to @rossmoody, leaving as is because only browser dev tools can force a width that results in this UI.

packages/accordions/src/styled/StyledIcon.ts
hzhu: Spoke to @rossmoody, leaving as is because only browser dev tools can force a width that results in this UI.

packages/accordions/src/styled/StyledContent.ts
hzhu: The spacing styles have been moved to a function. Let me know if you think further adjustments are needed.

packages/accordions/src/styled/StyledContent.ts
hzhu: A styled component feels good. I created a `StyledInnerContent` and moved transition styles over to it.

hzhu:
> A consumer can hide the labels, right? I don’t see a prop for it, but maybe it’s done another way. That wouldn’t be recommended for the vertical one, but we would definitely want it for the horizontal one, so having the option available would be good.

At this time labels cannot be hidden. However, I can add a `prop` to hide the label. 

I have some further thoughts that are related to the differences between the orientations. From what I recall, a horizontal `Stepper` by default does not have labels because a "current task" UI can communicate the information that a label would. Meanwhile, the vertical `Stepper` has labels by default and does not use a "current task" UI. The vertical orientation has expand and collapse behavior for its content while the horizontal orientation has content rendered immediately beneath it. I think having unique characteristics for one orientation and not the other is an indication that there could be two separate components:

1) `Stepper` can have a vertical or horizontal orientation. Stepper content lives outside the stepper component. A "current task" UI could be implemented by consumers indicating textual info about the current task. Label text is optional. These characteristics apply to _both_ vertical and horizontal orientations.

2) `AccordionStepper` is related to `Stepper`, but is a separate component. This component only supports vertical orientation and has expand and collapse behavior. Due to this behavior, the stepper content lives inside the `AccordionStepper`. Labels are always used and a "current task" UI isn't needed.

A third component that would live in this package would be the `Accordion`, which is based on the W3 spec for accordions. It is indirectly related to `Stepper` and `AccordionStepper`.

ginnywood:
@hzhu those definitions sound good to me. From a design pattern standpoint, an `AccordionStepper` would cover vertical use cases, and a vertical `Stepper` would not be needed at this time. However, if you think it’s best to include both orientations for `Stepper`, that is fine.

hzhu:
@ginnywood 

✅ Ability to hide label text.
✅ Transitions added to the icons.
✅ Step content margin should be `4px` on top and bottom.
✅ Increased the icon margin to `8px`.
✅Updated padding for `StepContent`.

> Is it possible to have 0px padding on the left side of the first step, and 0px padding on the right side of the last step? That way, the Stepper will appear to align to the content.

The starting step does not have padding. The visible spacing is calculated by the flex layout to render the steps (Step) so that they take the full width of the flex container (Stepper). The step line is part of each step.

Screen Shot 2020-03-08 at 7 57 00 PM

This technique is used to make the step lines flexible so that they grow to accommodate a given width the user is on. I wasn't able to create a layout that would have flexible lines and have them overlap the step, but I can see if @jzempel and @austin94 have some ideas.

packages/accordions/examples/basic.md
```diff
@@ -0,0 +1,247 @@
...
+                  const nextActiveIndex = parseInt(event.target.value, 10);
+                  setActiveIndex(nextActiveIndex);
+                }}
+              />
+            </Field>
+          </Col>
+        </Row>
+      </Well>
+      <Row className="u-mt">
+        <Col>
```
jzempel: There are many alignment, sizing, and word-wrap issues that need to be addressed when the stepper is forced to render under extreme widths (i.e. 100px).

packages/accordions/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface IStepperProps extends HTMLAttributes<HTMLDivElement> {
+  /** Used to show the current step and compute completed steps */
+  activeIndex?: number;
+  /** Displays the Stepper in a horizontal layout */
+  isHorizontal?: boolean;
+}
+
+/**
+ * Accepts all `<div>` attributes and events. Also accepts static properties:
```
jzempel: `<ol>`

ginnywood:
@hzhu this is awesome. The icon transition looks great!

Only other nitpick I have is to increase the vertical StepContent padding to 20px. Looks like it is 16px right now. https://share.getcloudapp.com/P8uR9Adk

packages/accordions/src/elements/components/Step.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+describe('Step', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByTestId } = render(
+      <Stepper>
+        <Step data-test-id="step" ref={ref} />
+      </Stepper>
+    );
+
+    expect(getByTestId('step')).toBe(ref.current);
```
jzempel: This ref test needs to be against `Stepper.Step` in order to be valid.

packages/accordions/src/elements/Stepper.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+ *  - `Stepper.Step`
+ *  - `Stepper.StepLabel`
+ *  - `Stepper.StepContent`
```
jzempel: Are these "static properties" as stated above? I think of them as sub-components.

packages/accordions/src/elements/components/StepContent.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByTestId } = render(
+      <Stepper>
+        <Step>
+          <StepContent ref={ref} data-test-id="step-content" />
+        </Step>
+      </Stepper>
+    );
+
+    expect(getByTestId('step-content')).toBe(ref.current);
```
jzempel: This ref test needs to run against `Stepper.Content` in order to be valid.

packages/accordions/src/elements/components/StepLabel.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByTestId } = render(
+      <Stepper>
+        <Step>
+          <StepLabel ref={ref} data-test-id="step-label" />
+        </Step>
+      </Stepper>
+    );
+
+    expect(getByTestId('step-label')).toBe(ref.current);
```
jzempel: This ref test needs to run against `Stepper.Label` in order to be valid.

packages/accordions/examples/basic.md
```diff
@@ -0,0 +1,247 @@
+### Steppers
+
+The `Stepper` component accepts an `activeIndex` prop and determines the current step and computes
+the completed steps. The `activeIndex` can be used to help communicate to assistive devices in
+which step the user is currently on using
+[ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
```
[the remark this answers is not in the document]
hzhu: What do you think about following?

> Features that implement a `Stepper` can use [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) to communicate what step the user is on.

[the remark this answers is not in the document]
jzempel: > `Stepper` implementations can use [ARIA live regions]() to communicate which step the user is on.

- or -

> Implementations can use ...

packages/accordions/src/styled/StyledLabelText.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import styled from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+interface IStyledLabelTextProps {
+  isHorizontal?: boolean;
+}
+
+export const StyledLabelText = styled.div<IStyledLabelTextProps>`
+  padding: ${props => props.isHorizontal && `0 ${props.theme.space.base * 3}px`};
```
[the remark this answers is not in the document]
hzhu: Ouch, you're right the text starts bleeding into the next step. I've updated the example to include a label with a very long word and have updated the styles to have `word-break` and `overflow-wrap` properties. Thank you for the suggestion!

packages/accordions/src/elements/Stepper.tsx
hzhu: That sounds good, updated 👍 .

packages/accordions/src/styled/StyledStepper.ts
```diff
@@ -0,0 +1,28 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'steppers.stepper';
+
+export interface IStyledStepper {
+  isHorizontal?: boolean;
+}
+
+export const StyledStepper = styled.div.attrs<IStyledStepper>({
```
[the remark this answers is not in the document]
hzhu: Great idea, thanks for the suggestion 👍 .

packages/accordions/src/elements/components/Step.spec.tsx
hzhu: Whoops, this isn't how the API is used. Thanks for the suggestion, I've updated the tests to use the `Stepper` sub-components where necessary 👍 .

hzhu:
> increase the vertical StepContent padding to 20px. Looks like it is 16px right now. https://share.getcloudapp.com/P8uR9Adk

@ginnywood thanks for the screen recording! This should be addressed now.

packages/accordions/src/styled/StyledLabel.ts
hzhu: I think the alignment issue has been addressed:

Screen Shot 2020-03-09 at 3 28 49 PM

I'll resolve this thread and move all other sizing issues regarding `Stepper` being rendered in small containers to [this thread](https://github.com/zendeskgarden/react-components/pull/658#discussion_r389900897).

packages/accordions/src/elements/components/StepContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const contentRef = useCombinedRefs<HTMLDivElement>(ref);
+    const { activeIndex, isHorizontal } = useStepperContext();
+    const { currentStepIndex } = useStepContext();
+    const isActive = currentStepIndex === activeIndex;
+
+    useEffect(() => {
+      if (contentRef.current && isHorizontal === false) {
+        const child = contentRef.current.children[0] as any;
+
+        child.style.height = `${child.scrollHeight}px`;
```
hzhu: Note: this `height` is set for animation purposes. This `height` is not updated when the window resizes at this time.

packages/accordions/examples/basic.md
hzhu: I've updated the sizing and word-wrap for `Stepper`'s that are rendered in a constrained width. 

@ginnywood could you take a look at `Stepper`'s in constrained (small) widths and see if it needs further adjustments?  One way to view a constrained width in the example is navigate to the "Edit Code" section to set a `md` prop on the `<Col>` that wraps the `<Stepper>`:

```jsx
<Col md={3}>
  <Stepper>...</Stepper>
</Col>
```

We can also go through this together over video call 👍 .

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,75 @@
...
+})<IStyledIcon>`
+  display: flex;
+  align-items: center;
+  align-self: ${props => !props.isHorizontal && 'self-start'};
+  justify-content: center;
+  transition: background 0.25s ease-in-out, color 0.25s ease-in-out;
+  /* stylelint-disable-next-line property-no-unknown */
+  margin-${props => (props.theme.rtl ? 'left' : 'right')}: ${props =>
+  !props.isHorizontal && `${props.theme.space.base * 3}px`};
+  margin-bottom: ${props => props.isHorizontal && props.theme.space.base * 2}px;
```
jzempel: Margin is size-related and should be moved to `sizeStyles` per https://github.com/zendeskgarden/react-components/blob/master/docs/api.md#rules.

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,75 @@
...
+  border-radius: 100%;
+  background: ${props =>
+    props.isActive
+      ? getColor('neutralHue', 600, props.theme)
+      : getColor('neutralHue', 200, props.theme)};
+  line-height: ${props => getLineHeight(props.theme.lineHeights.lg, props.theme.fontSizes.sm)};
```
jzempel: Do not use `theme.lineHeights` anywhere except for in `typography`. This should be moved to `sizeStyles` where it can leverage the calculated `size` together with `fontSize` in the `getLineHeight` call.

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,75 @@
...
+  color: ${props =>
+    props.isActive ? props.theme.colors.background : getColor('neutralHue', 800, props.theme)};
+  font-size: ${props => props.theme.fontSizes.sm};
```
jzempel: Move to `sizeStyles`.

packages/accordions/src/styled/StyledLabel.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const StyledLabel = styled.div.attrs<IStyledLabelProps>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledLabelProps>`
+  display: ${props => !props.isHorizontal && 'flex'};
+  align-items: ${props => !props.isHorizontal && 'center'};
+  text-align: ${props => props.isHorizontal && 'center'};
+  line-height: ${props => getLineHeight(props.theme.lineHeights.lg, props.theme.fontSizes.md)};
```
jzempel: Do not use `theme.lineHeights`; use `space.base` instead. Also, designs call for the label to be the standard line-height of 20px, not 24px. This won't display properly when the line wraps. You'll probably need to apply padding to make up for the difference between the 20px line-height and the 24px icon to which this is intended to be centered.

packages/accordions/src/styled/StyledLabelText.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const StyledLabelText = styled.div.attrs<IStyledLabelTextProps>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledLabelTextProps>`
+  display: ${props => props.isHidden && 'none'};
+  padding: ${props => props.isHorizontal && `0 ${props.theme.space.base * 3}px`};
+  word-break: ${props => props.isHorizontal && 'break-word'};
+  overflow-wrap: ${props => props.isHorizontal && 'break-word'};
+  word-wrap: ${props => props.isHorizontal && 'break-word'}; /* [1] */
```
jzempel: In `css-components` PostCSS compiles `overflow-wrap: break-word` down to `word-wrap: break-word`. So `word-wrap` should be the only property you need across all browsers. (Pity that styled-components doesn't take care of this post-compile for us.) Please update all uses accordingly.

packages/accordions/src/styled/StyledStep.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  &:first-of-type ${StyledLine} {
+    display: ${props => !props.theme.rtl && 'none'};
+  }
+
+  &:not(:last-of-type) ${StyledContent} {
+    /* stylelint-disable-next-line property-no-unknown */
+    border-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.borders.sm};
+    border-color: ${props => getColor('neutralHue', 300, props.theme)};
+  }
```
jzempel: Now that we see how line wrapping impacts vertical layouts, I wonder if you'll need a variation on `StyledLine` instead of this border treatment? 😬 

Screen Shot 2020-03-10 at 11 22 46 AM

/cc @ginnywood

hzhu: Another option would is the vertical stepper could keep the label on a single line with a `min-width`. This would cause the label to overflow the `Stepper`. In the meantime, I can see about another variation on `StyledLine`, but that will be difficult since the line will need to be sized based on the amount of content.

jzempel: Yeah. Get design direction on this before you chase down any particular solution.

packages/accordions/README.md
```diff
@@ -0,0 +1,50 @@
+# @zendeskgarden/react-accordions [![npm version][npm version badge]][npm version link]
```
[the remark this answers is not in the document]
hzhu: 🤔 I noticed this too and haven't found out why yet. Though, it seems like other packages also produce the same UI locally. For example, here's how modals look on the local environment:

Screen Shot 2020-03-10 at 11 39 37 AM

packages/accordions/src/styled/StyledStep.ts
ginnywood: > Another option would is the vertical stepper could keep the label on a single line with a `min-width`. This would cause the label to overflow the `Stepper`
Can I see what that would be like?

hzhu: @ginnywood an arbitrary `min-width` would look like this:

(`min-width: 600px` used):
Screen Shot 2020-03-10 at 1 42 03 PM

The label _would_ still wrap if the text overflows the arbitrary `min-width` (`600px` in the case of these screenshots). For example, here is a paragraph used as a label:

Screen Shot 2020-03-10 at 1 43 16 PM

jzempel: You would need to impose `white-space: nowrap` on these labels to prevent the wrapping.

ginnywood: I’m concerned about not allowing wrapping. It seems like it will obscure information.

I would rather have the lines break at these extreme sizes than obscure information. However, before resorting to that, could a styling change to the line help here? From a design perspective, it is negotiable. We could remove the gap between the line and the icon...we could explore removing the line...whatever would be helpful.

jzempel: @ginnywood I think line-wrapping is the right call. Can we agree to punt the 1px border spacing decision to a follow-on? There are a frustrating number of structural implications that need to be considered in order to handle that.

You should feel free to make the most optimal design decision. I think it would be best if we can postpone development under a separate PR.

ginnywood: Absolutely!

hzhu: Thanks for the discussion @ginnywood @jzempel! Let's reopen this thread when the follow-on occurs.

packages/accordions/README.md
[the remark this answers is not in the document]
jzempel: Don't worry about it for now.

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,80 @@
...
+    margin-bottom: ${props.isHorizontal && `${props.theme.space.base * 2}px`};
+    /* stylelint-disable-next-line property-no-unknown */
+    margin-${props.theme.rtl ? 'left' : 'right'}: ${!props.isHorizontal &&
+    `${props.theme.space.base * 3}px`};
+    width: ${size};
+    min-width: ${size};
+    height: ${size};
+    min-height: ${size};
+    line-height: ${getLineHeight(size, props.theme.fontSizes.sm)};
+    font-size: ${props.theme.fontSizes.sm};
```
jzempel: `fontSize` could move to a variable declared above.

packages/accordions/src/styled/StyledIcon.ts
```diff
@@ -0,0 +1,80 @@
...
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledIcon>`
+  display: flex;
+  align-items: center;
+  align-self: ${props => !props.isHorizontal && 'self-start'};
+  justify-content: center;
+  transition: background 0.25s ease-in-out, color 0.25s ease-in-out;
+  border-radius: 100%;
+  ${colorStyles}
+  ${sizeStyles}
```
jzempel: Switch these two. `colorStyles` always take priority over `sizeStyles`. See https://github.com/zendeskgarden/react-components/blob/master/docs/api.md#rules for details.

PR #659, opened 2020-03-02
fix(forms): allow controlled range to render background-size

hzhu:
## Description

While working on the `Steppers` demo, I noticed that the `Range` component does not render the correct background-size when used as a controlled component.

## Detail

I've added `props.value` to the `useEffect` dependency array which correctly updates the background-size when a controlled value changes. I've also added a test that verifies the background-size is updated correctly when `Range` is used as a controlled component.

## Screenshots

The following screen recordings are from a contrived example but demonstrates the fix.

## Checklist

### DIFF
--- a/packages/forms/src/elements/Range.spec.tsx
+++ b/packages/forms/src/elements/Range.spec.tsx
@@ -43,6 +43,28 @@ describe('Range', () => {
       expect(getByTestId('range').style.backgroundSize).toBe('25%');
     });
 
+    it('applies correct backgroundSize for a controlled component', () => {
+      const ControlledExample = () => {
+        const [value, setValue] = React.useState(25);
+
+        return (
+          <>
+            <Field>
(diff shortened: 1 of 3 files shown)


packages/forms/src/elements/Range.spec.tsx
```diff
@@ -6,7 +6,7 @@
  */
 
 import React from 'react';
-import { render, fireEvent } from 'garden-test-utils';
+import { render, fireEvent, prettyDOM } from 'garden-test-utils';
```
austin94: `prettyDOM` import should be removed

austin94: Also, the `debug` command returned from `render` uses `prettyDOM` internally if you're looking to use it in the future.

hzhu: 🤦‍♀ TY!

PR #663, opened 2020-03-05
fix(modals): include useModal args as props

hzhu:
## Description

This PR allows consumers of `Modal` to determine whether a modal should focus on mount and restore focus when the modal is dismissed. 

## Detail

The `container-modal` dependency has been updated which forwards `restoreFocus` to the `useFocusJail` hook. `<Modal>` now exposes two new props `focusOnMount` and `restoreFocus` which forwards boolean values to the underlying `useModal` and `useFocusJail` hooks.

Note: `focusOnMount` isn't a new feature, however it was never exposed to consumers as a prop.

## Demo

The new props can be tested with "Edit Code" in the demo example, and providing a boolean value to the `Modal`'s `restoreFocus` or `focusOnMount` prop.

The demo can be found at: [https://modal-updated-props.netlify.com/modals/](https://modal-updated-props.netlify.com/modals/#basic)

## Checklist

### DIFF
--- a/packages/modals/src/elements/Modal.tsx
+++ b/packages/modals/src/elements/Modal.tsx
@@ -71,6 +71,14 @@ export interface IModalProps {
    * Enable large modal styling
    */
   isLarge?: boolean;
+  /**
+   * Determines whether to apply keyboard focus on the modal on mount
+   */
+  focusOnMount?: boolean;
+  /**
+   * Determines whether to return keyboard focus to the element that triggered the modal
+   */
+  restoreFocus?: boolean;
 }
 
 /**
@@ -92,6 +100,8 @@ export const Modal = withTheme(
         id,
         appendToNode,
         theme,
+        focusOnMount,
+        restoreFocus,
         ...modalProps
       },
       ref
@@ -105,7 +115,13 @@ export const Modal = withTheme(
         getTitleProps,
         getContentProps,
         getCloseProps
-      } = useModal({ id, onClose, modalRef });
+      } = useModal({
+        id,
+        onClose,
+        modalRef,
+        focusOnMount,
+        restoreFocus
+      });
 
       useFocusVisible({ scope: modalRef });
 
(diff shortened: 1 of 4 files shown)


packages/modals/src/elements/Modal.tsx
```diff
@@ -71,6 +71,14 @@ export interface IModalProps {
    * Enable large modal styling
    */
   isLarge?: boolean;
+  /**
+   * Determines whether to put keyboard focus on the modal on mount
+   */
+  hasFocusOnMount?: boolean;
```
austin94: With `focus/restore` being actions could we leave theme as `focusOnMount`/`restoreFocus`? Feels clunky to with the `has/is` prefix.

packages/modals/src/elements/Modal.tsx
```diff
@@ -71,6 +71,14 @@ export interface IModalProps {
    * Enable large modal styling
    */
   isLarge?: boolean;
+  /**
+   * Determines whether to put keyboard focus on the modal on mount
```
austin94: > to ~put~ apply keyboard

Does this make any sense?

packages/modals/src/elements/Modal.tsx
hzhu: These variables use `has/is` prefix to follow the heuristic that all non-native boolean attributes should use those prefixes. I could rename these if "action variables" don't get `has/is` prefixes. Would these be the first booleans in Garden that don't have those prefixes?

packages/modals/src/elements/Modal.tsx
hzhu: Yeah that wording makes sense 👍, thanks!

packages/modals/src/elements/Modal.tsx
hzhu: Looks like I need to update my heuristics when it comes to boolean naming as this is already an established practice: `focusInset`.

hzhu:
> `hasApproved` 😄

🤣🤣🤣

PR #664, opened 2020-03-06
chore: update to use CircleCI

jzempel:
## Description

This migrates the repo from Travis to CircleCI for continuous integration/delivery. Additionally, update all badges to use https://badgen.net/.

## Detail

https://app.circleci.com/pipelines/github/zendeskgarden/react-components


.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+    executor: node/default
+    steps:
+      - checkout
+      - node/with-cache:
+          cache-key: yarn.lock
+          dir: ~/.cache/yarn
+          steps:
+            - run: yarn --frozen-lockfile --ignore-scripts
+            - run: yarn lerna bootstrap
+            - run: yarn lerna run build --concurrency=2
```
austin94: Could we include an inline comment on why we limit this concurrency?

.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+      - persist_to_workspace:
+          root: *workspace_root
+          paths: .
+
+  test:
+    executor: node/default
+    steps:
+      - *attach_workspace
+      - run: yarn lint
+      - run: yarn format
```
austin94: Does CircleCI break the build if any changes are found? If this is handled different than Travis we may be able to remove the `yarn format` command.

.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+      - *attach_workspace
+      - run: npm set //registry.npmjs.org/:_authToken=$NPM_TOKEN
+      - run: yarn lerna publish from-git --ignore-scripts --no-verify-access --yes
+
+workflows:
+  main:
+    jobs:
+      - build:
+          <<: *filters_tags
```
austin94: Why do we have to filter tags for the `build/test` jobs only?

jzempel: > if a job requires any other jobs (directly or indirectly), you must specify tag filters for those jobs

https://circleci.com/docs/2.0/configuration-reference/#tags

jzempel: Oops, this is carry-over from the linting repos. We don't need tag filtering in the lerna workflows.

.circleci/config.yml
jzempel: As I understand it, the clean working directory check has always been an `npm` step (on version) not Travis. As such, we should leave this intact just as before.

PR #674, opened 2020-03-12
chore: update gh-pages deploy

jzempel:
- [ ] **BREAKING CHANGE?** 

## Description

- Update commit user
- Skip CI on gh-pages branch
- Persist coverage for deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,9 @@ workflows:
           requires:
             - test
           context: maintainer
+          filters:
+            branches:
+              only: master
       - publish:
           requires:
             - test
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -39,6 +42,7 @@ jobs:
       - run: yarn tsc
       - run: yarn test:all --coverage --runInBand
       - run: '[ $COVERALLS_REPO_TOKEN ] && yarn coveralls < demo/coverage/lcov.info || true'
+      - *persist_to_workspace
```
austin94: What are we persisting from the `test` job?

jzempel: We were missing the `demo/coverage` folder on deploy. See https://github.com/zendeskgarden/react-components/tree/gh-pages

austin94: Aw, forgot we were publishing those. 👍

PR #690, opened 2020-03-28
fix(dropdowns, datepickers): restore menu theming and prevent autocomplete pixel nudge

jzempel:
## Description

This PR addresses several styling issues:
- calls to `menuStyles` failed to pass the `theme` object, preventing menus from receiving theming
- the `FauxInput` wrapped by `StyledSelect` did not have access to the current theme prop 🤷‍♂ 
- restore `vertical-align: baseline` styling for any child `Input` of `FauxInput` in order to retain expected 40px height (without this, the `FauxInput > Input` height is calculated to be 41.33px in Chrome)

## Detail

Using the following "dark" mode theming object:

```js
const theme = {
  ...DEFAULT_THEME,
  colors: {
    ...DEFAULT_THEME.colors,
    background: PALETTE.kale[800],
    foreground: PALETTE.white,
    chromeHue: PALETTE.black,
    neutralHue: PALETTE.kale[400]
  }
};
```

### Before

Screen Shot 2020-03-28 at 10 26 06 AM

Screen Shot 2020-03-28 at 9 39 03 AM

### After

Screen Shot 2020-03-28 at 10 25 09 AM

Screen Shot 2020-03-28 at 9 41 32 AM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/styled/field/StyledSelect.ts
+++ b/packages/dropdowns/src/styled/field/StyledSelect.ts
@@ -83,7 +83,8 @@ export interface IStyledSelectProps {
 export const StyledSelect = styled(FauxInput).attrs<IStyledSelectProps>(props => ({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  'aria-invalid': isInvalid(props.validation)
+  'aria-invalid': isInvalid(props.validation),
+  theme: props.theme
 }))<IStyledSelectProps>`
   position: relative;
   cursor: ${props => (props.disabled ? 'default' : 'pointer')};
--- a/packages/datepickers/.size-snapshot.json
+++ b/packages/datepickers/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/index.cjs.js": {
-    "bundled": 130048,
-    "minified": 72880,
-    "gzipped": 16129
+    "bundled": 130072,
+    "minified": 72894,
+    "gzipped": 16133
   },
   "dist/index.esm.js": {
-    "bundled": 128833,
-    "minified": 71716,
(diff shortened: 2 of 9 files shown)


packages/dropdowns/src/styled/field/StyledSelect.ts
```diff
@@ -83,7 +83,8 @@ export interface IStyledSelectProps {
 export const StyledSelect = styled(FauxInput).attrs<IStyledSelectProps>(props => ({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  'aria-invalid': isInvalid(props.validation)
+  'aria-invalid': isInvalid(props.validation),
+  theme: props.theme
```
jzempel: I'm a bit stumped by this one. Is it because the object being extended is a high-abstraction vs. styled component? Either way, this seems to fix the issue of theme not being applied to the underlying `StyledFauxInput`.

hzhu: I tried removing `props.theme` (removal of line 87). I console logged `theme` throughout this file and the `theme` was available. Locally, after I removed that line 87, the UI seems to be the same as the "after" screenshots posted in the PR. Would you be able to confirm if adding `props.theme` here is needed?

jzempel: Did you alter the theme and wrap the `Select` with a custom theme provider? Without this addition, you would get a result similar to the **Before** shown in the PR description.

hzhu: I only removed `props.theme` on line 87, and I got a similar result to the *After* screenshot.

PR #692, opened 2020-04-03
feat(pagination): create CursorPagination component

hzhu:
## Description

This pull request introduces a new component, `<CursorPagination>`.

## Detail

The `CursorPagination` component is used to implement a pagination UX that is backed by [cursor-based pagination](https://blog.apollographql.com/understanding-pagination-rest-graphql-and-relay-b10f835549e7) APIs. It is a controlled component and has the following API structure (thanks @rossmoody):

```jsx
<CursorPagination aria-label="Cursor pagination">
  <CursorPagination.First>First</CursorPagination.First>
  <CursorPagination.Previous>Previous</CursorPagination.Previous>
  <CursorPagination.Next>Next</CursorPagination.Next>
  <CursorPagination.Last>Last</CursorPagination.Last>
</CursorPagination>
```

In addition to adding a new component, I have re-organized the package structure ~and renamed the original pagination component to `OffsetPagination`~:

```
src
├── index.ts         
└── elements         
  └── CursorPagination
    ├── CursorPagination.tsx
    └── components
      ├── First.tsx
      ├── Previous.tsx
      ├── Next.tsx
      └── Last.tsx
  └── Pagination
    └── Pagination.tsx
└── styled          
    ├── index.ts
  └── CursorPagination
    ├── StyledCursorPagination.ts
    ├── StyledIcon.ts
    ├── StyledFirst.ts
    ├── StyledPrevious.ts
    ├── StyledNext.ts
    └── StyledLast.ts
  └── Pagination 
    ├── StyledGap.ts
    ├── StyledNavigation.ts        
    ├── StyledPage.ts    
    └── StyledPagination.ts
```

✏️🗒I tried to limit the changes to the original `<Pagination>` component to file organization. There can be a follow-up PR to tidy up if need. One such cleanup could be to include a `<nav>` along with the `OffsetPagination` component.

## Demo

The demo for this pull request can be found here: https://cursor-pagination.netlify.com/pagination/

## Checklist

      designer as a reviewer)


packages/pagination/src/styled/OffsetPagination/StyledPage.ts
```diff
@@ -91,7 +91,7 @@ export const StyledPage = styled.li.attrs({
   visibility: ${props => props.hidden && 'hidden'};
   margin-left: ${props => `${props.theme.space.base}px`};
   border-radius: ${props => props.theme.borderRadii.md};
-  cursor: pointer;
+  cursor: ${props => !(props as any).disabled && 'pointer'};
```
hzhu, line 94: These props are typed `as any` because `StyledCursor` (button) composes from `StyledPage` (not button):

```jsx
const StyledCursor = styled(StyledPage)
```

An alternative solution is to use Garden’s buttons from the `react-buttons` package. I went with what was discussed in person, but am open to suggestions.

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,112 @@ initialState = {
...
+          <StyledFlexContainer>
+            {circles.map((circle, index) =>
+              index === cursor ? (
+                <StyledCircleFill key={index} />
+              ) : (
+                <StyledCircleStroke key={index} />
+              )
+            )}
+          </StyledFlexContainer>
+          <VisuallyHidden aria-live="polite">{`The cursor is ${cursor}`}</VisuallyHidden>
```
hzhu, line 180: This visually hidden element provides audio affordance for users who visit the example page on assistive devices.

packages/pagination/src/styled/pagination/StyledPage.ts
```diff
@@ -91,7 +94,7 @@ export const StyledPage = styled.li.attrs({
   visibility: ${props => props.hidden && 'hidden'};
   margin-left: ${props => `${props.theme.space.base}px`};
   border-radius: ${props => props.theme.borderRadii.md};
-  cursor: pointer;
+  cursor: ${props => !(props as any).disabled && 'pointer'};
```
hzhu, line 97: These props are typed `as any` because `StyledCursor` (button) composes from `StyledPage` (not button):

```jsx
const StyledCursor = styled(StyledPage)
```

An alternative solution is to use Garden’s buttons from the `react-buttons` package. I went with what was discussed in person, but I am open to suggestions.

packages/pagination/README.md
```diff
@@ -14,6 +14,8 @@ npm install react react-dom prop-types styled-components @zendeskgarden/react-th
 
 ## Usage
 
+### Offset Pagination
+
 ```jsx static
 import { ThemeProvider } from '@zendeskgarden/react-theming';
 import { Pagination } from '@zendeskgarden/react-pagination';
```
rossmoody, line 21: I know it's a larger refactor, but if we're going to commit to this naming for this component, shouldn't it be <OffsetPagination> for the component name (instead of <Pagination>)? This will stop us from having the same effect we have for Notifications where we have a Notification component inside the Notifications package and it can be confusing to tell what you're referring to.

rossmoody:
From a style standpoint, things are looking great. My only remaining questions would be how we handle the ancillary items:
1. The renaming of Pagination [(See my comment)](https://github.com/zendeskgarden/react-components/pull/692#discussion_r405003616)
2. The disabled styling/functionality consistency between Cursor and Offset (instead of disappearing)

packages/pagination/README.md
hzhu: I originally had renamed this to `OffsetPagination` because I think the updated name helps clarify the `react-pagination` package. It makes it explicit that there two patterns to choose from depending on the use case.

Unfortunately renaming `Patination` to `OffsetPagination` would result in a breaking change that would require a major version bump — which isn't planned for the upcoming releases. I really like the idea of the rename though, and would love to suggest discussing it for the next major version.

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,181 @@ initialState = {
...
+      <Row alignItems="center">
+        <Col alignSelf="center" className="u-mb u-mt">
+          <FlexContainer>
+            {circles.map((circle, index) =>
+              index === cursor ? <CircleFill key={index} /> : <CircleStroke key={index} />
+            )}
+          </FlexContainer>
+          <VisuallyHidden aria-live="polite">{`The cursor is ${cursor}`}</VisuallyHidden>
+          <CursorPagination aria-label="Cursor pagination">
+            <CursorPagination.First onClick={onFirst} disabled={cursor === 0}>
```
hzhu, line 255: `CursorPagination.First`, `CursorPagination.Previous`, `CursorPagination.Next`, and `CursorPagination.Last` are rendered as `<button>` elements within a `<nav>`. I often render a list in a `nav` but decided on using buttons.

> A nav element doesn't have to contain a list, it can contain other kinds of content as well.

https://html.spec.whatwg.org/multipage/sections.html#the-nav-element

hzhu:
> 2. The disabled styling/functionality consistency between Cursor and Offset (instead of disappearing)

@rossmoody I aimed to scope this pull request to create the new cursor pagination component. I think we could do that as a follow-on PR or can put this PR on hold while that gets worked on.

Reconciling the differences between the two pagination component's "disabled" state will require some work. The challenge is that `Pagination` (offset) "buttons" are rendered as [`li` elements](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element), and I believe `li` elements [does not have](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLIElement) the `disabled` attribute. 

Ideally, we could refactor the component so that the `li` elements contain a `button` element. Then that can be the element that is disabled when the user reaches the first page (or last).

packages/pagination/README.md
rossmoody: Ah, interesting. Do we have a running list of planned major version changes?

rossmoody: I'm actually quite interested in this workflow. It seems a real shame to indefinitely punt refactor changes like this when they play a rather important role in how everyone learns to talk about the components. If it's a necessity than it is what it is but maybe we start a Jira epic?

rossmoody:
> > 1. The disabled styling/functionality consistency between Cursor and Offset (instead of disappearing)
> 
> @rossmoody I aimed to scope this pull request to create the new cursor pagination component. I think we could do that as a follow-on PR or can put this PR on hold while that gets worked on. I agree it should be done 👍.
> 
> Reconciling the differences between the two pagination component's "disabled" state will require some work. The challenge is that `Pagination` (offset) "buttons" are rendered as [`li` elements](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element), and I believe `li` elements [does not have](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLIElement) the `disabled` attribute.
> 
> Ideally, we could refactor the component so that the `li` elements contain a `button` element. Then that can be the element that is disabled when the user reaches the first page (or last).

Sounds good. My biggest concern is that it gets punted as a follow-on and then not addressed because it isn't an initiative that would typically warrant priority.

packages/pagination/README.md
jzempel: @rossmoody we'll track and commit to this in a staggered manner. Something like first deprecating `Pagination` and providing a non-deprecated `OffsetPagination` alias. And then subsequently removing `Pagination` in a future, major v9.

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,181 @@ initialState = {
...
+const StyledCircle = styled.div`
+  margin: 0 ${props => props.theme.space.base}px;
+  border-radius: 100%;
+  height: ${props => props.theme.space.base * 3}px;
+  width: ${props => props.theme.space.base * 3}px;
+`;
+
+const CircleStroke = styled(StyledCircle)`
+  border: ${props =>
+    `0.6px ${props.theme.borderStyles.solid} ${getColor('grey', 300, props.theme)}`};
```
jzempel, line 116: Add `box-sizing: border-box;` to `StyledCircle` and change the border-size to `1px`. Even better, add:

```css
box-sizing: border-box;
border: ${props => props.theme.borders.sm} transparent;
```
to `StyledCircle` and `CircleStroke` would only be:

```css
border-color: ${props => getColor('grey', 300, props.theme)};
```

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,181 @@ initialState = {
...
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+  white-space: nowrap;
+  word-wrap: normal;
+`;
```
jzempel, line 140: Prefer the more restrained styling provided by https://github.com/zendeskgarden/css-components/blob/cff98fbf952f118ac60413ce4d9a24351872777d/packages/utilities/src/visibility.css#L18-L27 or just replace below with `<div className="u-visibility-screenreader" aria-live="polite">`, as the utility classes are available in Styleguidist.

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,181 @@ initialState = {
...
+  margin: 0 ${props => props.theme.space.base}px;
+  border-radius: 100%;
+  height: ${props => props.theme.space.base * 3}px;
+  width: ${props => props.theme.space.base * 3}px;
+`;
+
+const CircleStroke = styled(StyledCircle)`
+  border: ${props =>
+    `0.6px ${props.theme.borderStyles.solid} ${getColor('grey', 300, props.theme)}`};
+`;
```
jzempel, line 202: Use simplified border treatment mentioned above.

packages/pagination/examples/basic.md
```diff
@@ -93,3 +95,181 @@ initialState = {
...
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+  white-space: nowrap;
+  word-wrap: normal;
+`;
```
jzempel, line 225: Used simplified utility treatment mentioned above.

packages/pagination/examples/basic.md
jzempel: I think this makes sense. Although, we may want to think about reconciling the unexpected structural difference between `Pagination` and `CursorPagination`.

packages/pagination/src/elements/CursorPagination/components/Previous.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { HTMLAttributes } from 'react';
+import ChevronLeftIcon from '@zendeskgarden/svg-icons/src/16/chevron-left-stroke.svg';
+import { StyledIcon, StyledCursor } from '../../../styled';
+
+export const Previous = React.forwardRef<HTMLButtonElement, HTMLAttributes<HTMLButtonElement>>(
+  ({ children, ...restProps }, ref) => {
```
jzempel, line 20: [nit] we tend to use `...other` vs `...restProps` throughout the codebase.

packages/pagination/src/elements/CursorPagination/components/Last.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
```
jzempel, line 13: Remove duplicate.

packages/pagination/src/elements/CursorPagination/components/First.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
```
jzempel, line 14: Remove duplicate.

packages/pagination/README.md
rossmoody: Niceeeeeeeee

packages/pagination/src/styled/Pagination/StyledPage.ts
```diff
@@ -111,7 +114,7 @@ export const StyledPage = styled.li.attrs({
     font-weight: ${props => props.theme.fontWeights.semibold};
   }
 
-  ${props => colorStyles(props)};
+  ${props => !(props as any).disabled && colorStyles(props)};
```
jzempel, line 117: Can we take a different approach? Let's extract out a common set of styling values to a `StyledBase` (or some improved name) component. Then both `StyledPage` and `StyledCursor` would extend that base component. This should end up being more intuitive vs. the fighting resets that are happening between StyledPage and StyledCursor. Also, the `disabled` prop should be properly typed within the scope of `StyledCursor`.

packages/pagination/src/styled/CursorPagination/StyledCursor.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledPage } from '../Pagination/StyledPage';
+
+const COMPONENT_ID = 'cursor_pagination.cursor';
+
+/**
+ * 1. Reset CSS from <button>.
+ * 2. Reset CSS from <StyledPage>.
```
jzempel, line 16: See comment on `StyledPage`. The `/* [2] */` resets should not be needed here.

packages/pagination/src/styled/Pagination/StyledPage.ts
jzempel: btw, if `disabled` is retained in `StyledPage` (for future use), we should take a different approach. Do not rely on UA styling for the color. Rather, all disabled-related styling should be applied at the end of the `colorStyles` function – probably to both the `:disabled` psuedo and the `[aria-disabled='true']` attributes. See https://github.com/zendeskgarden/react-components/blob/ba6076878aa729383b285e76ba37273054bfe215/packages/buttons/src/styled/StyledButton.ts#L132-L136 for details.

packages/pagination/examples/basic.md
hzhu: Awesome, I went with the utility classes provided by Styleguidist. Thanks for the suggestion 👍.

packages/pagination/examples/basic.md
hzhu: Added. Thanks for the suggestion 👍.

packages/pagination/src/styled/Pagination/StyledPage.ts
hzhu: > Can we take a different approach?

Yeah, I would love to! I originally had something like this in mind, but steered away from modifying `StyledPage`. 

I've moved all shared styling into a file called `StyledPageBase` (open to name ideas) and have both `StyledCursor` and `StyledBase` extending from the base styles. I've pushed up a new commit for this and updated the [demo](https://cursor-pagination.netlify.com/pagination/). Let me know what you think of these changes.

packages/pagination/src/styled/Pagination/StyledPageBase.ts
```diff
@@ -5,8 +5,12 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
+// removed from shared css file
+// margin: 0; /* [2] */
+// min-width: unset; /* [2] */
+// max-width: none; /* [2] */
+
```
jzempel, line 12: Remove leftover comments.

packages/pagination/src/styled/Pagination/StyledPageBase.ts
```diff
@@ -111,15 +111,11 @@ export const StyledPage = styled.li.attrs({
     font-weight: ${props => props.theme.fontWeights.semibold};
   }
 
-  ${props => colorStyles(props)};
-
-  &${props => (props.theme.rtl ? ':last-of-type' : ':first-of-type')} {
-    margin-left: 0;
-  }
+  ${props => !(props as any).disabled && colorStyles(props)};
```
jzempel, line 114: Always call `colorStyles`. Add the following to the end of the `colorStyles` function:

```css
:disabled,
[aria-disabled='true'] {
  background-color: transparent;
  color: /* applicable color overrides go here */;
}
```

Otherwise the code is relying on UA color treatment – which isn't right.

packages/pagination/src/styled/Pagination/StyledPage.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * 1. Text truncation.
+ */
+export const StyledPage = styled(StyledPageBase).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  margin-left: ${props => `${props.theme.space.base}px`};
+  cursor: ${props => !(props as any).disabled && 'pointer'};
```
jzempel, line 36: Cursor is already handled by the `StyledPageBase`.

packages/pagination/src/styled/Pagination/StyledPage.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    &[aria-current='true'] {
+      max-width: none;
+    }
+  `;
+};
+
+/**
+ * 1. Text truncation.
+ */
```
jzempel, line 30: Move this comment to `sizeStyles` where it is referenced.

packages/pagination/src/styled/Pagination/StyledPageBase.ts
```diff
@@ -66,32 +73,25 @@ const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
...
-
-    &[aria-current='true'] {
-      max-width: none;
-    }
   `;
 };
 
 /**
  * 1. Text truncation.
  */
```
jzempel, line 84: Remove unused reference comment.

PR #698, opened 2020-04-08
feat(menu): revise menu spacing

rossmoody:
## Description

This PR revises and simplifies the menu item spacing between compact and default sizes in preparation for Select Media Items. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/examples/tree-menu.md
+++ b/packages/dropdowns/examples/tree-menu.md
@@ -65,7 +65,7 @@ const renderItems = () => {
   <Trigger>
     <Button>Tree Layout</Button>
   </Trigger>
-  <Menu placement="end" hasArrow style={{ width: 200, height: 258 }}>
+  <Menu placement="end" hasArrow>
     {renderItems()}
   </Menu>
 </Dropdown>;
--- a/packages/dropdowns/src/styled/items/StyledItemMeta.ts
+++ b/packages/dropdowns/src/styled/items/StyledItemMeta.ts
@@ -23,7 +23,7 @@ export const StyledItemMeta = styled.span.attrs({
   'data-garden-version': PACKAGE_VERSION
 })<IStyledItemMetaProps>`
   display: ${props => (props.isCompact ? 'none' : 'block')};
-  line-height: ${props => props.theme.space.base * 4}px;
+  line-height: ${props => props.theme.lineHeights.sm};
   color: ${props => getColor('neutralHue', props.isDisabled ? 400 : 600, props.theme)};
   font-size: ${props => props.theme.fontSizes.sm};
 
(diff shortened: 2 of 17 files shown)


packages/dropdowns/examples/tree-menu.md
```diff
@@ -65,7 +65,7 @@ const renderItems = () => {
   <Trigger>
     <Button>Tree Layout</Button>
   </Trigger>
-  <Menu placement="end" hasArrow style={{ width: 200, height: 258 }}>
+  <Menu placement="end" hasArrow>
```
austin94, line 68: Why is this removed? Without a fixed height, the size of the tree will change based on the number of items.

rossmoody: When the padding on the list items was revised it resulted in a bunch of empty space at the bottom on the menu. I was going to try to do the math to figure out the new menu height but it didn't seem like a very good practice from a responsive standpoint to hard code widths and heights in the examples we provide.

In general glad you called this out, had a question: is a reason why we are choosing to replace the menu in nested navigation vs showing a new one? Something like [this example](https://share.getcloudapp.com/z8uxA492)? That's what I have typically seen for this behavior and it would eliminate this height jumping.

austin94: That pattern only works on desktop where there is enough space to nest relative to the other elements. This was the design decision chosen at the time of development.

> I was going to try to do the math to figure out the new menu height but it didn't seem like a very good practice

With the ability for the menu content to overflow this is still responsive. Since this is a separate design decision I'm not sure we should be adding this here.

rossmoody: I'm unsure who to defer to here. I would advocate for not having explicit heights inline in the example if we don't have to but if design has traditionally said to do that then I guess I need to file a Jira ticket for it and appeal to someone?

jzempel: @rossmoody, I'll work together with you on this one to help get it right.

packages/dropdowns/src/styled/items/header/StyledHeaderIcon.ts
```diff
@@ -34,13 +24,12 @@ export const StyledHeaderIcon = styled.div.attrs({
   bottom: 0;
   align-items: center;
   justify-content: center;
-  ${props => (props.theme.rtl ? 'right' : 'left')}: ${props =>
-    `calc(${math(`${getItemPaddingHorizontal(props)} - ${getIconSize(props)}`)} / 2)`};
+  ${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 3}px;
```
jzempel, line 27: Nice fix here. Previous resulted in `left: calc(8px);`

packages/dropdowns/src/styled/items/StyledItemMeta.ts
```diff
@@ -23,7 +23,7 @@ export const StyledItemMeta = styled.span.attrs({
   'data-garden-version': PACKAGE_VERSION
 })<IStyledItemMetaProps>`
   display: ${props => (props.isCompact ? 'none' : 'block')};
-  line-height: ${props => props.theme.space.base * 4}px;
+  line-height: ${props => props.theme.lineHeights.sm};
```
jzempel, line 26: @rossmoody outside of `typography` we never reference the `lineHeights` theming object. This allows `space.base` to take maximum effect across the component base.

PR #699, opened 2020-04-08
chore(loaders): update Dots API descriptions

rossmoody:
## Description
This is the first submission as part of the Propstravaganza initiative to align API descriptions to be consistent in structure and generally understood by a designer audience.

I think future PRs might include more than one component at a time, but I wanted to see if the structure of this PR is correct or if anything could be improved before continuing on. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/loaders/src/elements/Dots.tsx
+++ b/packages/loaders/src/elements/Dots.tsx
@@ -21,20 +21,19 @@ const COMPONENT_ID = 'loaders.dots';
 
 interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Size of the loader. Can inherit from `font-size` styling.
+   * Sets the `height` and `width` in pixels. Inherits the parent `font-size` by default
    **/
   size?: string | number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill `color`. Inherits the parent font `color` by default
    **/
   color?: string;
   /**
-   * Duration (ms) of the animation. Default is 1250ms.
+   * Sets the speed of the loading animation in milliseconds
    **/
   duration?: number;
   /**
-   * Delay in MS to begin loader rendering. This helps prevent
-   * quick flashes of the loader during normal loading times.
+   * Delays showing the loader to prevent a quick render flash during normal loading times. Time is in milliseconds
    **/
   delayMS?: number;
 }


packages/loaders/src/elements/Dots.tsx
```diff
@@ -21,20 +21,19 @@ const COMPONENT_ID = 'loaders.dots';
...
    **/
   size?: string | number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill color
    **/
   color?: string;
   /**
-   * Duration (ms) of the animation. Default is 1250ms.
+   * Increases or decreases the speed of the loading animation in milliseconds
```
hzhu, line 32: Alternatively could be something along the lines of _"Sets a new loading animation duration in milliseconds"_. I think from a dev perspective this would be more accurate, since adding a duration prop is replacing the default duration value and not mutating it.

rossmoody: For me, what I found most interesting about this prop after fiddling is that it's not actually changing the total duration of time the loader runs, but rather how fast or slow the loader animates once. This is tricky cause I realize `duration` from a css perspective is how fast or slow it takes to complete once, but from a plain language perspective I thought it specified how long it would run in general. 

Maybe: 
1. Sets the speed of the loading animation in milliseconds
2. Sets the time to complete one loading animation in milliseconds

jzempel:
- Structure the title per https://www.conventionalcommits.org/en/v1.0.0/. In this case `fix(loaders): update Dots API descriptions`. Or, maybe more appropriately: `chore(loaders): update Dots API descriptions` (since this isn't a bug fix).
- Always be sure to add a PR label as prompted for in the PR template. These should be `PR: Docs`.

packages/loaders/src/elements/Dots.tsx
```diff
@@ -21,20 +21,19 @@ const COMPONENT_ID = 'loaders.dots';
 
 interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Size of the loader. Can inherit from `font-size` styling.
+   * Sets the height and width in pixels. Inherits the parent `font-size` by default
    **/
   size?: string | number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill color
```
jzempel, line 28: It's helpful engineering documentation to indicate that the fill inherits from font `color` by default.

rossmoody: Nice, will amend.

rossmoody:
@jzempel Holy smokes I butchered that title originally.

packages/loaders/src/elements/Dots.tsx
rossmoody: Went with 1, lemme know if we should give it another go

hzhu: Oh yeah that (1) sounds good!

packages/loaders/src/elements/Dots.tsx
```diff
@@ -21,20 +21,19 @@ const COMPONENT_ID = 'loaders.dots';
 
 interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Size of the loader. Can inherit from `font-size` styling.
+   * Sets the height and width in pixels. Inherits the parent `font-size` by default
    **/
   size?: string | number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill color. Inherits the parent font color by default
```
jzempel, line 28: Use backtick `color` for consistency with ☝️

PR #700, opened 2020-04-08
chore(loaders): update remaining Loaders API descriptions

rossmoody:
## Description
The remaining Loader API descriptions for Skeleton, Progress, Inline, and Spinner. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/loaders/src/elements/Inline.tsx
+++ b/packages/loaders/src/elements/Inline.tsx
@@ -20,8 +20,9 @@ const InlineTypingIndicator: React.FC<IStyledTypingSvgProps> = props => {
 };
 
 interface IInlineProps extends React.HTMLAttributes<SVGSVGElement> {
-  /* Width of the loader in px */
+  /** Sets the width in pixels and scales the loader proportionally */
   size?: number;
+  /** Sets the fill color. Inherits the parent font `color` by default */
   color?: string;
 }
 
--- a/packages/loaders/src/elements/Progress.tsx
+++ b/packages/loaders/src/elements/Progress.tsx
@@ -12,9 +12,11 @@ import { StyledProgressBackground, StyledProgressIndicator } from '../styled';
 const COMPONENT_ID = 'loaders.progress';
 
 interface IProgressProps extends HTMLAttributes<HTMLDivElement> {
-  /** The progress as a value between 0 and 100 */
+  /** Sets the progress as a value between 0 and 100 */
   value?: number;
+  /** Sets the foreground bar fill color. Defaults to the `successHue` theme value */
   color?: string;
+  /** Sets the height of the progress bar. Can be: “small”, “medium”, or “large” */
   size?: 'small' | 'medium' | 'large';
 }
 
--- a/packages/loaders/src/elements/Skeleton.tsx
+++ b/packages/loaders/src/elements/Skeleton.tsx
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
 import { StyledSkeleton } from '../styled';
 
 interface ISkeletonProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** Sets the width as a percentage of the parent element */
   width?: string;
+  /** Sets the height of the loader as a percentage of parent element when not inherited by `line-height`. Default is .6 of `line-height` */
   height?: string;
-  /** Apply light styling for use on dark backgrounds */
+  /** Inverts the loader color for use on dark backgrounds */
   isLight?: boolean;
 }
 
--- a/packages/loaders/src/elements/Spinner.tsx
+++ b/packages/loaders/src/elements/Spinner.tsx
@@ -47,20 +47,19 @@ const computeFrames = (
 
 interface ISpinnerProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Size of the loader. Can inherit from `font-size` styling.
+   * Sets the height and width in pixels. Inherits the parent `font-size` by default
    **/
   size?: string;
   /**
-   * Duration (ms) of the animation. Default is 1250ms.
+   * Sets the speed of the loading animation in milliseconds
    **/
   duration?: number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill color. Inherits the parent font `color` by default
    **/
   color?: string;
   /**
-   * Delay in MS to begin loader rendering. This helps prevent
-   * quick flashes of the loader during normal loading times.
+   * Delays showing the loader to prevent a quick render flash during normal loading times. Time is in milliseconds
    **/
   delayMS?: number;
 }
--- a/packages/loaders/src/elements/Dots.tsx
+++ b/packages/loaders/src/elements/Dots.tsx
@@ -21,11 +21,11 @@ const COMPONENT_ID = 'loaders.dots';
 
 interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Sets the `height` and `width` in pixels. Inherits the parent `font-size` by default
+   * Sets the height and width in pixels. Inherits the parent `font-size` by default
    **/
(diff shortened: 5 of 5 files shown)


packages/loaders/src/elements/Inline.tsx
```diff
@@ -20,8 +20,9 @@ const InlineTypingIndicator: React.FC<IStyledTypingSvgProps> = props => {
 };
 
 interface IInlineProps extends React.HTMLAttributes<SVGSVGElement> {
-  /* Width of the loader in px */
+  /** Sets the size proportionally based on width designation. Number is in pixels */
```
rossmoody, line 23: I don't love this one. Maybe:

1. Sets the size. Number specifies width in pixels
2. Sets the width in pixels and scales the loader proportionally

packages/loaders/src/elements/Progress.tsx
```diff
@@ -12,9 +12,11 @@ import { StyledProgressBackground, StyledProgressIndicator } from '../styled';
 const COMPONENT_ID = 'loaders.progress';
 
 interface IProgressProps extends HTMLAttributes<HTMLDivElement> {
-  /** The progress as a value between 0 and 100 */
+  /** Sets the progress as a value between 0 and 100 */
   value?: number;
+  /** Sets the foreground bar fill `color`. Defaults to the `successHue` theme value */
   color?: string;
+  /** Sets the `height` of the progress bar. Can be: “small”, “medium”, or “large” */
```
hzhu, line 19: The usage of the backticks in this description confused me initially. Does the word “height” in this description refer to the CSS attribute or the characteristic of the `Progress` element? 

I read this as it referring the characteristic of an element, particularly because “small”, “medium”, and “large” aren’t valid values for CSS `height` attribute.

hzhu: On alternative for this description could be:

> Sets the height of the progress bar.

This re-phrase does two things. 1) height now refers to a characteristic of the element, rather than a CSS code value. 2) "small", "medium", and "large" is not communicated twice since it will be communicated in the `type` section of this prop. Curious what ya'lls thoughts are.

rossmoody: I think in this case it actually does set the `height` but I agree, I like prioritizing the characteristic over the property. Will amend. 

The second part you called out is something I wanted to ask the group (from the [API Template doc](https://docs.google.com/document/d/19QvlF4MDKFfUKzHeUfrKIS9856GyTWvAXJwoSKRKX-Q/edit#heading=h.70kaxhzb08a8)). It's a little forward thinking and will result in some initial redundancy but[ this is what I'm proposing the formatting in the long game might be.](https://share.getcloudapp.com/P8uE0wLk) Would be curious if we could make it better.

hzhu: > I think in this case it actually does set the height

Oh yes, it does set the `height`. However `large`, `medium`, etc isn't a valid `height` value. I think the prop description can communicate that the visual height changed, and not necessarily refer to the CSS attribute `height` — which initially confused me since it sentence read like `height` was a characteristic of the element but it also had code backticks. For example, the visual height can be changed using the `transform` property. Then it could be

> Sets the `transform: scaleY` of the progress bar.

but instead of exposing the implementation details in the description, it could be phrased as:

> Sets the height of the progress bar.

And the underlying implementation can use `height`, `transform`, etc.

packages/loaders/src/elements/Inline.tsx
```diff
@@ -20,8 +20,9 @@ const InlineTypingIndicator: React.FC<IStyledTypingSvgProps> = props => {
 };
 
 interface IInlineProps extends React.HTMLAttributes<SVGSVGElement> {
-  /* Width of the loader in px */
+  /** Sets the width in pixels and scales the loader proportionally */
   size?: number;
+  /** Sets the fill `color`. Inherits the parent font `color` by default */
```
jzempel, line 25: The use of code backticks feels a bit inconsistent. I think they make sense where we are explicitly referring to a CSS property – less so when there is an implicit reference. To mitigate, I'd suggest the following:

"Sets the fill _color_. Inherits the parent font `color` by default"

packages/loaders/src/elements/Progress.tsx
```diff
@@ -12,9 +12,11 @@ import { StyledProgressBackground, StyledProgressIndicator } from '../styled';
 const COMPONENT_ID = 'loaders.progress';
 
 interface IProgressProps extends HTMLAttributes<HTMLDivElement> {
-  /** The progress as a value between 0 and 100 */
+  /** Sets the progress as a value between 0 and 100 */
   value?: number;
+  /** Sets the foreground bar fill `color`. Defaults to the `successHue` theme value */
```
jzempel, line 17: "Sets the foreground bar fill _color_."

packages/loaders/src/elements/Skeleton.tsx
```diff
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
 import { StyledSkeleton } from '../styled';
 
 interface ISkeletonProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** Sets the `width` as a percentage of the parent element */
```
jzempel, line 13: _width_ vs `width`

packages/loaders/src/elements/Skeleton.tsx
```diff
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
...
   width?: string;
+  /** Sets the `height` of the loader as a percentage of parent element when not inherited by `line-height`. Default is .6 of `line-height` */
```
jzempel, line 15: _height_ vs `height`

packages/loaders/src/elements/Skeleton.tsx
```diff
@@ -10,9 +10,11 @@ import PropTypes from 'prop-types';
...
   height?: string;
-  /** Apply light styling for use on dark backgrounds */
+  /** Inverts the loader `color` for use on dark backgrounds */
```
jzempel, line 17: _color_ vs `color`

packages/loaders/src/elements/Spinner.tsx
```diff
@@ -47,20 +47,19 @@ const computeFrames = (
...
    **/
   size?: string;
   /**
-   * Duration (ms) of the animation. Default is 1250ms.
+   * Sets the speed of the loading animation in milliseconds
    **/
   duration?: number;
   /**
-   * Color of the loader. Can inherit from `color` styling.
+   * Sets the fill `color`. Inherits the parent font `color` by default
```
jzempel, line 58: "Sets the fill _color_. Inherits the parent font `color` by default"

packages/loaders/src/elements/Inline.tsx
rossmoody: I like this way better.

packages/loaders/src/elements/Spinner.tsx
```diff
@@ -47,20 +47,19 @@ const computeFrames = (
 
 interface ISpinnerProps extends React.HTMLAttributes<SVGSVGElement> {
   /**
-   * Size of the loader. Can inherit from `font-size` styling.
+   * Sets the `height` and `width` in pixels. Inherits the parent `font-size` by default
```
jzempel, line 50: Should this be "Sets the _height_ and _width_ in pixels. Inherits the parent `font-size` by default"?

jzempel: ...since we're not committing to using `height` and `width` CSS properties for sizing.

rossmoody: Thanks for the fine tooth comb on this one. My eyes are going blind to back ticks.

rossmoody: With fresh eyes, fixed up `Dots` to align with this as well.

PR #712, opened 2020-04-09
fix(accordions): update height when content scrollHeight changes

hzhu:
## Description

This PR updates the `maxHeight` of the vertical `Stepper`'s inner content when the inner content's `scrollHeight` changes.

### Problem

The Stepper’s inner content height is [calculated and applied](https://github.com/zendeskgarden/react-components/blob/master/packages/accordions/src/elements/stepper/components/Content.tsx#L23) in order to animate the Stepper’s inner content. However, **when the inner content’s `scrollHeight` changes, the height is not updated, resulting in a UI bug where the Stepper content has the old height value**. In the following screenshot examples, this problem is demonstrated by adjusting the browser window, which in turn changes the inner content’s `scrollHeight` as its text content readjusts to fit the new width.

### Solution

**The inner content’s `maxHeight` needs to be updated to the inner content’s `scrollHeight` when the inner content’s `scrollHeight` changes**. There are two known scenarios where the inner content’s `scrollHeight` changes: 1) When the window width is adjusted, and text flow adjusts to fill in the new container width. This is demonstrated in the screenshots above. 2) When the content of the inner content changes. For example, if new elements were added or removed from the inner content.

### Screenshots of The Problem

When the `Stepper` is initially rendered (small window width size), the `maxHeight` on the inner content is dynamically and correctly set to the inner content’s `scrollHeight`:

<kbd>
Screen Shot 2020-04-08 at 10 51 19 AM
</kbd>

When the inner content's text re-flows and the element's `scrollHeight` grows larger due to resizing the window, the inner content’s `maxHeight` is not updated, which results in more height than necessary. Notice that the height is too large, and there is empty space in the bottom:

<kbd>
Screen Shot 2020-04-08 at 10 50 42 AM
</kbd>

## Demo of Solution

The demo for the solution can be found here: https://stepper-content-fix.netlify.com/accordions/#basic

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/accordions/package.json
+++ b/packages/accordions/package.json
@@ -33,8 +33,10 @@
     "styled-components": "^4.2.0"
   },
   "devDependencies": {
+    "@types/lodash.debounce": "4.0.6",
     "@zendeskgarden/react-theming": "^8.4.1",
-    "@zendeskgarden/svg-icons": "6.15.0"
+    "@zendeskgarden/svg-icons": "6.15.0",
+    "lodash.debounce": "4.0.8"
   },
   "keywords": [
     "accordions",
--- a/packages/accordions/.size-snapshot.json
+++ b/packages/accordions/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/index.cjs.js": {
-    "bundled": 16291,
-    "minified": 11501,
-    "gzipped": 2855
+    "bundled": 21234,
+    "minified": 13893,
+    "gzipped": 3828
   },
   "dist/index.esm.js": {
-    "bundled": 15599,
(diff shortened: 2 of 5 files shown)


packages/accordions/package.json
```diff
@@ -33,8 +33,10 @@
     "styled-components": "^4.2.0"
   },
   "devDependencies": {
+    "@types/lodash.debounce": "^4.0.6",
```
austin94, line 36: Should the `lodash.debounce` dependencies be pinned?

hzhu: Oops, I'll go ahead and update this, thanks for catching this!

PR #722, opened 2020-04-14
feat(typography): add bold option

jzempel:
## Description

All typography t-shirt sizes now expose an `isBold` prop for applying a semibold font-weight to the text content.

## Detail

Demo pre-published for review at https://affectionate-curran-d65162.netlify.com/typography/#typography

FWIW, this is how semibold treatment looks on Android (Samsung Galaxy S10):

Screen Shot 2020-04-14 at 1 03 01 PM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/typography/.size-snapshot.json
+++ b/packages/typography/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/index.cjs.js": {
-    "bundled": 16531,
-    "minified": 12033,
-    "gzipped": 2666
+    "bundled": 16687,
+    "minified": 12165,
+    "gzipped": 2674
   },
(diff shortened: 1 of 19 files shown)


jzempel:
@austin94 this is only stylistic. Garden's only treatment for `bold` is the semibold weight number of 600. That weight number may change in the future, but the prop should not. Btw, Garden `react-components` will not offer an italics equivalent as we don't use that in products.

We definitely do not want to couple semantic tags such as `strong` or `em` with stylistic treatment. The user should have the freedom to use strong or em against any style of text. With design approval, we'll be updating `css-bedrock` accordingly.

PR #723, opened 2020-04-14
chore(tabs): update Tabs API descriptions

rossmoody:
## Description
The new Tabs API descriptions. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/tabs/src/elements/Tab.tsx
+++ b/packages/tabs/src/elements/Tab.tsx
@@ -12,15 +12,18 @@ import { StyledTab } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 export interface ITabProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Disables the Tab and prevents user interaction
+   */
   disabled?: boolean;
   /**
-   * A value to uniquely identify a Tab. Provided to the `onChange` event.
+   * Sets a value to uniquely identify a Tab. Provided to the `onChange` event
    */
   item?: any;
 }
 
 /**
- * Accepts all `<div>` props
+ * Accepts all `<div>` props.
  */
 export const Tab = React.forwardRef<HTMLDivElement, ITabProps>(
   ({ disabled, item, ...otherProps }, ref) => {
--- a/packages/tabs/src/elements/Tab.tsx
+++ b/packages/tabs/src/elements/Tab.tsx
@@ -13,11 +13,11 @@ import { useTabsContext } from '../utils/useTabsContext';
 
 export interface ITabProps extends HTMLAttributes<HTMLDivElement> {
   /**
-   * Disables the Tab and prevents user interaction
+   * Disables the `Tab` and prevents user interaction
    */
   disabled?: boolean;
   /**
-   * Sets a value to uniquely identify a Tab. Provided to the `onChange` event
+   * Defines a value to uniquely identify the `Tab`. Provided to the `onChange` event in `Tabs`
    */
   item?: any;
 }
--- a/packages/tabs/src/elements/TabList.tsx
+++ b/packages/tabs/src/elements/TabList.tsx
@@ -10,7 +10,7 @@ import { StyledTabList } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 /**
- * Accepts all `<div>` props
+ * Provides styling to contained Tab items. Accepts all `<div>` props.
  */
 export const TabList = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
   (props, ref) => {
--- a/packages/tabs/src/elements/TabList.tsx
+++ b/packages/tabs/src/elements/TabList.tsx
@@ -10,7 +10,7 @@ import { StyledTabList } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 /**
- * Provides styling to child `Tab` components. Accepts all `<div>` props.
+ * Accepts all `<div>` props.
  */
 export const TabList = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
   (props, ref) => {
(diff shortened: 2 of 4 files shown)


packages/tabs/src/elements/TabList.tsx
```diff
@@ -10,7 +10,7 @@ import { StyledTabList } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 /**
- * Accepts all `<div>` props
+ * Provides styling to child `Tab` components. Accepts all `<div>` props.
```
jzempel, line 13: Hmm, feels a bit limiting to say "Provides styling" as there's also a11y wrapped-up in this component.

packages/tabs/src/elements/Tabs.tsx
```diff
@@ -17,16 +17,16 @@ import { StyledTabs } from '../styled/StyledTabs';
...
   /**
-   * Currently selected tab to display
+   * Sets the currently selected `Tab` to display
    */
   selectedItem?: any;
   /**
-   * Callback for when a tab has been selected by keyboard or mouse
-   * @param {String} selectedItem - The item of the selected tab
+   * Provides a callback function that returns the `selectedItem` when a `Tab` has been selected by keyboard or mouse
+   * @param {String} selectedItem: The `item` value of the selected `Tab`
```
jzempel, line 29: What happens with this syntax? Usually, proper jsdoc syntax would be `@param {String} name Description...` with no intervening colon or dash – just space.

rossmoody: The thinking here was that the sentence following the parameter seemed like an explanation/definition of the word, so a colon seemed appropriate. I suppose since I changed it my preference is colon but dash works too. Was thinking having nothing to separate the word from the explanation might look like one complete sentence.

packages/tabs/src/elements/TabList.tsx
rossmoody: Nice catch. Maybe: 

`Provides structure and a11y attributes to...`
`Provides structures and necessary a11y to...`
`Provides styling and accessibility attributes to...`

hzhu: I think a generic description would be most suitable since HTML attributes can be applied for various reasons (styling, a11y, general web behavior). For example (very contrived), one could pass `contentEditable` as an attribute.

I think the original description communicates that well.

rossmoody: I think my initial thought was that this wrapper is more than just a `div` that accepts all `div` props. It provides a utility to intended child components within it. Does that make sense? Maybe the description I'm proposing still doesn't get that across. 

As a new user, if I ended up here without going through the code to inspect the structure of how it's used in context, I don't really know what it does. So, the hope here is to give a reader some semblance of what it's being used for. I think my goal is a reader could leave the description saying "Oh this is the wrapper component for a list of `Tab` elements."

jzempel: @rossmoody it might be helpful for you to understand why this element exists. It speaks directly to the WAI-ARIA pattern described here: https://www.w3.org/TR/wai-aria-practices/#tabpanel. The link to `tablist` contains helpful text: "A list of tab elements, which are references to tabpanel elements."

packages/tabs/src/elements/Tab.tsx
```diff
@@ -12,15 +12,18 @@ import { StyledTab } from '../styled';
...
   /**
-   * A value to uniquely identify a Tab. Provided to the `onChange` event.
+   * Defines a value to uniquely identify the `Tab`. Provided to the `onChange` event in `Tabs`
    */
   item?: any;
 }
 
 /**
- * Accepts all `<div>` props
+ * Accepts all `<div>` props.
```
jzempel, line 26: If you're going to amplify `TabList`, should this also align?

packages/tabs/src/elements/TabPanel.tsx
```diff
@@ -12,13 +12,13 @@ import { useTabsContext } from '../utils/useTabsContext';
...
   /**
-   * A value used to match a `TabPanel` with its associated Tab.
+   * Defines a value used to match a `TabPanel` with its associated `Tab`
    */
   item?: any;
 }
 
 /**
- * Accepts all `<div>` props
+ * Accepts all `<div>` props.
```
jzempel, line 21: If you're going to amplify `TabList`, should this also align?

packages/tabs/src/elements/Tabs.tsx
jzempel: I pulled your branch. We should drop the invalid `:` and worry about structural output of the data when it renders on `website`.

jzempel: Valid jsdoc here is 

```js
@param {String} selectedItem The `item` value of the selected `Tab`
```

packages/tabs/src/elements/TabList.tsx
rossmoody: Yeah! This is the missing context for me, personally. I think it seemed a shame initially (with the absence of any properties to infer use) to say nothing about the component utility except it accepts div props. It seems like this is obvious info from a dev standpoint so I'll revert this to original.

packages/tabs/src/elements/Tabs.tsx
rossmoody: Perfect, ty

packages/tabs/src/elements/Tab.tsx
```diff
@@ -12,15 +12,18 @@ import { StyledTab } from '../styled';
 import { useTabsContext } from '../utils/useTabsContext';
 
 export interface ITabProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Disables the `Tab` and prevents user interaction
```
hzhu, line 16: 👍

packages/tabs/src/elements/TabList.tsx
rossmoody: Henry was proactive and awesome and helped me get a more appropriate/rounded sense of goals for API takeaways. This type of granular commentary would cause problems. ty ty team

PR #725, opened 2020-04-21
feat(accordions): add new accordion component

hzhu:
## Description

This pull request introduces a new `Accordion` component.

### Styleguidist Accordion Examples

The pull request proposes to demonstrate two examples in Styleguidist: a basic and an advanced example. The advanced demo shows the flexibility of the `Accordion.Header` component. It demonstrates the ability to add and style additional textual elements to the underlying button element. It also demonstrates the ability to add menu buttons (or any other elements) into the header.

The examples strive not to couple form controls to composable children because it may complicate the demo code (among other reasons). The Styleguidist example shows additional menu buttons and header text elements in a separate advanced example, rather than being toggled with a "knob" control.

✨After creating this PR, I realized that the demo should show uncontrolled usage as the basic example and **controlled usage as an advanced example**. In this type of demonstration, the controlled component _could_ demonstrates accordion behaviors such as an "Expand All" or "Collapse All" toggle, which isn't possible with an uncontrolled accordion. Even without toggles such as "Expand All" or "Collapse All", the advanced example can be driven by controlled props.

**Edit: I have updated the "Advanced" accordion example to demonstrate controlled usage.**

### Accordion Arrow Key Behavior

Some accordion widgets can have their header buttons navigated using `ArrowUp` and `ArrowDown` keys. The W3 WAI-ARIA authoring practices suggest that behavior [is optional](https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction). Although optional, that behavior is something that @steue and I think is important to have. This behavior isn't implemented in this version of the `Accordion`.

## Demo (Updated 4/27)

The demo preview can be found at https://youthful-bartik-aca6f7.netlify.app/accordions/

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/accordions/examples/advanced.md
+++ b/packages/accordions/examples/advanced.md
@@ -4,22 +4,23 @@ The Accordion component is implemented using the [W3C Accordion pattern](https:/
 
 #### Controlled Usage
 
-This advanced example demonstrates
-[controlled component](https://reactjs.org/docs/forms.html#controlled-components)
-usage. It also demonstrates additional header text and menu buttons in the `Accordion.Header`.
+This advanced example demonstrates custom behavior using a
+[controlled](https://reactjs.org/docs/forms.html#controlled-components) `Accordion`
+component. It also demonstrates additional header text and menu buttons in the `Accordion.Header`.
 
 ```jsx
+const { SM } = require('@zendeskgarden/react-typography/src');
 const { Well } = require('@zendeskgarden/react-notifications/src');
 const { Toggle, Field, Label } = require('@zendeskgarden/react-forms/src');
 const { IconButton } = require('@zendeskgarden/react-buttons/src');
 const GearIcon = require('@zendeskgarden/svg-icons/src/16/gear-stroke.svg').default;
 const FolderIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
 const { Dropdown, Menu, Item, Trigger } = require('@zendeskgarden/react-dropdowns/src');
+const { Button } = require('@zendeskgarden/react-buttons/src');
 
-const StyledHint = styled.span`
-  font-size: ${props => props.theme.fontSizes.sm};
-  font-weight: normal;
-  font-style: italic;
+const StyledSM = styled(SM)`
+  margin-left: ${props => props.theme.space.base}px;
+  font-weight: ${props => props.theme.fontWeights.regular};
 `;
 
 const StyledButtonGroup = styled.div`
@@ -54,7 +55,7 @@ const accordionSections = [
   {
     id: 'section-2',
     header: 'Celery quandong swiss',
-    headerDetail: 'chard chicory earthnut pea potato. Salsify taro catsear',
+    headerDetail: 'chard chicory earthnut pea potato salsify taro catsear',
     panelContent: `Celery quandong swiss chard chicory earthnut pea potato. Salsify taro catsear
     garlic gram celery bitterleaf wattle seed collard greens nori. Grape wattle seed kombu beetroot
     horseradish carrot squash brussels sprout chard. Pea horseradish azuki bean lettuce avocado
@@ -89,29 +90,22 @@ const AdvancedExample = () => {
   const [expandedSections, setExpandedSections] = React.useState([0]);
   const [isBare, setIsBare] = React.useState(false);
   const [isCompact, setIsCompact] = React.useState(false);
-  const [isExpandAll, setIsExpandAll] = React.useState(false);
-  const [isCollapseAll, setIsCollapseAll] = React.useState(false);
-  const onExpandAll = () => {
-    setIsExpandAll(!isExpandAll);
-    setIsCollapseAll(false);
-    setExpandedSections([0, 1, 2, 3, 4]);
-  };
-  const onCollapseAll = () => {
-    setIsCollapseAll(!isCollapseAll);
-    setIsExpandAll(false);
-    setExpandedSections([]);
-  };
+  const isExpandAll = expandedSections.length === 5;
+  const isCollapseAll = expandedSections.length === 0;
+  const onExpandAll = () => setExpandedSections([0, 1, 2, 3, 4]);
+  const onCollapseAll = () => setExpandedSections([]);
   const onChange = index => {
-    setExpandedSections(sections => [index]);
-    setIsExpandAll(false);
-    setIsCollapseAll(false);
+    setExpandedSections(sections =>
+      expandedSections.includes(index)
+        ? sections.filter(section => section !== index)
+        : [...expandedSections, index]
+    );
   };
-
-  const buttonGroup = (
+  const ButtonGroup = ({ isCompact }) => (
     <StyledButtonGroup>
       <Dropdown onSelect={item => alert(item)}>
         <Trigger>
-          <IconButton aria-label="Settings" title="Settings">
+          <IconButton size={isCompact ? 'small' : 'medium'} aria-label="Settings" title="Settings">
             <GearIcon />
           </IconButton>
         </Trigger>
@@ -123,7 +117,7 @@ const AdvancedExample = () => {
       </Dropdown>
       <Dropdown onSelect={item => alert(item)}>
         <Trigger>
-          <IconButton aria-label="Settings" title="Settings">
+          <IconButton size={isCompact ? 'small' : 'medium'} aria-label="Files" title="Files">
             <FolderIcon />
           </IconButton>
         </Trigger>
@@ -142,20 +136,30 @@ const AdvancedExample = () => {
         <Row>
           <Col>
             <Field>
+              <Toggle checked={isCompact} onChange={event => setIsCompact(!isCompact)}>
+                <Label>Compact</Label>
+              </Toggle>
+            </Field>
+            <Field className="u-mt">
               <Toggle checked={isBare} onChange={event => setIsBare(!isBare)}>
                 <Label>Bare</Label>
               </Toggle>
             </Field>
-            <Field className="u-mt">
-              <Toggle checked={isExpandAll} onChange={onExpandAll}>
-                <Label>Expand All</Label>
-              </Toggle>
-            </Field>
-            <Field className="u-mt">
-              <Toggle checked={isCollapseAll} onChange={onCollapseAll}>
-                <Label>Collapse All</Label>
-              </Toggle>
-            </Field>
+            <div>
+              <Button size="small" onClick={onExpandAll} disabled={isExpandAll} className="u-mt">
+                Expand All
+              </Button>
+            </div>
+            <div>
+              <Button
+                size="small"
+                onClick={onCollapseAll}
+                disabled={isCollapseAll}
+                className="u-mt"
+              >
+                Collapse All
+              </Button>
+            </div>
           </Col>
         </Row>
       </Well>
@@ -165,6 +169,7 @@ const AdvancedExample = () => {
             level={3}
             isBare={isBare}
             onChange={onChange}
+            isCompact={isCompact}
             expandedSections={expandedSections}
           >
             {accordionSections.map(section => (
@@ -172,10 +177,9 @@ const AdvancedExample = () => {
                 <Accordion.Header>
                   <Accordion.Label>
                     {section.header}
-                    {'\u00A0'}
-                    <StyledHint>{section.headerDetail}</StyledHint>
+                    <StyledSM forwardedAs="span">{section.headerDetail}</StyledSM>
                   </Accordion.Label>
-                  {buttonGroup}
+                  <ButtonGroup isCompact={isCompact} />
                 </Accordion.Header>
                 <Accordion.Panel>{section.panelContent}</Accordion.Panel>
               </Accordion.Section>
--- a/packages/accordions/examples/advanced.md
+++ b/packages/accordions/examples/advanced.md
@@ -1,0 +1,205 @@
+### Accordion
+
+The Accordion component is implemented using the [W3C Accordion pattern](https://www.w3.org/TR/wai-aria-practices/#accordion).
+
+#### With custom header elements
+
+This advanced example demonstrates additional header text and menu buttons in the `Accordion.Header`.
+
+```jsx
+const { Well } = require('@zendeskgarden/react-notifications/src');
+const { Toggle, Field, Label } = require('@zendeskgarden/react-forms/src');
+const { IconButton } = require('@zendeskgarden/react-buttons/src');
+const GearIcon = require('@zendeskgarden/svg-icons/src/16/gear-stroke.svg').default;
+const FolderIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
+const { Dropdown, Menu, Item, Trigger } = require('@zendeskgarden/react-dropdowns/src');
+
+const StyledHint = styled.span`
+  font-size: ${props => props.theme.fontSizes.sm};
+  font-weight: normal;
+  font-style: italic;
+`;
+
+const StyledButtonGroup = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-around;
+  width: ${props => props.theme.space.base * 28}px;
+`;
+
+const AdvancedExample = () => {
+  const [isCollapsible, setIsCollapsible] = React.useState(false);
+  const [isExpandable, setIsExpandable] = React.useState(false);
+  const [isBare, setIsBare] = React.useState(false);
+  const [isCompact, setIsCompact] = React.useState(false);
+
+  const buttonGroup = (
+    <StyledButtonGroup>
+      <Dropdown onSelect={item => alert(item)}>
+        <Trigger>
+          <IconButton aria-label="Settings" title="Settings">
+            <GearIcon />
+          </IconButton>
+        </Trigger>
+        <Menu placement="top" hasArrow>
+          <Item value="option-1">Option 1</Item>
+          <Item value="option-2">Option 2</Item>
+          <Item value="option-3">Option 3</Item>
+        </Menu>
+      </Dropdown>
+      <Dropdown onSelect={item => alert(item)}>
+        <Trigger>
+          <IconButton aria-label="Settings" title="Settings">
+            <FolderIcon />
+          </IconButton>
+        </Trigger>
+        <Menu placement="top" hasArrow>
+          <Item value="option-1">Option 1</Item>
+          <Item value="option-2">Option 2</Item>
+          <Item value="option-3">Option 3</Item>
+        </Menu>
+      </Dropdown>
+    </StyledButtonGroup>
+  );
+
+  return (
+    <Grid>
+      <Well isRecessed>
+        <Row>
+          <Col>
+            <Field>
+              <Toggle checked={isCollapsible} onChange={event => setIsCollapsible(!isCollapsible)}>
+                <Label>Collapsible</Label>
+              </Toggle>
+            </Field>
+            <Field className="u-mt">
+              <Toggle checked={isExpandable} onChange={event => setIsExpandable(!isExpandable)}>
+                <Label>Expandable</Label>
+              </Toggle>
+            </Field>
+            <Field className="u-mt">
+              <Toggle checked={isBare} onChange={event => setIsBare(!isBare)}>
+                <Label>Bare</Label>
+              </Toggle>
+            </Field>
+          </Col>
+        </Row>
+      </Well>
+      <Row className="u-mt">
+        <Col>
+          <Accordion
+            level={3}
+            isBare={isBare}
+            isExpandable={isExpandable}
+            isCollapsible={isCollapsible}
+          >
+            <Accordion.Section>
+              <Accordion.Header>
+                <Accordion.Label>
+                  Turnip greens yarrow
+                  <StyledHint>ricebean rutabaga endive cauliflower sea lettuce kohlrabi</StyledHint>
+                </Accordion.Label>
+                {buttonGroup}
+              </Accordion.Header>
+              <Accordion.Panel>
+                <div>
+                  Turnip greens yarrow ricebean rutabaga endive cauliflower sea lettuce kohlrabi
+                  amaranth water spinach avocado daikon napa cabbage asparagus winter purslane kale.
+                  Celery potato scallion desert raisin horseradish spinach carrot soko. Lotus root
+                  water spinach fennel kombu maize bamboo shoot green bean swiss chard seakale
+                  pumpkin onion chickpea gram corn pea. Brussels sprout coriander water chestnut
+                  gourd swiss chard wakame kohlrabi beetroot carrot watercress.
+                </div>
+              </Accordion.Panel>
+            </Accordion.Section>
+            <Accordion.Section>
+              <Accordion.Header>
+                <Accordion.Label>
+                  Corn amaranth salsify
+                  <StyledHint>bunya nuts nori azuki bean chickweed potato bell pepper</StyledHint>
+                </Accordion.Label>
+                {buttonGroup}
+              </Accordion.Header>
+              <Accordion.Panel>
+                <div>
+                  Corn amaranth salsify bunya nuts nori azuki bean chickweed potato bell pepper
+                  artichoke. Nori grape silver beet broccoli kombu beet greens fava bean potato
+                  quandong celery. Bunya nuts black-eyed pea prairie turnip leek lentil turnip
+                  greens parsnip. Sea lettuce lettuce water chestnut eggplant winter purslane fennel
+                  azuki bean earthnut pea sierra leone bologi leek soko chicory celtuce parsley
+                  jícama.
+                </div>
+              </Accordion.Panel>
+            </Accordion.Section>
+            <Accordion.Section>
+              <Accordion.Header>
+                <Accordion.Label>
+                  Celery quandong swiss
+                  <StyledHint>chard chicory earthnut pea potato salsify taro catsear</StyledHint>
+                </Accordion.Label>
+                {buttonGroup}
+              </Accordion.Header>
+              <Accordion.Panel>
+                <div>
+                  Celery quandong swiss chard chicory earthnut pea potato. Salsify taro catsear
+                  garlic gram celery bitterleaf wattle seed collard greens nori. Grape wattle seed
+                  kombu beetroot horseradish carrot squash brussels sprout chard. Pea horseradish
+                  azuki bean lettuce avocado asparagus okra. Kohlrabi radish okra azuki bean corn
+                  fava bean mustard tigernut jícama green bean celtuce collard greens avocado
+                  quandong fennel gumbo black-eyed pea.
+                </div>
+              </Accordion.Panel>
+            </Accordion.Section>
+            <Accordion.Section>
+              <Accordion.Header>
+                <Accordion.Label>
+                  Grape silver beet
+                  <StyledHint>
+                    watercress potato tigernut corn groundnut chickweed okra winter
+                  </StyledHint>
+                </Accordion.Label>
+                {buttonGroup}
+              </Accordion.Header>
+              <Accordion.Panel>
+                <div>
+                  Grape silver beet watercress potato tigernut corn groundnut. Chickweed okra winter
+                  purslane coriander yarrow sweet pepper radish garlic brussels sprout pea groundnut
+                  summer purslane earthnut pea tomato spring onion azuki bean gourd. Gumbo kakadu
+                  plum komatsuna black-eyed pea green bean zucchini gourd winter purslane silver
+                  beet rock melon radish asparagus spinach.
+                </div>
+              </Accordion.Panel>
+            </Accordion.Section>
+            <Accordion.Section>
+              <Accordion.Header>
+                <Accordion.Label>
+                  Soko radicchio bunya
+                  <StyledHint>
+                    nuts gram dulse silver beet parsnip napa cabbage lotus root
+                  </StyledHint>
+                </Accordion.Label>
+                {buttonGroup}
+              </Accordion.Header>
+              <Accordion.Panel>
+                <div>
+                  Soko radicchio bunya nuts gram dulse silver beet parsnip napa cabbage lotus root
+                  sea lettuce brussels sprout cabbage. Catsear cauliflower garbanzo yarrow salsify
+                  chicory garlic bell pepper napa cabbage lettuce tomato kale arugula melon sierra
+                  leone bologi rutabaga tigernut. Sea lettuce gumbo grape kale kombu cauliflower
+                  salsify kohlrabi okra sea lettuce broccoli celery lotus root carrot winter
+                  purslane turnip greens garlic. Jícama garlic courgette coriander radicchio
+                  plantain scallion cauliflower fava bean desert raisin spring onion chicory bunya
+                  nuts. Sea lettuce water spinach gram fava bean leek dandelion silver beet eggplant
+                  bush.
+                </div>
+              </Accordion.Panel>
+            </Accordion.Section>
+          </Accordion>
+        </Col>
+      </Row>
+    </Grid>
+  );
+};
+
+<AdvancedExample />;
+```
--- a/packages/accordions/examples/advanced.md
+++ b/packages/accordions/examples/advanced.md
@@ -20,7 +20,6 @@ const { Button } = require('@zendeskgarden/react-buttons/src');
 
 const StyledSM = styled(SM)`
   margin-left: ${props => props.theme.space.base}px;
-  font-weight: ${props => props.theme.fontWeights.regular};
 `;
 
 const StyledButtonGroup = styled.div`
@@ -105,7 +104,12 @@ const AdvancedExample = () => {
     <StyledButtonGroup>
       <Dropdown onSelect={item => alert(item)}>
         <Trigger>
-          <IconButton size={isCompact ? 'small' : 'medium'} aria-label="Settings" title="Settings">
+          <IconButton
+            title="Settings"
+            aria-label="Settings"
+            focusInset={isCompact}
+            size={isCompact ? 'small' : 'medium'}
+          >
             <GearIcon />
           </IconButton>
         </Trigger>
@@ -117,7 +121,12 @@ const AdvancedExample = () => {
       </Dropdown>
       <Dropdown onSelect={item => alert(item)}>
         <Trigger>
-          <IconButton size={isCompact ? 'small' : 'medium'} aria-label="Files" title="Files">
+          <IconButton
+            title="Settings"
+            aria-label="Settings"
+            focusInset={isCompact}
+            size={isCompact ? 'small' : 'medium'}
+          >
             <FolderIcon />
           </IconButton>
         </Trigger>
--- a/packages/accordions/examples/advanced.md
+++ b/packages/accordions/examples/advanced.md
@@ -2,9 +2,11 @@
 
 The Accordion component is implemented using the [W3C Accordion pattern](https://www.w3.org/TR/wai-aria-practices/#accordion).
 
-#### With custom header elements
+#### Controlled Usage
 
-This advanced example demonstrates additional header text and menu buttons in the `Accordion.Header`.
+This advanced example demonstrates
+[controlled component](https://reactjs.org/docs/forms.html#controlled-components)
+usage. It also demonstrates additional header text and menu buttons in the `Accordion.Header`.
 
 ```jsx
 const { Well } = require('@zendeskgarden/react-notifications/src');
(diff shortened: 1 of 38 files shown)


packages/accordions/src/elements/accordion/components/Label.tsx
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef, HTMLAttributes } from 'react';
+import { StyledButton } from '../../../styled';
+import { useAccordionContext, useHeaderContext } from '../../../utils';
+
+export const Label = forwardRef<HTMLButtonElement, HTMLAttributes<HTMLButtonElement>>(
```
hzhu, line 12: Would it be more clear if this was renamed from "Label" to "Button" since it is the header button? I kept "Label" since that was a suggestion during a pair session. Open to changing this.

hzhu: Would it be more clear if this component was renamed from "Label" to "Button" since it is the header button? I kept "Label" since that was a suggestion during a pair session. Open to changing this.

packages/accordions/src/styled/accordion/StyledAccordion.ts
```diff
@@ -0,0 +1,22 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'accordions.accordion';
+
+export const StyledAccordion = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
```
jzempel, line 17: Will it help consumers down the line if you set `display: flex;` on this outer element?

packages/accordions/src/styled/accordion/StyledHeader.ts
```diff
@@ -0,0 +1,40 @@
...
+  isExpanded?: boolean;
+}
+
+export const StyledHeader = styled.div.attrs<IStyledHeader>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledHeader>`
+  display: flex;
+  align-items: center;
+  box-shadow: ${props =>
```
jzempel, line 26: Should focus be inset here? It's not often that Garden allows focus to extend beyond border boundaries. See https://garden.zendesk.com/react-components/tables/#selectable or https://garden.zendesk.com/react-components/tabs/#default-usage for examples. /cc @steue

packages/accordions/src/styled/accordion/StyledHeader.ts
```diff
@@ -0,0 +1,40 @@
...
+import styled from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'accordions.header';
+
+export interface IStyledHeader {
+  label?: string;
+  isFocused?: boolean;
+  isCompact?: boolean;
+  isExpanded?: boolean;
```
jzempel, line 17: Are `label`, `isCompact`,  and `isExpanded` needed here?

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
```diff
@@ -0,0 +1,30 @@
...
+  isExpanded?: boolean;
+}
+
+export const StyledInnerPanel = styled.div.attrs<IStyledInnerPanel>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledInnerPanel>`
+  transition: max-height 0.25s ease-in-out;
+  overflow: hidden;
+  max-height: ${props => !props.isExpanded && '0 !important'}; /* stylelint-disable-line */
```
jzempel, line 23: Please use a construct like this to disable stylelint and document why the `!important` is necessary: https://github.com/zendeskgarden/react-components/blob/348329cfed710b389be93ce5aa0d26f5edb59d5a/packages/dropdowns/src/styled/menu/StyledMenu.ts#L31-L32

packages/accordions/src/styled/accordion/StyledPanel.ts
```diff
@@ -0,0 +1,55 @@
...
+    paddingBottom = 0;
+  }
+
+  return css`
+    transition: padding 0.25s ease-in-out;
+    padding: ${paddingTop}px ${paddingHorizontal}px ${paddingBottom}px;
+  `;
+};
+
+export const StyledPanel = styled.div.attrs<IStyledPanel>({
```
jzempel, line 42: Prefer using semantic `section` element here and knocking out the unnecessary `role="region"` attribute in the associated `Accordion.Panel` component. See the https://garden.zendesk.com/react-containers/storybook/?path=/story/accordion-container--useaccordion story for details.

packages/accordions/src/elements/accordion/Accordion.tsx
```diff
@@ -0,0 +1,118 @@
...
+  extends ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>> {
+  Section: typeof Section;
+  Header: typeof Header;
+  Label: typeof Label;
+  Panel: typeof Panel;
+}
+
+interface IAccordionProps extends Omit<HTMLAttributes<HTMLDivElement>, 'onChange'> {
+  /** The `aria-level` used to apply heading rank on accordion headers */
+  level: number;
```
rossmoody, line 35: Is this a practice we typically do? To take an aria attribute available to the element and make it part of the API? Could we advise this attribute to be added to to the component using the actual aria attribute name?

Additionally, I'm curious, is there any way possible we could take care of this programmatically somehow based on children qty?

rossmoody: On the other hand, I'm seeing now it's required so it's sort of nice to control the accessibility implementation in that way, my concern is it might set a precedent we could apply to most components where we start making aria attributes mandatory in the API.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+export const StyledRotateIcon = styled(({ children, isRotated, isHovered, isCompact, ...props }) =>
+  cloneElement(Children.only(children), props)
+).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledRotateIcon>`
+  transform: ${props => props.isRotated && `rotate(${props.theme.rtl ? '-' : '+'}180deg)`};
+  transition: transform 0.25s ease-in-out;
```
jzempel, line 27: Please add color `0.25s` to the transition.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import { cloneElement, Children } from 'react';
+import styled from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'accordions.rotate_icon';
+
+export interface IStyledRotateIcon {
+  isExpanded?: boolean;
```
jzempel, line 15: Is `isExpanded` used here?

packages/accordions/src/styled/accordion/StyledSection.ts
```diff
@@ -0,0 +1,22 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'accordions.section';
+
+export const StyledSection = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
```
jzempel, line 17: Is a flex display treatment helpful here (depending on container flex)?

packages/accordions/src/elements/accordion/Accordion.tsx
```diff
@@ -0,0 +1,118 @@
...
+Accordion.Header = Header;
+Accordion.Label = Label;
+Accordion.Panel = Panel;
+
+Accordion.displayName = 'Accordion';
+
+Accordion.defaultProps = {
+  isBare: false,
+  isCompact: false,
+  isCollapsible: false,
```
rossmoody, line 114: @steue I wonder, should we consider the `true` by default? I'm tinkering with the knobs in the examples and I'm finding my inability to collapse the accordions kind of frustrating. It seems more likely that we would want to allow the user to open and close accordions by default and then mandate them open in special cases. The functionality seems the same by default, it just give the user the ability to close the accordion to save space.

jzempel: If this is the decision, we should set the `isCollapsible` prop to `true` by default along with the demo toggle. Totally fine and makes sense to me.

packages/accordions/src/elements/accordion/Accordion.tsx
hzhu: > Controlled usage allows consumers to have full control of the accordion state and drive custom behavior (such as expand all, or expanding a single accordion panel).

This was my initial approach, however @austin94 pointed out some problems* with adding heading rank to the accordion headers programmatically. Austin proposed the current implementation of the `level` prop as most suitable, and I agree that it was much better than the original (programmatic solution) that I had.

> my concern is it might set a precedent we could apply to most components where we start making aria attributes mandatory in the API.

I agree! I think most accessibility should be abstracted away from the consumer. [These](https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#accessible) are some of the person guiding principles I follow. I think this particular prop is a special case, and should not set a precedent of not abstracting away accessibility when possible.

_* The problem with programmatically determining header rank is that rank can sometimes depend on the context of the content. It's not a sure thing that an `h2` rank belongs under an `h1` rank. This was difficult for me to grok, and Austin had to go to school on me to get this across my head haha._

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,191 @@
...
+            <Field className="u-mt">
+              <Toggle checked={isExpandAll} onChange={onExpandAll}>
+                <Label>Expand All</Label>
+              </Toggle>
+            </Field>
+            <Field className="u-mt">
+              <Toggle checked={isCollapseAll} onChange={onCollapseAll}>
+                <Label>Collapse All</Label>
+              </Toggle>
+            </Field>
```
jzempel, line 158: These two should probably be buttons – feels weird as toggle controls.

hzhu: Buttons are much more suitable for this 👍.

packages/accordions/src/styled/accordion/StyledButton.ts
```diff
@@ -0,0 +1,52 @@
...
+  outline: none;
+  border: none;
+  background: transparent;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 1.5}px ${props.theme.space.base * 3}px`
+      : `${props.theme.space.base * 5}px`};
+  width: 100%;
+  text-align: ${props => (props.theme.rtl ? 'right' : 'left')};
+  color: ${props => props.isHovered && getColor('primaryHue', 600, props.theme)};
```
jzempel, line 34: There should be a `transition` property to finesse the color on hover.

packages/accordions/src/elements/accordion/Accordion.tsx
jzempel: > is there any way possible we could take care of this programmatically somehow based on children qty

This prop isn't telling us how many children are within the component. It's telling us where the component falls within the ancestoral DOM hierarchy. This is information we can't know, so we need the user to tell us :wink:

packages/accordions/src/elements/accordion/Accordion.tsx
hzhu: > it just gives the user the ability to close the accordion to save space.

An accordion that has one expanded section still saves space. According to WAI-ARIA, accordions are _"commonly used to reduce the need to scroll when presenting multiple sections of content on a single page"_. Here's their demo:

https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

I'm unsure that "saving" space is a strong enough justification for having this prop be defaulted to `true`. Saving space is context-specific (how much space? what kind of space?). The nature of the accordion already saves space (for scrolling on a page). I think it is nice to have the default for `isCollapsible" as "off" to match the WAI-ARIA example and other accordion examples I've seen. However, this isn't mandatory, and Garden can have an opinionated view here.

jzempel: Furthermore, the inability to collapse on the "Advanced" controlled example doesn't map with the description provided on the `isCollapsible` component: "Determines if panels can be collapsed in an uncontrolled accordion". Setting a reasonable default might help with this.

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,191 @@
...
+            isBare={isBare}
+            onChange={onChange}
+            expandedSections={expandedSections}
+          >
+            {accordionSections.map(section => (
+              <Accordion.Section key={section.id}>
+                <Accordion.Header>
+                  <Accordion.Label>
+                    {section.header}
+                    {'\u00A0'}
```
rossmoody, line 175: I don't think this necessity for a non-breaking space is a practice we can advocate for in implementation of this component. I'm unsure of the best solve but we should space correctly out of the box. Maybe margin-left on the span element?

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,191 @@
...
+
+```jsx
+const { Well } = require('@zendeskgarden/react-notifications/src');
+const { Toggle, Field, Label } = require('@zendeskgarden/react-forms/src');
+const { IconButton } = require('@zendeskgarden/react-buttons/src');
+const GearIcon = require('@zendeskgarden/svg-icons/src/16/gear-stroke.svg').default;
+const FolderIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
+const { Dropdown, Menu, Item, Trigger } = require('@zendeskgarden/react-dropdowns/src');
+
+const StyledHint = styled.span`
```
rossmoody, line 19: I'd be curious if we could advocate for using out-of-the-box component form Garden as opposed to styled component elements within the examples? At the very least, I believe UXCS advises against italics, but I wonder if we could just use typography components?

packages/accordions/src/styled/accordion/StyledButton.ts
```diff
@@ -0,0 +1,52 @@
...
+export const StyledButton = styled.button.attrs<IStyledButton>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledButton>`
+  outline: none;
+  border: none;
+  background: transparent;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 1.5}px ${props.theme.space.base * 3}px`
```
rossmoody, line 30: We would want this to follow base 4 spacing. 4px or 8px

packages/accordions/examples/advanced.md
jzempel: Good call. The most recent typography components support this case – but @hzhu will need to release the updated v8.6.0 before they can be used here.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
```diff
@@ -0,0 +1,44 @@
...
+  cloneElement(Children.only(children), props)
+).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledRotateIcon>`
+  transform: ${props => props.isRotated && `rotate(${props.theme.rtl ? '-' : '+'}180deg)`};
+  transition: transform 0.25s ease-in-out;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 1.5}px ${props.theme.space.base * 3}px`
```
rossmoody, line 30: Also base-4 sizing. 4px or 8px

packages/accordions/src/styled/accordion/StyledPanel.ts
```diff
@@ -0,0 +1,55 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledPanel>`
+  ${paddingStyles};
+  border-bottom: ${props =>
```
rossmoody, line 47: Do we want to programmatically add a top border to `first-child` by any chance? It seems a smidge odd to me visually that the first list item doesn't have a border but the last has one on the bottom. https://share.getcloudapp.com/NQuDZQl5

I think I'd advocate for ditching the last border or adding a top border.

packages/accordions/src/elements/accordion/Accordion.tsx
rossmoody: Yes yes yes, I see now. Very awesome.

packages/accordions/src/elements/accordion/Accordion.tsx
rossmoody: My thinking from a fundamental point of view is an accordion component opens and closes, so to have it unable to close by default feels odd to me considering the intended functionality. I'd defer on this particular point but giving the user the ability to close seems like a benefit with zero drawback.

jzempel: Another point to make here is that we would never show disabled in a component’s default state. Having these props for asserting control is great. But I think we almost have to set `isCollapsed = true`

packages/accordions/src/styled/accordion/StyledHeader.ts
hzhu: 👀 I missed these during clean up. These aren't needed.

packages/accordions/src/styled/accordion/StyledButton.ts
hzhu: Sounds good. I can make these adjustments once the artboards are updated accordingly 👍.

@rossmoody @steue

packages/accordions/examples/advanced.md
hzhu: 👍 in the meantime, I've removed the italics.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
rossmoody, line 27: Should we add a min width/height designation here of some sort to account for the dropdown chevron getting smaller as it's scaled down? Doesn't seem to happen unless more elements are added inside the `Header`: https://share.getcloudapp.com/xQuWWPd8

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,205 @@
...
+            <FolderIcon />
+          </IconButton>
+        </Trigger>
+        <Menu placement="top" hasArrow>
+          <Item value="option-1">Option 1</Item>
+          <Item value="option-2">Option 2</Item>
+          <Item value="option-3">Option 3</Item>
+        </Menu>
+      </Dropdown>
+    </StyledButtonGroup>
```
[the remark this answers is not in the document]
hzhu: I think the advanced example leaves out the `isCompact` toggle because of the spacing limitations it causes for the menu buttons. If a consumer wants to add Garden menu buttons in the header, the `isCompact` prop should not be used as doing so constrains the space. 

IMO, the `isCompact` prop is somewhat inflexible, and I was hesitant about adding this prop to the `Accordion`. I think it makes the most sense for the `Accordion` to come with default padding. If consumers decide to render large menu buttons, they can add more padding. If they want less padding, they can customize that too.

@steue what do you think? How should the menu buttons respond to `isCompact={true}`?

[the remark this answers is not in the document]
jzempel: @hzhu, the system should work together here. If it doesn't, then compact sizing would need to be re-thought. I would expect you to use `size="small"` icon buttons.

packages/accordions/src/elements/accordion/Accordion.tsx
hzhu: @rossmoody 

> to have it unable to close by default feels odd to me 

It's not clear to me if there are fundamental reasons for defaulting `isCollapsible` to `true` besides "it feels odd". Does the WAI-ARIA [example](https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html) feel odd too?

> but giving the user the ability to close seems like a benefit without any drawbacks

There are various implementations of Accordion behaviors — each suitable to a different UX context. Each of these implementations can be achieved with various combinations of the `isCollapsible` and `isExpandable` props. Maybe the reason for defaulting `isCollapsible` to `true` is that research has shown that our users consistently use that implementation most?

hzhu: @rossmoody 

> to have it unable to close by default feels odd to me 

It's not clear to me if there are fundamental reasons for defaulting `isCollapsible` to `true` besides "it feels odd". Does the WAI-ARIA [example](https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html) feel odd too?

> but giving the user the ability to close seems like a benefit without any drawbacks

There are various implementations of Accordion behaviors — each suitable to a different UX context. Each of these implementations can be achieved with various combinations of the `isCollapsible` and `isExpandable` props. Maybe the reason for defaulting `isCollapsible` to `true` is that research has shown that our users consistently use that implementation most? 

@jzempel 

> we would never show disabled in a component’s default state.

Could you clarify what this means? Is it the case that the basic example would never be used since it always has one panel expanded, and one header button with `aria-disabled="true"`?

jzempel: > Could you clarify what this means?

@steue is going to provide disabled styling for `isCollapsible={false}` sections. That means the most basic form of an `<Accordion>` (with no props added) would be disabled. We can't have that. It makes sense to go with Ross' design instincts on this and set the default for `isCollapsible` to `true`.

jzempel: > Does the WAI-ARIA example feel odd too?

Yes, I have always thought it did.

hzhu: > Furthermore, the inability to collapse on the "Advanced" controlled example doesn't map with the description provided on the isCollapsible component: "Determines if panels can be collapsed in an uncontrolled accordion". Setting a reasonable default might help with this.

I'll go ahead and add the ability to collapse on the "Advanced" controlled example. The `isCollapsible` prop and description doesn't apply to a controlled accordion. A controlled has custom behavior.

packages/accordions/src/styled/accordion/StyledAccordion.ts
hzhu: This element could be set to `display: flex; flex-direction: column;` but I'm unaware if this helps consumers down the line. Do you think it would be helpful? I can add it if it does 👍 .

jzempel: If the world around the accordion is `flex`, then this could be a natural benefit as the flex parent can make helpful assertions about the layout of the child. But I'll let you apply the imaginative scenario stress to determine whether this ends up providing helpful value or not.

hzhu: I ran through a couple of scenarios and concluded to leave `flex` off the outer element for the time being.

packages/accordions/src/styled/accordion/StyledSection.ts
hzhu: I tried a couple of renderings and found that a flex display probably isn't necessary as a default style.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
hzhu: Added.

packages/accordions/src/styled/accordion/StyledButton.ts
hzhu: Added a transition for `color`. The common transition duration values used in Garden are `0.1s`, `0.15s`, and `0.25s`. Since `0.25s` felt slow, I went with `0.1s` transition on the color.

packages/accordions/examples/advanced.md
hzhu: @rossmoody I can use CSS spacing instead, but could you help me understand the consequences of using an HTML entity here?

packages/accordions/src/styled/accordion/StyledButton.ts
steue: @rossmoody i remember this conversation. did we end up changing the `isCompact` padding on the rest of our components to base-4? currently i still see them at `6px` and `12px` for vertical and horizontal padding respectively.

to clarify, i'm completely onboard on updating this to base-4, i just want some clarification on the whole system's perspective on this spacing so the accordions can match.

packages/accordions/src/styled/accordion/StyledPanel.ts
steue: @rossmoody @hzhu great point, i'm for ditching the border for `last-child`!

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
steue: >i remember this conversation. did we end up changing the isCompact padding on the rest of our components to base-4? currently i still see them at 6px and 12px for vertical and horizontal padding respectively.

to clarify, i'm completely onboard on updating this to base-4, i just want some clarification on the whole system's perspective on this spacing so the accordions can match.

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
steue: @rossmoody @hzhu not exactly sure what is happening here. designs did not intend for the chevron to change in size.

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,205 @@
...
+              </Toggle>
+            </Field>
+          </Col>
+        </Row>
+      </Well>
+      <Row className="u-mt">
+        <Col>
+          <Accordion
+            level={3}
+            isBare={isBare}
```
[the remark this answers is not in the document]
steue: @jzempel @hzhu for the sake of visual proof, i am not opposed to adding a toggle option for `isCompact`. 

as a design guideline, we don't want to encourage `isCompact` usage when there are many elements within the header i.e. icons and media items. we initially decided not to include it as it might unintentionally give the wrong idea about this guideline.

packages/accordions/examples/advanced.md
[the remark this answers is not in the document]
steue: @hzhu @jzempel it makes sense to me that icon buttons scale with `isCompact`

there's [a different thread here](https://github.com/zendeskgarden/react-components/pull/725#discussion_r414067549) about `isCompact` padding. I believe this can be fixed by changing the vertical padding so there is `8px` of spacing between icon and the header container. [Visual explorations of icon button scaling + spacing on Abstract.](https://share.goabstract.com/eee22b64-e2ff-4bed-8e7f-6588b57a4c13?collectionLayerId=79334da2-e6e6-47a9-99cb-38d26a3292da&mode=design)

packages/accordions/examples/advanced.md
jzempel: with the release of v8.6.0, this can just use the `<SM>` typography component (which `isBold={false}` by default).

packages/accordions/examples/advanced.md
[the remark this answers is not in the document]
jzempel: On these pages, we need to stress test all combinations. We'll reserve the `website` (needed so badly) for guard-railing expected usage. Please add an `isCompact` toggle here.

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
```diff
@@ -0,0 +1,31 @@
...
+  isExpanded?: boolean;
+}
+
+export const StyledInnerPanel = styled.div.attrs<IStyledInnerPanel>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledInnerPanel>`
+  transition: max-height 0.25s ease-in-out;
+  /* stylelint-disable-next-line declaration-no-important */
+  max-height: ${props => !props.isExpanded && '0 !important'};
```
jzempel, line 23: "...and document why the `!important` is necessary". Aka, add a `/* [1] */` with doc added to the header.

https://github.com/zendeskgarden/react-components/blob/348329cfed710b389be93ce5aa0d26f5edb59d5a/packages/dropdowns/src/styled/menu/StyledMenu.ts#L31-L32

packages/accordions/src/styled/accordion/StyledHeader.ts
jzempel: Why was this resolved?

packages/accordions/src/styled/accordion/StyledPanel.ts
jzempel: Why was this resolved?

packages/accordions/src/styled/accordion/StyledButton.ts
jzempel: It's helpful to note that Garden's `<Anchor>` uses a `0.25s` transition. Although the stacking of these elements may necessitate the quicker transition you selected. @steue please double-check that the fine-tuned result fits the Garden system and makes sense in this context.

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
hzhu: > "...and document why the !important is necessary"

Yeah I noticed this comment. I went and looked at the codebase, and places where I saw `!important` used, no comments were added, which sort of confused me. I'll go ahead an add a comment here.

jzempel: I've linked to a location ☝️ that shows exactly what I'm looking for.

hzhu: Awesome, thanks 👍 .

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
hzhu: Yeah the element needs `min-width` / `min-height` 👍 .

packages/accordions/src/styled/accordion/StyledPanel.ts
hzhu: This was resolved because I had made the change on my local machine and was using these comments as a checklist. I was queuing the changes up for a single push to reduce CI builds, but I can keep a separate checklist and hold off on resolving these until the commit is pushed.

hzhu: I pushed up the commit this morning and marked this thread resolved. Feel free to re-open the thread if the `Accordion.Panel` needs further adjustments.

packages/accordions/src/styled/accordion/StyledHeader.ts
hzhu: Sorry about that, I was using these comments as a checklist and had resolved it on locally. I'll need to be careful about marking these as resolved if the commits aren't pushed up since it can be confusing.

I pushed the commit up now and marked this as resolved. Feel free to re-open this if further adjustments are needed.

packages/accordions/examples/advanced.md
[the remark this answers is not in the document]
hzhu: The `isCompact` toggle has been added.

packages/accordions/examples/advanced.md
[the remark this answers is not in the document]
hzhu: The `isCompact` toggle has been added to the advanced example.

packages/accordions/src/styled/accordion/StyledButton.ts
hzhu: Update: connected with @steue and concluded that `0.1s` is a suitable duration for transitioning this `color` property.

packages/accordions/examples/advanced.md
hzhu: The text now uses `<SM>` from `react-typography`.

packages/accordions/src/styled/accordion/StyledPanel.ts
hzhu: Update: removed the border on the last panel of the `Accordion`.

packages/accordions/src/styled/accordion/StyledButton.ts
rossmoody: Yeah, there are a couple outstanding tickets in queue to update `Dropdowns` and `Forms` components to base-4 spacing. `Menus` spacing just recently merged to follow. I wouldn't say you have to follow their spacing increments, just that we as a system adhere to base-4.

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,194 @@
...
+const { Well } = require('@zendeskgarden/react-notifications/src');
+const { Toggle, Field, Label } = require('@zendeskgarden/react-forms/src');
+const { IconButton } = require('@zendeskgarden/react-buttons/src');
+const GearIcon = require('@zendeskgarden/svg-icons/src/16/gear-stroke.svg').default;
+const FolderIcon = require('@zendeskgarden/svg-icons/src/16/folder-open-stroke.svg').default;
+const { Dropdown, Menu, Item, Trigger } = require('@zendeskgarden/react-dropdowns/src');
+const { Button } = require('@zendeskgarden/react-buttons/src');
+
+const StyledSM = styled(SM)`
+  margin-left: ${props => props.theme.space.base}px;
```
rossmoody, line 22: `font-weight: ${props => props.theme.fontWeights.regular};`

So that we don't have two bold type treatments next to each other in the example. This might be considered subjective, but the examples I've seen in product don't have descriptor/hint copy in bold.

packages/accordions/src/styled/accordion/StyledButton.ts
steue: Awesome. I'll make those adjustments. @hzhu

packages/accordions/package.json
```diff
@@ -22,7 +22,9 @@
   "sideEffects": false,
   "types": "dist/typings/index.d.ts",
   "dependencies": {
-    "@zendeskgarden/container-utilities": "^0.5.1",
+    "@zendeskgarden/container-accordion": "^1.0.0",
+    "@zendeskgarden/container-utilities": "^0.5.3",
+    "@zendeskgarden/react-typography": "^8.6.0",
```
austin94, line 27: Is this only used in the markdown examples? This should be a `devDependency` if so.

packages/accordions/src/elements/accordion/Accordion.tsx
```diff
@@ -0,0 +1,119 @@
...
+  Section: typeof Section;
+  Header: typeof Header;
+  Label: typeof Label;
+  Panel: typeof Panel;
+}
+
+interface IAccordionProps extends Omit<HTMLAttributes<HTMLDivElement>, 'onChange'> {
+  /** The `aria-level` used to apply heading rank on accordion headers */
+  level: number;
+  /** A controlled prop that determines which sections are expanded */
```
austin94, line 36: > a controlled prop

Do we need to specify that this is "controlled"? I'm used to the pattern of just describing what the prop is related to. i.e. "Determines which sections are expanded"

packages/accordions/src/elements/accordion/components/Header.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import React, { useState, FocusEvent, forwardRef, HTMLAttributes } from 'react';
+import { composeEventHandlers } from '@zendeskgarden/container-utilities';
+import ChevronDown from '@zendeskgarden/svg-icons/src/16/chevron-down-stroke.svg';
+import { useAccordionContext, useSectionContext, HeaderContext } from '../../../utils';
+import { StyledHeader, StyledRotateIcon, COMPONENT_ID as buttonGardenId } from '../../../styled';
+
+export interface IHeaderProps extends HTMLAttributes<HTMLDivElement> {
+  /** Apply props to the wrapping `Accordion.Header` element */
+  wrapperProps?: any;
```
austin94, line 24: Why is it necessary to pass `wrapperProps` and use context with the current separation?

Can we have a standard DOM mapping of:
- `Accordion.Header` - `getHeaderProps()`
- `Accordion.Label` - `getTriggerProps()`

packages/accordions/src/utils/useAccordionContext.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { createContext, useContext, HTMLAttributes, MutableRefObject } from 'react';
+
+export interface IAccordionContext {
```
austin94, line 10: Could this extend `IUseAccordionPropGetters` from `container-accordions` to get some of these values?

packages/accordions/src/elements/accordion/Accordion.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * Accepts all `<div>` attributes and events. Also accepts sub-components:
+
+ *  - `Accordion.Section`
+ *  - `Accordion.Header`
+ *  - `Accordion.Panel`
+ * 
+ * Note: The `Accordion.Label` is a sub-component of `Accordion.Header`.
```
austin94, line 57: Personally, it makes sense to me for the `Label` component to be namespaced the same as `Section/Header/Panel`. This will help with discoverability.

packages/accordions/src/elements/accordion/Accordion.tsx
hzhu: Probably not, I'll update the prop description to "Determines which sections are expanded".

packages/accordions/examples/advanced.md
hzhu: Sounds good, the font-weight has been updated.

packages/accordions/src/styled/accordion/StyledButton.ts
hzhu: Sounds good, the vertical padding has been updated.

packages/accordions/package.json
jzempel: Do we even add `devDependency` for other markdown uses? For example we have `Grid`, `Well`, and `Input` all over the demos without needing to import.

packages/accordions/src/elements/accordion/components/Header.tsx
hzhu: > Why is it necessary to pass `wrapperProps`

Passing `wrapperProps` allows consumers to style the wrapping heading element. Do you think `wrapperProps` should be passed to the button element instead? Or to eliminate this prop completely?

> Can we have a standard DOM mapping of:

The current code is similar to this mapping, except the `onClick` from `getTriggerProps` is plucked off, and passed along with the `getHeaderProps` object. This is done to allow clicking the header to trigger expanding / collapsing the accordion sections. Is there a better alternative to achieve this behavior?

packages/accordions/package.json
hzhu: @jzempel it looks like this isn't needed. I'll remove this and give it a try, and update accordingly.

packages/accordions/src/utils/useAccordionContext.ts
hzhu: Yeah that's a great idea, I will bundle the change in a new commit shortly. Thanks!

packages/accordions/src/styled/accordion/StyledPanel.ts
```diff
@@ -0,0 +1,55 @@
...
+  `;
+};
+
+export const StyledPanel = styled.section.attrs<IStyledPanel>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledPanel>`
+  ${paddingStyles};
+  border-bottom: ${props =>
+    `1px solid ${props.isBare ? 'transparent' : getColor('neutralHue', 300, props.theme)}`};
```
jzempel, line 48: Border `1px solid` should be getting pulled from `props.theme.borders.sm`.

packages/accordions/src/styled/accordion/StyledAccordion.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const COMPONENT_ID = 'accordions.accordion';
+
+export const StyledAccordion = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  & ${StyledSection}:last-child ${StyledPanel} {
+    border: none;
+  }
```
jzempel, line 21: Feels like this should've been added to `StyledSection` as:

```
&:last-child ${StyledPanel} {
  border: none;
}
```

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
```diff
@@ -0,0 +1,56 @@
...
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledRotateIcon>`
+  transform: ${props => props.isRotated && `rotate(${props.theme.rtl ? '-' : '+'}180deg)`};
+  transition: transform 0.25s ease-in-out, color 0.1s ease-in-out;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 1.5}px ${props.theme.space.base * 3}px`
+      : `${props.theme.space.base * 5}px`};
+  min-width: ${props => props.theme.space.base * 4}px;
+  min-height: ${props => props.theme.space.base * 4}px;
```
jzempel, line 46: Shouldn't these just be `width` and `height`? Also, the dimensions should be pulling from `props.theme.iconSizes.md`.

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,195 @@
...
+  font-weight: ${props => props.theme.fontWeights.regular};
```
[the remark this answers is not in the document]
hzhu: Thanks, removed the `font-weight`. Will push new commit up, update demo, re-tag reviewers shortly.

packages/accordions/src/styled/accordion/StyledAccordion.ts
hzhu: Thanks, that's a more suitable place to style the border.

packages/accordions/src/elements/accordion/Accordion.tsx
hzhu: Sounds good, I've pushed up a new commit and updated the demo.

packages/accordions/src/elements/accordion/components/Header.tsx
hzhu: Updated 🎉 .

packages/accordions/examples/advanced.md
```diff
@@ -0,0 +1,195 @@
...
+          <Item value="option-1">Option 1</Item>
+          <Item value="option-2">Option 2</Item>
+          <Item value="option-3">Option 3</Item>
+        </Menu>
+      </Dropdown>
+      <Dropdown onSelect={item => alert(item)}>
+        <Trigger>
+          <IconButton size={isCompact ? 'small' : 'medium'} aria-label="Files" title="Files">
+            <FolderIcon />
+          </IconButton>
```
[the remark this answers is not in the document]
hzhu: Updated 🎉 .

packages/accordions/src/styled/accordion/StyledPanel.ts
hzhu: Thanks, updated 🎉 .

packages/accordions/src/styled/accordion/StyledRotateIcon.ts
hzhu: Thanks, the height and width have been updated.

packages/accordions/src/styled/accordion/StyledButton.ts
```diff
@@ -0,0 +1,69 @@
...
+  border: none;
+  background: transparent;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 2}px ${props.theme.space.base * 3}px`
+      : `${props.theme.space.base * 5}px`};
+  width: 100%;
+  text-align: ${props => (props.theme.rtl ? 'right' : 'left')};
+  font-size: ${props => props.theme.fontSizes.md};
+  font-weight: ${props => props.theme.fontWeights.semibold};
```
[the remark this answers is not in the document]
hzhu: Added line-heights, thanks 👍.

packages/accordions/src/styled/accordion/StyledButton.ts
```diff
@@ -0,0 +1,75 @@
...
+  outline: none;
+  border: none;
+  background: transparent;
+  padding: ${props =>
+    props.isCompact
+      ? `${props.theme.space.base * 2}px ${props.theme.space.base * 3}px`
+      : `${props.theme.space.base * 5}px`};
+  width: 100%;
+  text-align: ${props => (props.theme.rtl ? 'right' : 'left')};
+  line-height: ${props => getLineHeight(props.theme.lineHeights.lg, props.theme.fontSizes.md)};
```
jzempel, line 56: Reminder: we do not leverage `theme.lineHeights` outside of the `typography` package – opting rather to impart component theming power to `theme.space.base`.

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export const StyledInnerPanel = styled.div.attrs<IStyledInnerPanel>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledInnerPanel>`
+  transition: max-height 0.25s ease-in-out;
+  /* stylelint-disable-next-line declaration-no-important */
+  max-height: ${props => !props.isExpanded && '0 !important'}; /* [1] */
+  overflow: hidden;
+  line-height: ${props => getLineHeight(props.theme.lineHeights.md, props.theme.fontSizes.md)};
```
jzempel, line 32: Prefer using `space.base` calculation rather than `lineHeights.md`.

packages/accordions/src/styled/accordion/StyledButton.ts
hzhu: Thanks for the reminder 👍.

PR #726, opened 2020-04-23
chore: enhanced scripting workflow

jzempel:
## Description

- Introduces Netlify preview deployments to the PR workflow. The updated scripting will deploy PRs to Netlify. The `master` branch will continue to deploy to `gh-pages` – although now with branch protection limited to the CI user.

- Refine `yarn run tag` workflow. All the steps remain, but have been reordered and enhanced with `garden` scripts as follows:
  - `sync` – ensure that the master branch and tags are sync'd up with GitHub
  - `version` – run `yarn lerna version` (commits since current tag are listed for ease of reference); enhancements include:
    - use conventional commits logic to guess the right version bump (the maintainer can execute `yarn run tag <semver>` to override)
    - prevent push to GH – all potentially destructive actions are queued for the final step
    - the expensive pre-test run is outsourced to the `preversion` lifecycle hook, ensuring the maintainer gets an early look at the version bump before wasting precious time testing against the wrong version
  - `changelog` – generate, edit, and commit the updated CHANGELOG.md (now using the OS temporary folder with auto-cleanup)
  - `release` or `rollback` – final prompt to double-check if everything is ready to go.
    - on `release` the two commits (version and changelog) and tag get pushed, triggering a CI npm publish, and a draft release is created on GitHub.
    - on `rollback` the version tag is deleted and two commits reset.

This workflow aligns more closely with all the other Garden repos (albeit with improved automation), easing the mental mapping for devs moving from one repo to the next.

## Detail

Screen Shot 2020-04-23 at 9 02 18 AM

`yarn run tag` is a CLI command that accepts master branch overrides for running within major version branches. As mentioned above, it also accepts a semantic `[version]` override.


utils/scripts/tag.js
```diff
@@ -0,0 +1,214 @@
...
+ * Bump and publish package versions.
+ *
+ * @param {String} bump Semantic or explicit versioning bump.
+ * @param {String} preid Prerelease ID as defined by https://semver.org/#spec-item-9.
+ * @param {String} master Name for the master branch being used for tagging.
+ * @param {Ora} spinner Terminal spinner.
+ *
+ * @returns New version tag.
+ */
+const version = async (bump, preid, master, spinner) => {
```
hzhu, line 133: > use conventional commits logic to guess the right version bump

Could you clarify how this works? Does the `version` function look for the strings "fix" "feat" "BREAKING CHANGE" "!" from a set of commits from the last release to determine the right version bump?

jzempel: via [`lerna version`](https://github.com/lerna/lerna/tree/master/commands/version)

jzempel:
@austin94 agreed. Solve for https://github.com/zendeskgarden/react-components/blob/272491ddbad1e5269fe0f5b110fa34b4d329c426/.circleci/config.yml#L49-L50 and we're in a better place. Thanks for the call-out.

PR #728, opened 2020-04-24
feat(dropdowns): add start slot to autocomplete and select components

austin94:
## Description

This PR adds the ability to provide a `start` icon to `Select` and `Autocomplete` components. Due to the interaction of `MultiSelect` I have not added the `start` option at this time for that components. 

## Detail

I was originally hoping to mimic (or use directly) the styling provided by the `MediaInput` in `react-forms`. But, due to our use of `FauxInput` and necessity to dynamically render the internal `<input />` of the component I was unable to match the styling directly.

I was able to reuse the styling for the chevron icon though. The `start` slot will style it's child to match the `sm` and `md` icon sizes based on `isCompact`. It also plays well with the overflow styling already defined for both components.

## Other

I have also removed the `translateY(-1px)` styling from the chevron rotation. Let me know if this is still something that we want to add in!

## Demo

Take a look at the "Field Variants" section to view a new toggle for this prop.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/index.cjs.js": {
-    "bundled": 75276,
-    "minified": 48384,
-    "gzipped": 10298
+    "bundled": 76824,
+    "minified": 49579,
+    "gzipped": 10412
   },
   "dist/index.esm.js": {
-    "bundled": 72646,
-    "minified": 45866,
-    "gzipped": 10120,
+    "bundled": 74159,
+    "minified": 47026,
+    "gzipped": 10250,
     "treeshaked": {
       "rollup": {
-        "code": 35390,
+        "code": 36276,
         "import_statements": 807
       },
       "webpack": {
-        "code": 39271
+        "code": 40190
       }
     }
   }
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "dist/index.cjs.js": {
-    "bundled": 76824,
-    "minified": 49579,
-    "gzipped": 10412
+    "bundled": 76907,
+    "minified": 49630,
+    "gzipped": 10431
   },
   "dist/index.esm.js": {
-    "bundled": 74159,
-    "minified": 47026,
-    "gzipped": 10250,
+    "bundled": 74242,
+    "minified": 47077,
+    "gzipped": 10268,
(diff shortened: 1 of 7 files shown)


rossmoody:
Nice! Looking good. A few questions: 
1. Do we post a followup PR for `Multiselect` implementation? My thinking is we would want to make a decision to take it on or not but not necessarily punt it indefinitely. Or if we file a follow-on ticket, commit to it in the next sprint. 
2. I'm seeing now that the Jira tickets coupled styling and the implementation of the icon component in the same ticket. That might be a bit much for one PR. Were you thinking of doing a follow-on for the styling portion? I could break the Jira tickets out as well if it helps. That might be smart as we'll want to update the Form `text-input` styling to match as well (which could be an additional ticket). 
3. Do we want to tackle the ability to specify a color of the icon in this PR? By default, we should update it to `grey-600`.

jzempel:
@rossmoody the [MediaInput](https://garden.zendesk.com/react-components/forms/#basic) icons are `grey-400`. Do those need to change to `grey-600` for system consistency?

austin94:
> Do we post a followup PR for Multiselect implementation

It was my understanding that this addition wasn't intended for the `Multiselect`. If we know we want it I can add it to this PR otherwise, let's hold off until we know it's a necessity. 

> styling and color

This styling is already defined in our [MediaInput examples](https://garden.zendesk.com/react-components/forms/#!/Basic/1). I've matched the current size/colors. Should these match the chevron icon across all states?

rossmoody:
> @rossmoody the [MediaInput](https://garden.zendesk.com/react-components/forms/#basic) icons are `grey-400`. Do those need to change to `grey-600` for system consistency?

That would get my vote. I did the thing where I batch up a bunch of unrelated changes into one Jira ticket and I'm seeing that now. Depending on how difficult this change might be coupled with the ability to set a color on the icon, we could break it off this scope.

rossmoody:
My original thinking was we would extend this functionality across the components in the `Dropdown` package but if Multiselect implementation brings a bunch of baggage or work I wouldn't mind dropping it. The original ask for the functionality didn't have a Multiselect as part of it, after all. 

Per styling, I'm seeing that changes in that regard will effect `Forms` and `Dropdowns` so a follow-on probably does make more sense. There is quite a bit of spacing nuance in the [Jira ticket example](https://share.getcloudapp.com/xQuWANdO) that isn't outwardly apparent. Like updating the icons sizes, spacing and default color of the icon (I was thinking it would follow chevron behavior of `grey-600` unless disabled it's `grey-400`)

PR #733, opened 2020-05-04
feat(forms/dropdowns): improve start icon sizing

austin94:
## Description

This PR:
- Updates the default vertical padding for all `Input`, `Select`, `Autocomplete`, and `Multiselect` components.
- Updates default color of media items to `grey-600` (`grey-400` when disabled)
- Add consistent icon and padding to `compact` and `default` sizes across all inputs and dropdowns

## Details

After refactoring there are several styling assertions in our tests that break stylelint, even when stylelint is disabled. I've removed these tests as they were only asserting styles.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/styled/text/StyledTextInput.ts
+++ b/packages/forms/src/styled/text/StyledTextInput.ts
@@ -129,8 +129,8 @@ const sizeStyles = (props: IStyledTextInputProps & ThemeProps<DefaultTheme>) => 
   const swatchMarginHorizontal = math(
     `${paddingVertical} + ${swatchMarginVertical} - ${paddingHorizontal}`
   );
-  const figureMarginFirst = `auto ${props.theme.space.base * 1.5}px auto 0`;
-  const figureMarginLast = `auto 0 auto ${props.theme.space.base * 1.5}px`;
+  const figureMarginFirst = `auto ${props.theme.space.base * 2}px auto 0`;
+  const figureMarginLast = `auto 0 auto ${props.theme.space.base * 2}px`;
 
   return css`
     padding: ${padding};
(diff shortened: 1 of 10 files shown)


packages/forms/src/styled/text/StyledTextInput.ts
```diff
@@ -131,8 +129,8 @@ const sizeStyles = (props: IStyledTextInputProps & ThemeProps<DefaultTheme>) =>
   const swatchMarginHorizontal = math(
     `${paddingVertical} + ${swatchMarginVertical} - ${paddingHorizontal}`
   );
-  const figureMarginFirst = `auto ${math(`${paddingHorizontal} * 3 / 4`)} auto 0`;
-  const figureMarginLast = `auto 0 auto ${math(`${paddingHorizontal} * 3 / 4`)}`;
+  const figureMarginFirst = `auto ${math(`${paddingHorizontal} * 1 / 2`)} auto 0`;
```
rossmoody, line 132: Is the math logic necessary here? I think the margins for these in `default` and `isCompact` sizes can be 8px.

austin94: I believe @jzempel handled this logic originally. Is it ok to still perform this `math()` operation?

jzempel: It's probably worth retaining the calculated relationship between these spacing properties. But it looks like most instances of `math` could be replaced with interpolated calculations with `px` appended.

jzempel: Although, I think @rossmoody is right that it is no longer worth feeding an unchanging value into this particular calc. It's always meant to be `${props.theme.space.base * 2}`.

austin94: I've updated some of the `math()` values to use interpolation where possible.

austin94: With the updated spacing it came out to `${props.theme.space.base * 1.5}px` for me. @rossmoody could you double-check the visuals on this once it's deployed?

austin94: It looks like the first round of approved logic was slightly off. Moving to 8px.

PR #734, opened 2020-05-04
fix(dropdowns): use correct aria attributes for menu dropdown

hzhu:
## Description

This pull request allows the `Dropdown` component to be configured using a `role` prop so that it can use `menu` ARIA attributes.

## Detail

The `getMenuProps` prop getter from Downshift always returns a `role="listbox"` dropdown list and `role="option"` for the dropdown list items. For comboboxes and [select dropdowns](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html), this works ✅. **However, Garden also supports [menu dropdowns](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions-active-descendant.html)**, so there needs to be a way to configure the ARIA attributes to align with the menu button widget (e.g. `role="menu"`, `role="menuitem"`) as outlined in [WAI ARIA 1.1](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions-active-descendant.html).

To achieve this, a `hasMenuRef` is initialized at the `Dropdown` level and is set to `true` within the `Trigger`. This is because only menu dropdowns have a `Trigger`. The boolean is then used via context in `Item` and `Menu` to set the correct menu ARIA attributes.

Now, Dropdown Menu's has menu ARIA attributes:

<kbd>
menu
</kbd>

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -33,7 +33,7 @@ export interface IDropdownProps {
   ) => void;
   downshiftProps?: object;
   /* Determines whether the Dropdown implementation uses listbox or menu ARIA attributes */
-  role?: 'listbox' | 'menu';
+  role?: 'listbox' | 'menu' | undefined;
 }
 
 /**
@@ -52,7 +52,7 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
     onStateChange,
     onInputValueChange,
     downshiftProps,
-    role = 'listbox'
+    role
   } = props;
   // Refs used to handle custom Garden keyboard nav
   const itemIndexRef = useRef<number>(0);
@@ -213,6 +213,10 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
       </Downshift>
     </Manager>
   );
+};
+
+Dropdown.defaultProps = {
+  role: 'listbox'
 };
 
 Dropdown.propTypes = {
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -32,6 +32,8 @@ export interface IDropdownProps {
     stateAndHelpers: ControllerStateAndHelpers<any>
   ) => void;
   downshiftProps?: object;
+  /* Determines whether the Dropdown implementation uses listbox or menu ARIA attributes */
+  role?: 'listbox' | 'menu';
 }
 
 /**
@@ -49,7 +51,8 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
     onSelect,
     onStateChange,
     onInputValueChange,
-    downshiftProps
+    downshiftProps,
+    role = 'listbox'
   } = props;
   // Refs used to handle custom Garden keyboard nav
   const itemIndexRef = useRef<number>(0);
@@ -193,6 +196,7 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
         {downshift => (
           <DropdownContext.Provider
             value={{
+              role,
               itemIndexRef,
               previousItemRef,
               previousIndexRef,
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -32,8 +32,6 @@ export interface IDropdownProps {
     stateAndHelpers: ControllerStateAndHelpers<any>
   ) => void;
   downshiftProps?: object;
-  /* Determines whether the Dropdown implementation uses listbox or menu ARIA attributes */
-  role?: 'listbox' | 'menu' | undefined;
 }
 
 /**
@@ -51,8 +49,7 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
     onSelect,
     onStateChange,
     onInputValueChange,
-    downshiftProps,
-    role
+    downshiftProps
   } = props;
   // Refs used to handle custom Garden keyboard nav
   const itemIndexRef = useRef<number>(0);
@@ -60,6 +57,9 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
   const previousIndexRef = useRef<number | undefined>(undefined);
   const nextItemsHashRef = useRef<object>({});
   const containsMultiselectRef = useRef(false);
+
+  // Ref used to determine ARIA attributes for menu dropdowns
+  const hasMenuRef = useRef(false);
 
   // Used to inform Menu (Popper) that a full-width menu is needed
   const popperReferenceElementRef = useRef<any>(null);
@@ -196,7 +196,7 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
         {downshift => (
           <DropdownContext.Provider
             value={{
-              role,
+              hasMenuRef,
               itemIndexRef,
               previousItemRef,
               previousIndexRef,
@@ -213,10 +213,6 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
       </Downshift>
     </Manager>
   );
-};
-
-Dropdown.defaultProps = {
-  role: 'listbox'
 };
 
 Dropdown.propTypes = {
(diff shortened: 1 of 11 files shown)


hzhu:
# Screenshots

Differences are demonstrated on macOS (Mojave & Catalina) VoiceOver on Safari.

## Difference 1

**BEFORE (macOS Catalina):** Screen reader incorrectly calls the menu button's dropdown a "listbox pop up, Menu pop-up combo box". Screen reader incorrectly says aloud "Default Men".
Screen Shot 2020-05-04 at 10 33 11 AM

**AFTER (macOS Catalina):** Notice the trigger is no longer treated as a combo box by the screen reader. Screen reader no longer says aloud "Default Men".
Screen Shot 2020-05-04 at 10 45 54 AM

--------------------------------------------------------------------------------

## Difference 2

**BEFORE (macOS Mojave):** Screen reader incorrectly calls the menu button's dropdown a "Menu pop-up combo box". Screen reader incorrectly says aloud "Default Men".
Screen Shot 2020-05-04 at 9 45 56 AM

**AFTER (macOS Mojave):** Notice the trigger is no longer treated as a combo box by the screen reader. Screen reader no longer says aloud "Default Men".
Screen Shot 2020-05-04 at 9 46 30 AM

--------------------------------------------------------------------------------

## Difference 3

**BEFORE (macOS Mojave):** Notice the dropdown menu is treated as a listbox
Screen Shot 2020-05-04 at 9 57 32 AM
Screen Shot 2020-05-04 at 9 59 55 AM

**AFTER (macOS Mojave):** Notice the dropdown menu is treated as a menu
Screen Shot 2020-05-04 at 10 00 36 AM

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -49,7 +51,8 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
     onSelect,
     onStateChange,
     onInputValueChange,
-    downshiftProps
+    downshiftProps,
+    role = 'listbox'
```
austin94, line 55: Should we use `defaultProps` to provide this value? That way it is shown in the prop-sheet.

packages/dropdowns/src/elements/Trigger/Trigger.tsx
```diff
@@ -45,19 +46,35 @@ const Trigger: React.FunctionComponent<ITriggerProps> = ({ children, refKey, ...
...
+      ...triggerProps,
+      ...(children as any).props
+    });
+
+    const menuToggleProps = {
+      ...listboxToggleProps,
+      'aria-haspopup': 'true',
+      'aria-controls': listboxToggleProps['aria-owns'],
+      'aria-owns': null
+    };
```
austin94, line 64: All props should be passed through `getToggleButtonProps()` in some way before being applied to the element. I think this would be more clear if the conditional (listbox vs menu) logic is moved above here somehow.

packages/dropdowns/src/elements/Trigger/Trigger.tsx
```diff
@@ -45,19 +46,35 @@ const Trigger: React.FunctionComponent<ITriggerProps> = ({ children, refKey, ...
...
+
+    // Some ARIA attributes from Downshift's `getMenuProps` props are overwritten depending on
+    // whether the dropdown renders a `role="menu"` or `role="listbox"`. This override is intended
+    // to align a11y with examples demonstrated in WAI ARIA 1.1.
```
austin94, line 68: Use a mult-line comment for this.

packages/dropdowns/src/elements/Menu/Items/Item.tsx
```diff
@@ -87,6 +88,11 @@ export const Item = React.forwardRef<HTMLLIElement, IItemProps>(
       );
     }
 
+    const menuItemProps = {
+      role: 'menuitem',
+      'aria-selected': null
+    };
```
austin94, line 94: Could this just be inlined at line 109?

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
hzhu: Yeah this sounds good 👍 .

packages/dropdowns/src/elements/Menu/Items/Item.tsx
hzhu: It could.

packages/dropdowns/src/elements/Trigger/Trigger.tsx
hzhu: > All props should be passed through getToggleButtonProps() in some way before being applied to the element.

The props are passed through `getToggleButtonProps()`. The spread (line 60) points directly to a `getToggleButtonProps` invocation. 

<kbd>
Screen Shot 2020-05-05 at 12 37 37 PM
</kbd>

hzhu: > I think this would be more clear if the conditional (listbox vs menu) logic is moved above here somehow.

Could you clarify how your suggestion might look? Specifically, which bit of logic should be moved above? It would help me understand why what I have isn't clear.

hzhu:
Thanks @jzempel 👍. I decided not to feed the role into `<Select>`, `<Autocomplete>`, or `<Multiselect>` because Downshift provides the correct ARIA attributes to those components.

jzempel:
@hzhu, ah interesting. In that case, why wouldn't you default the role to `menu` for the non-field cases?

hzhu:
@jzempel I noticed that "Advanced Menu usage" example is a non-field case and I didn't pass a `role` into that `Dropdown`. I pushed up a [new commit](https://github.com/zendeskgarden/react-components/pull/734/commits/4682ae791a5c0672e0b5a82297ee46408f70e901#diff-bbef7bd7035dfef159c34973bfba6189R99) which adds `role="menu"` for that dropdown. If I misunderstood, would you be able to point out the files/lines I should be looking at?

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -211,6 +215,10 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
   );
 };
 
+Dropdown.defaultProps = {
+  role: 'listbox'
```
jzempel, line 219: I think `role` should be `undefined` by default. Rules:
- if provided by the user, then apply the user's value
- if used in the context of a `Select`, `Multiselect`, or `Autocomplete`, then apply `listbox`
- otherwise, apply a role of `menu`

austin94: > if used in the context of a Select, Multiselect, or Autocomplete, then apply listbox

With `Dropdown` being instantiated before `Select/Multiselect/Autocomplete/Trigger` there may need to be some additional context work needed to apply the role.

packages/dropdowns/src/elements/Trigger/Trigger.tsx
austin94: `aria-haspopup`, `aria-control`, and `aria-owns` are not being passed through `getToggleButtonProps()` in this example.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
jzempel: Exactly ☝️ This is what I meant earlier: https://github.com/zendeskgarden/react-components/pull/734#pullrequestreview-406784839

packages/dropdowns/src/elements/Trigger/Trigger.tsx
hzhu: `aria-haspopup`, `aria-control`, and `aria-owns` aren't passed through because we want Downshift to take care of listbox props.

We want to override these props when Downshift is used to render menu dropdowns. This is why I chose the names "listboxToggleProps" versus "menuToggleProps". Note, Downshift does not support menu dropdowns out of the box — only listbox dropdowns.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
hzhu: It appears I misunderstood these comments. My understanding is that `Trigger` is only used in menu dropdowns, and that can be the heuristic to determine the appropriate ARIA attributes.

In the latest commit, I have removed all component `role` props. Consumers no longer need to pass in a `role`. The proper ARIA attributes are now determined by rendering a `Trigger` element, which will set a ref passed down from the dropdown context.

packages/dropdowns/src/elements/Trigger/Trigger.tsx
```diff
@@ -39,25 +40,48 @@ const Trigger: React.FunctionComponent<ITriggerProps> = ({ children, refKey, ...
     }
 
     previousIsOpenRef.current = isOpen;
-  }, [isOpen]);
+
+    if (hasMenuRef.current === false) {
+      hasMenuRef.current = true;
+    }
```
austin94, line 46: Does the `hasMenuRef` logic have anything to do with the `isOpen` state? This may need to be moved to its own `useEffect`.

hzhu: Ah yes, good catch. Since `hasMenuRef` has nothing to do with `isOpen`, I've moved it into its own `useEffect`.

PR #735, opened 2020-05-05
fix(breadcrumbs): change chevron color from black to grey-600

li-kai:
## Description

The chevrons in the `Breadcrumbs` component should be displaying the secondary color (`grey-600`) instead of `currentColor`, which is black.

https://zendesk.atlassian.net/browse/GARDEN-1191

### Before
Screenshot 2020-05-05 at 4 51 36 pm

### After
Screenshot 2020-05-05 at 4 29 20 pm

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
+++ b/packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
@@ -8,7 +8,7 @@
 import React, { HTMLAttributes } from 'react';
 import styled, { ThemeProps, DefaultTheme } from 'styled-components';
 import { em } from 'polished';
-import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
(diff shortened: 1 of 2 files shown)


packages/breadcrumbs/src/styled/StyledChevronIcon.tsx
```diff
@@ -28,7 +28,8 @@ const ValidChevronIcon: React.FC<HTMLAttributes<SVGElement> & ThemeProps<Default
  */
 export const StyledChevronIcon = styled(ValidChevronIcon).attrs({
   role: 'presentation',
-  'aria-hidden': 'true'
+  'aria-hidden': 'true',
+  color: PALETTE.grey[600]
```
austin94, line 32: To allow this to be themed dynamically, could we use `props.theme` within the styled component?

Something like:

```
import { getColor } from '@zendeskgarden/react-theming';

styled(...).attrs(...)`
  color: ${props => getColor('neutralHue', 600, props.theme)};
`
```

li-kai: Thanks for that tip. I've updated it!

PR #746, opened 2020-05-13
feat(grid): add `textAlign` prop to `Col` component

jzempel:
## Description

We continue to see a desire to control horizontal alignment within `Col` components. This PR adds that capability (with RTL awareness) and will eradicate many of the `StyledCol` usages that are continuing to proliferate – even in Garden's own codebase.

## Detail

See https://github.com/zendeskgarden/website/blob/master/src/examples/button/ButtonSizes.tsx for example (should not have a `margin-bottom` within a single `Row`).

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/grid/src/styled/StyledCol.spec.tsx
+++ b/packages/grid/src/styled/StyledCol.spec.tsx
@@ -134,7 +134,7 @@ describe('StyledCol', () => {
 
   describe('Text Align', () => {
     ARRAY_TEXT_ALIGN.forEach(textAlign => {
-      let expected: any;
+      let expected: string;
 
       if (textAlign === 'start') {
         expected = 'left';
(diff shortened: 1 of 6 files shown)


packages/grid/src/styled/StyledCol.spec.tsx
```diff
@@ -132,6 +132,54 @@ describe('StyledCol', () => {
     });
   });
 
+  describe('Text Align', () => {
+    ARRAY_TEXT_ALIGN.forEach(textAlign => {
+      let expected: any;
```
hzhu, line 137: Should this `any` be upgraded to `string`?

jzempel:
@austin94 I have never encountered a design where `left` shouldn't be `right` in RTL and vice versa. My instinct is to encourage best practice by only supplying `start` and `end`. I'm willing to be wrong. But, at that point, it's a patch release with the simple addition of `left` & `right` to the types constants.

PR #750, opened 2020-05-16
feat(typography): add Span component and allow Code to inherit

jzempel:
## Description

This PR adds a new `<Span>` component that inherits surrounding `font` and `line-height` CSS, but can control `font-weight` via an `isBold` prop.

The `<Code>` component was enhanced to allow font inheritance via a `size="inherit"` prop.

## Detail

**Reviewers**

The updated `Code` API seen below is not ideal. The way it should work is that the default is `undefined` (meaning `font-size` is calculated via inheritance) and there is no specific `inherit` prop.

Screen Shot 2020-05-16 at 2 30 25 PM

What do we think about changing the API default and considering it to be more of a "bug fix" than a "breaking change"? My sense is that any existing usage would not be affected because the new `<Code>` which inherits from its surroundings is likely to match any old `<Code>` which defaults to `medium` sizing.

The only problem with this assertion would be existing usage where someone is planting a `<Code>` inside a `<LG>` font-sized block, for example, and expecting it to render at `medium` size. Again, I really doubt this is happening or intended from a design POV.

## Checklist

      designer as a reviewer)


austin94:
> What do we think about changing the API default and considering it to be more of a "bug fix" than a "breaking change"?

I think this is up to your discretion,  I see value in having an 'inherit' property. If we move to `undefined` this doesn't show consumers that the component is now inheriting its size. An explicit value helps clear this up.

jzempel:
Good points, @austin94. My instinct is that prop documentation would alleviate the confusion. And the fact that current `undefined` means "medium size" for a `<Code>` component is just odd. Makes the API feel unintentional.

rossmoody:
I feel a smidge confused. Maybe it would help (since I'm fresh to this) to explain what I think would happen from the API sheet. The `Code` component will render `medium` in size because it's the default value unless I use the `inherit` prop to have it inherit the parent sizing? Is that right?

jzempel:
@rossmoody sorry for the confusion. My proposal is to make the prop sheet look like this:

Screen Shot 2020-05-19 at 12 11 58 PM

Compare to the current deployment for differences: https://5ec06f1f68f98626be01fc29--zendeskgarden-react-components.netlify.app/typography/#code-1

rossmoody:
Ah, yes that helps me. Well, my +1 would be towards the direction you just posted. If I have the right idea, I find the current deployment to be odd because it makes it seem like it doesn't inherit unless you specify it to via the property. It seems more preferable to me to inherit by default unless you specify a size.

jzempel:
Yeah, @rossmoody. Perhaps an approach that takes this into account yet balances with @austin94's concern would look like this:

Screen Shot 2020-05-19 at 12 25 36 PM

rossmoody:
Super +1

allisonacs:
+1 to the latest version. This is clearest to me. Sweet sauce.

PR #751, opened 2020-05-20
feat(buttons): revise button padding

rossmoody:
## Description

This pull request refines the horizontal padding to more appropriately space the button content depending on size and be friendlier in applications with limited space.

## Detail

Padding increments have been refined to the following: 

```
large buttons: 28px -> 20px horizontal padding
medium buttons: 28px -> 16px horizontal padding
small buttons: 28px -> 12px horizontal padding
```

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/buttons/src/styled/StyledButton.ts
+++ b/packages/buttons/src/styled/StyledButton.ts
@@ -40,6 +40,16 @@ export const getHeight = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) 
   return `${props.theme.space.base * 10}px`;
 };
 
+const getHorizontalPadding = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
+  if (props.size === SIZE.SMALL) {
+    return `${props.theme.space.base * 3}px`;
+  } else if (props.size === SIZE.LARGE) {
+    return `${props.theme.space.base * 5}px`;
+  }
+
+  return `${props.theme.space.base * 4}px`;
+};
+
 const colorStyles = (
   props: IStyledButtonProps & ThemeProps<DefaultTheme> & HTMLAttributes<HTMLButtonElement>
 ) => {
@@ -209,7 +219,7 @@ const sizeStyles = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
   } else {
     const height = getHeight(props);
     const lineHeight = math(`${height} - (${props.theme.borderWidths.sm} * 2)`);
-    const padding = props.theme.space.base * 7;
+    const padding = getHorizontalPadding(props);
     let fontSize;
 
     if (props.size === 'small') {
--- a/packages/buttons/src/styled/StyledButton.ts
+++ b/packages/buttons/src/styled/StyledButton.ts
@@ -40,16 +40,6 @@ export const getHeight = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) 
   return `${props.theme.space.base * 10}px`;
 };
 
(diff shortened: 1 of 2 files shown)


packages/buttons/src/styled/StyledButton.ts
```diff
@@ -40,6 +40,16 @@ export const getHeight = (props: IStyledButtonProps & ThemeProps<DefaultTheme>)
...
 
+const getHorizontalPadding = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
+  if (props.size === SIZE.SMALL) {
+    return `${props.theme.space.base * 3}px`;
+  } else if (props.size === SIZE.LARGE) {
+    return `${props.theme.space.base * 5}px`;
+  }
+
+  return `${props.theme.space.base * 4}px`;
+};
```
jzempel, line 51: I would prefer seeing this logic added to the setup for a `padding` variable in the `sizeStyles` function rather than adding a one-off support function.

rossmoody: I'm trying my best to infer patterns and I gave this change a stab. I feel like there is a lot of small things that could go different a different way so I'll be curious your thoughts.

jzempel: The `getBorderRadius` is an outlier because it is applied in the base styles where there is no space for declaring local variables. The `getHeight` is an outlier because it is exported. I'll add a comment on your improved code with a follow-on that can push the code quality just a bit further.

rossmoody: Oh, just for posterity, this comment was referencing the new changes I just pushed that moved the code into `sizeStyles`, not a commentary on this.

rossmoody: Nvmd I see you were also commenting on the new stuff. I'll resolve this comment.

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -209,9 +209,17 @@ const sizeStyles = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) => {
...
+    let padding;
     let fontSize;
 
+    if (props.size === 'small') {
+      padding = `${props.theme.space.base * 3}px`;
+    } else if (props.size === 'large') {
+      padding = `${props.theme.space.base * 5}px`;
+    } else {
+      padding = `${props.theme.space.base * 4}px`;
+    }
```
jzempel, line 221: Now that this is in-context, we see a further opportunity for optimization. There's no need to travel through the conditionals twice when both `padding` and `fontSize` can be resolved like this:

```js
if (props.size === SIZE.SMALL) {
  fontSize = props.theme.fontSizes.sm;
  padding = `${props.theme.space.base * 3}px`;
} else {
  fontSize = props.theme.fontSizes.md;

  if (props.size === SIZE.LARGE) {
    padding = `${props.theme.space.base * 5}px`;
  } else {
    padding = `${props.theme.space.base * 4}px`;
  }
}
```

Also notice I used the `SIZE` constants vs strings.

rossmoody: The strings part was one thing I was so curious about! Is that a preference thing or is one a better option?

Will push up this update right meow. Thanks for working through this with me. Learning a lot.

jzempel: If they are easily available, I think using constants is always preferable. The good news is TS has us covered when they aren't. I think the original code here was just an oversight re: constants vs strings.

PR #752, opened 2020-05-20
chore(scripts): update tag script to force install

jzempel:
## Description

See https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-force

## Detail

I encountered problems with a linked lint config package that inadvertently bombed the release script. This will fix for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "build:analyze": "ANALYZE_BUNDLE=true yarn build:single",
     "build:demo": "lerna run build:demo --stream",
     "build:single": "utils/scripts/scoped-npm-command.js --script build",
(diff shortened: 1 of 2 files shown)


austin94:
> with a linked lint config package

Was this due to a local `yarn link`? I'm not sure we should be doing a full `git clean` on every commit. We have local caches in `.changelog` that are beneficial to keep around.

jzempel:
@austin94 I'm not following. Yes it was a local `yarn link`. But this change doesn't clean on every commit – only as a setup for release. And there is no `.changelog` that I'm aware of. See the script's `temp` for details.

PR #756, opened 2020-05-27
fix(dropdowns): prevent icon box-sizing inheritance

jzempel:
## Description

Along with abstracting common code (1k+ reduction in minified size) this PR fixes a list of styling problems:
- hover color not being applied to chevron
- focus color not being applied to chevron
- color transition timing not applied
- prevent icon wrapper height from exceeding container

## Detail

Fixes icon sizing when CSS Bedrock is being used.

https://github.com/zendeskgarden/css-components/blob/9a80f2d1f053d17d8447b96feb46e01f2d378397/packages/bedrock/src/_base.css#L33-L35

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/styled/field/StyledSelect.ts
+++ b/packages/dropdowns/src/styled/field/StyledSelect.ts
@@ -49,6 +49,7 @@ const iconStyles = (props: IStyledSelectIconProps & ThemeProps<DefaultTheme>, is
     align-items: center;
     justify-content: ${justifyContent};
     transition: color 0.25s ease-in-out;
+    box-sizing: content-box;
     /* stylelint-disable-next-line property-no-unknown */
     padding-${position}: ${padding};
     width: ${size};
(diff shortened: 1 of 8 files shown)


packages/dropdowns/src/styled/field/StyledSelect.ts
```diff
@@ -17,72 +17,83 @@ const isInvalid = (validation?: VALIDATION) => {
...
-  isBare?: boolean;
-}
+const iconStyles = (props: IStyledSelectIconProps & ThemeProps<DefaultTheme>, isStart: boolean) => {
+  const maxHeight = `${props.theme.space.base * (props.isCompact ? 8 : 10)}px`;
+  const padding = `${props.theme.space.base * 3}px`;
+  const size = props.theme.iconSizes.md;
+  let position;
+  let justifyContent;
+  let shade;
+
```
rossmoody, line 27: This is crafty. It took me like 20 minutes to understand why this is working but I think I finally got it. I kept thinking it was redundant. The RTL logic makes my brain hurt.

PR #757, opened 2020-06-01
feat(buttons): new toggle button components

jzempel:
## Description

Provides component support for https://www.w3.org/TR/wai-aria-1.1/#aria-pressed toggle buttons, including:

- new `ToggleButton` component with `isPressed` prop
- new `ToggleIconButton` component with `isPressed` prop
- styling connected with `[aria-pressed]` attribute applied to any button

## Detail

@rossmoody I realized I ran a bit in front of official designs for this one. The styling for `[aria-pressed="true"]` is exactly the same as the existing `:active` state. If this is not desired, let's talk. It will be trivial to modify the CSS.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/buttons/src/elements/ToggleButton.tsx
+++ b/packages/buttons/src/elements/ToggleButton.tsx
@@ -26,7 +26,7 @@ interface IToggleButtonProps extends IButtonProps {
 const ToggleButton: React.FunctionComponent<
   IToggleButtonProps & React.RefAttributes<HTMLButtonElement>
 > = React.forwardRef<HTMLButtonElement, IToggleButtonProps>(({ isPressed, ...otherProps }, ref) => (
-  <Button aria-pressed={isPressed} ref={ref} {...otherProps} />
+  <Button aria-pressed={isPressed === undefined ? false : isPressed} ref={ref} {...otherProps} />
 ));
 
 ToggleButton.displayName = 'ToggleButton';
--- a/packages/buttons/src/elements/ToggleButton.tsx
+++ b/packages/buttons/src/elements/ToggleButton.tsx
@@ -26,7 +26,7 @@ interface IToggleButtonProps extends IButtonProps {
 const ToggleButton: React.FunctionComponent<
   IToggleButtonProps & React.RefAttributes<HTMLButtonElement>
 > = React.forwardRef<HTMLButtonElement, IToggleButtonProps>(({ isPressed, ...otherProps }, ref) => (
-  <Button aria-pressed={isPressed === undefined ? false : isPressed} ref={ref} {...otherProps} />
+  <Button aria-pressed={isPressed} ref={ref} {...otherProps} />
 ));
 
 ToggleButton.displayName = 'ToggleButton';
@@ -37,6 +37,7 @@ ToggleButton.propTypes = {
 };
 
 ToggleButton.defaultProps = {
+  isPressed: false,
   size: 'medium'
 };
 
--- a/packages/buttons/src/elements/ToggleButton.tsx
+++ b/packages/buttons/src/elements/ToggleButton.tsx
@@ -7,7 +7,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { default as Button, IButtonProps } from './Button';
+import Button, { IButtonProps } from './Button';
 
 interface IToggleButtonProps extends IButtonProps {
   /**
@@ -36,5 +36,8 @@ ToggleButton.propTypes = {
   isPressed: PropTypes.bool
 };
 
-/** @component */
+ToggleButton.defaultProps = {
+  size: 'medium'
+};
+
 export default ToggleButton;
--- a/packages/buttons/src/elements/ToggleButton.tsx
+++ b/packages/buttons/src/elements/ToggleButton.tsx
@@ -33,7 +33,7 @@ ToggleButton.displayName = 'ToggleButton';
 
 ToggleButton.propTypes = {
   ...Button.propTypes,
-  isPressed: PropTypes.bool
+  isPressed: PropTypes.oneOf([true, false, 'mixed'])
 };
 
 ToggleButton.defaultProps = {
--- a/packages/buttons/src/elements/ToggleButton.tsx
+++ b/packages/buttons/src/elements/ToggleButton.tsx
@@ -1,0 +1,40 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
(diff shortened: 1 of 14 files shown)


jzempel:
@austin94 sure. We can't do what you're proposing. As soon as `aria-pressed` is applied to a button – with any value – it turns the button into what is considered a [Toggle Button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Toggle_buttons).

One way to achieve what you're proposing would be to add two props to the existing `Button`:
- `isToggle`
- `isPressed` (which is only valid when `isToggle` is true)

But that's a gross API.

austin94:
The Toggle Button pattern is possible with a single attribute. `buttin[aria-pressed=""]`, could you explain why `isToggle` would be necessary?

packages/buttons/src/elements/ToggleButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+import PropTypes from 'prop-types';
+import Button, { IButtonProps } from './Button';
+
+interface IToggleButtonProps extends IButtonProps {
+  /**
+   * Determine if the button is pressed or not. Use `'mixed'` to indicate
+   * whether the toggle controls other elements which do not share the same
+   * value.
+   */
+  isPressed?: boolean | 'mixed';
```
austin94, line 18: As this is a native attribute, I believe `boolean` is incorrect here. This may need to be a strings for all values: `'true' | 'false' | 'mixed'`

Could we reference the `ButtonHtmlProps` interface to get the matching values from `aria-pressed`?

jzempel:
I don't think I'm following you. We don't want to have to explain the complexities of ARIA usage when it's not necessary. The goal here is to answer the question "does the button toggle?" Then use a `ToggleButton`. If the button doesn't toggle, use a `Button`. The fact that there are fundamentally different VO/a11y experiences doesn't need to be spelled out in detail.

packages/buttons/src/elements/ToggleButton.tsx
jzempel: It's not a native attribute, it is a prop which is eventually mapped to a corresponding attribute. As the types here are just a subset, I felt it simplified the API and there wasn't an overwhelming need to support the corresponding boolean strings.

Screen Shot 2020-06-01 at 9 59 03 AM

packages/buttons/src/elements/ToggleIconButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+ */
+const ToggleIconButton: React.FunctionComponent<
+  IToggleIconButtonProps & React.RefAttributes<HTMLButtonElement>
+> = React.forwardRef<HTMLButtonElement, IToggleIconButtonProps>(
+  ({ isPressed, ...otherProps }, ref) => (
+    <IconButton aria-pressed={isPressed} ref={ref} {...otherProps} />
+  )
+);
+
+ToggleIconButton.displayName = 'ChevronButton';
```
austin94, line 34: Should this be `ChevronButton`?

jzempel: Oops, fixed.

packages/buttons/src/elements/ToggleButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * A `Button` with the [ARIA
+ * attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Toggle_buttons)
+ * to indicate a pressed state.
+ */
+const ToggleButton: React.FunctionComponent<
```
hzhu, line 26: Since toggle buttons are buttons with `aria-pressed` state, what do you think about adding `aria-pressed={isPressed}` directly to the `Button` component and removing this component (`ToggleButton`)? 

So when a user wants to use a toggle button the API would be:

```jsx
<Button isPressed={isPressed} />
```

jzempel: @hzhu, adding `isPressed` to `Button` tends to have the unintended affect of turning all Garden buttons into toggle buttons. Due to the important accessibility distinction, I elected to make `ToggleButton` its own component.

The most recent deployment will have a dedicated `ToggleButton` example. Please have a look at that and the linked documentation (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Toggle_buttons), turn on VO and observe the difference between how a `Button` is read vs. a `ToggleButton`.

jzempel:
@rossmoody @ginnywood will you be able to respond to @allisonacs's concern? I'm fine to take on a different styling direction. Although, for the purposes of this PR, I'd encourage us to steer clear of any tweaks to `<ButtonGroup>` (that treatment is a bit gnarly and might need to be fully reconsidered).

rossmoody:
I didn't notice the ButtonGroup change. Good spot. Since this is effecting other components stylistically my 2 cents would be: 
1. Have a follow-on for the functionality of a `Tile` state and whether it should follow button styling. That seems like something that requires more consideration but could be done on its own.
2. Take a brief pause on this PR for design definition for Pressed state logic of buttons. Right now we are changing the appearance of Pressed states in ButtonGroups so we'd need an Abstract/Sketch library update for this anyway. I investigated the logic that is causing this a little bit so I'll put together an Abstract branch with a proposal.

jzempel:
@rossmoody the styling change to `<ButtonGroup>` was unintended. 🤦  Unwinding that now.

packages/buttons/src/elements/ToggleButton.tsx
austin94: @jzempel how would adding `isPressed` make all buttons register as toggle buttons?

The default case of `<Button>foo</Button>` would have the value of `<button aria-pressed={undefined}>foo</button>`.

[I don't believe that would register as a toggle button.](https://www.w3.org/TR/wai-aria-1.1/#aria-pressed)

jzempel: @austin94 true, but now you've encountered the flip side of the problem. That is that a consumer would need to understand the difference between `<Button>` and `<Button isPressed={false}>` in order to provide a successful toggle button.

Given the fact that a toggle button is not a flavor variety, but a fundamentally different type of button, I think that providing a `<ToggleButton>` is a more human-centered API.

packages/buttons/src/elements/ToggleButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  IToggleButtonProps & React.RefAttributes<HTMLButtonElement>
+> = React.forwardRef<HTMLButtonElement, IToggleButtonProps>(({ isPressed, ...otherProps }, ref) => (
+  <Button aria-pressed={isPressed === undefined ? false : isPressed} ref={ref} {...otherProps} />
```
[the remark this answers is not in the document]
jzempel: Yep, for sure.

PR #758, opened 2020-06-01
feat(buttons): add support for media button

jzempel:
## Description

Adds props for a `start` and `end` icon slot to `<Button>` along with the ability to rotate the icons if desired.

## Detail

This PR is currently based off of and on hold for #757. Let's see how that plays out. May need to rebase this feature off of `master`.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/buttons/src/elements/Button.tsx
+++ b/packages/buttons/src/elements/Button.tsx
@@ -7,7 +7,7 @@
 
 import React, { ButtonHTMLAttributes } from 'react';
 import PropTypes from 'prop-types';
-import { StyledButton } from '../styled';
+import { StyledButton, StyledIcon } from '../styled';
 import { useButtonGroupContext } from '../utils/useButtonGroupContext';
 import { useSplitButtonContext } from '../utils/useSplitButtonContext';
 
@@ -27,6 +27,14 @@ export interface IButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   isPill?: boolean;
   /** Applies inset `box-shadow` styling on focus */
   focusInset?: boolean;
+  /** Slot for "start" icon */
+  start?: any;
+  /** Rotates "start" icon 180 degrees */
+  isStartRotated?: boolean;
+  /** Slot for "end" icon */
+  end?: any;
+  /** Rotates "start" icon 180 degrees */
+  isEndRotated?: boolean;
   /** @ignore prop used by `ButtonGroup` */
   isSelected?: boolean;
 }
@@ -36,7 +44,7 @@ export interface IButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
  */
 const Button: React.FunctionComponent<
   IButtonProps & React.RefAttributes<HTMLButtonElement>
-> = React.forwardRef<HTMLButtonElement, IButtonProps>((props, ref) => {
+> = React.forwardRef<HTMLButtonElement, IButtonProps>(({ start, end, children, ...props }, ref) => {
   const buttonGroupContext = useButtonGroupContext();
   const splitButtonContext = useSplitButtonContext();
 
@@ -58,7 +66,21 @@ const Button: React.FunctionComponent<
     });
   }
 
-  return <StyledButton ref={ref} {...computedProps} />;
+  return (
+    <StyledButton ref={ref} {...computedProps}>
+      {start && (
+        <StyledIcon position="start" isRotated={props.isStartRotated}>
+          {start}
+        </StyledIcon>
+      )}
+      {children}
+      {end && (
+        <StyledIcon position="end" isRotated={props.isEndRotated}>
+          {end}
+        </StyledIcon>
+      )}
+    </StyledButton>
+  );
 });
 
 Button.propTypes = {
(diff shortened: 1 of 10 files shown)


austin94:
@jzempel did you look at providing a separate component to help with this styling?

The prop name duplication seems unique to me:
-  `start`
- `isStartRotated`
- `end`
- `isEndRotated`

Would it be possible to have a static accessor that accomplishes this?

```
<Button>
    <Button.Start isRotated={boolean}>
        <Icon />
    </Button.Start>
    Content
    <Button.End isRotated={boolean}>
        <Icon />
    </Button.End>
</Button>
```

packages/buttons/src/styled/StyledIcon.ts
```diff
@@ -5,16 +5,34 @@
...
 }
 
+const sizeStyles = (props: IStyledIconProps & ThemeProps<DefaultTheme>) => {
+  let marginProperty;
+
+  if (props.position === 'start') {
+    marginProperty = `margin-${props.theme.rtl ? 'left' : 'right'}`;
+  } else if (props.position === 'end') {
+    marginProperty = `margin-${props.theme.rtl ? 'right' : 'left'}`;
+  }
```
austin94, line 26: Is it worthwhile for the `start/end` logic to be handled at the `<Button>` level (maybe with :first-of-type or something)? This may drop the requirement for a `position` prop and allow a single `Button.Icon` accessor or simplified API.

jzempel: If you look at the underlying DOM structure, there's not a way to reliably target the difference between:
- start icon
- end icon
- normal button icon

That's why I'm using the prop to help here.

jzempel:
@austin94, I tend to prefer your `<Button.Start>` static accessor idea. But this deviates from existing APIs provided by [`MediaInput`](https://garden.zendesk.com/react-components/forms/#mediainput), [`Select`](https://garden.zendesk.com/react-components/dropdowns/#select) and [`Autocomplete`](https://garden.zendesk.com/react-components/dropdowns/#autocomplete).

The backing issue calls for API consistency. But I can see making an exception in order to clean up the `isRotated` props. Also, I think we need to be concerned about how static components (especially those with props) don't show up in the generated API documentation.

Thoughts?

packages/buttons/src/styled/StyledIcon.ts
jzempel: You probably already noted, but the `position` prop is effectively private – only exists at the `StyledIcon` level and not exposed in the public element API.

austin94:
For `MediaInput`, `Select`, and `Autocomplete` there isn't a 1-1 mapping of React component to DOM.

Specifically for `MediaInput` the API:

```
<MediaInput {...inputProps} start={S} end={E} />
```

maps to

```
<FauxInput>
  <Start>
    {S}
  </Start>
  <input {...inputProps} />
</FauxInput>
```

With `Button` mapping directly to a `<button>` element I think this allows us to follow this path more easily.

packages/buttons/src/elements/Button.tsx
```diff
@@ -77,5 +77,21 @@ Button.defaultProps = {
...
+(Button as any).StartIcon = StartIcon;
+(Button as any).EndIcon = EndIcon;
+
 /** @component */
-export default Button;
+export default Button as React.FunctionComponent<
+  IButtonProps & React.RefAttributes<HTMLButtonElement>
+> & {
+  StartIcon: typeof StartIcon;
+  EndIcon: typeof EndIcon;
```
jzempel, line 96: @austin94 I have a commit ready that renames these to `Start` and `End` if we decide that's the preferred API.

packages/buttons/src/elements/Button.tsx
```diff
@@ -77,5 +77,21 @@ Button.defaultProps = {
   size: 'medium'
 };
 
+interface IIconProps {
+  isRotated?: boolean;
+  children: any;
+}
+
+const StartIcon = (props: IIconProps) => <StyledIcon position="start" {...props} />;
+const EndIcon = (props: IIconProps) => <StyledIcon position="end" {...props} />;
```
austin94, line 86: Should these elements use `React.forwardRef`?

packages/buttons/src/elements/Button.tsx
austin94: This seems fine to me. Whichever feels best to you.

packages/buttons/src/elements/Button.tsx
jzempel: Testing shows that we can't reference function components due to the child cloning: https://github.com/zendeskgarden/react-components/blob/992c41ff83f4f6349bd491b6cdbe3dc9373b0fae/packages/buttons/src/styled/StyledIcon.ts#L19-L21

I think we'll expect consumers to put the `ref` on the `<Button>` and drill into the DOM if necessary.

jzempel:
Backing this one off to `draft` as I've received new (to me) styling information. Be back soon :wink:

m-lai:
@jzempel What @allisonacs described my initial intent with the icon & text alignment but under suggestion from Ross, the icon & text are centered using a `stretched `property we can apply to Garden. [Demo GIF](https://share.getcloudapp.com/5zuX7Q9J). It seems logical to follow if the buttons are being stretched responsively over having a chunk of empty space on one end for `media.start` and `media.end`? Thoughts?

PR #759, opened 2020-06-03
fix(dropdowns): remove type attribute from div

hzhu:
## Description

There exists a bug that causes the Select Dropdown to render `appearance: button` in Safari. The problem can be seen be visiting this [Codesandbox example](https://codesandbox.io/s/dropdownbedrock-0i0rd) in Safari. This bug blocks https://github.com/zendeskgarden/website/pull/52#discussion_r433376778.

## Detail

The Select Dropdown forwards an unnecessary `type=button` attribute that is applied to the Select Dropdown `div`. CSS bedrock correctly [resets styles](https://github.com/necolas/normalize.css/blob/master/normalize.css#L196) based on `type="button"` on valid elements, but the Select Dropdown incorrectly applies `type=button` onto the `div` element.

I've decided to pluck off the `type="button` and not forward the attribute to be spread onto the `div` based on these findings:

1) I didn't see `type` as a [global attribute](https://html.spec.whatwg.org/#global-attributes) here in the HTML spec.

2) It does not pass the W3 validator:

Screen Shot 2020-06-02 at 7 07 00 PM

## Checklist

      designer as a reviewer)~
- [x] :memo: tested in Chrome, Firefox, Safari, Edge, and ~IE11~

### DIFF
--- a/packages/dropdowns/src/styled/field/StyledSelect.ts
+++ b/packages/dropdowns/src/styled/field/StyledSelect.ts
@@ -135,7 +135,6 @@ export const StyledSelect = styled(FauxInput).attrs<IStyledSelectProps>(props =>
 }))<IStyledSelectProps>`
   position: relative;
   cursor: ${props => (props.disabled ? 'default' : 'pointer')};
-  appearance: none;
   /* stylelint-disable property-no-unknown */
   padding-${props => (props.theme.rtl ? 'left' : 'right')}: ${props =>
   `${props.theme.space.base * 9}px`};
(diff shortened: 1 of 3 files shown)


packages/dropdowns/src/styled/field/StyledSelect.ts
```diff
@@ -135,7 +135,6 @@ export const StyledSelect = styled(FauxInput).attrs<IStyledSelectProps>(props =>
 }))<IStyledSelectProps>`
   position: relative;
   cursor: ${props => (props.disabled ? 'default' : 'pointer')};
-  appearance: none;
```
hzhu, line 138: Since `type="button` is not applied to the Select `div`, we don't have to apply this style.

hzhu: This style can be removed since `type="button` is no longer applied to the Select Dropdown `div`.

PR #786, opened 2020-06-25
feat(dropdowns): allow Select to receive alphanumeric selection

austin94:
## Description

This PR allows users to highlight `Select` values based on the W3C list pattern

> Type a character: focus moves to the next item with a name that starts with the typed character.
> 
> Type multiple characters in rapid succession: focus moves to the next item with a name that starts with the string of characters typed.

### Solution

I based this interaction from the [Collapsible Dropdown List W3C example](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) as well as the [Reach UI Listbox example](https://reacttraining.com/reach-ui/listbox).

The _type multiple characters in rapid succession_ feature adds some complexity to this implementation. To accomplish this I:
- Register the `innerText` of each `<Item>` on render
  - We are unable to rely on `itemToString()` as this is calculated from the value, not the text provided
- On each `Select` keyDown, start a timeout for any alphanumeric key
- Keep track of keys pressed within that timeout, while resetting the timeout each key press
- Searching through the item text value registry based off of the current index when the initial key was selected

## Deployment

As this is large net-positive for a11y and user experience, I have started this work in the v7 branch. With a majority of consumers still using v7 I wanted to validate that an approach could work across both v7 and v8 before continuing. This same work will be cherry-picked (as much as possible) onto v8 after approval.

## Remaining a11y Items

A common feature of the listbox pattern is allowing the alphanumeric key press to select a matching value when the dropdown is closed. This isn't currently possible with our Downshift implementation.

Due to #613 we are unable to fully render the `Item` elements. This makes it impossible for us to know the current Downshift index and `innerText` values.

## Checklist

      designer as a reviewer)
- [x] ~:arrow_left: renders as expected with reversed (RTL) direction~
- [x] ~:memo: tested in Chrome, Firefox, Safari, Edge, and IE11~

### DIFF
--- a/packages/dropdowns/src/Select/Select.tsx
+++ b/packages/dropdowns/src/Select/Select.tsx
@@ -6,7 +6,7 @@
  */
 
 import React, { useRef, useState, useEffect, useCallback, HTMLAttributes } from 'react';
-import { useCombinedRefs } from '@zendeskgarden/container-utilities';
+import { useCombinedRefs, KEY_CODES } from '@zendeskgarden/container-utilities';
 import PropTypes from 'prop-types';
 import { Reference } from 'react-popper';
 import { StyledInput, StyledSelect, VALIDATION } from '../styled';
@@ -111,8 +111,18 @@ export const Select = React.forwardRef<HTMLDivElement, ISelectProps>(
     const onInputKeyDown = useCallback(
       (e: React.KeyboardEvent) => {
         // Only search with alphanumeric keys
-        if ((e.keyCode < 48 || e.keyCode > 57) && (e.keyCode < 65 || e.keyCode > 90)) {
+        if (
+          (e.keyCode < 48 || e.keyCode > 57) &&
+          (e.keyCode < 65 || e.keyCode > 90) &&
+          e.keyCode !== KEY_CODES.SPACE
+        ) {
           return;
+        }
+
+        // Prevent space from closing Menu
+        if (e.keyCode === KEY_CODES.SPACE) {
+          e.preventDefault();
+          e.stopPropagation();
         }
 
         const character = String.fromCharCode(e.which || e.keyCode);
--- a/packages/dropdowns/src/Select/Select.tsx
+++ b/packages/dropdowns/src/Select/Select.tsx
@@ -78,9 +78,9 @@ export const Select = React.forwardRef<HTMLDivElement, ISelectProps>(
       }
 
       // Reset search string after delay
-      searchTimeoutRef.current = (setTimeout(() => {
+      searchTimeoutRef.current = window.setTimeout(() => {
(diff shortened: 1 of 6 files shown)


packages/dropdowns/src/Select/Select.tsx
```diff
@@ -58,6 +68,85 @@ export const Select = React.forwardRef<HTMLDivElement, ISelectProps>(
...
+    useEffect(() => {
+      // Cancel existing timeout if keys are pressed rapidly
+      if (searchTimeoutRef.current) {
+        clearTimeout(searchTimeoutRef.current);
+      }
+
+      // Reset search string after delay
+      searchTimeoutRef.current = (setTimeout(() => {
+        setSearchString('');
+      }, 500) as unknown) as number;
```
hzhu, line 83: An alternative to casting the timeout's return value as `unknown` and then to `number` would be to call `setTimeout` off of the window property. I think this tells TypeScript its a browser environment timeout. It then uses the correct browser timeout type definitions, rather than the NodeJS timeout type definitions.

```tsx
searchTimeoutRef.current = window.setTimeout(() => {
  setSearchString('');
}, 500);
```

austin94: TIL that's really nice. Updated.

austin94:
Here is a draft publish for review: https://5ef634b66a1798021448604b--garden-preview-v1.netlify.app/dropdowns/#select-usage

I also just realized that we intercept `space` commands to open/close Dropdowns. Changing this feature just for `Select` seems odd, but end-users are limited to only searching continuous strings. Is this something you think we should enable?

## Testing Update

It seems [jsdom doesn't implement `innerText`](https://5ef634b66a1798021448604b--garden-preview-v1.netlify.app/dropdowns/#select-usage). So this feature is not testable currently.

jzempel:
> Is this something you think we should enable?

Ref http://nativeformelements.com/. It appears that a timeout is incorporated into the native `<select>` to distinguish between space used for open/close/select and space used for search. We should model the same behavior.

austin94:
@jzempel here is an updated draft deploy that allows `space` in searches https://5efa1b107960e8763efa58bc--garden-preview-v1.netlify.app/dropdowns/

austin94:
Here is another draft: https://5efa2e5dc3b0af02161b8a42--garden-preview-v1.netlify.app/dropdowns/#menu-usage

This includes:
- Allowing `space` to interact with a searchable text
- If `space` is pressed in isolation it will select/close the highlighted item
- All filtering logic is now added to the `Menu` element as well to align with the [W3C Menu Button spec](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html)

PR #788, opened 2020-06-29
fix(notifications): add support for regular weight title

hzhu:
## Description

Add the ability to show a Notification's title in regular font-weight.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/notifications/src/elements/content/Title.tsx
+++ b/packages/notifications/src/elements/content/Title.tsx
@@ -8,14 +8,19 @@
 import React, { HTMLAttributes } from 'react';
 import { StyledTitle } from '../../styled';
 
+export interface ITitleProps extends HTMLAttributes<HTMLDivElement> {
+  /** Style using regular (non-bold) font weight */
+  isRegular?: boolean;
+}
+
 /**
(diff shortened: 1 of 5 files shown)


packages/notifications/src/elements/content/Title.tsx
```diff
@@ -8,14 +8,19 @@
 import React, { HTMLAttributes } from 'react';
 import { StyledTitle } from '../../styled';
 
+export interface ITitleProps extends HTMLAttributes<HTMLDivElement> {
+  /** Style using regular (non-bold) font weight */
+  isRegular?: boolean;
```
austin94, line 13: Even if it's optional, should we include a `defaultProp` of `false` show it's shown correctly in the generated prop-sheets?

jzempel: @austin94 that isn't consistent with any of our other boolean props. Consider https://garden.zendesk.com/react-components/forms/#label, for example.

austin94: 👍 Sounds good to me

PR #790, opened 2020-07-01
fix(chrome): add fixed positioning to html element

hzhu:
## Description

Prevent fixed positioning from being applied to the <html> element

## Detail

This [CodeSandbox](https://codesandbox.io/s/loving-fast-86t3l?file=/src/examples/Example.js) demonstrates the issue. Notice that there is extra white space on the bottom. Also, click the `Checkbox` and notice the screen jumps to the bottom of the white space. Applying a fixed position to the `html` element resolves the issue.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/chrome/src/elements/Chrome.spec.tsx
+++ b/packages/chrome/src/elements/Chrome.spec.tsx
@@ -18,14 +18,14 @@ describe('Chrome', () => {
     expect(container.firstChild).toBe(ref.current);
   });
 
-  it('does not render fluid style on <html> element by default', () => {
-    render(<Chrome />);
+  it('renders fluid html element', () => {
+    render(<Chrome isFluid />);
 
-    expect(document.querySelector('html')).not.toHaveAttribute('style');
+    expect(document.querySelector('html')).toHaveAttribute('style', '');
   });
 
-  it('renders fluid style on <html> element and removes it when Chrome is unmounted', () => {
-    const { unmount } = render(<Chrome isFluid />);
+  it('does not render fluid html element', () => {
+    const { unmount } = render(<Chrome />);
 
     expect(document.querySelector('html')).toHaveAttribute('style', 'position: fixed;');
     unmount();
--- a/packages/chrome/src/elements/Chrome.spec.tsx
+++ b/packages/chrome/src/elements/Chrome.spec.tsx
@@ -18,6 +18,20 @@ describe('Chrome', () => {
     expect(container.firstChild).toBe(ref.current);
   });
 
+  it('does not render fluid style on <html> element by default', () => {
+    render(<Chrome />);
+
+    expect(document.querySelector('html')).not.toHaveAttribute('style');
+  });
+
+  it('renders fluid style on <html> element and removes it when Chrome is unmounted', () => {
+    const { unmount } = render(<Chrome isFluid />);
+
+    expect(document.querySelector('html')).toHaveAttribute('style', 'position: fixed;');
+    unmount();
+    expect(document.querySelector('html')).toHaveAttribute('style', '');
+  });
+
   describe('Hue', () => {
     it('applies light styling if hue is below luminance threshold', () => {
       const { container } = render(<Chrome hue={PALETTE.green[100]} />);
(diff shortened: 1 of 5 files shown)


packages/chrome/src/elements/Chrome.tsx
```diff
@@ -38,6 +40,27 @@ const Chrome = React.forwardRef<HTMLDivElement, IChromeProps & ThemeProps<Defaul
...
+    const environment = useDocument(theme);
+
+    useEffect(() => {
+      if (!environment) {
+        return undefined;
+      }
+
+      const htmlElement = environment.querySelector('html');
+
+      if (htmlElement && isFluid) {
```
jzempel, line 52: "fluid" is the opposite of "fixed". The prop should stay defined as-is. However if `isFluid={true}` then no `position` style is added. Otherwise, `position: fixed` is added to the `<html>` element by default.

Consequently `isFluid` would be added to all styleguidist examples so that the page continues to scroll.

jzempel: Recommend changing the title of this PR to "fix(chrome): add fixed positioning to html element"

packages/chrome/src/elements/Chrome.spec.tsx
```diff
@@ -18,6 +18,20 @@ describe('Chrome', () => {
     expect(container.firstChild).toBe(ref.current);
   });
 
+  it('renders fluid html element', () => {
+    render(<Chrome isFluid />);
+
+    expect(document.querySelector('html')).toHaveAttribute('style', '');
```
jzempel, line 24: [nit] I think you could also use `...'html')).not.toHaveAttribute('style');`

packages/chrome/src/elements/Chrome.tsx
```diff
@@ -5,24 +5,26 @@
...
 import { readableColor } from 'polished';
-import { withTheme, DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
+import { withTheme, DEFAULT_THEME, getColor, useDocument } from '@zendeskgarden/react-theming';
 import { ChromeContext } from '../utils/useChromeContext';
 import { StyledChrome } from '../styled';
 
 interface IChromeProps extends HTMLAttributes<HTMLDivElement> {
   /** Apply a custom hue to chrome navigation */
   hue?: string;
+  /** Apply a fixed position to the <html> element */
```
jzempel, line 19: Please update this description. Something like `Prevent fixed positioning from being applied to the <html> element`

packages/chrome/src/elements/Chrome.tsx
```diff
@@ -38,6 +40,27 @@ const Chrome = React.forwardRef<HTMLDivElement, IChromeProps & ThemeProps<Defaul
...
+      const defaultHtmlPosition = htmlElement.style.position;
+
+      if (!isFluid) {
+        htmlElement.style.position = 'fixed';
+      }
+
+      return () => {
+        htmlElement.style.position = defaultHtmlPosition;
+      };
+    }, [environment, isFluid]);
```
jzempel, line 63: Falsy logic tends to be difficult to parse and is missing the opportunity to short-circuit. Alternative:

```js
if (environment && !isFluid) {
  const htmlElement = environment.querySelector('html');

  if (htmlElement) {
    const defaultHtmlPosition = htmlElement.style.position;

    htmlElement.style.position = defaultHtmlPosition;

    return () => {
      htmlElement.style.position = defaultHtmlPosition;
    }
  }
}
```

hzhu: Thanks that's easier to read.

packages/chrome/src/elements/Chrome.spec.tsx
hzhu: Hmm that didn't work.

jzempel: 🤷 fine as-is.

packages/chrome/src/elements/Chrome.tsx
jzempel: ☝️

hzhu: Doh. In some universe, I read this as update the PR description.

jzempel: 😅

PR #811, opened 2020-07-16
chore(docs): update Table selection example to allow keyboard

austin94:
## Description

This PR adds some keyboard logic to our selection example. This logic follows the logic shown in [ag-grid](https://www.ag-grid.com/example-runner/grid-vanilla.php?section=javascript-grid-selection&example=multiple-row-selection&generated=1&grid=%7B%22noStyle%22%3A0%2C%22height%22%3A%22100%25%22%2C%22width%22%3A%22100%25%22%2C%22enterprise%22%3Anull%7D)

There are some edge-cases since we provide checkboxes for selection rather than a grid pattern, but overall this feels like a good middle-ground.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/tables/examples/selectable-table.md
+++ b/packages/tables/examples/selectable-table.md
@@ -89,8 +89,6 @@ const isSelectAllChecked = rows => {
                 setState({ shiftEnabled: false });
               }}
               onChange={e => {
-                console.log('onChange', state.shiftEnabled, state.focusedRowIndex);
-
                 const updatedRows = [...state.rows];
 
                 if (state.shiftEnabled && state.focusedRowIndex !== undefined) {
@@ -100,8 +98,6 @@ const isSelectAllChecked = rows => {
                   const isAllChecked = updatedRows
                     .slice(startIndex, endIndex + 1)
                     .every(row => row.selected);
-
-                  console.log(isAllChecked);
 
                   for (let x = startIndex; x <= endIndex; x++) {
                     if (x === index && isAllChecked) {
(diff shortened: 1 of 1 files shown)


hzhu:
@austin94 could you clarify on the instructions for interacting with the selectable table using the keyboard? I was unable to replicate behavior on ag-grid.

I tried holding `shift` on a checkbox (also tried holding `shift` after clicking a table row), followed by pressing arrow keys. I wasn't able to get selection to move around.

packages/tables/examples/selectable-table.md
```diff
@@ -66,22 +74,48 @@ const isSelectAllChecked = (selectedRows, rows) => {
...
   <Body>
-    {state.rows.map(row => (
-      <Row key={row.id} isSelected={state.selectedRows[row.id]}>
+    {state.rows.map((row, index) => (
+      <Row key={row.id} isSelected={row.selected}>
         <Cell isMinimum>
           <Field>
             <Checkbox
-              checked={state.selectedRows[row.id] ? true : false}
+              checked={row.selected ? true : false}
```
hzhu, line 82: Since `row.selected` is a boolean, could this line be simplified by removing the ternary operation?

```jsx
checked={row.selected}
```

austin94: Updated.

austin94:
@hzhu @steue I've updated the example to include some basic instructions for select-ability.

I only used `ag-grid` as a basis for the selection interaction. That component follows the [W3C Grid Pattern](https://www.w3.org/TR/wai-aria-practices/#grid) which differs from our native `<table>` implementation. We aren't replicating the ability to navigate between individual cells/rows with the keyboard.

> is there a keyboard input only equivalent

This would only be applicable if we were following the Grid pattern where an entire row or cell group could be selected (usually with the `SHIFT + ARROW KEY` interaction).

Since we are using standalone checkboxes keyboard users could just select them as they go through the tab order.

PR #812, opened 2020-07-16
fix(forms): implement label behavior for MultiThumbRange

hzhu:
## Description

This pull request implements label (visual) behavior for the `<MultiThumbRange>`.

## Detail

The `MultiThumbRange` label new behavior allows users to interact with the multi-thumb range label using a mouse. 

1) Hovering over the label visually indicates the min thumb
2) Pressing (and holding) the label visually indicates the min thumb in it's activated state
3) Clicking the label sets the min thumb as the active element allowing for keyboard navigation

## Checklist

      designer as a reviewer)


.stylelintrc
```diff
@@ -13,5 +13,6 @@
     "number-leading-zero": null,
     "selector-type-case": null,
     "selector-type-no-unknown": null
-  }
+  },
+  ignoreFiles: ["packages/**/*.spec.{js,ts,tsx}"]
```
hzhu, line 17: Test files are ignored in the [yarn lint:css](https://github.com/zendeskgarden/react-components/blob/main/package.json#L17) command, but not ignored in the [.lintstagedrc](https://github.com/zendeskgarden/react-components/blob/main/.lintstagedrc#L3) file. Because I ran into some wild stylelint issues when writing tests using css modifiers, I added this line to ignore `stylelint` when husky commands are ran.

Alternatively, I also tried modifying `.lintstaged` file with something like `stylelint --ignore-files "packages/**/*.spec.{js,ts,tsx}"`, but had no luck. Open to suggestions for getting around css modifier linting issues.

jzempel: Seems fine. I think we tend to run `git commit --no-verify` in order to get around issues like this.

packages/forms/src/elements/MultiThumbRange.spec.tsx
```diff
@@ -57,11 +67,107 @@ describe('MultiThumbRange', () => {
...
+      const label = getByText('MultiThumbRange');
+      const minThumb = getAllByRole('slider')[0];
+
+      expect(minThumb).toHaveAttribute('data-garden-hover', 'false');
+      expect(minThumb).toHaveStyleRule('border-color', getColor('blue', 600));
+      expect(minThumb).toHaveStyleRule('background-color', getColor('blue', 600));
+
+      fireEvent.mouseEnter(label);
+
+      expect(minThumb).toHaveAttribute('data-garden-hover', 'true');
```
hzhu, line 120: Removing this assertion as it is implicitly tested below

.stylelintrc
hzhu: That sounds good. I ended up leaving this since `--no-verify` would skip the eslint, jest, and prettier too. But I can manually run those commands and use `--no-verify` if that is preferred 👍 .

austin94: I think this makes sense. We `stylelint` is for production styling and CSS. Tests should be fine without it.

package.json
```diff
@@ -45,6 +45,7 @@
     "@testing-library/jest-dom": "5.11.0",
     "@testing-library/react": "10.4.6",
     "@testing-library/react-hooks": "3.3.0",
+    "@testing-library/user-event": "^12.0.11",
```
austin94, line 48: What is the requirement for this library? We currently use `fireEvent.click` for our interaction testing.

If required we should lock the dependency range as well.

packages/forms/src/elements/MultiThumbRange.spec.tsx
```diff
@@ -57,11 +67,103 @@ describe('MultiThumbRange', () => {
...
+        <Field>
+          <Label>MultiThumbRange</Label>
+          <MultiThumbRange />
+        </Field>
+      );
+      const label = getByText('MultiThumbRange');
+      const minThumb = getAllByRole('slider')[0];
+
+      expect(minThumb).not.toHaveFocus();
+      userEvent.click(label);
```
austin94, line 82: Should this use our standard `fireEvent` strategy?

packages/forms/src/elements/MultiThumbRange.spec.tsx
```diff
@@ -57,11 +67,103 @@ describe('MultiThumbRange', () => {
     });
   });
 
+  describe('MultiThumbRange label', () => {
+    test('clicking the label focuses the min thumb', () => {
```
austin94, line 71: We don't use the `test()` helper in other test files. Let's stick with our standard `it()` usage.

packages/forms/src/elements/MultiThumbRange.spec.tsx
```diff
@@ -98,7 +206,10 @@ describe('MultiThumbRange', () => {
     it('updates max value if slider is clicked near max thumb', () => {
       const onChangeSpy = jest.fn();
       const { getByTestId } = render(
-        <MultiThumbRange minValue={15} maxValue={75} step={5} onChange={onChangeSpy} />
+        <Field>
+          <Label>MultiThumbRange</Label>
```
austin94, line 210: Do all of these existing tests require a `Label`?

packages/forms/src/elements/MultiThumbRange.tsx
```diff
@@ -65,6 +66,13 @@ const MultiThumbRange: React.FC<IMultiThumbRangeProps & ThemeProps<DefaultTheme>
   const trackRailRef = useRef<HTMLDivElement>(null);
   const minThumbRef = useRef<HTMLDivElement>(null);
   const maxThumbRef = useRef<HTMLDivElement>(null);
+  const fieldContext = useFieldContext();
+
+  if (fieldContext === undefined) {
+    throw new Error('A MultiThumbRange must be rendered within a form Field component.');
+  }
```
austin94, line 73: I'm not sure this structure should be required. Other elements in the `forms` package mark the `Field` wrapping as optional. 

This would technically be a breaking change as well since it requires a new DOM ordering.

packages/forms/src/elements/MultiThumbRange.spec.tsx
```diff
@@ -57,11 +67,103 @@ describe('MultiThumbRange', () => {
...
+          <MultiThumbRange />
+        </Field>
+      );
+      const label = getByText('MultiThumbRange');
+      const minThumb = getAllByRole('slider')[0];
+
+      expect(minThumb).toHaveStyleRule('border-color', getColor('blue', 600));
+      expect(minThumb).toHaveStyleRule('background-color', getColor('blue', 600));
```
austin94, line 115: We apply `data-garden-*` testing values in the DOM. Could we reference these instead of comparing CSS values?

packages/forms/src/elements/common/Label.tsx
```diff
@@ -37,6 +38,26 @@ export const Label = React.forwardRef<HTMLLabelElement, ILabelProps>((props, ref
...
+      const { setIsLabelActive, setIsLabelHovered } = fieldContext;
+
+      combinedProps = {
+        ...combinedProps,
+        onMouseUp: composeEventHandlers(props.onMouseUp, () => {
+          setIsLabelActive(false);
+        }),
+        onMouseDown: composeEventHandlers(props.onMouseDown, () => {
+          setIsLabelActive(true);
+        }),
```
austin94, line 52: By controlling focus from the `mouseDown` callback, I'm not sure this matches native functionality as expected.

If someone were to `mouseDown` the label, but then drag away and not fully click, this would still trigger the minimum thumb to be selected.

From our offline discussion, it may be easier for the `MinThumb` to provide its `ref` via context. This `Label` could then have something like:

```
onClick: () => minThumbRef.current.focus()
```

packages/forms/src/elements/MultiThumbRange.spec.tsx
hzhu: Opps, old habit. Will push new commit that uses `it` shortly.

allisonacs:
Not sure if this has been caught, but it looks like the minimum thumb is catching hover styling when focusing/clicking on the label. When you click and hold on the label, the control gets active styling (which feels super weird, bc I'm explicitly not clicking on that control). Tabbing through does not remove hover (nor does it apply focus styling), although moving the mouse does. https://share.getcloudapp.com/Z4uY0qRB

IMO that interaction needs to be fixed. If something can be controlled via keyboard, I'd expect it to get focus treatment, not hover treatment. Focus only appears on the min thumb once you start manipulating it with the arrow keys (and it keeps the hover color treatment), which again I wouldn't expect.

packages/forms/src/elements/MultiThumbRange.tsx
hzhu: I can update the code so that this new behavior is opt-in and not a breaking change. I could also create a v9 task that enforces this change pending what the team thinks.

austin94:
> When you click and hold on the label, the control gets active styling

@allisonacs I believe this is actually intended. This matches native input (including range) functionality.

package.json
hzhu: The `user-event` library isn't strictly required. I think it could removed, but here my rationale: its APIs simulate more accurate user interaction than manually using a single `fireEvent` API. For example, to assert against hovering, `user.hover(element)` does a lot more than simply `fireEvent.mouseEnter(element)`. It fires other events that would occur in an actual hover:

```jsx
function hover(element, init) {
  if (isLabelWithInternallyDisabledControl(element)) return

  fireEvent.pointerOver(element, init)
  fireEvent.pointerEnter(element, init)
  if (!element.disabled) {
    fireEvent.mouseOver(element, getMouseEventOptions('mouseover', init))
    fireEvent.mouseEnter(element, getMouseEventOptions('mouseenter', init))
  }
  fireEvent.pointerMove(element, init)
  if (!element.disabled) {
    fireEvent.mouseMove(element, getMouseEventOptions('mousemove', init))
  }
}
```

The same applies to the two other APIs I wanted to use in this PR: [`user.click(element)`](https://github.com/testing-library/user-event/blob/master/src/click.js#L81) and [`user.tab(element)`](https://github.com/testing-library/user-event/blob/master/src/tab.js#L22). Also, the `user-event` package is included in the default `create-react-app` now. Looks like it's getting wide adoption 🎉 .

_Note: I forgot to use `user.hover(element)` in my [original PR](https://github.com/zendeskgarden/react-components/pull/812/files#diff-a0cecbd65bdd8164d6cb5cdc9acf81c5R117)._

hzhu: We could slowly migrate to moving towards using `user-events` APIs over-time. An introduction of this package could include a wider team discussion. So I leave this package out of this PR.

packages/forms/src/elements/MultiThumbRange.spec.tsx
hzhu: Jumping from the comment about breaking changes -> I will update these tests so that multi-thumb range usages do not require a `Label`. New commits coming shortly 👍 .

jzempel:
> I believe this is actually intended

Yes, this is intended. We want the first thumb of the multi-range to match the existing `Range` (which is a fully native input).

packages/forms/src/elements/MultiThumbRange.spec.tsx
hzhu: Wouldn't asserting against `data-garden-*` make this a test brittle which encourages assertions against implementation details over user experience?

What if a developer accidentally mistyped the CSS property? Or maybe they provided an incorrect color? Or the CSS properties were accidentally removed? The `data-garden-*` attributes would still be valid, the test still passes ✅ , but the user would see a broken UI.

packages/forms/src/elements/common/Label.tsx
hzhu: > If someone were to mouseDown the label, but then drag away and not fully click, this would still trigger the minimum thumb to be selected.

After some pair programming with @austin94 and some exploration, I've concluded that the only way to achieve such an interaction is leveraging an event listener on the `document` that listens for a `mouseup` event:

```jsx
useEffect(() => {
  if (fieldContext) {
    if (isLabelActive && minThumbRef.current) {
      minThumbRef.current.focus();
    }

    const listener = () => {
      if (!fieldContext.isLabelHovered) {
        minThumbRef.current && minThumbRef.current.blur();
        setIsLabelActive && setIsLabelActive(false);
        document.body.focus();
      }
    };

    window.document.documentElement.addEventListener('mouseup', listener);

    return () => {
      window.document.documentElement.removeEventListener('mouseup', listener);
    };
  }

  return undefined;
  }, [fieldContext, isLabelActive, setIsLabelActive]);
```

We've concluded that the above code is invasive due to the global event listener and decided to keep the interaction as-is.

jzempel: [nit] something about this interaction seems to be preventing `user-select` on the label. In other words, I can't drag to select label text for copy like I can with the native control.

hzhu: Awesome, thanks for the catch. I'll take a look at this issue.

hzhu: It looks like when the `mousedown` event is fired (and the min thumb is simultaneously focused) text selection is interrupted. 

I've decoupled the min thumb focus from the `mousedown` event by focusing on the min thumb on a `click` event instead which fixes the text selection issue.

hzhu: It looks like when the `mousedown` event is fired (and the min thumb is simultaneously focused) text selection is interrupted. 

I've decoupled the min thumb focus from the `mousedown` event by focusing on the min thumb on a `click` event instead which fixes the text selection issue.

Edit: @jzempel I've re-requested a review for the [new commit](https://github.com/zendeskgarden/react-components/pull/812/commits/8caf0eeb64667d05b3b44669db3830624751abff) that addresses the text selection issue.

PR #820, opened 2020-07-30
feat(forms): allow Textarea to dynamically change height

austin94:
## Description

This PR introduces the concept of auto-resizing to the `Textarea` component. This functionality is disabled by default.

### Props

This feature is enabled with two new props, `minRows` and `maxRows`. These new props define the upper and lower bounds for overflow when resizing and only enable auto-resizing if the `isResizable` props is not provided.

### Size Calculations

After looking at several other implementations, the most common way of achieving this functionality in a cross-browser compatible way is to use hidden "shadow" input with the same styling to calculate when an overflow occurs. This shadow input is only rendered if the auto-resizing props are provided.

Resizing is handled on initial render, when the screen is resized (with debounce), and when the input value is changed. It also handles both controlled and uncontrolled usages.

The use of the shadow input has some great results in mobile browsers when compared against simpler size calculations.

### Testing

JSDOM makes it difficult to mock scrollHeight and other computed style properties. I was able to test basic functionality, but some of the more complex edge-cases are only reproducible in a browser environment.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/elements/Textarea.tsx
+++ b/packages/forms/src/elements/Textarea.tsx
@@ -67,7 +67,7 @@ export const Textarea = React.forwardRef<HTMLTextAreaElement, ITextareaProps>(
         parseStyleValue(computedStyle.paddingBottom) + parseStyleValue(computedStyle.paddingTop);
       const border =
         parseStyleValue(computedStyle.borderTopWidth) +
-        parseStyleValue(computedStyle.borderTopWidth);
+        parseStyleValue(computedStyle.borderBottomWidth);
 
       const innerHeight = shadowTextArea.scrollHeight - padding;
 
--- a/packages/forms/src/elements/Textarea.tsx
+++ b/packages/forms/src/elements/Textarea.tsx
@@ -169,6 +169,8 @@ export const Textarea = React.forwardRef<HTMLTextAreaElement, ITextareaProps>(
             className={props.className}
             ref={shadowTextAreaRef}
             tabIndex={-1}
+            isBare={props.isBare}
+            isCompact={props.isCompact}
             style={style}
           />
         )}
--- a/packages/forms/src/elements/Textarea.tsx
+++ b/packages/forms/src/elements/Textarea.tsx
@@ -5,8 +5,10 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { TextareaHTMLAttributes } from 'react';
+import React, { TextareaHTMLAttributes, useRef, useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


packages/forms/src/elements/Textarea.tsx
```diff
@@ -20,27 +22,160 @@ export interface ITextareaProps extends TextareaHTMLAttributes<HTMLTextAreaEleme
...
+          <StyledTextarea
+            aria-hidden
+            readOnly
+            isHidden
+            className={props.className}
+            ref={shadowTextAreaRef}
+            tabIndex={-1}
+            style={style}
+          />
+        ) : null}
```
jzempel, line 174: Is this more readable as `{isAutoResizable && <StyledTextarea... />}`?

austin94: I just stumbled onto https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx and it had some good points against `&&` in render functions.

I can update this if you feel strongly against it.

jzempel: The article is guarding against array checking. As this is a boolean, I'd like to think `&&` is a reliable choice.

austin94: Updated.

packages/forms/src/elements/Textarea.tsx
```diff
@@ -20,27 +22,160 @@ export interface ITextareaProps extends TextareaHTMLAttributes<HTMLTextAreaEleme
...
+      if (!isAutoResizable) {
+        return;
+      }
+
+      const textarea = textAreaRef.current!;
+      const computedStyle = window.getComputedStyle(textarea);
+      const shadowTextArea = shadowTextAreaRef.current!;
+
+      shadowTextArea.style.width = computedStyle.width;
+      shadowTextArea.value = textarea.value || textarea.placeholder || ' ';
```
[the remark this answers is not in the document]
austin94: There was an issue relating to this where I wasn't passing `isBare` and `isCompact` to the shadow element. This is now resolved, but doesn't address this specific issue.

The textarea inherits it's min-height from the `StyledTextInput` component. This min-height is causing the shadow element to pin it's height when the font-size is smaller. If you increase the font-size the value is calculated correctly. When the `isBare` prop is provided the font-size is handled well.

Would we want to remove the `min-height` value for the `Textarea` to support this use-case?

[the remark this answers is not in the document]
jzempel: No, removing `min-height` is not desirable. I think it's ok to get close on this one, with the expectation that folks won't typically be manipulating the `font-size`.

PR #821, opened 2020-08-03
feat(buttons): default external links to use safe referrer attributes

austin94:
## Description

Adds default values for the `target` and `rel` attributes for Anchors that include the `isExternal` prop. These values are customizable for advanced use-cases.

https://web.dev/external-anchors-use-rel-noopener/

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/buttons/src/elements/Anchor.spec.tsx
+++ b/packages/buttons/src/elements/Anchor.spec.tsx
@@ -15,4 +15,26 @@ describe('Anchor', () => {
 
     expect(container.querySelector('svg')).not.toBeNull();
   });
+
+  it('renders link security attributes while external', () => {
+    const { getByText } = render(<Anchor isExternal>link</Anchor>);
+
+    const anchor = getByText('link');
+
+    expect(anchor).toHaveAttribute('target', '_blank');
+    expect(anchor).toHaveAttribute('rel', 'noopener noreferrer');
+  });
+
+  it('renders custom link attributes if provided while external', () => {
+    const { getByText } = render(
+      <Anchor isExternal rel="bookmark" target="_parent">
+        link
+      </Anchor>
+    );
+
+    const anchor = getByText('link');
+
+    expect(anchor).toHaveAttribute('target', '_parent');
+    expect(anchor).toHaveAttribute('rel', 'bookmark');
+  });
 });
--- a/packages/buttons/.size-snapshot.json
+++ b/packages/buttons/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 25386,
-    "minified": 18207,
-    "gzipped": 4570
+    "bundled": 25556,
+    "minified": 18276,
+    "gzipped": 4612
   },
   "index.esm.js": {
-    "bundled": 24495,
-    "minified": 17385,
-    "gzipped": 4450,
+    "bundled": 24665,
+    "minified": 17454,
+    "gzipped": 4491,
     "treeshaked": {
       "rollup": {
-        "code": 13488,
+        "code": 13544,
         "import_statements": 383
       },
       "webpack": {
-        "code": 15359
+        "code": 15415
       }
     }
   }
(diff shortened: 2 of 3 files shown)


packages/buttons/src/elements/Anchor.spec.tsx
```diff
@@ -15,4 +15,26 @@ describe('Anchor', () => {
 
     expect(container.querySelector('svg')).not.toBeNull();
   });
+
+  it('renders link security attributes while external', () => {
+    const { getByText } = render(<Anchor isExternal>link</Anchor>);
+
+    const anchor = getByText('link');
+
+    expect(anchor).toHaveAttribute('target', '_blank');
```
alanhogan, line 24: I’m not sure that we want all external links to open in new windows. Do we? 

In my understanding, the `rel` precautions only apply to new windows. Then we'd want to auto-set them if and only if the user of this component already wanted `target=_blank`.

That said, if it were a product recommendation at Zendesk for our products to always open third-party links in a new window, well, that is fair enough and in that case this line LGTM!

allisonacs: Good call out @alanhogan. It is Garden's recommendation to always open third-arty links in a new window; we anticipate this mimics user intention most of the time. Feels like a good, smart default (while still allowing for customized use cases where we might not want this).

allisonacs:
@austin94 it doesn't appear that any of the docs have been updated to reference these improvements. I'd recommend we do that either in this branch or an immediate follow-up.

austin94:
With these docs living in https://github.com/zendeskgarden/website now we will need to wait until `react-buttons` is updated there.

packages/buttons/src/elements/Anchor.spec.tsx
austin94: I think the `external` icon is most often used to denote that a link is opening in a new context (tab in this case), not that the link is going to a third-party.

Another option would be to only add the `rel` attribute if `target="_blank"` is provided. This would be more specific, but starts to feel complicated for consumers to figure out.

austin94: But, to @allisonacs 's point, most third-party links are opened in a new tab from this guidance. Even for first-party properties this seems like a good benefit and is customizable if tweaks are necessary.

bence-toth:
Copying my comment here from Slack (sorry if not relevant):

My suggestion would be to default to `target="_blank"` and `rel="noopener roferrer"` only if `isExternal` is set.

If `isExternal` is not set, I would probably expect no change in functionality.

Whether a link is external should depend on the individual use-case, and should not depend on the `href` value (and whether that belongs to the same domain or another domain).

In some cases we want internal (same-domain) links to behave as external links (just so that they open on a new tab), however in these cases we could to manually override `rel` if we do actually rely on the browsing context -- opting in instead of opting out.

Examples for clarity:
- `<Anchor href="url_from_another_domain" isExternal>Text</Anchor>`
would render
`<a href="url_from_another_domain" target="_blank" rel="noopener roferrer">Text <i class="externalLinkIcon"></i></a>`

- `<Anchor href="url_from_another_domain" target="_self" isExternal>Text</Anchor>`
would render
`<a href="url_from_another_domain" target="_self" rel="noopener roferrer">Text <i class="externalLinkIcon"></i></a>`

- `<Anchor href="url_from_another_domain" rel="noopener" isExternal>Text</Anchor>`
would render
`<a href="url_from_another_domain" target="_blank" rel="noopener">Text <i class="externalLinkIcon"></i></a>`

- `<Anchor href="url_from_another_domain" target="_self">Text</Anchor>`
would render
`<a href="url_from_another_domain" target="_self">Text</a>`,
could probably also show a warning or notice reminding you about the `isExternal` prop (because `target="_self"` was set)

- `<Anchor href="url_from_another_domain" rel="noopener">Text</Anchor>`
would render
`<a href="url_from_another_domain" rel="noopener">Text</a>`,
could probably also show a warning or notice reminding you about the `isExternal` prop (because `rel="noopener"` was set, the same for `noreferrer`, and any combination of `noopener` and `noreferrer`)

On the other hand, I see how warnings about another-domain links that don’t have `isExternal` would be valuable, I’m working with the assumption that it’s a rare use-case (and I may be totally wrong here), and I would also argue that it's something which does not belong to the responsibilities of a component library.

austin94:
> My suggestion would be to default to target="_blank" and rel="noopener roferrer" only if isExternal is set.

With this being the current implementation, I'm going to go ahead and merge this. Thanks for the insights!

PR #822, opened 2020-08-04
chore(tests): add user-event testing library

hzhu:
## Description

This pull request integrates the [@testing-library/user-event](https://www.npmjs.com/package/@testing-library/user-event) package to help better simulate real events. From the docs:

> user-event tries to simulate the real events that would happen in the browser as the user interacts with it. 

## Detail

* This PR only touches `package.json` and `.spec` files
* Most of this PR is a simple migration from `fireEvent` to `userEvent`
* Some tests will still require `fireEvent` as `userEvent` is not a 1 to 1 replacement
* I rewrote two tests as they were false positives which have been highlighted in the comments

## Resources

* The testing-lib docs have a [great section](https://testing-library.com/docs/guide-events) on events and testing
* A list of available `userEvent` APIs from [the docs](https://www.npmjs.com/package/@testing-library/user-event#api)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@testing-library/jest-dom": "5.11.0",
     "@testing-library/react": "10.4.6",
     "@testing-library/react-hooks": "3.3.0",
-    "@testing-library/user-event": "^12.1.0",
+    "@testing-library/user-event": "12.1.0",
     "@types/jest": "26.0.4",
     "@types/prop-types": "15.7.3",
     "@types/react": "^16.9.33",
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@testing-library/jest-dom": "5.11.0",
     "@testing-library/react": "10.4.6",
     "@testing-library/react-hooks": "3.3.0",
+    "@testing-library/user-event": "^12.1.0",
     "@types/jest": "26.0.4",
     "@types/prop-types": "15.7.3",
     "@types/react": "^16.9.33",
--- a/packages/dropdowns/src/elements/Multiselect/Multiselect.spec.tsx
+++ b/packages/dropdowns/src/elements/Multiselect/Multiselect.spec.tsx
@@ -6,6 +6,7 @@
  */
 
 import React from 'react';
+import userEvent from '@testing-library/user-event';
 import { render, fireEvent, renderRtl, act } from 'garden-test-utils';
 import { Dropdown, Multiselect, Field, Menu, Item, Label } from '../..';
 import { IDropdownProps } from '../Dropdown/Dropdown';
@@ -75,7 +76,7 @@ describe('Multiselect', () => {
       </ExampleWrapper>
     );
 
-    fireEvent.click(getByTestId('multiselect'));
+    userEvent.click(getByTestId('multiselect'));
 
     expect(document.activeElement!.nodeName).toBe('INPUT');
   });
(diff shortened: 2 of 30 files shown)


packages/dropdowns/src/elements/Multiselect/Multiselect.spec.tsx
```diff
@@ -530,19 +534,25 @@ describe('Multiselect', () => {
...
 
-      fireEvent.focus(input!);
+      expect(tags[0]).not.toHaveFocus();
+
       fireEvent.keyDown(input!, { key: 'Home', keyCode: KEY_CODES.HOME });
 
       expect(tags[0]).toHaveFocus();
     });
 
     it('does not focus first tag on home keydown with input value', () => {
```
hzhu, line 544: This test was a false positive because the first `tag` did not have focus throughout the test lifecycle. I've added assertions to ensure that the first tag doesn't have focus prior to a user pressing "home" with the input value. Then I added assertions to ensure that the first tag does have focus after user interaction.

packages/tooltips/src/elements/Tooltip.spec.tsx
```diff
@@ -35,47 +36,59 @@ describe('Tooltip', () => {
...
+      userEvent.tab();
       jest.runOnlyPendingTimers();
     });
 
+    expect(getByText('Trigger')).toHaveFocus();
     expect(getByTestId('tooltip')).toHaveStyleRule('direction', 'rtl');
   });
 
   describe('Focusable Tooltip Content', () => {
     it('remains visible if tooltip content is focused', () => {
```
hzhu, line 53: The original test did not have focusable content. The updated test more accurately asserts when a user focuses into the tooltip, and then focuses on a node that is part of the content.

hzhu, line 53: I've rewritten this test to better test real-world scenarios. It's uncommon for tooltip to contain text and be focusable. A more common scenario is that a tooltip has focusable content, say a button, as demonstrated [here](https://zendeskgarden.github.io/react-containers/storybook/?path=/story/tooltip-container--focusable-tooltip).

This test is modified so that it asserts against a more common scenario.

package.json
```diff
@@ -45,6 +45,7 @@
     "@testing-library/jest-dom": "5.11.0",
     "@testing-library/react": "10.4.6",
     "@testing-library/react-hooks": "3.3.0",
+    "@testing-library/user-event": "^12.1.0",
```
austin94, line 48: Should we pin this dependency?

packages/tables/src/elements/Row.spec.tsx
```diff
@@ -63,8 +64,8 @@ describe('Row', () => {
     );
     const row = getByTestId('row');
 
-    fireEvent.focus(row);
-    fireEvent.blur(row);
+    userEvent.click(row);
+    row.blur();
```
austin94, line 68: Is it ok to directly call the `blur` event?

hzhu: Yeah it's okay and encouraged:

https://testing-library.com/docs/guide-events#focusblur

https://github.com/testing-library/user-event/issues/350

austin94: TIL

austin94:
@jzempel that custom module resolver is only needed due to our [custom RTL rendering implementations](https://testing-library.com/docs/react-testing-library/setup#custom-render). If we aren't customizing anything I'd vote to leave it as is.

hzhu:
@jzempel my preference was to import `user-event` directly because it's a standalone package separate from `@testing-library/react`. garden-test-utils seems to be an export of `@testing-library/react`. But I could see it being exported from our testing utils too if garden-test-utils includes all things testing. I am fine with either decision 👍 .

PR #828, opened 2020-08-10
feat(modals): introduce new TooltipModal component

austin94:
## Description

This PR introduces the new `TooltipModal` component. Overall, the structure and semantics align with the existing `Modal` implementation, with an updated focus on relative positioning.

### Testing

We have begun seeing `ts-jest` warnings due to our Node version. I've updated the `tsconfig.test.json` file to resolve these warnings.

### Popper.js v2

This is the first package to consume Popper v2 and the updated `usePopper` hook. Overall this feels like a significant improvement to Popper v1. There are some differences in how modifiers are applied, including default offsets. We now include a default `8px` offset in all directions (this is customizable by the consumer).

The required testing syntax also differs from Popper v1, but seems more robust and requires no mocking.

### Re-Positioning and maintaining focus

A key requirement for this component is to be re-positioned without a full DOM re-render. This allows focus to be retained within the focus-jail and stay located on any navigation buttons.

## Checklist

      designer as a reviewer)


packages/modals/examples/tooltip-modal.md
```diff
@@ -0,0 +1,129 @@
...
+            </Button>
+          </TooltipModal.FooterItem>
+        </TooltipModal.Footer>
+        <TooltipModal.Close aria-label="Close" />
+      </TooltipModal>
+    </Grid>
+  );
+};
+
+<Example />;
```
hzhu, line 128: This is looking really good! I noticed that when the user presses `esc` in the example, the first trigger always gets focused. Do you think the example can have focus returned to the trigger element which caused the tooltip modal to open?

austin94: I’m not sure if this is possible with the logic provided by `useModal`. It’s the same modal throughout all the steps, so the focus is returned to that first triggered element. 

Would it make more sense to set the `restoreFocus` prop as `false` by default?

hzhu: Setting `restoreFocus` to `false` seems like a temporary solution and sounds good to me 👍 .

However, I think setting `restoreFocus` to `false` would not be ideal because the [spec](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7) says "When a dialog closes, focus returns to the element that invoked the dialog ". But this is an interesting situation because the "element that invoked the dialog" could be located in the most recently closed dialog.

When a user clicks "Step 1" and navigates to the "Step 3" modal tooltip by selecting the "Next" button, then presses `esc` while the "Step 3" modal tooltip is opened, where should the focus be after the modal tooltip is closed?

@m-lai would love to get your opinion on this too!

austin94: Another wrinkle is that these may be positioned against elements which are not focusable. Like sidebars and the like.

jzempel: Does it make sense to add an optional `restoreFocusRef` prop so that the consumer can determine how restored focus is handled?

austin94: Possibly, if we are able to disable the default functionality in `useModal`. Most of the "wizard" flows where these are repositioned are programmatically triggered as well. There may not be a sane way to find the correct focus-point without it being provided.

jzempel: Perhaps `useModal` needs to be enhanced to accept a `restoreFocusRef`?

austin94: I would go against that. For their most common usage, modals are triggered from a single item. This aligns with how `useModal` works now.

The cleanest solution to be feels like we should do the following:
- disable `restoreFocus` in the `useModal()` hook to remove the default functionality
- use `restoreFocus` to apply focus to the current `referenceElement` when the component is unmounted
  - Probably have to do this in a `useEffect()` cleanup

This would still allow consumers to disable the functionality if positioned against a non-focusable element.

jzempel: If you're able to achieve the desired behavior, then this sounds like a preferred approach.

austin94: The latest commit adds this functionality. Wasn't as bad as I was expecting.

packages/modals/src/elements/TooltipModal/TooltipModal.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <TooltipModal referenceElement={container}>
+          <TooltipModal.Close />
+        </TooltipModal>
+      );
+    });
+
+    expect(getByRole('button')).toHaveAttribute('aria-label', 'Close modal');
+  });
+
+  describe('onClose()', () => {
```
hzhu, line 132: I think asserting against function calls and using [`rerender`](https://testing-library.com/docs/react-testing-library/api#rerender) goes against Testing Library's [guiding principles](https://testing-library.com/docs/guiding-principles) as their guiding princples favor tests "that closely resemble how your web pages are used". 

From what I understand of the guiding principles, asserting against function calls and re-renders rank lower in terms of importance when compared to asserting against UI elements and simulating user events. 

What do you think about rewriting these tests to more align with the Testing Library guiding principles? Here's how I think a test that aligns with the guiding principles might look like. It is a test for asserting that the tooltip modal closes when clicking the close element.

```tsx
describe('Tooltip Modal', () => {
  const BasicTooltipModal = () => {
    const [isOpen, setIsOpen] = React.useState(false);
    const buttonRef = React.useRef<HTMLButtonElement>(null);

    return (
      <>
        <button ref={buttonRef} onClick={() => setIsOpen(!isOpen)}>
          Open Tooltip Modal
        </button>
        <TooltipModal
          hasArrow={true}
          placement="bottom-start"
          onClose={() => setIsOpen(false)}
          referenceElement={isOpen && buttonRef.current ? buttonRef.current : undefined}
        >
          <TooltipModal.Title>Header content</TooltipModal.Title>
          <TooltipModal.Body>Body content</TooltipModal.Body>
          <TooltipModal.Footer>Footer content</TooltipModal.Footer>
          <TooltipModal.Close aria-label="Close" />
        </TooltipModal>
      </>
    );
  };

  it('can close the tooltip modal by clicking the close element', async () => {
    const { getByText, queryByText, getByLabelText } = render(<BasicTooltipModal />);
    const button = getByText('Open Tooltip Modal');

    // Given: a tooltip modal is not rendered
    expect(queryByText('Header content')).toBeNull();
    expect(queryByText('Body content')).toBeNull();
    expect(queryByText('Footer content')).toBeNull();

    // When: the user clicks a button to open the tooltip modal
    await act(async () => {
      userEvent.click(button);
    });

    // Then: the tooltip modal is rendered in the document
    expect(getByText('Header content')).toBeInTheDocument();
    expect(getByText('Body content')).toBeInTheDocument();
    expect(getByText('Footer content')).toBeInTheDocument();

    // And when: the user closes the tooltip modal
    const close = getByLabelText('Close');
    userEvent.click(close);

    // Then: the tooltip modal is not rendered
    expect(queryByText('Header content')).toBeNull();
    expect(queryByText('Body content')).toBeNull();
    expect(queryByText('Footer content')).toBeNull();
  });
});
```

The test above could be written in different flavors. For example, with different matchers. Or can be abstracted differently. I think the main goal is that it aligns with the guiding principles by asserting against UI elements (not function calls) and simulating events (not manually calling `rerender`).

packages/modals/examples/tooltip-modal.md
hzhu: Wow, looks great!

packages/modals/src/styled/StyledTooltipModalFooter.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.tooltip_modal.footer';
+
+export const StyledTooltipModalFooter = styled.div.attrs({
```
jzempel, line 13: Looks like https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledFooter.ts could be factored to extend this styled component.

packages/modals/src/styled/StyledTooltipModalClose.ts
```diff
@@ -0,0 +1,90 @@
...
+      background-color: ${getColor(backgroundColor, 600, props.theme, 0.2)};
+      color: ${getColor(foregroundColor, 800, props.theme)};
+    }
+  `;
+};
+
+/**
+ * 1. Remove dotted outline from Firefox on focus.
+ */
+export const StyledTooltipModalClose = styled.button.attrs({
```
jzempel, line 46: Is this a near copy/paste of https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledClose.ts? How might this be refactored to take advantage of extended, reduced, and common code?

packages/modals/src/styled/StyledTooltipModalBody.ts
```diff
@@ -0,0 +1,33 @@
...
+import styled from 'styled-components';
+import {
+  getLineHeight,
+  retrieveComponentStyles,
+  DEFAULT_THEME
+} from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.tooltip_modal.body';
+
+export const StyledTooltipModalBody = styled.div.attrs({
```
jzempel, line 17: Looks like https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledBody.ts could be factored to extend this styled component.

packages/modals/src/styled/StyledTooltipModalBackdrop.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import styled from 'styled-components';
+import { DEFAULT_THEME, retrieveComponentStyles } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.tooltip_modal.backdrop';
+
+/**
+ * 1. Smooth iOS scrolling.
+ */
+export const StyledTooltipModalBackdrop = styled.div.attrs({
```
jzempel, line 16: Looks like https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledBackdrop.ts could be factored to extend this styled component.

packages/modals/src/styled/StyledTooltipModal.ts
```diff
@@ -0,0 +1,51 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  className: 'is-animated'
+})<IStyledTooltipModalProps>`
+  /* stylelint-disable-next-line declaration-no-important */
+  position: static !important; /* [1] */
+
+  padding: ${props => props.theme.space.base * 5}px;
+  width: 400px;
+  min-height: 150px; /* TODO themify */
```
jzempel, line 36: Leftover TODO?

packages/modals/src/styled/StyledTooltipWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+  zIndex?: number;
+  placement?: Placement;
+}>`
+  ${props =>
+    menuStyles(getMenuPosition(props.placement), {
+      theme: props.theme,
+      hidden: false,
+      margin: `${props.theme.space.base}px`,
+      zIndex: props.zIndex,
+      animationModifier: '.is-animated'
```
jzempel, line 29: I think every other Garden usage sets the `animationModifier` based on a `isAnimated` prop surfaced at the element-level API. Would be good to be consistent here.

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+   * [Popper.js v2 modifiers](https://popper.js.org/docs/v2/modifiers/)
+   * to customize positioning logic.
+   */
+  popperModifiers?: Array<Partial<Modifier<any, any>>>;
+  /**
+   * These placements differ from the default naming of Popper placements to help
+   * assist with RTL layouts.
+   **/
+  placement?: GARDEN_PLACEMENT;
+  hasArrow?: boolean;
```
jzempel, line 65: As mentioned, this component should expose `isAnimated` for Garden API consistency.

packages/modals/examples/tooltip-modal.md
```diff
@@ -0,0 +1,129 @@
...
+    <Grid>
+      <Row className="u-mb-md">
+        <Col>
+          <Well isRecessed>
+            <Field>
+              <Toggle checked={hasArrow} onChange={e => setHasArrow(e.target.checked)}>
+                <Label>
+                  <Code>hasArrow</Code>
+                </Label>
+              </Toggle>
```
jzempel, line 58: Include `isAnimated` toggle once added to the API.

packages/modals/examples/tooltip-modal.md
```diff
@@ -0,0 +1,129 @@
...
+          Gumbo beet greens corn soko endive gumbo gourd. Parsley shallot courgette tatsoi pea
+          sprouts fava bean collard greens dandelion okra wakame tomato. Dandelion cucumber earthnut
+          pea peanut soko zucchini.
+        </TooltipModal.Body>
+        <TooltipModal.Footer>
+          <StyledProgress>
+            <span>{step} of 4</span>
+          </StyledProgress>
+          <TooltipModal.FooterItem>
+            <Button size="small" disabled={step === 1} onClick={() => setStep(step - 1)}>
```
jzempel, line 102: Isn't it preferable to not show the button rather than setting `disabled` on step 1?

packages/modals/examples/tooltip-modal.md
```diff
@@ -0,0 +1,129 @@
...
+          </Button>
+        </Col>
+      </Row>
+      <TooltipModal
+        referenceElement={referenceElement}
+        placement={placement}
+        hasArrow={hasArrow}
+        onClose={() => setStep(undefined)}
+      >
+        <TooltipModal.Title>Header content</TooltipModal.Title>
```
jzempel, line 91: Would be nice if the header title corresponded with the step in question, i.e. **Title for step 1** or something similar.

packages/modals/src/styled/StyledTooltipModal.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Accepts all `<div>` props
+ */
+export const StyledTooltipModal = styled.div.attrs<IStyledTooltipModalProps>({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  className: 'is-animated'
+})<IStyledTooltipModalProps>`
+  /* stylelint-disable-next-line declaration-no-important */
+  position: static !important; /* [1] */
```
hzhu, line 32: Should `/* [1] */` refer to a comment?

packages/modals/src/styled/StyledTooltipModal.ts
hzhu, line 36: Was it intended to keep this TODO around? Should it be documented in a task somewhere?

m-lai:
@austin94 Looking good so far on the visual front. Do we have to retain the disabled button in Step 1? Feels a bit odd.

Also, what would happen if the tooltip modal is implemented near the edge of a screen? Presumably it would adjust its position instead of cutting off?

austin94:
> Do we have to retain the disabled button in Step 1? Feels a bit odd

This was modeled after some in-product usages. I'll remove it for the first step.

> Also, what would happen if the tooltip modal is implemented near the edge of a screen? 

Depending on the position value provided it would automatically change position or adjust itself against the body size to avoid a cutoff.

packages/modals/src/utils/useTooltipModalContext.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const Example = () => <ModalContextConsumer />;
+
+    expect(() => {
+      render(<Example />);
+    }).toThrow();
+
+    console.error = originalError;
+  });
+
+  it('does not throw if called within Modal component', () => {
```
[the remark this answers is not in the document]
austin94: Updated.

packages/modals/src/utils/useTooltipModalContext.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render } from 'garden-test-utils';
+import { TooltipModal } from '../elements/TooltipModal/TooltipModal';
+import { useTooltipModalContext } from './useTooltipModalContext';
+
+describe('useTooltipModalContext', () => {
+  const ModalContextConsumer = () => {
```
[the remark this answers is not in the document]
austin94: Update.

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+  zIndex?: number;
```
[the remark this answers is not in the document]
austin94: These props aren't documented in `Tooltip` or `Menu` at this time. I'll include some rough copy.

packages/modals/src/styled/StyledTooltipWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+    })};
+
+  margin: 0; /* [1] */
```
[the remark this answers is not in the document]
austin94: I've shuffled this around to include `0` directly.

packages/modals/src/styled/StyledTooltipModalTitle.ts
```diff
@@ -0,0 +1,35 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  margin: 0;
+  /* stylelint-disable-next-line property-no-unknown */
+  padding-${props => (props.theme.rtl ? 'left' : 'right')}: ${props =>
+  props.theme.space.base * 8}px;
+  line-height: ${props => getLineHeight(props.theme.lineHeights.md, props.theme.fontSizes.md)};
+  color: ${props => props.theme.colors.foreground};
+  font-size: ${props => props.theme.fontSizes.md};
```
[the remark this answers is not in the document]
austin94: I mimicked the styling provided by `StyledHeader` (Modal.Header) which didn't include this style utility. I've added it here though.

packages/modals/src/styled/StyledTooltipModalFooterItem.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.tooltip_modal.footer_item';
+
+export const StyledTooltipModalFooterItem = styled.span.attrs({
```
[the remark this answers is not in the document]
austin94: My original thought is that the `TooltipModal` isn't a variant of `Modal`. With them being exclusive I didn't want to tie styling between the two.

I've updated this based on your suggestion.

packages/modals/src/styled/StyledTooltipModalTitle.ts
```diff
@@ -0,0 +1,35 @@
...
+import styled from 'styled-components';
+import {
+  getLineHeight,
+  retrieveComponentStyles,
+  DEFAULT_THEME
+} from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'modals.tooltip_modal.title';
+
+export const StyledTooltipModalTitle = styled.div.attrs({
```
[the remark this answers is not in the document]
austin94: I'm not sure these align as well as `FooterItem` does. `TooltipModal` isn't a variant of `Modal`. We would have to override several styles to make it fit. i.e. as `Modal` evolves we don't want `TooltipModal` to change with it

packages/modals/src/styled/StyledTooltipModalFooter.ts
austin94: Same comment as https://github.com/zendeskgarden/react-components/pull/828#discussion_r469426145

packages/modals/src/styled/StyledTooltipModalBody.ts
austin94: Same comment as https://github.com/zendeskgarden/react-components/pull/828#discussion_r469426145

packages/modals/src/styled/StyledTooltipModalBackdrop.ts
austin94: Same comment as https://github.com/zendeskgarden/react-components/pull/828#discussion_r469426145

packages/modals/src/styled/StyledTooltipModal.ts
austin94: Was waiting on some design input here. Removed.

packages/modals/examples/tooltip-modal.md
austin94: Yes. This was based off of an older Abstract mock.

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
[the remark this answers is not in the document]
jzempel: If you want to attempt to comply with forthcoming doc guidelines, lead off all prop descriptions with a 3rd person present verb. aka "Applies", "Defines", "Determines". See updates made to https://github.com/zendeskgarden/react-components/blob/main/packages/dropdowns/src/elements/Multiselect/Multiselect.tsx for examples. The idea is that the prop name itself becomes the implicit noun of the description.

packages/modals/src/elements/TooltipModal/TooltipModal.spec.tsx
austin94: The technical reason for this testing structure is to allow `forwardRef` and `usePopper` to accomplish state updates properly. The implementation above produces failures and warnings similar to:

> Warning: An update to ForwardRef inside a test was not wrapped in act(...)

https://github.com/popperjs/react-popper/issues/368 includes some additional info

austin94:
@jzempel @hzhu I think I've addressed a good chunk of the comments.

Regarding the double-scrollbars, I've included logic similar to `Modal` by adding `overflow: hidden` styling to the `body` element whenever a `TooltipModal` is displayed.

packages/modals/src/styled/StyledTooltipModalTitle.ts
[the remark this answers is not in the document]
jzempel: For all of these comments, I was suggesting that the `TooltipModal` becomes the new baseline and `Modal` extends from that. I might've missed something in the comparison, but it seemed that all properties represented in `TooltipModal` were also represented (and nearly 100% the same) in `Modal`.

packages/modals/src/elements/TooltipModal/TooltipModal.spec.tsx
hzhu: I ran the test implementation above and it didn't produce failures or warnings. The test passes and makes the correct assertions before and after each user event simulation ✅.

However, I noticed that there is a warning (not error) regarding an update to `ForwardRef` when the tooltip modal implementation broke. But the warning is caused by a bug in `react-popper` and is only visible when the Tooltip Modal implementation is broken. I don't think that warrants the test to be structured differently than what the Testing Library guiding principles suggest. Also, as long as the test is passing, the warning won't be encountered.

Alternatively, I think the test can be kept as-is with a code comment pointing to popperjs/react-popper#368 so that the tests in this file can be rewritten when `react-popper` fixes the associated bug.

austin94: Leaving a note for posterity... from our offline chat I needed to change the order of some `await/async` commands for this to pass correctly. Tests are now updated.

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+      </TooltipModalContext.Provider>
+    ) : null;
+  }
+) as IStaticTooltipModalExport<HTMLDivElement, ITooltipModalProps>;
+
+TooltipModal.Title = Title;
+TooltipModal.Body = Body;
+TooltipModal.Close = Close;
+TooltipModal.Footer = StyledTooltipModalFooter;
+TooltipModal.FooterItem = StyledTooltipModalFooterItem;
```
jzempel, line 195: Looks like some of these are still showing up in Styleguidist. Should they be removed?

m-lai:
Thanks @austin94 ! One other question - I noticed the padding between the start of the arrows and the edge of the tooltip modal is smaller than what we have in the Sketch files - was this intentional? It looks kind of tight IMO compared to Sketch but I'd like to update the Sketch file to be consistent with code either way

https://share.getcloudapp.com/E0unzmA0

austin94:
@m-lai I based the original sizing off of the large `Tooltip`, but it looks like I should have used the extra-large. I've updated the arrow sizing and margin to align with the sizes shown in Abstract.

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
austin94: I've updated the glob to remove these.

m-lai:
> @m-lai I based the original sizing off of the large `Tooltip`, but it looks like I should have used the extra-large. I've updated the arrow sizing and margin to align with the sizes shown in Abstract.

Cheers!!

packages/modals/src/styled/StyledTooltipModal.ts
```diff
@@ -0,0 +1,54 @@
...
+export const StyledTooltipModal = styled.div.attrs<IStyledTooltipModalProps>(props => ({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  className: props.isAnimated && 'is-animated'
+}))<IStyledTooltipModalProps>`
+  /* stylelint-disable-next-line declaration-no-important */
+  position: static !important; /* [1] */
+
+  padding: ${props => props.theme.space.base * 5}px;
+  width: 400px;
```
hzhu, line 39: Hard coding a width here forces consumers to override CSS anything `400px` doesn't work for their context. Currently running into this situation using the tooltip modal in the color picker and color swatch dialogs.

jzempel: I believe the solution is to `!important` as we do with https://github.com/zendeskgarden/react-components/blob/2ad0042eb3bea0ed6a93be02b3b2e2333aff183d/packages/colorpickers/src/styled/ColorpickerDialog/StyledTooltipModal.ts#L23.

PR #829, opened 2020-08-10
feat(forms, dropdowns): use FauxInput as the fundamental styling basis for Select, Autocomplete, and Multiselect

jzempel:
## Description

This is precursor work to smooth the path for a follow-on native `<select>` component. Notable features/fixes for this PR include:

- the `FauxInput` is enhanced to take either a `FauxInput.StartIcon` or `FauxInput.EndIcon` – eradicating field-level styled-components in `react-dropdowns`
  - fixes chevron disabled styling on hover
  - fixes chevron pixel shift on multiselect
- added `wrapperRef` to `FauxInput`, targeting the outer `<div>`
- fixed `MediaInput` to receive expected styling on label hover
- `Multiselect` updated to receive an optional `start` icon (parallel with Select and Autocomplete)
- enhanced the `Multiselect` "more" anchor with padding in order to increase the tap target and prevent phantom clicks
- the `Tag` items within a `Multiselect` have the correct `size` applied whether the multiselect is default or `isCompact`
- prop documentation updated in `Select`, `Autocomplete` and `Multiselect` according to documentation guidelines

## Checklist

      designer as a reviewer)


packages/dropdowns/src/styled/multiselect/StyledMultiselectInput.ts
```diff
@@ -0,0 +1,51 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { css, ThemeProps, DefaultTheme } from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledInput } from '../select/StyledInput';
+
+const COMPONENT_ID = 'dropdowns.multiselect_input';
+
+interface IStyledMultiselectInputProps extends ThemeProps<DefaultTheme> {
```
austin94, line 14: This should pass type-checking, but it may be cleaner to follow the syntax we use in other components.

`ThemeProps<DefaultTheme>` is added through the `styled-components` types. Rather than add them to the styled element props it may be more accurate to include it as `& ThemeProps<DefaultTheme>` to your styling helpers.

packages/forms/src/elements/FauxInput.tsx
```diff
@@ -67,3 +69,24 @@ FauxInput.propTypes = {
...
+  children: any;
+}
+
+const StartIcon = (props: IIconProps) => <StyledTextMediaFigure position="start" {...props} />;
+const EndIcon = (props: IIconProps) => <StyledTextMediaFigure position="end" {...props} />;
+
+(FauxInput as any).StartIcon = StartIcon;
+(FauxInput as any).EndIcon = EndIcon;
+
+export default FauxInput as React.FunctionComponent<
```
austin94, line 87: Rather than using a `default` export. Could we apply this type cast to something like `export const FauxInput`?

austin94: There is also a standardized way of applying state component properties shown in `Accordion`. This may help with the typing situation.

packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx
```diff
@@ -58,18 +66,24 @@ const Autocomplete = React.forwardRef<HTMLDivElement, IAutocompleteProps>(
           (e.nativeEvent as any).preventDownshiftDefault = true;
         }
       },
+      onMouseEnter: composeEventHandlers(props.onMouseEnter, () => setIsHovered(true)),
+      onMouseLeave: composeEventHandlers(props.onMouseLeave, () => setIsHovered(false)),
       ...props
     } as any);
 
+    const isContainerHovered = isLabelHovered && !isOpen;
+    const isContainerFocused = isOpen ? true : isFocused;
```
hzhu, line 75: Would `isOpen || isFocused` be a concise substitute for the ternary expression?

packages/dropdowns/src/elements/Multiselect/Multiselect.spec.tsx
```diff
@@ -162,6 +162,31 @@ describe('Multiselect', () => {
...
+    const { getByTestId } = render(
+      <ExampleWrapper>
+        <Label data-test-id="label">Label</Label>
+        <Multiselect
+          data-test-id="multiselect"
+          start={<svg data-test-id="icon" />}
+          renderItem={({ value, removeValue }) => (
+            <div data-test-id="tag">
+              {value}
+              <button data-test-id="remove" onClick={() => removeValue()} tabIndex={-1}>
```
hzhu, line 175: I noticed some of the `data-test` attributes in this render aren't used such as `data-test-id="label"`, `data-test-id="tag"`, `data-test-id="remove"`, and `data-test-id="multiselect"`. Could these attributes be safely removed?

packages/dropdowns/src/elements/Multiselect/Multiselect.tsx
```diff
@@ -269,16 +304,20 @@ const Multiselect = React.forwardRef<HTMLDivElement, IMultiselectProps & ThemePr
       props.isCompact
     ]);
 
+    const isContainerHovered = isLabelHovered && !isOpen;
+    const isContainerFocused = isOpen ? true : isFocused;
```
hzhu, line 308: Not a request for change, but if the earlier concise substitute is safe (and preferred), then here's another location where  `isOpen || isFocused` could be used.

packages/dropdowns/src/elements/Select/Select.tsx
```diff
@@ -185,17 +194,22 @@ export const Select = React.forwardRef<HTMLDivElement, ISelectProps>(
...
 
+    const isContainerHovered = isLabelHovered && !isOpen;
+
     return (
       <Reference>
         {({ ref: popperReference }) => (
-          <SelectWrapper
-            isHovered={isLabelHovered && !isOpen}
+          <StyledFauxInput
+            data-test-is-open={isOpen}
```
hzhu, line 208: I couldn't find where this `data-test` attribute was used. Are there plans to use this `data-test` attribute in subsequent select work?

jzempel: It's used by existing tests within `Select.spec.tsx`; for example:
https://github.com/zendeskgarden/react-components/blob/3fa59f9a29842b65e317d69fc6fa217bb00ac3e3/packages/dropdowns/src/elements/Select/Select.spec.tsx#L97

hzhu: Oh whoops, I totally missed this. Thanks for pointing where the usage is.

PR #830, opened 2020-08-10
fix(datepickers): format day with NumberFormat utility

austin94:
## Description

The `Datepicker` components are currently formatting individual days using the `DateTimeFormat` utility. This is creating invalid localizations for some locales.

_Invalid Japanese Days_
image

The value `DateTimeFormat` makes sense in the context of a date string, but not as standalone numbers. To format these values correctly I've modified our day formatting logic to use the `Intl.NumberFormat` utility. This uses the same BCP tag as `DateTimeFormat` and provides a valid localization for our main locales.

## Detail

_English_
image

_Japanese_
image

_Arabic_
image

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/datepickers/.size-snapshot.json
+++ b/packages/datepickers/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 131420,
-    "minified": 74065,
-    "gzipped": 16506
+    "bundled": 131231,
+    "minified": 73980,
+    "gzipped": 16493
   },
   "index.esm.js": {
-    "bundled": 130205,
-    "minified": 72901,
-    "gzipped": 16448,
+    "bundled": 130022,
+    "minified": 72822,
+    "gzipped": 16434,
     "treeshaked": {
       "rollup": {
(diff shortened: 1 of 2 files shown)


olenavelykanova:
@austin94 thanks for this, JA looks great, I am checking AR as it currently looks differently in Google calendar [https://share.getcloudapp.com/o0u8D4jD](https://share.getcloudapp.com/o0u8D4jD)
Will get back to you on this one today.

austin94:
> was this a bug

I think so. The core issue is that `DateTimeFormat` is used to format date strings, not standalone days. If we were trying to format a single day in a sentence then our format would be correct. For a calendar we are formatting a standalone number so the `NumberFormat` utility is better suited.

> should there be a new test

I'm not sure if a test would have helped here. The value is determined by the locale so we just trust the value provided by the browser. This was more of a technical design failure than a bug.

austin94:
After discussing with @olenavelykanova I've updated the `Datepicker` to show Gregorian day labels for _all_ locales. This is the default provided by `date-fns` so I've using their `getDay()` utility directly.

PR #835, opened 2020-08-12
fix(modals): increase padding for overflown header text

austin94:
## Description

For localized Modals we've seen some text overflow issues where the `Close` button overlaps `Header` content.

This PR updates the padding to give it some more space.

### Before

image

### After

image

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/styled/StyledHeader.ts
+++ b/packages/modals/src/styled/StyledHeader.ts
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled from 'styled-components';
+import styled, { css, ThemeProps, DefaultTheme } from 'styled-components';
 import {
   getLineHeight,
   getColor,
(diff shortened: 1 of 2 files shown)


packages/modals/src/styled/StyledHeader.ts
```diff
@@ -22,21 +22,36 @@ export interface IStyledHeaderProps {
   isDanger?: boolean;
 }
 
+const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
+  const verticalPadding = `${props.theme.space.base * 5}px`;
+  const startPadding = `${props.theme.space.base * 10}px`;
+  const endPadding = `${props.theme.space.base * 15}px`;
```
jzempel, line 28: The end padding should accommodate the entirety of the close button, which is 40px wide. Edge padding (40) + button padding (40) = 80px.

austin94: > Is it reasonable to think that every single modal will contain a Close button? If not, this applies a rather severe padding imbalance to the header that can't be rationalized in the absence of a Close.

Oh snap. Didn't think of that. With this being positioned relative to the modal, is there even a way to do this accurately (without some JS scenanigans)?

jzempel: Feed it through context? Alternatively, it seems like the description use-case should be applying custom styling.

m-lai: Dunno if it helps but from a UX perspective, I don't think we'd recommend implementing modals without the Close button. Ideally there would be a non-janky fix for the severe padding if that scenario comes up but will leave that up to y'all

austin94:
Closing for now

PR #837, opened 2020-08-17
fix(buttons): apply min-width to icon buttons

rossmoody:
## Description

Icon buttons in `flex` environments are squishing to accommodate for other elements inline. This PR adds a min-width designation to make them retain their size. 

### Current

Screen Shot 2020-08-17 at 9 33 36 AM

### Proposed

Screen Shot 2020-08-17 at 9 33 40 AM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/buttons/src/styled/StyledIconButton.ts
+++ b/packages/buttons/src/styled/StyledIconButton.ts
@@ -38,6 +38,7 @@ const iconButtonStyles = (props: IStyledButtonProps & ThemeProps<DefaultTheme>) 
     border: ${props.isBasic && 'none'};
     padding: 0;
     width: ${getHeight(props)};
+    min-width: ${getHeight(props)};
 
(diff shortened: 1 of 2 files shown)


packages/buttons/src/styled/StyledIconButton.ts
```diff
@@ -38,6 +38,7 @@ const iconButtonStyles = (props: IStyledButtonProps & ThemeProps<DefaultTheme>)
     border: ${props.isBasic && 'none'};
     padding: 0;
     width: ${getHeight(props)};
+    min-width: ${getHeight(props)};
```
jzempel, line 41: [nit] it might be slightly more efficient/maintainable to pull the two `getHeight` function calls into a `const width = getHeight(props);` variable substitution declared above the `css` block.

PR #838, opened 2020-08-17
fix(accordions): remove default button behavior

hzhu:
## Description

When an accordion is inside a `<form>` the accordion buttons cause a form submission due to the default button behavior. This PR removes the default button behavior on accordions by adding a `type=button` to the accordion buttons.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/accordions/src/elements/accordion/components/Label.spec.tsx
+++ b/packages/accordions/src/elements/accordion/components/Label.spec.tsx
@@ -25,4 +25,18 @@ describe('Label', () => {
 
     expect(getByRole('button')).toBe(ref.current);
   });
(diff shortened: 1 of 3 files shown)


packages/accordions/src/elements/accordion/components/Label.spec.tsx
```diff
@@ -25,4 +25,18 @@ describe('Label', () => {
 
     expect(getByRole('button')).toBe(ref.current);
   });
+
+  it('uses a button with no defaul behavior', () => {
```
austin94, line 29: `defaul` -> `default`

PR #842, opened 2020-08-18
fix(forms): workaround shift+click checkbox labels on Firefox

hzhu:
## Description

There is a bug in Firefox which prevents checkboxes from being selected when a user performs a `shift + click` on the checkbox label. 

## Detail

A small Firefox-specific snippet of code has been added to manually call events when a checkbox label is clicked which allows the checkbox to become checked when a user performs `shift + click` on a checkbox label.

## Screenshots

❌ **Before (`react-forms`)**: `shift + click` on the label does not select the checkbox in Firefox

✅ **After (`react-forms`)**: `shift + click` on the label selects the checkbox in Firefox

---

❌ **Before (`react-tables`)**: `shift + click` does not select multiple rows in Firefox

✅ **After (`react-tabels`)**: `shift + click` selects multiple rows in Firefox

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/forms/src/elements/common/Label.tsx
+++ b/packages/forms/src/elements/common/Label.tsx
@@ -71,6 +71,32 @@ export const Label = React.forwardRef<HTMLLabelElement, ILabelProps>((props, ref
       </StyledRadioLabel>
     );
   } else if (type === 'checkbox') {
+    /**
+     * `onLabelSelect` is a workaround for checkbox label `shift + click` bug in Firefox
+     * See: https://bugzilla.mozilla.org/show_bug.cgi?id=559506
+     */
+    const onLabelSelect = (e: React.KeyboardEvent) => {
+      const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+
+      if (fieldContext && isFirefox) {
+        const inputProps = fieldContext.getInputProps();
+        const input = document.getElementById(inputProps.id as string) as HTMLInputElement;
+
+        if (input && input.type === 'checkbox') {
+          if (e.shiftKey) {
+            input.click();
+            input.checked = true;
+          }
+          input.focus();
+        }
+      }
+    };
+
+    combinedProps = {
+      ...combinedProps,
+      onClick: composeEventHandlers(combinedProps.onClick, onLabelSelect)
+    };
+
     return (
       <StyledCheckLabel ref={ref} {...(combinedProps as any)}>
         <StyledCheckSvg />
--- a/packages/forms/src/elements/common/Label.tsx
+++ b/packages/forms/src/elements/common/Label.tsx
@@ -75,12 +75,15 @@ export const Label = React.forwardRef<HTMLLabelElement, ILabelProps>((props, ref
      * `onLabelSelect` is a workaround for checkbox label `shift + click` bug in Firefox
      * See: https://bugzilla.mozilla.org/show_bug.cgi?id=559506
      */
-    const onLabelSelect = (e: React.KeyboardEvent) => {
+    const onLabelSelect = (e: React.KeyboardEvent<HTMLInputElement>) => {
       const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
 
-      if (fieldContext && isFirefox) {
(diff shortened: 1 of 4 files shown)


packages/forms/src/elements/common/Label.tsx
```diff
@@ -71,6 +71,32 @@ export const Label = React.forwardRef<HTMLLabelElement, ILabelProps>((props, ref
...
+    /**
+     * `onLabelSelect` is a workaround for checkbox label `shift + click` bug in Firefox
+     * See: https://bugzilla.mozilla.org/show_bug.cgi?id=559506
+     */
+    const onLabelSelect = (e: React.KeyboardEvent) => {
+      const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+
+      if (fieldContext && isFirefox) {
+        const inputProps = fieldContext.getInputProps();
+        const input = document.getElementById(inputProps.id as string) as HTMLInputElement;
```
austin94, line 83: Rather than call `getInputProps()` for the ID, is it possible to retrieve the ID from the `htmlFor` attribute on the current label? Should be available under `e.target`

packages/forms/src/elements/common/Label.tsx
```diff
@@ -71,6 +71,32 @@ export const Label = React.forwardRef<HTMLLabelElement, ILabelProps>((props, ref
...
+
+        if (input && input.type === 'checkbox') {
+          if (e.shiftKey) {
+            input.click();
+            input.checked = true;
```
austin94, line 88: How does this work when an element is controlled?

jzempel: It _seems_ to hold up: https://5f3c3e29d6b3cf01999d887e--zendeskgarden-react-components.netlify.app/forms/#advanced

jzempel: Although, the prop for `indeterminate` doesn't seem to be getting respected here.

hzhu: I'll need to consider the implications of a controlled checkbox. I believe there may need to be some changes. Please hold 👍  .

hzhu: Currently, controlled usages do not work in Firefox. The controlled example should have the ability to have the checkbox set to indeterminate, _but this does not work_. Thanks for catching this @austin94 🙏 . 

My new (slightly modified) proposal: only address the Firefox-specific bug in _uncontrolled_ checkboxes. I could add an additional boolean check (e.g `isControlled`) and only apply the fix for _uncontrolled_ checkboxes. The big win here is that the selectable `Table` components will receive the fix and controlled checkboxes work as they did before.

New commit coming shortly.

hzhu: **Edit**: I went back and tried various controlled examples (custom set checked buttons, etc). It looks like when a checkbox is controlled, the controlled state will be applied to the checkbox. Here's how I verified this:

I added the following to the controlled example:

```jsx
<button
  onClick={() => {
    setState({
      checkboxChecked: true,
      checkboxIndeterminate: false
    });
  }}
>
  Set checked
</button>
<button
  onClick={() => {
    setState({
      checkboxIndeterminate: true
    });
  }}
>
  Set indeterminate
</button>
```

Then I performed `shift + click` the checkbox label to check and uncheck the checkbox. Then to verify that controlled usages worked, I clicked the "Set checked" and "Set indeterminate" buttons.



@austin94, do you see any issues regarding controlled checkboxes?

hzhu: > Although, the prop for indeterminate doesn't seem to be getting respected here.

I noticed this too during the controlled checkbox usage investigation. It seems like indeterminate wasn't respected prior to this PR.

packages/forms/src/elements/common/Label.tsx
hzhu: I've updated this to retrieve the ID from the `htmlFor` attribute. This is great, thanks for the suggestion. 

Along with this change, I had to set `pointer-events: none;` to the label's children SVG elements. Otherwise, `onLabelSelect` would be fired for SVG elements and not the label element.

PR #846, opened 2020-08-21
fix(accordions): workaround for Chromium autoscroll

hzhu:
## Description

There is an auto-scroll issue with the `Accordion` component in Chromium browsers. The issue seems to stem from a browser quirk.

## Detail

The combination of focusable element (accordion header button), and collapsing height (accordion panel) causes Chromium to auto-scroll the window. Chromium auto-scrolls the equivalent amount of pixels of the collapsed element. In this particular case, that would be the height of the collapsed accordion panel.

This auto-scroll issue occurs given these factors: 

* User is on a Chrome or Edge (Chromium)
* There is enough page content to make the page scrollable
* An Accordion panel _above_ an active accordion header is collapsed
  - e.g. `expandable` prop is set to `false` and `collapsible` prop is set to `true`

## Solution

To work around the browser quirk, I've added a snippet of code that temporarily disables scrolling when an accordion header button is selected. Scrolling is restored when the accordion panel completes its height transition.

## Screenshots

❌ **Before:** Chrome auto-scrolls each time a height transition occurs as a panel is collapsing

---

✅ **After:** Chrome does not auto-scroll for the duration of the transition

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/accordions/src/utils/scroll.ts
+++ b/packages/accordions/src/utils/scroll.ts
@@ -1,0 +1,26 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export function disableScroll() {
+  document.body.style.overflow = 'hidden';
+
+  const x = window.scrollX;
+  const y = window.scrollY;
+
+  window.onscroll = () => {
+    window.scrollTo(x, y);
+  };
+}
+
+export const enableScroll = () => {
+  setTimeout(() => {
+    // Set overflow on the next tick of event loop so Chromium does not add scrollbars.
+    document.body.style.overflow = '';
+  }, 0);
+
+  window.onscroll = null;
+};
--- a/packages/accordions/.size-snapshot.json
+++ b/packages/accordions/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 41402,
-    "minified": 27438,
-    "gzipped": 6568
+    "bundled": 42229,
+    "minified": 27959,
+    "gzipped": 6707
   },
   "index.esm.js": {
-    "bundled": 40094,
-    "minified": 26181,
-    "gzipped": 6473,
+    "bundled": 40921,
+    "minified": 26702,
+    "gzipped": 6610,
     "treeshaked": {
       "rollup": {
-        "code": 20715,
+        "code": 21158,
         "import_statements": 535
       },
       "webpack": {
-        "code": 23568
+        "code": 24015
       }
     }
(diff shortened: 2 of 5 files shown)


packages/accordions/src/utils/scroll.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  window.onscroll = () => {
+    window.scrollTo(x, y);
+  };
+}
+
+export const enableScroll = () => {
+  setTimeout(() => {
+    // Set overflow on the next tick of event loop so Chromium does not add scrollbars.
+    document.body.style.overflow = '';
```
jzempel, line 22: I'll leave it to further review for comments on the overall approach, but don't we commit to pulling `document` off of the current theme? https://github.com/zendeskgarden/react-components/blob/main/packages/theming/src/utils/getDocument.ts

austin94: We provide a `useDocument` hook that can be used with the `ThemeContext` provided by styled-components.

hzhu:
@austin94 what are your thoughts on the overall approach of this solution? Does it seem reasonable to tap into the global scroll event via `onscroll` to address a Chromium quirk?

P.S. The auto-scroll issue also affects the accordion hook Storybook examples in `react-containers`. I've documented that [here](https://github.com/zendeskgarden/react-containers/pull/226).

jzempel:
I'll be honest, this feels a bit heavy-handed. Although, I truly appreciate the instinct to get the component behavior just right. I'm torn, but yield to the team to assert overall confidence in the solution.

austin94:
I agree with @jzempel. This feels more like a symptom than a solution. Might want to look at the expand/collapse logic instead of overriding this scroll mechanism.

jzempel:
@hzhu is the Accordion `website` [PR](https://github.com/zendeskgarden/website/pull/163) telling a different story?

It seems from https://5f46eaacbd7b19a0cee0da03--zendeskgarden.netlify.app/components/accordion that when a section's line count changes, the size of the surrounding container (that is, the `CodeExample` component) is modified, which in turn modifies the length of the page.

If that's the case, then it would be up to the consumer to set a fixed height on the accordion container to prevent movement (if that is, indeed, desirable).

hzhu:
@jzempel I don't think the Accordion website PR is telling a different story, but we can sync up to see if that is the case. Would it be difficult for consumers to set a fixed height because the inner accordion panel height can have dynamic and/or localized text?

@austin94 I have taken a look at the expand/collapse logic and does not seem to be related to that. Check out https://github.com/zendeskgarden/react-containers/pull/226 which shows the issue `react-components` agnostic.

I'll be closing this PR shortly, but will keep track of the discussions within it.

hzhu:
Closing this PR for now. Will investigate more and open the PR again when an alternative solution is found.

PR #850, opened 2020-08-31
feat(forms): introduce FileUpload component

austin94:
## Description

This PR introduces a new `FileUpload` component. 

File uploads are dependent on business logic and backend support for common features like thumbnails, upload progress, and file restrictions. [react-dropzone](https://react-dropzone.js.org/) is a popular library that helps abstract some of this complexity.

To support as many use-cases as possible this component is a simple, styled `<div>` that is able to be used with `react-dropzone` or any other file upload abstraction.

## Detail

For a11y considerations this is built within the `react-forms` package and allows it to be used along our existing `Field/Label/Hint/Message/Input` components.

The example includes an implementation of a file list. It includes some mocked random file upload speeds to demonstrate a common upload pattern.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/examples/file-upload.md
+++ b/packages/forms/examples/file-upload.md
@@ -6,6 +6,8 @@ accepts any file type or size.
 
 ```jsx
 const { useDropzone } = require('react-dropzone');
+const { Well } = require('@zendeskgarden/react-notifications/src');
+const { Code } = require('@zendeskgarden/react-typography/src');
 const { getColor } = require('@zendeskgarden/react-theming/src');
 const { IconButton } = require('@zendeskgarden/react-buttons/src');
 const { Progress } = require('@zendeskgarden/react-loaders/src');
@@ -93,6 +95,7 @@ const File = React.memo(({ name, onRemove }) => {
 });
 
 const Example = () => {
+  const [isCompact, setIsCompact] = React.useState(false);
   const [files, setFiles] = React.useState(['squash.jpg', 'soybean.pdf']);
 
   const onDrop = React.useCallback(
@@ -117,11 +120,24 @@ const Example = () => {
 
   return (
     <>
+      <Well isRecessed className="u-mb-sm">
+        <Field>
+          <Toggle checked={isCompact} onChange={event => setIsCompact(event.target.checked)}>
+            <Label>
+              <Code>isCompact</Code>
+            </Label>
+          </Toggle>
+        </Field>
+      </Well>
       <Field>
         <Label>File upload</Label>
         <Hint>Works with react-dropzone</Hint>
-        <FileUpload {...getRootProps()} isDragging={isDragActive}>
-          {isDragActive ? <p>Drop files here</p> : <p>Drag files here or click to upload</p>}
+        <FileUpload {...getRootProps()} isDragging={isDragActive} isCompact={isCompact}>
+          {isDragActive ? (
+            <span>Drop files here</span>
+          ) : (
+            <span>Drag files here or click to upload</span>
+          )}
           <Input {...getInputProps()} />
         </FileUpload>
       </Field>
--- a/packages/forms/examples/file-upload.md
+++ b/packages/forms/examples/file-upload.md
@@ -1,0 +1,145 @@
+The `FileUpload` component is a styled `<div>` which can be used
+with 3rd-party file upload libraries like [react-dropzone](https://github.com/react-dropzone/react-dropzone/).
+
+The example below includes an a mock file upload implementation that
+accepts any file type or size.
+
+```jsx
+const { useDropzone } = require('react-dropzone');
+const { getColor } = require('@zendeskgarden/react-theming/src');
+const { IconButton } = require('@zendeskgarden/react-buttons/src');
+const { Progress } = require('@zendeskgarden/react-loaders/src');
+const FileImageStroke = require('@zendeskgarden/svg-icons/src/16/file-image-stroke.svg').default;
+const CloseStroke = require('@zendeskgarden/svg-icons/src/16/x-stroke.svg').default;
+
+const StyledFileWrapper = styled.ul`
+  direction: ${p => (p.theme.rtl ? 'rtl' : 'ltr')};
+  margin-top: ${p => p.theme.space.sm};
+  list-style: none;
+  padding: 0;
+
+  & > *:not(:first-child) {
+    margin-top: ${p => p.theme.space.xs};
+  }
+`;
+
+const StyledFile = styled.div`
+  position: relative;
+  border: ${p => p.theme.borders.sm};
+  border-color: ${p => getColor('neutralHue', 300, p.theme)};
+  border-radius: ${p => p.theme.borderRadii.md};
+  padding: ${p => p.theme.space.xxs} ${p => p.theme.space.xs};
+  display: inline-flex;
+  min-width: 200px;
+  flex-wrap: nowrap;
+`;
+
+const StyledFileName = styled.span`
+  display: flex;
+  align-items: center;
+  flex-grow: 1;
+  padding: 0 ${p => p.theme.space.xs};
+`;
+
+const StyledFileIcon = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+const StyledProgress = styled(Progress)`
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  margin: 0;
+`;
+
+const File = React.memo(({ name, onRemove }) => {
+  const [uploadProgress, setUploadProgress] = React.useState(0);
+
+  React.useEffect(() => {
+    const randomUploadInterval = Math.random() * (600 - 200) + 600;
+
+    const uploadInterval = setInterval(() => {
+      setUploadProgress(prevProgress => {
+        if (prevProgress >= 100) {
+          clearInterval(uploadInterval);
+          return 100;
+        }
+
+        return prevProgress + 10;
+      });
+    }, randomUploadInterval);
+
+    return () => {
+      clearInterval(uploadInterval);
+    };
+  }, []);
+
+  return (
+    <StyledFile aria-label="File">
+      <StyledFileIcon>
+        <FileImageStroke />
+      </StyledFileIcon>
+      <StyledFileName aria-label="File name">{name}</StyledFileName>
+      <IconButton size="small" onClick={onRemove} aria-label="Remove file">
+        <CloseStroke />
+      </IconButton>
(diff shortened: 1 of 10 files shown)


packages/forms/src/elements/FileUpload.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+  it('renders correct direction when RTL', () => {
+    const { container } = renderRtl(<FileUpload />);
+
+    expect(container.firstChild).toHaveStyleRule('direction', 'rtl');
+  });
+
+  it('renders correct styling when isDragging is active', () => {
+    const { container } = renderRtl(<FileUpload isDragging />);
+
+    expect(container.firstChild).toHaveStyleRule('color', '#0f3554');
```
hzhu, line 35: I noticed that in some [places](https://github.com/zendeskgarden/react-components/blob/main/packages/forms/src/styled/common/StyledMessage.spec.tsx#L31), tests use helpers like `getColor` to retrieve a color. Should these also use a similar method?

packages/forms/src/elements/FileUpload.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { HTMLAttributes } from 'react';
+import PropTypes from 'prop-types';
+import { StyledFileUpload } from '../styled';
+
+export interface IFileUploadProps extends HTMLAttributes<HTMLDivElement> {
+  /** Apply drag styling programmatically */
```
hzhu, line 13: What do you think about "Determines whether to apply drag styling"? Would leaving out the word "programmatically" cause the sentence to lose critical information?

austin94: Much better. Updated.

packages/forms/src/elements/FileUpload.spec.tsx
austin94: Updated to now use style utilities.

austin94:
Our `<p>` tags have top and bottom margin applied in our styleguidist examples. This isn't actually semantically correct so I've replaced them with standard `<span>` elements. I've also applied our medium line-height to clean this up more.

From our offline discussion I've also included a new `isCompact` prop with a new knob-style example.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
```diff
@@ -0,0 +1,110 @@
...
+    background-color 0.25s ease-in-out,
+    color 0.25s ease-in-out;
+  border: dashed ${props => props.theme.borderWidths.sm};
+  border-radius: ${props => props.theme.borderRadii.md};
+  cursor: pointer;
+  padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 15}px`};
+  text-align: center;
+  line-height: ${props => props.theme.lineHeights.md};
+  font-size: ${props => props.theme.fontSizes.md};
+
```
jzempel, line 93: This would benefit from the addition of a `user-select: none` to prevent mouse selection of the internal span text.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
```diff
@@ -0,0 +1,110 @@
...
+    border-color 0.25s ease-in-out,
+    box-shadow 0.1s ease-in-out,
+    background-color 0.25s ease-in-out,
+    color 0.25s ease-in-out;
+  border: dashed ${props => props.theme.borderWidths.sm};
+  border-radius: ${props => props.theme.borderRadii.md};
+  cursor: pointer;
+  padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 15}px`};
+  text-align: center;
+  line-height: ${props => props.theme.lineHeights.md};
```
jzempel, line 91: Using the `getLineHeight` utility along with `space.base * 5` is preferable here. We tend to reserve `lineHeights` theme values for `typography` only.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
```diff
@@ -0,0 +1,110 @@
...
+  /* prettier-ignore */
+  transition:
+    border-color 0.25s ease-in-out,
+    box-shadow 0.1s ease-in-out,
+    background-color 0.25s ease-in-out,
+    color 0.25s ease-in-out;
+  border: dashed ${props => props.theme.borderWidths.sm};
+  border-radius: ${props => props.theme.borderRadii.md};
+  cursor: pointer;
+  padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 15}px`};
```
jzempel, line 89: At worst, the component should be given a `min-width` that is the sum of (border-right + padding-right + padding-left + border-right) in order to protect styling breakage at extreme widths.

@steue may want to comment about adding additional `em` space for the minimum amount of text to show in the center.

Moving all of this to a `sizeStyles` as [directed](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules) would have the added benefit of cleaning up the combinations & inline calculations.

packages/forms/examples/file-upload.md
```diff
@@ -0,0 +1,161 @@
...
+        </Field>
+      </Well>
+      <Field>
+        <Label>File upload</Label>
+        <Hint>Works with react-dropzone</Hint>
+        <FileUpload {...getRootProps()} isDragging={isDragActive} isCompact={isCompact}>
+          {isDragActive ? (
+            <span>Drop files here</span>
+          ) : (
+            <span>Drag files here or click to upload</span>
```
jzempel, line 139: Recommend replacing these `<span>` with `<Ellipsis>` as we'd prefer to not wrap at extreme sizes.

packages/forms/examples/file-upload.md
```diff
@@ -0,0 +1,161 @@
...
+
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop });
+
+  return (
+    <>
+      <Well isRecessed className="u-mb-sm">
+        <Field>
+          <Toggle checked={isCompact} onChange={event => setIsCompact(event.target.checked)}>
+            <Label>
+              <Code>isCompact</Code>
```
jzempel, line 127: Can we see a toggle for disabled?

packages/forms/examples/file-upload.md
```diff
@@ -0,0 +1,161 @@
...
+
+  return (
+    <StyledFile aria-label="File">
+      <StyledFileIcon>
+        <FileImageStroke />
+      </StyledFileIcon>
+      <StyledFileName aria-label="File name">{name}</StyledFileName>
+      <IconButton size="small" onClick={onRemove} aria-label="Remove file">
+        <CloseStroke />
+      </IconButton>
```
jzempel, line 89: [nit] given the constrained space, these might benefit from `focusInset`. What do you think, @steue?

jzempel (changes requested):

packages/forms/examples/file-upload.md
```diff
@@ -0,0 +1,161 @@
...
+      clearInterval(uploadInterval);
+    };
+  }, []);
+
+  return (
+    <StyledFile aria-label="File">
+      <StyledFileIcon>
+        <FileImageStroke />
+      </StyledFileIcon>
+      <StyledFileName aria-label="File name">{name}</StyledFileName>
```
jzempel, line 86: Recommend re-using the [`Span`](https://garden.zendesk.com/components/span#icons) component for this icon-text combo.

packages/forms/src/elements/FileUpload.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it('renders correct direction when RTL', () => {
+    const { container } = renderRtl(<FileUpload />);
+
+    expect(container.firstChild).toHaveStyleRule('direction', 'rtl');
+  });
+
+  it('renders correct styling when isDragging is active', () => {
+    const { container } = renderRtl(<FileUpload isDragging />);
```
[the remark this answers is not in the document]
austin94: Correct. Updated.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
austin94: Still in website mode 😓 Updated.

packages/forms/examples/file-upload.md
austin94: With this being instructional text, I don't believe this should be using an ellipsis. Similar to label/hint/message this should wrap to display the full content. I don't think there is anything positioned relative to this that would require that styling.

Here is the design mocks which would almost definitely be truncated if it was localized with this styling due to the string length:
image

packages/forms/examples/file-upload.md
austin94: Toggle added.

packages/forms/examples/file-upload.md
austin94: Updated.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
austin94: Do we provide a min-width for other block elements? I don't think `Input` includes one

jzempel: For other form elements? Yes. See dropdowns. We want to protect against broken layout styling seen here.

Screen Shot 2020-09-03 at 6 14 20 PM

Even a small `min-width: 4em` to accommodate a longer word will help here.

packages/forms/examples/file-upload.md
jzempel: Actually, we already know from the overflow button treatment on https://zendeskgarden.github.io/react-components/tables/#overflow-menu that this should be `focusInset`.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
jzempel: Please note that `box-sizing` will need to be set in order for the chosen `min-width` to play nice with CSS Bedrock.

packages/forms/examples/file-upload.md
austin94: Updated.

packages/forms/src/styled/file-upload/StyledFileUpload.ts
austin94: `4em` seems to work well with this combination. Updated.

image

jzempel (approved):

PR #853, opened 2020-09-01
fix: correct keyboard navigation for nested multiselect menus

louisscruz:
## Description

The left (or right in RTL) arrow key ends up focusing the last tag in the input for Multi-select components with nested menus. This PR makes it such that pressing the left (or right in RTL) arrow key does not end up focusing the last tag in the input for Multi-select component with nested menus.

## Checklist

      designer as a reviewer)


hzhu:
What do you think about adding a test for the bug? @louisscruz

PR #854, opened 2020-09-03
feat(typography): introduce CodeBlock component

jzempel:
## Description

The new `CodeBlock` component statically renders blocks of code with language syntax highlighting based on [Prism](https://prismjs.com/) and with Garden color palette theming.

## Detail

After careful evaluation, I selected [prism-react-renderer](https://github.com/FormidableLabs/prism-react-renderer) due to flexibility of the model, structural semantics (block -> line -> token), and reduced interference with the underlying Prism tokenization. While more popular, [highlight.js](https://highlightjs.org/) proved to be problematic on each of these points.

There are still plenty of styling nits with the final result, but we obtained the best result that the underlying technology is able to give us. Ideally, Prism improves over time, and our `CodeBlock` will improve with it.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/typography/src/elements/CodeBlock.tsx
+++ b/packages/typography/src/elements/CodeBlock.tsx
@@ -10,9 +10,13 @@ import Highlight, { Language, Prism } from 'prism-react-renderer';
 import { StyledCodeBlock, StyledCodeBlockLine, StyledCodeBlockToken } from '../styled';
 
 export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
+  /** Selects the language used by the Prism tokenizer */
   language?: Language;
+  /** Specifies the font size */
   size?: 'small' | 'medium' | 'large';
+  /** Determines light mode styling */
   isLight?: boolean;
+  /** Determines whether line numbers are displayed */
   isNumbered?: boolean;
 }
 
--- a/packages/typography/src/elements/CodeBlock.tsx
+++ b/packages/typography/src/elements/CodeBlock.tsx
@@ -9,14 +9,10 @@ import React, { HTMLAttributes } from 'react';
 import Highlight, { Language, Prism } from 'prism-react-renderer';
 import { StyledCodeBlock, StyledCodeBlockLine, StyledCodeBlockToken } from '../styled';
 
-interface IRendererArgs {
-  rows: any;
-  stylesheet: any;
-}
-
 export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
   language?: Language;
   size?: 'small' | 'medium' | 'large';
+  isNumbered?: boolean;
 }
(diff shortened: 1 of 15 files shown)


packages/typography/src/styled/StyledCodeBlock.ts
```diff
@@ -0,0 +1,47 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { ThemeProps, DefaultTheme, css } from 'styled-components';
+import { DEFAULT_THEME, getColor, retrieveComponentStyles } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'typography.codeblock';
+
+const colorStyles = (props: IStyledCodeBlockProps & ThemeProps<DefaultTheme>) => {
+  const backgroundColor = getColor('neutralHue', props.isLight ? 100 : 1000, props.theme);
```
jzempel, line 14: @steue note that the use of `1000` here awaits the decision on `grey-900`.

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -0,0 +1,72 @@
...
+  isLight?: boolean;
+  isNumbered?: boolean;
+}
+
+/**
+ * @extends HTMLAttributes<HTMLPreElement>
+ */
+const CodeBlock: React.FunctionComponent<
+  ICodeBlockProps & React.RefAttributes<HTMLPreElement>
+> = React.forwardRef(({ children, isLight, isNumbered, language, size, ...other }, ref) => {
```
austin94, line 24: `React.forwardRef<T, P>()` takes two generics which can remove the need for `React.RefAttributes` and `React.FunctionComponent`.

austin94 (changes requested):

packages/typography/src/styled/StyledCodeBlock.ts
steue: @jzempel design team's consensus is to use `getColor`, as `grey-900` one off isn't ideal and should be visited holistically.

jzempel: Awesome. I'm a bit relieved.

packages/typography/src/elements/CodeBlock.tsx
jzempel: At present, it seems we have dozens of declarations scattered throughout the codebase that don't line up with the simplified syntax. Would probably be good to set a task to clean those up so folks (like me 😅 ) don't keep following suboptimal examples.

austin94 (approved):

hzhu (approved):

steue (approved):
so good! thank you @jzempel !!!!

PR #857, opened 2020-09-03
fix(datepickers): allow event composition on Datepicker inputs

austin94:
## Description

After playing around with the `Datepicker` and `DatepickerRange` components I've found that we are unable to compose event handlers in a similar way to other components.

This can make it difficult to override default functionality and have more granular control over the datepicker state. To help with this control I've added our `composeEventHandler` utility to all events within the components.

## Detail

An example use-case where this could be helpful is if a consumer wants to remove the date value when an empty string is provided. With these new event handlers it could be accomplished with a simple `onChange` event.

```jsx
<Field>
  <Label>Select a date</Label>
  <Datepicker value={state.value} onChange={newDate => setState({ value: newDate })}>
    <Input
      onChange={e => {
        if (e.target.value === '') {
          setState({ value: undefined });
        }
      }}
    />
  </Datepicker>
</Field>
```

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/datepickers/.size-snapshot.json
+++ b/packages/datepickers/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 131231,
-    "minified": 73980,
-    "gzipped": 16493
+    "bundled": 132266,
+    "minified": 74802,
+    "gzipped": 16577
   },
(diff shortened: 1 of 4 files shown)


jzempel (approved):
:tada:

PR #876, opened 2020-09-10
fix(modals): allow Modal to be used in SSR environments

austin94:
## Description

While implementing some `TooltipModal` documentation in the `website` repo I found a server-side rendering issue with our `Modal` component.

Our usage of `document.body` in the `propTypes` definition for `Modal` makes it unable to run within SSR environments. To resolve this I've made the implementation use the updated `useDocument` hook which provides this element safely.

## Detail

Now that we are conditionally rendering the `Modal` content based on `useDocument` this has found some issues with the `useCombinedRefs` hook that we use throughout the components. It works well for our common use-cases, but fell apart here.

After looking into it some more, there are now some popular solutions that are more graceful 😄 https://github.com/gregberge/react-merge-refs stood out as the best option and gives us more flexibility than the current implementation.

I've only added it to `react-modals` for now, but it may be useful to use this exclusively within `react-components` in the future.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/modals/package.json
+++ b/packages/modals/package.json
@@ -27,6 +27,7 @@
     "@zendeskgarden/container-modal": "^0.8.0",
     "@zendeskgarden/container-utilities": "^0.5.1",
     "dom-helpers": "^5.1.0",
+    "react-merge-refs": "^1.1.0",
     "react-popper": "^2.2.3"
   },
   "peerDependencies": {
--- a/packages/modals/.size-snapshot.json
+++ b/packages/modals/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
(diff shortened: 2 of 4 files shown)


packages/modals/package.json
```diff
@@ -27,6 +27,7 @@
     "@zendeskgarden/container-modal": "^0.8.0",
     "@zendeskgarden/container-utilities": "^0.5.1",
     "dom-helpers": "^5.1.0",
+    "react-merge-refs": "^1.1.0",
```
jzempel, line 30: Nice. Let's definitely capture this as a TODO for exploring the `useCombinedRefs` replacement.

jzempel (approved):

hzhu (approved):

PR #878, opened 2020-09-14
chore(docs): migrate react-accordions examples to storybook

austin94:
## Description

This PR migrates the `Stepper` and `Accordion` examples from styleguidist to storybook.

## Detail

This one of the first "complex" examples that is being brought over. I've created some custom `argTypes` for the story controls to match our knobs example. Let me know what you think.

## Checklist

      designer as a reviewer)~


packages/accordions/stories/2-Accordion.stories.tsx
```diff
@@ -0,0 +1,320 @@
+/**
```
hzhu, line 1: Is there a particular reason files under `/accordions/stories` are prefixed with a number like `1-filename.ext`? Should files under `/tabs/stories` follow a number prefix naming?

hzhu (approved):

packages/accordions/stories/2-Accordion.stories.tsx
austin94: This is to ensure the correct ordering of stories:
- `README` first
- Alphabetical groups after that

I've seen this method used in other storybooks, but if there is a cleaner way I would like to get rid of these.

packages/accordions/stories/2-Accordion.stories.tsx
```diff
@@ -0,0 +1,320 @@
...
+Advanced.args = {
+  isCompact: false,
+  isBare: false
+};
+
+Advanced.parameters = {
+  docs: {
+    description: {
+      story: `
+  The \`Accordion\` component is implemented using the [W3C Accordion pattern](https://www.w3.org/TR/wai-aria-practices/#accordion).
```
jzempel, line 310: This repeat from the section above might not be necessary.

jzempel (commented):
What do we think about renaming the sections under accordion to "Uncontrolled" and "Controlled"?

packages/accordions/stories/2-Accordion.stories.tsx
austin94: Removed.

austin94:
@jzempel I've renamed these to "Uncontrolled" and "Controlled". I also split the examples into two separate files to help with readability.

jzempel (approved):
purdy.

packages/accordions/stories/examples/Controlled.tsx
```diff
@@ -0,0 +1,207 @@
...
+Controlled.args = {
+  isCompact: false,
+  isBare: false
+};
+
+Controlled.parameters = {
+  docs: {
+    description: {
+      story: `
+  #### Controlled usage
```
jzempel, line 199: [nit] this line might be superfluous.

austin94: Removed.

PR #881, opened 2020-09-15
chore(docs): migrate react-avatars examples to storybook

austingreendev:
## Description

Migrate the `react-avatars` documentation to Storybook.

## Checklist

      designer as a reviewer)~


packages/avatars/stories/examples/ChromeUsage.tsx
```diff
@@ -0,0 +1,56 @@
...
+import GridIcon from '@zendeskgarden/svg-icons/src/16/grid-2x2-stroke.svg';
+
+export const ChromeUsage: Story = () => {
+  const CustomHeaderItem = HeaderItem as any;
+
+  return (
+    <Chrome isFluid style={{ height: 'auto' }}>
+      <Body>
+        <Header>
+          <CustomHeaderItem as="button">
```
hzhu, line 21: It looks like there's an a11y violation shown in the addon-a11y tab. Since this element is rendered as a button, could it have an `aria-label`? I think that might address the violation.

packages/avatars/stories/examples/ChromeUsage.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <HeaderItemIcon>
+              <GridIcon />
+            </HeaderItemIcon>
+          </CustomHeaderItem>
+          <CustomHeaderItem as="button" isRound>
+            <Avatar size="extrasmall">
```
jzempel, line 27: Why is this Avatar being styled as `isSystem` square? It needs to be round for this combo example.

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,233 @@
...
+          </Row>
+          <Row alignItems="center" style={{ backgroundColor: surfaceColor, marginBottom: '8px' }}>
+            <Col style={{ textAlign: 'center' }}>
+              <Avatar
+                size={size}
+                status={computedStatus}
+                badge={computedBadge}
+                surfaceColor={surfaceColor}
+              >
+                <img alt="" src={`images/avatar-3.png`} />
```
hzhu, line 119: Could `src` attribute be a string since it doesn't need string template features?

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const computedStatus = useMemo(() => {
+    if (status === 'none') {
+      return undefined;
+    }
+
+    return status;
+  }, [status]);
+
+  return (
+    <Grid>
```
jzempel, line 55: Can we get this grid to layout similar to the example at https://zendeskgarden.github.io/react-components/avatars/#basic? More like a "box" of avatars, and the surface color should apply throughout and beyond the extents of the grid of Avatars.

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,233 @@
...
+      <Row>
+        <Col size="1">
+          <Row style={{ height: '10%' }}></Row>
+          <Row alignItems="center" style={{ height: '30%' }}>
+            <Col style={{ textAlign: 'center' }}>
```
jzempel, line 60: All the `textAlign` styles can be replaced with `Col` prop.

packages/avatars/stories/examples/Default.tsx
hzhu, line 119: Could src attribute be a string since it doesn't need string template features? I think there are a few other src attributes that also use a string template without using string template features.

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,233 @@
...
+              </Avatar>
+            </Col>
+          </Row>
+        </Col>
+      </Row>
+    </Grid>
+  );
+};
+
+Default.argTypes = {
```
jzempel, line 167: Is there a way to get these controls to display in the same order as https://zendeskgarden.github.io/react-components/avatars/#basic?

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,233 @@
...
+  surfaceColor: PALETTE.white,
+  backgroundColor: PALETTE.kale[800],
+  foregroundColor: PALETTE.white
+};
+
+Default.parameters = {
+  docs: {
+    description: {
+      component: `
+  ### Basic
```
jzempel, line 209: This heading probably isn't needed. Just have the initial paragraph live under "Avatars".

packages/avatars/stories/examples/Default.tsx
austin94: Correct. Prettier got aggressive here.

packages/avatars/stories/examples/MenuUsage.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <Col>
+          <Dropdown onStateChange={changes => setHighlightedItem(changes.highlightedIndex)}>
+            <Trigger>
+              <Button>Default menu</Button>
+            </Trigger>
+            <Menu hasArrow>
+              <MediaItem value="linden">
+                <MediaFigure>
+                  <Avatar
+                    size="small"
```
jzempel, line 39: There is a rendering problem with this example. Why are the Avatars not being shown as small?
Screen Shot 2020-09-16 at 9 48 34 AM

jzempel (changes requested):
Why are the "Chrome usage" and "Menu usage" stories shown with "No code available"? Is there a way to fix that?

packages/avatars/stories/examples/ChromeUsage.tsx
austin94: Updated.

packages/avatars/stories/examples/Default.tsx
austin94: Updated.

packages/avatars/stories/examples/MenuUsage.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <Col>
+          <Dropdown onStateChange={changes => setHighlightedItem(changes.highlightedIndex)}>
+            <Trigger>
+              <Button>Small menu</Button>
+            </Trigger>
+            <Menu hasArrow isCompact>
+              <MediaItem value="clove">
+                <MediaFigure>
+                  <Avatar
+                    size="extrasmall"
```
jzempel, line 93: Per https://zendeskgarden.github.io/react-components/dropdowns/#advanced-menu-usage these should've been updated to be `extraextrasmall`. Again, there appears to be a rendering issue with storybook.

packages/avatars/stories/examples/Default.tsx
austin94: Looks like _all_ controls must have a default value for ordering to be applied. Looks better now.

packages/avatars/stories/examples/Default.tsx
austin94: Updated.

packages/avatars/stories/examples/MenuUsage.tsx
austin94: These are rendering correctly for me in development and production builds

image

packages/avatars/stories/examples/MenuUsage.tsx
austin94: I've updated the size and they seem to be positioned correctly for me now. What do you think?

packages/avatars/stories/examples/ChromeUsage.tsx
austin94: This is randomly happening based depending on styled-components specificity. Going to keep digging into this one.

packages/avatars/stories/examples/Default.tsx
austin94: I've rewritten this grid layout to use a [2-dimensional table](https://www.w3.org/WAI/tutorials/tables/two-headers/). This is now semantically correct and solves most of the layout issues that I was having.

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const StyledTable = styled.table.attrs<Partial<IStoryProps>>(p => ({
+  style: { backgroundColor: p.surfaceColor }
+}))<Partial<IStoryProps>>`
+  width: 100%;
+  border-spacing: 0;
+`;
+
+const StyledAvatarTd = styled.td`
+  padding: ${p => p.theme.space.xs} 0;
```
jzempel, line 25: Might feel better OOTB with large/xl padding.

packages/avatars/stories/examples/ChromeUsage.tsx
jzempel: yeah, still weird on https://5f62659c6de450374ab825a5--zendeskgarden-react-components.netlify.app/storybook/?path=/story/components-avatars--chrome-usage

packages/avatars/stories/examples/MenuUsage.tsx
jzempel: Hmm, deployment still off: https://5f62659c6de450374ab825a5--zendeskgarden-react-components.netlify.app/storybook/?path=/story/components-avatars--menu-usage

hzhu: I'm seeing the same rendering issues @jzempel is seeing with the Deployment link in an incognito browser window 🤔 .

packages/avatars/stories/examples/Default.tsx
austin94: Update to include large padding.

packages/avatars/stories/examples/MenuUsage.tsx
austin94: I found the issue! Both `react-menus` and `react-chrome` include styling rules for `img` elements. `Avatar` needs to increase it's specificity to beat out these competing styles.

This wasn't occurring in stylguidist as the order of style insertion was handled differently which avoided the specificity issues.

packages/avatars/stories/examples/ChromeUsage.tsx
austin94: I've increased Avatar styling specificity to beat the `img` styles supplied by `react-chrome`

austin94:
@jzempel I'm still seeing some issues with the menu usage when this is built in production mode. Will need to dig into this further.

packages/avatars/stories/examples/MenuUsage.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { Button } from '@zendeskgarden/react-buttons';
+import { PALETTE } from '@zendeskgarden/react-theming';
+
+export const MenuUsage: Story = () => {
+  const [highlightedItem, setHighlightedItem] = useState<number | null>();
+
+  return (
+    <Grid>
+      <Row>
+        <Col>
```
hzhu, line 30: Would `<Col="center">` center the menu buttons? I think in the Buttons stories they might have [centered buttons](https://github.com/zendeskgarden/react-components/pull/889/files#diff-6110424735f7b4751a85edd0cc84b336R36) so it might be nice to the menu buttons centered too.

hzhu (approved):

packages/avatars/stories/examples/MenuUsage.tsx
austingreendev: Updated.

austingreendev:
@jzempel I've added some additional `!important` specificity to help with some of the class ordering issues. Does this look good?

packages/avatars/src/styled/StyledAvatar.ts
```diff
@@ -194,8 +194,10 @@ const sizeStyles = (props: IStyledAvatarProps & ThemeProps<DefaultTheme>) => {
 
   return css`
     border-radius: ${borderRadius};
-    width: ${size};
-    height: ${size};
+    /* stylelint-disable declaration-no-important */
+    width: ${size} !important;
+    height: ${size} !important;
```
jzempel, line 199: These should get a `/* [1] */` head note to explain why the `!important` is necessary.

packages/avatars/src/styled/StyledAvatar.ts
```diff
@@ -250,19 +252,19 @@ export const StyledAvatar = styled.figure.attrs({
...
 
-  & > img {
+  && > img {
     box-sizing: inherit;
     vertical-align: bottom;
     object-fit: cover;
   }
 
-  & > svg {
+  && > svg {
```
jzempel, line 267: Can these double `&&` hacks be avoided by applying `!important` to the width/height? In this case I think that would be preferable, and more bullet-proof styling.

packages/dropdowns/src/styled/items/media/StyledMediaFigure.ts
```diff
@@ -30,7 +30,9 @@ export const StyledMediaFigure = styled(
   'data-garden-version': PACKAGE_VERSION
 })<IStyledMediaFigureProps>`
   float: ${props => (props.theme.rtl ? 'right' : 'left')};
-  margin-top: ${props => props.theme.space.base * 0.5}px;
+  /* stylelint-disable declaration-no-important */
+  margin-top: ${props => props.theme.space.base * 0.5}px !important;
```
jzempel, line 34: Please `/* [1] */` document the reason for `!important` here.

packages/avatars/stories/examples/Default.tsx
```diff
@@ -0,0 +1,232 @@
...
+`;
+
+const StyledAvatarTd = styled.td`
+  padding: ${p => p.theme.space.lg} 0;
+  text-align: center;
+`;
+
+interface IStoryProps {
+  size: 'extraextrasmall' | 'extrasmall' | 'small' | 'medium' | 'large';
+  status: 'none' | 'away' | 'available';
```
jzempel, line 31: Will the lack of `status` animation transitions be corrected when this PR is merged with @hzhu's fix? https://github.com/zendeskgarden/react-components/pull/889/commits/fc68647ba29ef28d5f57be618c82cd44f7a58f82

jzempel (changes requested):
Curious if you looked at `chrome` to see what the differences might be between this and that storybook's ability to "Show code" across the examples?

packages/avatars/stories/examples/Default.tsx
austingreendev: Is this not showing animations for you? I've already merged `main` into this branch and the animations are working well for me.

packages/avatars/src/styled/StyledAvatar.ts
austingreendev: Updated with a comment

packages/dropdowns/src/styled/items/media/StyledMediaFigure.ts
austingreendev: Updated with comment.

packages/avatars/src/styled/StyledAvatar.ts
austingreendev: Since this is applied to nested elements, I don't think we would have to worry about these specificity issues being beaten. With `<figure />` being the wrapping element I think we should limit `!important` overrides to that element for now.

austingreendev:
@jzempel I'm not sure why those specific stories aren't showing their source. I've updated them to only return native elements and the source still isn't being produced.

packages/avatars/stories/examples/Default.tsx
jzempel: Oh, interesting. They are animating as expected on the Canvas tab, but not on the Docs tab. See https://5f84ace4f41bdb249d521b09--zendeskgarden-react-components.netlify.app/storybook/?path=/docs/components-avatars--default

packages/avatars/src/styled/StyledAvatar.ts
jzempel: The problem is that we've learned that double/triple specificity hacks are a real pain for customization. In this case, I suspect the hack is being applied to items where it is not needed (`box-sizing`, `vertical-align`, `object-fit`?). It turns out `!important offers a cleaner path for customization overrides, should those ever be attempted.

packages/avatars/src/styled/StyledAvatar.ts
```diff
@@ -192,10 +192,16 @@ const sizeStyles = (props: IStyledAvatarProps & ThemeProps<DefaultTheme>) => {
     svgSize = `${props.theme.space.base * 4}px`;
   }
 
+  /**
+   * 1. Ensure width and height are applied when used in combation with
```
jzempel, line 196: [nit] *combination

packages/dropdowns/src/styled/items/media/StyledMediaFigure.ts
```diff
@@ -30,7 +33,9 @@ export const StyledMediaFigure = styled(
   'data-garden-version': PACKAGE_VERSION
 })<IStyledMediaFigureProps>`
   float: ${props => (props.theme.rtl ? 'right' : 'left')};
-  margin-top: ${props => props.theme.space.base * 0.5}px;
+  /* stylelint-disable declaration-no-important */
+  margin-top: ${props => props.theme.space.base * 0.5}px !important; /* [1] */
```
jzempel, line 37: I'm curious, where is there a `margin-top` in StyledAvatar? I'm not seeing it.

jzempel (approved):
Bummer about the "Show code". I still have a few comments/curiosities.

packages/dropdowns/src/styled/items/media/StyledMediaFigure.ts
austingreendev: The `margin: 0; /* <figure> reset */` is what is winning right now

packages/avatars/src/styled/StyledAvatar.ts
austingreendev: Fixed!

packages/dropdowns/src/styled/items/media/StyledMediaFigure.ts
jzempel: I keep thinking `margin-top` beats out `margin` in terms of specificity. But, alas, this is not true.

hzhu (approved):

hzhu:
It looks like there is a console warning for the "Default" story for avatars:

Screen Shot 2020-10-13 at 10 29 29 AM

austingreendev:
@hzhu Weird, not seeing that locally for me until I do a hard-refresh. Updated.

hzhu (approved):

PR #884, opened 2020-09-17
feat(modals): introduce drawer modal

hzhu:
## Description

This PR introduces the new drawer modal component, `<DrawerModal>`, to the `react-modals` package.

## Detail

The `<DrawerModal>` component opens and transitions a dialog from one side (right to left) at this time. The component leverages existing `useModal`, `useModalContext`, `ModalsContext`, and some existing `styled` modal components.

### Transition state

To achieve the intended CSS transitions, the component sets style values and controls an internal (private and local) React state called `showDrawer`. This internal state is updated accordingly based on `transitionend` event which determines when the `DrawerModal` component can be unmounted. The transition must finish before the dialog can be unmounted. This means that `onClose` function is asynchronous and invokes pending the `transitionend` event.

### Transition approach (alternative)

ℹ️ **After discussing with the team, this alternative has been selected.**

An alternative to the approach described above would be to use [ReactTransitionGroup](https://reactjs.org/docs/animation.html#low-level-api-reacttransitiongroup) which provides API access to component transition lifecycle events. This is helpful to animate elements that leave the DOM (unmounted by React). An element that isn't in the DOM cannot be transitioned. It could be valuable to leverage the helper library so that Garden components don't have to worry about syncing transition state with React mounts/unmounts.

The [minified + gzipped](https://bundlephobia.com/result?p=react-transition-group@4.4.1) bundle size for `ReactTransitionGroup` is `4.3kB`. 

### Transition performance
The CSS transitions the `transform` property which should be [performant](https://csstriggers.com/transform).

> Changing transform does not trigger any geometry changes or painting, which is very good. This means that the operation can likely be carried out by the compositor thread with the help of the GPU.

This would be opposite to transitioning on CSS properties which trigger layout and painting such as [`left`](https://csstriggers.com/left) or `right`.

## Checklist

      designer as a reviewer)


packages/modals/src/styled/StyledDrawerModalHeader.ts
```diff
@@ -0,0 +1,25 @@
...
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledHeader } from './StyledHeader';
+
+const COMPONENT_ID = 'modals.drawer_modal.header';
+
+export const StyledDrawerModalHeader = styled(StyledHeader).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  padding: 20px;
```
hzhu, line 18: 👀

packages/modals/src/elements/DrawerModal/Header.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useModalContext } from '../../utils/useModalContext';
+import { StyledDrawerModalHeader } from '../../styled';
+
+/**
+ * Accepts all `<div>` props
+ */
+export const Header = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
+    const { getTitleProps } = useModalContext();
+
```
rossmoody, line 18: Like we chatted about, in lieu of a possibly more comprehensive fix for the close button functionality in Modals, maybe [we follow Tooltip Modal's lead](https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledTooltipModalTitle.ts#L19) of padding the Header container so the content doesn't [run under the absolutely positioned Close button.](https://share.getcloudapp.com/wbu72y4Z)

rossmoody (changes requested):

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+      if (!environment) {
+        return undefined;
+      }
+
+      const bodyElement = environment.querySelector('body');
+
+      if (bodyElement && showDrawer) {
+        const previousBodyOverflow = bodyElement.style.overflow;
+
+        bodyElement.style.overflow = 'hidden';
```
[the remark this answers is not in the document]
hzhu: I haven't been able to recreate this issue both locally and on Netlify; both in regular and incognito browser windows.

Anyone else experiencing this issue? @jzempel @austin94

packages/modals/src/styled/StyledDrawerModalHeader.ts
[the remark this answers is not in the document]
jzempel: yes, plz `space.base * 5`

packages/modals/src/styled/StyledDrawerModalFooter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const StyledDrawerModalFooter = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  flex-shrink: 0;
+  align-items: center;
+  justify-content: flex-end;
+  margin-top: auto;
```
jzempel, line 21: What is `margin-top` providing? Doesn't see like it is needed.

packages/modals/src/styled/StyledDrawerModalFooter.ts
```diff
@@ -0,0 +1,30 @@
...
+  border-top: ${props => props.theme.borders.sm} ${getColor('neutralHue', 200)};
```
jzempel, line 22: Please provide the `theme` parameter to `getColor`.

packages/modals/src/styled/StyledDrawerModalBackdrop.ts
```diff
@@ -0,0 +1,25 @@
...
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledBackdrop } from './StyledBackdrop';
+
+const COMPONENT_ID = 'modals.drawer_modal.backdrop';
+
+export const StyledDrawerModalBackdrop = styled(StyledBackdrop).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  background-color: transparent;
```
jzempel, line 18: The design decision to apply `overflow: hidden` to the `<body>` on Garden's `<Modal>` is based, in large part, on the colored backdrop taking over the screen.

With transparency, it doesn't feel right that the background content is unable to scroll. The implementation should be adjusted to _not_ apply body `overflow: hidden` for the `<Drawer>`-styled modal.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * Callback when a close action has been completed.
+   * Can be triggered from the backdrop and Close icon.
+   * @param {Object} event - DOM event that triggered the close action
+   */
+  onClose?: (event: TransitionEvent) => void;
+  /**
+   * Enable drawer modal animation
+   */
+  isAnimated?: boolean;
```
jzempel, line 69: `isAnimated={false}` does not appear to be working as expected.

packages/modals/src/styled/StyledDrawerModal.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const COMPONENT_ID = 'modals.drawer.modal';
+
+export const StyledDrawerModal = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  position: fixed;
+  right: -10000px;
```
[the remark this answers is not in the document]
jzempel: Why is this needed? The modal is unmounted when hidden.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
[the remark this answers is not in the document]
jzempel: I'm not seeing it. I tried with the dev tools open as seen in the screen grab. Got any extensions enabled, Ross?

packages/modals/src/styled/StyledDrawerModal.ts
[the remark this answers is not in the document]
hzhu: It looks like I was trying to do something that wasn't needed. I'll remove this.

packages/modals/src/styled/StyledDrawerModalClose.ts
```diff
@@ -0,0 +1,28 @@
...
+import { StyledClose } from './StyledClose';
+
+const COMPONENT_ID = 'modals.drawer_modal.close';
+
+export const StyledDrawerModalClose = styled(StyledClose).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  top: ${props => props.theme.space.base * 3.5}px;
+  ${props => (props.theme.rtl ? 'left' : 'right')}: ${props => `${props.theme.space.base * 3}px`};
```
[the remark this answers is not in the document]
hzhu: Oh nice, looks like it could probably be removed from `StyledTooltipModalClose.ts` too.

packages/modals/src/elements/DrawerModal/Header.tsx
hzhu: Unfortunately, `TooltipModal` is designed differently enough that `DrawerModal` can't have similar styles here. 

`TooltipModal` has padding on the dialog element and no horizontal rules. The `TooltipModal` header padding would reduce each other sides to `0px`, and apply them instead to the dialog which would then put spacing at the start and end of the horizontal rules. The end padding could be adjusted to, say `60px`, to compensate for the `Close` button. 

If this is the path chosen, `Modal`'s header should be updated since this was mostly a copy 🍝 job based on what's in production.

packages/modals/src/styled/StyledDrawerModalFooter.ts
hzhu: I think at one point during spike it was used to keep the footer stickied to the bottom. However, it looks like it isn't needed and should be removed.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: Whoops, yeah this prop isn't passed through to `StyledBackdrop`. 

I'm not sure if there is a non-transitioned `DrawerModal` (e.g transitioning the `transform`). @rossmoody should the `DrawerModal` have an option to disable the "slide-in" transition? If so, I could bake the ability to enable/disable transition into the `isAnimated` prop.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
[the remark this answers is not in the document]
rossmoody: I don't see how it's possible, but today it isn't happening. I didn't update or change anything. Looks like this is all good. 👍

packages/modals/src/elements/DrawerModal/Header.tsx
rossmoody: I would defer to y'all on the correct approach holistically but I think it's pretty important the title content doesn't overlap the close button.

packages/modals/src/styled/StyledDrawerModalBackdrop.ts
rossmoody: The hope is to be less jarring than a standard modal and not block visual access to the page unnecessarily. If a user wants to start interacting with the page content again, it means they no longer want to perform the action they prompted the Drawer with (and it will automatically close on click).

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
rossmoody: I would be curious the rationale, but considering this functionality exists on all other components, I think it's safe to assume we'd want it here too.

packages/modals/src/styled/StyledDrawerModalBackdrop.ts
hzhu: > The hope is to be less jarring than a standard modal 

I'm not sure what that means; could you clarify how the "standard modal" is "jarring"? Is the current Garden [Modal](https://zendeskgarden.github.io/react-components/modals/#basic) jarring? And how would full transparency address that issue?

>  The hope is to...not block visual access to the page unnecessarily

I've outlined a couple of reasons why I think it could be necessary above. It seems like you may think those are invalid reasons? I would like to understand why.

> If a user wants to start interacting with the page content again, it means they no longer want to perform the action they prompted the Drawer with (and it will automatically close on click).

Yeah, I understand this. But I'm not sure how this is relevant to the concerns I've posed [here](https://github.com/zendeskgarden/react-components/pull/884#discussion_r492251198)? Also, it isn't clear what _"it will automatically close on click"_ means. How is it automatic, if the user has to initiate the event?

rossmoody: By jarring I meant how the standard modal interrupts by blocking vision to the page content and appears in the center of the page. 

I think those concerns have merit and I've shared them at various points while writing the design spec (early iterations had dark backdrops). Users will see the Drawer close when they click outside it and since it's elevated towards to the user on top of the page, the Drawer closing will be an immediate response to a click so I don't foresee any confusion there. 

This interaction is really similar to the Tooltip modal. If a user opens a Tooltip modal and tries to interact with anything else on the page, the first click closes the Tooltip. If this interaction proves confusing to users in practice, I imagine we'll need to revisit and fix both scenarios.

hzhu: > By jarring I meant how the standard modal interrupts by blocking vision to the page content and appears in the center of the page.

Thanks for clarifying as I would not have interpretted this as jarring. Isn't interruption the [purpose](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) of a modal dialog though? To interrupt the user by presenting UI they must interact with before moving on? 

> This interaction is really similar to the Tooltip modal

This interaction is also similar to the `Modal`. It is difficult to make the code changes when the engineering reason I must provide is "Polaris does it" or "X component does it". I'll go ahead and leave this fully transparent and remove the overflow hidden, but I’m still concerned there aren’t concrete reasons to do so.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+  focusOnMount: PropTypes.bool,
+  restoreFocus: PropTypes.bool,
+  onClose: PropTypes.func,
+  appendToNode: PropTypes.any,
+  id: PropTypes.string
+};
+
+DrawerModal.defaultProps = {
+  isAnimated: true,
+  appendToNode: document.body
```
hzhu, line 251: I need to update this to reflect https://github.com/zendeskgarden/react-components/pull/876

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: The `Modal` animates the backdrop background color. `DrawerModal` does not have a colored backdrop to animate or transition. But other components such as `Accordion` and `Steppers` **don't** have the ability to turn off transitions.

rossmoody: Nice. Do we know why it was included initially?

austin94: The original `isAnimated` prop also handles some entrance transitions for `Modal` I believe. We've had some areas disable animation from designer guidance, but I don't remember a rhyme or reason behind it.

In general, having the option to reduce animation globally is an [important a11y characteristic](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion). Normally this is handled by a global media query, but doing it at the component level could be helpful 🤷

packages/modals/examples/drawer-modal.md
```diff
@@ -0,0 +1,68 @@
...
+  const [isOpen, setIsOpen] = React.useState(false);
+  const open = () => setIsOpen(true);
+  const close = () => setIsOpen(false);
+
+  return (
+    <>
+      <Button onClick={open}>Open Drawer</Button>
+      <DrawerModal isOpen={isOpen} onClose={close}>
+        <DrawerModal.Header>Example Title</DrawerModal.Header>
+        <DrawerModal.Close />
```
austin94, line 15: For other Modal examples we've included the `Close` component at the end of the tab order. Should we do the same in this example?

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const Example = forwardRef<HTMLDivElement, IDrawerModalProps>((props, ref) => {
+    const [isOpen, setIsOpen] = useState(false);
+    const buttonRef = useRef<HTMLButtonElement>(null);
+
+    return (
+      <>
+        <button ref={buttonRef} onClick={() => setIsOpen(true)}>
+          Open Drawer
+        </button>
+
```
austin94, line 25: [nit] extra line?

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+  appendToNode?: Element;
+  /**
+   * Determines whether to apply keyboard focus on the drawer modal on mount
+   */
+  focusOnMount?: boolean;
+  /**
+   * Determines whether to return keyboard focus to the element that triggered the drawer modal
+   */
+  restoreFocus?: boolean;
+  children?: any;
```
austin94, line 83: Is this necessary?

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+}
+
+export const DrawerModal = forwardRef<
+  HTMLDivElement,
+  IDrawerModalProps & ThemeProps<DefaultTheme> & HTMLAttributes<HTMLDivElement>
```
austin94, line 88: With this using the new `ThemeContext` object we can remove `ThemeProps<DefaultTheme>` I think.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+          modalRef.current.addEventListener('transitionend', unmountDrawer);
+
+          if (rtl) {
+            modalRef.current.style.transform = 'translateX(0px)';
+          } else {
+            modalRef.current.style.right = `-${width}px`;
+            modalRef.current.style.transform = `translateX(${width}px)`;
+          }
+        }
+      }
```
austin94, line 139: There is a lot logic here for the unmount logic. We currently don't animate the unmount of our other Modal components. @rossmoody are we sure that the fade-out is a large user benefit?

If so we may need to think about create a new type of API for this type of interaction. 

Some first thoughts are not basing the visibility on the `DrawerModal` mount/unmount cycle. We could include an `isVisible` prop which would allow the value to still be controlled via `onClose`, but helps us avoid the event listening. This could also allow us to start using react-transition-group.

But, my first hunch is to align with the other Modal animation patterns we have.

packages/modals/src/styled/StyledDrawerModalBackdrop.ts
austin94: If the user is able to interact with the background content, that would make the scroll-bar visible. Especially in Windows environments we will start seeing a sidebar sandwich that may not look valid with a full-height modal.

austin94: From the [W3C design docs](https://www.w3.org/TR/wai-aria-practices/#dialog_modal)

> Windows under a modal dialog are inert. That is, users cannot interact with content outside an active dialog window. Inert content outside an active dialog is typically visually obscured or dimmed 

If their inert scrolling wouldn't be allowed, similar to `Modal` and `TooltipModal`

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
rossmoody: I would say it's definitely a more pleasant user experience to witness the Drawer animate out as opposed to vanish.

hzhu: @austin94 with `react-transition-group` the unmount/mount logic can be removed 💯 . If the primary reason to kill the transition upon unmount is the logic, then it can keep it since `react-transition-group` will help remove that logic. 

@rossmoody I think a good reason to have the unmount transition for the `DrawerModal` is that the transition on mount is more emphasized than on a `Modal`. E.g. the transition duration is longer and the CSS property being transitioned is more apparent (`opacity` vs `translateX`). So transitioning on unmount for the `DrawerModal` makes it feel more natural.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: I'm using this type definition to tell TS that `children` can be a prop. I could make this more specific and use `children?: React.ReactNode` too. Is there another way around these TS compile errors?

Screen Shot 2020-09-24 at 12 09 51 PM

packages/modals/examples/drawer-modal.md
hzhu: Yeah that sounds good.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
austin94: Weird, I'm not seeing this issue in the `TooltipModal` PR which includes a similar static interface. Do you see any differences there?

package.json
```diff
@@ -59,6 +59,7 @@
     "@types/prop-types": "15.7.3",
     "@types/react": "^16.9.33",
     "@types/react-dom": "^16.9.6",
+    "@types/react-transition-group": "^4.4.0",
```
austin94, line 62: If this isn't a global package (scoped to `react-modals`) should we have this `devDependency` in `modals/package.json`?

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+            <DrawerModal.FooterItem>
+              <button>Click</button>
+            </DrawerModal.FooterItem>
+          </DrawerModal.Footer>
+        </DrawerModal>
+      </>
+    );
+  });
+
+  Example.displayName = 'Example';
```
austin94, line 39: Should we disable this eslint rule for spec files?

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return null;
+    }
+
+    const modalProps = isOpen
+      ? getModalProps({ ref: mergeRefs([ref, modalRef]), ...props } as any)
+      : props;
+
+    return ReactDOM.createPortal(
+      <ModalsContext.Provider value={value}>
+        <CSSTransition in={isOpen} timeout={250} classNames="drawer-transition" unmountOnExit>
```
austin94, line 159: We may want to namespace this class to `garden-*` or a data-attribute to avoid any styling collision.

package.json
hzhu: I initially had it under `packages/modals/package.json`, but got a huge explosion of TS errors. I'm going to give it another attempt though.

Screen Shot 2020-09-24 at 12 46 13 PM

austin94: We may need to introduce an `@types/react` resolution in the root `package.json` if the versions aren't compatible (this probably has to do with the older `@types/styled-components` we're using)

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
hzhu: I copy 🍝 from [TooltipModal.spec.tsx](https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/TooltipModal/TooltipModal.spec.tsx#L50). I'm not sure, maybe it could be kept since it doesn't hurt and it's straight forward to do what the linter is telling us to do?

austin94: 👍 no problem if we want to leave it. This is only really useful for non-test components to show correctly in the stack-trace or React devtools

package.json
hzhu: >  introduce an @types/react resolution in the root package.json 

No luck so far 😔 .

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: Doh, fixed it!

```
export interface IDrawerModalProps extends HTMLAttributes<HTMLDivElement> {}
```

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: Updated

package.json
hzhu: This seems to work:

1) Install `@types/react-transition-group` into `packages/modal`
2) Add `@types/react` to yarn `resolutions`
3) Run `yarn install` which updates the root `yarn.lock` file

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  it('applies backdropProps to Backdrop element', () => {
+    const { getByTestId, getByText } = render(
+      <Example backdropProps={{ 'data-test-id': 'backdrop' }} />
+    );
+
+    userEvent.click(getByText('Open Drawer'));
+
+    expect(getByTestId('backdrop')).not.toBeNull();
```
hzhu, line 76: I think this could assert first that the backdrop doesn't exist, before making an assertion that it does. Also, it may be worth using the `toBeInTheDocument` matcher as that is more explicit than "this element is not null".

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
jzempel: We've also seen iframe contexts get wonky out of Garden's control and turn off animation. So we provide the prop, as a rule.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+interface IStaticDrawerModalExport<T, P>
+  extends ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>> {
+  Header: typeof Header;
+  Body: typeof Body;
+  Close: typeof Close;
+  Footer: typeof StyledDrawerModalFooter;
+  FooterItem: typeof StyledDrawerModalFooterItem;
+}
+
+export interface IDrawerModalProps extends HTMLAttributes<HTMLDivElement> {
```
jzempel, line 48: Looks like `isAnimated` didn't make the cut on this. Ok, I can buy it.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+      if (!environment) {
+        return undefined;
+      }
+
+      const bodyElement = environment.querySelector('body');
+
+      if (bodyElement && isOpen) {
+        const previousBodyOverflow = bodyElement.style.overflow;
+
+        bodyElement.style.overflow = 'hidden';
```
jzempel, line 120: I'm still tripping on this. But I'm willing to suspend belief in the interest of getting the component out into people's hands. For me: modal + transparent background ≠ locked scrolling. Let's see if there is kick-back.

jzempel (approved):
`react-transition-group` 🎉

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
jzempel: One more: the ability to knock-out animation is another feature of Garden that reinforces an open-source concept, allowing users to make it theirs by applying their own animation rubric (assuming they can hang with our sizes/shapes).

packages/modals/src/elements/DrawerModal/Header.tsx
hzhu: Updated the padding to compensate spacing for the close icon:

**BEFORE:**
Screen Shot 2020-09-29 at 8 57 46 AM

**AFTER:**
Screen Shot 2020-09-29 at 8 58 21 AM

packages/modals/src/styled/StyledDrawerModalBackdrop.ts
hzhu: Resolving thread, for now with the full transparency background and scroll lock, to get component released and into usage ✅ . I'll note some of JZ and my concerns and reopen this thread when the topic arises again.

packages/modals/src/elements/DrawerModal/Header.tsx
jzempel: @hzhu did you consider the conversation under https://github.com/zendeskgarden/react-components/pull/835? All elements of the modal are optional. In this case, removing the `Close` would result in a pretty severe visual imbalance for the `Header` title.

jzempel: In that previous PR, I think we came to the conclusion that overflow prevention was left as an exercise for the component consumer (i.e. add a fixed-width `span`), depending on modal elements included + string length.

hzhu: I had nearly forgotten a PR was opened for this. That sounds fair, `DrawerModal` should align with `Modal`. I'm going to revert this particular commit. If this issue needs to be revisited in the future, we could open a new PR.

rossmoody (approved):

PR #887, opened 2020-09-22
fix(modals): provide ref object for focusvisible

hzhu:
## Description

It was recently discovered that focusvisible behavior broke within modal components.

## Detail

It looks like the `environment` changes when the modal is mounted but the change is not captured by the `useFocusVisible` hook. 

In order to capture this change, the `environment` is passed to `useFocusVisible` as the `relativeDocument`. A `useEffect` within `useFocusVisible` re-runs the effect accordingly when the `relativeDocument` is updated. 

🙏  Thanks @austin94 for helping figure out the root problem.

## Screenshots

**Before:**

**After:**

## Checklist

### DIFF
--- a/packages/modals/src/elements/Modal.tsx
+++ b/packages/modals/src/elements/Modal.tsx
@@ -119,7 +119,7 @@ export const Modal = React.forwardRef<HTMLDivElement, IModalProps>(
       restoreFocus
     });
 
-    useFocusVisible({ scope: modalRef });
+    useFocusVisible({ scope: modalRef, relativeDocument: environment });
 
     useEffect(() => {
       if (!environment) {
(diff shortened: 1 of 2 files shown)


packages/modals/src/elements/Modal.tsx
```diff
@@ -119,7 +129,13 @@ export const Modal = React.forwardRef<HTMLDivElement, IModalProps>(
       restoreFocus
     });
 
-    useFocusVisible({ scope: modalRef });
+    useFocusVisible({ scope });
```
austin94: From our offline talk, lets take a look at the `relativeDocument` option to try and get this to recognize the conditional ref.

austin94 (approved):

PR #888, opened 2020-09-22
fix(table): revise cell type definition

rossmoody:
## Description

This PR is a little bit of a reach as my TS knowledge is non-existent but I took a stab at a fix. While doing the Table examples for the site, I was getting an error that colSpan didn't exist on the Cell component. It looks like `colSpan` doesn't exist on `HTMLAttributes` but it does on `ThHTMLAttributes`. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/tables/src/elements/Cell.tsx
+++ b/packages/tables/src/elements/Cell.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { ThHTMLAttributes } from 'react';
+import React, { TdHTMLAttributes } from 'react';
 import PropTypes from 'prop-types';
 
 import { StyledCell, IStyledCellProps } from '../styled';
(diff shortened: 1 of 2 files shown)


packages/tables/src/elements/Cell.tsx
```diff
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { HTMLAttributes } from 'react';
+import React, { ThHTMLAttributes } from 'react';
```
austin94, line 8: Should this be using `TdHTMLAttributes` as it's a table cell, not header cell?

Likewise `HeaderCell` should use `ThHTMLAttributes` like you have here.

austin94 (changes requested):

packages/tables/src/elements/Cell.tsx
rossmoody: Doh, nice spot on both accounts.

austin94 (approved):

PR #894, opened 2020-10-08
fix(modals): prevent html overflow

hzhu:
## Description

When Bedrock CSS is turned on and the `Modal` and `DrawerModal` components are opened the main window should be inert, but the user can scroll the main window.

**Steps to reproduce:**
1) Navigate to https://zendeskgarden.github.io/react-components/modals/
2) Turn on Bedrock CSS
3) Open a `Modal` or `DrawerModal`
4) Trigger some scroll events; **the main window behind the backdrop scrolls, but it shouldn't**

## Detail

The Bedrock CSS that normalizes CSS in the Garden ecosystem applies an `overflow-y:scroll;` to the `html` element. In some cases like modal components, the `overflow-y:scroll;` should be overwritten so that inert windows behind an open modal dialog cannot be scrolled. 

This PR proposes to modify overflow CSS for the `html` element instead of the `body` element. By doing so, the `overflow` from Bedrock CSS is correctly overwritten when a modal is opened.

## Screenshots

**Before:** when Bedrock CSS is turned on, the user **can** scroll an inert window behind open modals

----

**After:** when Bedrock CSS is turned on, the user **cannot** scroll an inert window behind open modals

## Checklist

### DIFF
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
@@ -113,34 +113,17 @@ export const DrawerModal = forwardRef<
       }
 
       const htmlElement = environment.querySelector('html');
-      const bodyElement = environment.querySelector('body');
-      let previousHtmlOverflowY: string;
-      let previousBodyOverflow: string;
-
-      if (bodyElement && isOpen) {
-        previousBodyOverflow = bodyElement.style.overflow;
-
-        bodyElement.style.overflow = 'hidden';
-      }
+      let previousHtmlOverflow: string;
 
       if (htmlElement && isOpen) {
-        previousHtmlOverflowY = htmlElement.style.overflowY;
+        previousHtmlOverflow = htmlElement.style.overflow;
 
-        // Safari treats overflowY differently than other browsers
-        if (navigator.userAgent.indexOf('Safari') > -1) {
-          htmlElement.style.overflowY = 'initial';
-        } else {
-          htmlElement.style.overflowY = 'hidden';
-        }
+        htmlElement.style.overflow = 'hidden';
       }
 
       return () => {
-        if (bodyElement) {
-          bodyElement.style.overflow = previousBodyOverflow;
-        }
-
         if (htmlElement) {
-          htmlElement.style.overflowY = previousHtmlOverflowY;
+          htmlElement.style.overflow = previousHtmlOverflow;
         }
       };
     }, [environment, isOpen]);
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
@@ -122,7 +122,7 @@ export const DrawerModal = forwardRef<
       }
 
       return () => {
-        if (htmlElement) {
+        if (htmlElement && isOpen) {
           htmlElement.style.overflow = previousHtmlOverflow;
         }
       };
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.tsx
@@ -112,19 +112,37 @@ export const DrawerModal = forwardRef<
         return undefined;
       }
 
+      const htmlElement = environment.querySelector('html');
       const bodyElement = environment.querySelector('body');
+      let previousHtmlOverflowY: string;
+      let previousBodyOverflow: string;
 
       if (bodyElement && isOpen) {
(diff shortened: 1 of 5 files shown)


packages/modals/src/elements/Modal.tsx
```diff
@@ -138,17 +141,32 @@ export const Modal = React.forwardRef<HTMLDivElement, IModalProps>(
...
         }
 
-        const previousBodyOverflow = bodyElement.style.overflow;
+        previousBodyOverflow = bodyElement.style.overflow;
 
         bodyElement.style.overflow = 'hidden';
+      }
+
+      if (htmlElement) {
+        previousHtmlOverflowY = htmlElement.style.overflowY;
```
jzempel, line 150: Overflow needs to be `hidden` on both axes, not just `y`.

packages/modals/src/elements/Modal.tsx
```diff
@@ -138,17 +141,32 @@ export const Modal = React.forwardRef<HTMLDivElement, IModalProps>(
           bodyElement.style.paddingRight = `${bodyPaddingRight + scrollbarSize}px`;
         }
 
-        const previousBodyOverflow = bodyElement.style.overflow;
+        previousBodyOverflow = bodyElement.style.overflow;
 
         bodyElement.style.overflow = 'hidden';
```
jzempel, line 146: Why is `body` overflow management being retained?

jzempel (changes requested):

packages/modals/src/elements/Modal.tsx
hzhu: Oh wow, it doesn't and the solution can be simplified a lot. Thanks!

jzempel (approved):
This maps to how the 1.0 version of modals was developed:

```js
  componentDidUpdate(prevProps) {
    const { hidden } = this.props;
    const { hidden: prevHidden } = prevProps;


    if (!hidden && prevHidden) {
      document.querySelector("html").style.overflow = "hidden";
    } else if (hidden && !prevHidden) {
      document.querySelector("html").style.overflow = "";
      this.tabJail = null;
    }
  }
```

...not sure why it was changed in version 2.0.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -112,19 +112,20 @@ export const DrawerModal = forwardRef<
...
-        return () => {
-          bodyElement.style.overflow = previousBodyOverflow;
-        };
+        htmlElement.style.overflow = 'hidden';
       }
 
-      return undefined;
+      return () => {
+        if (htmlElement) {
+          htmlElement.style.overflow = previousHtmlOverflow;
```
austingreendev, line 126: Could this unintentionally remove inline overflow styling on the `html`? The `return undefined` was intentional I think

austingreendev (commented):
@jzempel I think the original `html` -> `body` overflow change was due to iframed areas that were using `Modal`. Before we allowed `document` to be provided by the `ThemeProvider` this was a bigger issue.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
hzhu: > The return undefined was intentional I think

I figured it was intentionally added because the project has an ESLint rule that disallows implicit returns on arrow functions. 

> Could this unintentionally remove inline overflow styling on the html?

I don't think so, do you see a scenario where that might happen?

austingreendev: Since `previousHtmlOverflow` is undefined when the DrawerModal is closed:
1. Include `DrawerModal` in your page
2. Unmount the `DrawerModal` without applying the `isOpen` prop
3. The `useEffect` return statement is called during unmount. Since `htmlElement` is defined it will apply `htmlElement.style.overflow = undefined;`

Wouldn't this wipe out any inline overflow styling?

austingreendev: The purpose of the return is to undo the `hidden` styling applied when `isOpen=true`. Resetting it when `isOpen=false` doesn't feel right.

hzhu: > Since htmlElement is defined it will apply htmlElement.style.overflow = undefined;

I think it actually applies `previousHtmlOverflow` which is whatever the overflow styling was prior so there shouldn't be any noticeable difference from a user perspective. 

However, I added a new commit to only run the effect when the drawer modal `isOpen={true}` since that's the only time the cleanup is necessary 👍 .

austingreendev (approved):

PR #895, opened 2020-10-14
chore(docs): migrate react-typography to storybook

hzhu:
## Description

This PR migrates the `react-typography` package demos from Styleguidist to Storybook.

## Checklist


austingreendev (approved):
LGTM!

Would it be useful to nest the Type-related content under it's own folder? Would order XS->XXXL correctly

packages/typography/README.md
```diff
@@ -14,7 +14,7 @@ npm install react react-dom prop-types styled-components @zendeskgarden/react-th
 
 ## Usage
 
-```jsx static
+```jsx
```
jzempel, line 17: oy, this README needs some love. A job for another day.

packages/typography/stories/13-OrderedList.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    'OrderedList.Item': OrderedList.Item
+  }
+} as Meta;
+
+export const Default: Story = ({ size, type }) => (
+  <Grid>
+    <Row>
+      <Col textAlign="center">
+        <OrderedList size={size} type={type} style={{ display: 'inline-block' }}>
+          <OrderedList.Item>Garden trowel</OrderedList.Item>
```
jzempel, line 26: There is an indentation problem with this first item. Not sure what needs to change – could be the parent `inline-block` is a problem?

packages/typography/stories/14-UnorderedList.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+} as Meta;
+
+export const Default: Story = ({ size, type }) => (
+  <Grid>
+    <Row>
+      <Col textAlign="center">
+        <UnorderedList size={size} type={type} style={{ display: 'inline-block' }}>
+          <UnorderedList.Item>Plant</UnorderedList.Item>
+          <UnorderedList.Item>Water</UnorderedList.Item>
```
jzempel, line 27: The indentation problem appears on the first two items. Again, not sure what's going on.

packages/typography/stories/2-Code.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+);
+
+Default.argTypes = {
+  hue: {
+    control: { type: 'select', options: ['grey', 'red', 'green', 'yellow'] }
+  },
+  size: {
+    control: { type: 'select', options: ['inherit', 'small', 'medium', 'large'] }
+  }
+};
```
jzempel, line 37: I'm curious why some controls are defaulting and others (under lists, for example) indicate the initial "Select..."?

packages/typography/stories/4-Ellipsis.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+          Veggies es bonus vobis, proinde vos postulo essum magis.
+        </Ellipsis>
+      </Col>
+    </Row>
+  </Grid>
+);
+
+Default.argTypes = {
+  title: {
+    control: 'text'
```
jzempel, line 32: Is there a way to override the placeholder? If so, something like "Override the title attribute" would be helpful. Otherwise, it feels like this is controlling the Ellipsis children text.

Also, clearing the title control should restore the default `title`.

jzempel (changes requested):
Per Austin's comment, let's put `<SM>` through `<XXXL>` components under a "Type scale" section.

packages/typography/stories/4-Ellipsis.stories.tsx
hzhu: I don't think there's a way to override the placeholder as it looks like it's [hard coded](https://github.com/storybookjs/storybook/blob/next/lib/components/src/controls/Text.tsx#L25). However, I've updated the Story to use the `Default.args.` The input is prepopulated with the title and when the input is cleared, the default `title` is now restored.

packages/typography/stories/2-Code.stories.tsx
hzhu: I think this is due to this component rendering in Storybook using the `subcomponents` options. It does seem strange to not render the default props and controls on the initial tab. 

I've updated the `export default` for this Story to _not_ use the `subcomponents` attribute. I think the controls under lists should now show default props and use them for initial control values.

jzempel (approved):

austingreendev (approved):

PR #897, opened 2020-10-16
feat(theming): provide scoping <div> knock-out for ThemeProvider

jzempel:
## Description

Items of note for this PR:

- Adds support for `focusVisibleRef={null}` to prevent render of `:focus-visible` scoping div default
- Adds associated spec tests
- Fills out ThemeProvider prop documentation in prep for website.
- Restores `ThemeProvider` prop sheet to the example page

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/theming/src/elements/ThemeProvider.tsx
+++ b/packages/theming/src/elements/ThemeProvider.tsx
@@ -25,7 +25,7 @@ interface IGardenThemeProviderProps extends Partial<ThemeProps<DefaultTheme>> {
    * Assigning `null` (on a nested `ThemeProvider`, for example) prevents the
    * added polyfill and scoping `<div>`.
    */
-  focusVisibleRef?: React.RefObject<HTMLElement>;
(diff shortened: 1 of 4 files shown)


packages/theming/src/elements/ThemeProvider.tsx
```diff
@@ -13,7 +13,19 @@ import DEFAULT_THEME from './theme';
 import { useDocument } from '../utils/useDocument';
 
 interface IGardenThemeProviderProps extends Partial<ThemeProps<DefaultTheme>> {
-  focusVisibleRef?: React.RefObject<HTMLElement>;
+  /**
+   * Provides values for component styling. See styled-components
+   * [`ThemeProvider`](https://styled-components.com/docs/api#themeprovider)
+   * for details.
+   */
+  theme?: DefaultTheme;
```
jzempel, line 21: This repeats the definition from the extended interface. But without it, we aren't able to document the prop. Therefore, I think it's imperative in this one case to make an exception and include it.

austingreendev (approved):

PR #898, opened 2020-10-16
chore(pagination): get theme from context

hzhu:
## Description

This PR refactors the `Pagination` to receive the `theme` via context.

## Detail

I noticed that the prop type descriptions were not rendering in Storybook while migrating `Pagination` from Styleguidist to Storybook. It looks like the prop type and description information was being swallowed since the `Pagination` was wrapped in a `withTheme` higher-order component.

This situation felt like a good opportunity to refactor `Pagination` to use a more modern approach to getting the `theme` value, which also should allow Storybook to properly display the prop type and description text.

## Checklist

### DIFF
--- a/packages/pagination/.size-snapshot.json
+++ b/packages/pagination/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 27585,
-    "minified": 18321,
-    "gzipped": 4455
+    "bundled": 27548,
+    "minified": 18289,
+    "gzipped": 4454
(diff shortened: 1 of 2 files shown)


austingreendev:
@jzempel this is mentioned in the v4 and v5 documentation https://styled-components.com/docs/advanced#via-usecontext-react-hook

austingreendev (approved):

jzempel (approved):
Awesome – I was looking in the wrong place.

austingreendev (approved):

PR #905, opened 2020-10-23
chore(docs): migrate react-pagination examples to storybook

hzhu:
## Description

This PR migrates the `react-pagination` examples to Storybook.

## Checklist


packages/pagination/stories/2-Pagination.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+Default.args = {
+  currentPage: 1,
+  totalPages: 11,
+  pagePadding: 2,
+  pageGap: 2
+};
+
+Default.argTypes = {
+  currentPage: {
+    control: { type: 'number' }
```
hzhu, line 44: I went with a number control here because total pages can be changed to 1000 or 100000. It would allow a user to navigate to a specific page, e.g. 999 by typing it out. It also allows users to have negative numbers but that doesn't reflect in the UI and I thought that would be okay.

austingreendev (approved):

packages/pagination/stories/2-Pagination.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const Default: Story = ({ currentPage, totalPages, pagePadding, pageGap }) => (
+  <Grid>
+    <Row>
+      <Col textAlign="center">
+        <Pagination
+          totalPages={totalPages}
+          pagePadding={pagePadding}
+          pageGap={pageGap}
+          currentPage={currentPage}
```
jzempel, line 27: I don't think we're going to want to see `currentPage` controlled via knobs in this example. Maybe `Initial currentPage`? Otherwise, the example should control `currentPage` via `onChange` as seen with the current example: https://zendeskgarden.github.io/react-components/pagination/#basic

packages/pagination/stories/3-CursorPagination.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <CursorPagination.Previous onClick={action('onClick')} disabled={disabled}>
+            Previous
+          </CursorPagination.Previous>
+          <CursorPagination.Next onClick={action('onClick')} disabled={disabled}>
+            Next
+          </CursorPagination.Next>
+          <CursorPagination.Last onClick={action('onclick')} disabled={disabled}>
+            Last
+          </CursorPagination.Last>
+        </CursorPagination>
```
jzempel, line 36: The current examples seem like a more effective visual test: https://zendeskgarden.github.io/react-components/pagination/#basic

jzempel (changes requested):

packages/pagination/stories/3-CursorPagination.stories.tsx
hzhu: I've updated this Story to mirror the current website example because I wanted to keep the Story minimal to the component. If the dots are helpful, I can add those too.

packages/pagination/stories/3-CursorPagination.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default {
+  title: 'Components/Pagination/CursorPagination',
+  component: CursorPagination
+} as Meta;
+
+export const Default: Story = () => {
+  const [cursor, setCursor] = useState(0);
+
+  const pages = [0, 1, 2, 3, 4];
```
jzempel, line 21: Not seeing where this array is leveraged. Could it just be `const pageLength = 4`?

jzempel (approved):

packages/pagination/stories/3-CursorPagination.stories.tsx
hzhu: Updated, thanks!

hzhu: Updated, thanks!

PR #925, opened 2020-11-04
feat(typography): support anchor-styled code snippets

jzempel:
## Description

I have good news and bad news. Good news, this is a very simple styling tweak. Bad news, Garden's default anchor color doesn't meet AA contrast requirements over a `<Code>` element. But I've opened the PR in order to kick off a design discussion.

## Checklist

      designer as a reviewer)


allisonacs:
Well the good news is, the color system is working as spec'd: `600s` are only guaranteed to meet contrast ratios against `100s`, but the bad news is, we've got some `200` backgrounds now. 😐

allisonacs:
I might have an easy fix here. The type in `code` components is `700`, so maybe we shift the link color to the `700` shade as well. That would technically bork our hover/active states, but they could be shifted darker as well, and `getColor` could get us a `900` rendering…

jzempel:
Cool. This pushes us past the super-simple `color: inherit` solution. But I'll code it up based on `700` and we can see how it works out.

allisonacs:
> Cool. This pushes us past the super-simple `color: inherit` solution. But I'll code it up based on `700` and we can see how it works out.

Yea 👎 but short of reconsidering the background colors, I'm not sure what else we can do.

jzempel:
@allisonacs once the new commit deploys, you'll be able to checkout the use of darker blue colors. The good news is that there is AA compliance, even across the `hue` changes (use the storybook "Controls" tab to update). The bad news (for me) is that the color differentiation between regular and linked `<code>` is nearly imperceptible.

allisonacs (commented):
Seeing the example, I agree, link color vs. text color is undetectable. Both are `700` values, so it's understandable. I looked at adding an underline, but that's very off system. 

My recommendation is: 
- revert to `600` color. this is an extremely rare use case
- we should issue guidance to avoid `code` + `link` combos, both on our own site and across Zendesk
- we include this use case when we address our link formatting issue globally

allisonacs (approved):

austingreendev (approved):

hzhu (approved):

PR #936, opened 2020-11-16
fix(loaders): fix Dot loader for IE11

hzhu:
## Description

This pull request fixes the `Dot` loader animation in Internet Explorer 11.

## Detail

It looks like `useCSSSVGAnimation` is unreliable in determining if an SVG can be animated. **In both Chrome 86 and Internet Explorer, `useCSSSVGAnimation` returns the same result**. This is unexpected because Internet Explorer does not support SVG animations.

This PR removes `useCSSSVGAnimation` and replaces it with a working SVG animation check. It checks the `transform` property of the `svg` element to determine if SVG can be animated. In modern Chrome, Safari, Firefox, and Edge this check returns `true` indicating animation is supported. In IE11, the check returns `false` indicating the animation is not supported.

By using the correct boolean check, the `Dot` loader animates correctly across all browsers that are supported by Garden.

## Verification Steps

❌ **Before:** 
1) Open Internet Explorer 11
2) Navigate to the `Dot` loader Storybook example on the production site [here](https://zendeskgarden.github.io/react-components/storybook/?path=/story/components-loaders-dots--default)
3) The `Dot` loader **does not** animate  

✅ **After:** 
1) Open Internet Explorer 11
2) Navigate to the `Dot` loader Storybook example on the "View Deployment" site [here](https://5fb2e3baa4e39f00c772a999--zendeskgarden-react-components.netlify.app/storybook/?path=/story/components-loaders-dots--default)
3) The `Dot` loader **does** animate 

## Screenshots

❌ **Before:**

✅ **After:** 

## Checklist


packages/loaders/src/elements/Dots.tsx
```diff
@@ -38,15 +37,17 @@ export interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
   delayMS?: number;
 }
 
+const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
```
jzempel, line 40: Shouldn't `document` be accessed off the `theme` context? See https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/Modal.tsx

jzempel (approved):

packages/loaders/src/elements/Dots.tsx
```diff
@@ -39,14 +40,23 @@ export interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
...
-  const { delayComplete } = useSchedule({ duration, delayMS, loop: noAnimatedSVGSupport });
+  const theme = useContext(ThemeContext);
+  const environment = useDocument(theme);
+  const canTransformSVG = useRef<boolean | null>(null);
+
+  if (environment && canTransformSVG.current === null) {
+    const svg = environment.createElementNS('http://www.w3.org/2000/svg', 'svg');
+
+    canTransformSVG.current = 'transform' in svg;
+  }
```
austingreendev, line 51: For this to support SSR (and the website) wouldn't this need to be in a `useLayoutEffect` to time the calculation correctly?

packages/loaders/src/elements/Dots.tsx
```diff
@@ -39,14 +40,23 @@ export interface IDotsProps extends React.HTMLAttributes<SVGSVGElement> {
 }
 
 const Dots: React.FC<IDotsProps> = ({ size, color, duration, delayMS, ...other }) => {
-  const noAnimatedSVGSupport = useCSSSVGAnimation();
-  const { delayComplete } = useSchedule({ duration, delayMS, loop: noAnimatedSVGSupport });
+  const theme = useContext(ThemeContext);
+  const environment = useDocument(theme);
+  const canTransformSVG = useRef<boolean | null>(null);
```
austingreendev, line 45: For this to update the effect below, should this be `useState`? It could then be included as a dependency for `useEffect`.

austingreendev (changes requested):

packages/loaders/src/elements/Dots.tsx
hzhu: There should not be a need to sync this value with the effect below because the effect runs on each render.

packages/loaders/src/elements/Dots.tsx
hzhu: I tested the updated `Dot` component on the Garden website (and vanilla Next.js) app and it seems to work as expected.

Should SSR avoid `useLayoutEffect` completely? From the [docs](https://reactjs.org/docs/hooks-reference.html#uselayouteffect):

> If you use server rendering, keep in mind that neither useLayoutEffect nor useEffect can run until the JavaScript is downloaded. This is why React warns when a server-rendered component contains useLayoutEffect. To fix this, either move that logic to useEffect (if it isn’t necessary for the first render), or delay showing that component until after the client renders (if the HTML looks broken until useLayoutEffect runs).

austingreendev: Oh nice, haven't seen this warning before. I would revise my comment to just suggest `useEffect` then.

Did you run a full production build for the Website? [useDocument should return undefined originally](https://github.com/zendeskgarden/react-components/blob/main/packages/theming/src/utils/useDocument.ts)

packages/loaders/src/elements/Dots.tsx
austingreendev: Oh cool. It may be useful to still place the original calculation in an effect then. 

The component doesn't need to calculate `canTransformSVG` on every render. An environment shouldn't change this ability after the first calculation.

hzhu: Does the `null` check on line 47 prevent unnecessary calculations?

```jsx
if (environment && canTransformSVG.current === null) {
  const svg = environment.createElementNS('http://www.w3.org/2000/svg', 'svg');

  // Once ref is set to true or false, additional renders will not hit this branch again.
  canTransformSVG.current = 'transform' in svg;
}
```

packages/loaders/src/elements/Dots.tsx
hzhu: I only ran development builds of the website and Next.js app, but I can run a full production build for a sanity check.

It seems that wrapping this code in a `useEffect` is beneficial to prevent unnecessary calculations of creating an SVG on additional renders. But I think the `null` check already achieves this. It's why I opted for storing the boolean check into a `useRef`.

austingreendev: LGTM!

packages/loaders/src/elements/Dots.tsx
austingreendev: LGTM!

austingreendev (approved):

PR #942, opened 2020-11-18
fix(dropdowns): ensure Tab key selects highlighted items

austingreendev:
## Description

This PR enables the item selection in all Dropdown components using the `tab` key. This logic is similar to the current "filter" ability in `Menu` and `Select`.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -19,21 +19,21 @@
     }
   },
   "index.cjs.js": {
-    "bundled": 86463,
-    "minified": 53713,
-    "gzipped": 11012
+    "bundled": 87671,
+    "minified": 54235,
+    "gzipped": 11074
   },
   "index.esm.js": {
-    "bundled": 80062,
-    "minified": 48019,
-    "gzipped": 10758,
+    "bundled": 81188,
+    "minified": 48459,
+    "gzipped": 10824,
     "treeshaked": {
       "rollup": {
-        "code": 37827,
+        "code": 38234,
         "import_statements": 792
       },
       "webpack": {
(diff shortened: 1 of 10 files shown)


jzempel (commented):
Along with selection, it feels like the use of `tab` should navigate to the next/prev tabindex. In this way `tab` is slightly different than `enter`. See https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html for reference.

@allisonacs what do you think?

austingreendev:
To allow for portal usage, we return focus back to triggering elements with the `onClose` event. This is tied into Downshift logic and I'm not sure it's something we can differentiate between at this time.

allisonacs (commented):
Yea, I'm with @jzempel on this one. When we look at the ARIA spec example, we observe `tab` moving us to the next item in the tab order (or `shift + tab` to the go to the previous item in the tab order).

@austingreendev you sure you can't work your magic?

austingreendev:
We may be able to, but it will require some larger changes. The [current logic in all trigger elements](https://github.com/zendeskgarden/react-components/blob/main/packages/dropdowns/src/elements/Trigger/Trigger.tsx#L45-L57) only has access to the `isOpen` state. We don't have a way of know _how_ `isOpen` was toggled since that abstraction is in Downshift.

Definitely something that we should figure out how to do, but it may be safer to treat it as a separate issue since it's not a regression.

jzempel (approved):
Ok, after added offline discussion, I think this PR is effective for addressing the issue at hand. The desired user `tab` experience will need to be revisited along with proposed API changes for handling focus in v9.

PR #946, opened 2020-11-24
fix(forms): Auto resize textareas on visibility changes

li-kai:
## Description

If the textarea component mounts and its container is hidden with css properties such as `display: none`, the component will not auto resize correctly.

## Detail

You reproduce the bug by untoggling the "Hide form" checkbox. I tried to reproduce this in Storybook but was unable to due to re-rendering caused by Storybook's remounting behavior.
https://codesandbox.io/s/auto-resize-textarea-bug-7ucf2?file=/src/index.tsx

This fix attaches the visibility to the IntersectionObserver, which will make it resize when its visibility changes.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/.size-snapshot.json
+++ b/packages/forms/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 126094,
-    "minified": 83983,
-    "gzipped": 15911
+    "bundled": 126372,
+    "minified": 84080,
+    "gzipped": 15971
   },
   "index.esm.js": {
-    "bundled": 118895,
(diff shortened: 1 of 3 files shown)


austingreendev:
I believe `IntersectionObserver` is [not implemented in IE11]( https://caniuse.com/?search=IntersectionObserver) which is still a requirement for some of our consumers.

Let's [open an issue](https://github.com/zendeskgarden/react-components/issues/new) and we can discuss some of the different ways to approach this one.

jzempel:
Closing for now, as it seems this PR has stalled. Please re-open if necessary.

PR #953, opened 2020-12-15
chore(buttons): update prop descriptions

Mikaelia:
## Description

Updates props for button components

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/buttons/src/elements/Anchor.tsx
+++ b/packages/buttons/src/elements/Anchor.tsx
@@ -13,8 +13,8 @@ export interface IAnchorProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
   /** Applies danger styling */
   isDanger?: boolean;
   /**
-   * Attaches `target="_blank"` and `rel="noopener noreferrer"` to ensure
-   * that an anchor that navigates to an external resource is a
+   * Attaches `target="_blank"` and `rel="noopener noreferrer"` to an anchor that
+   * navigates to an external resource. This ensures ensures that the anchor is a
    * safe [cross-origin destination link](https://web.dev/external-anchors-use-rel-noopener/)
    **/
   isExternal?: boolean;
--- a/packages/buttons/src/elements/Anchor.tsx
+++ b/packages/buttons/src/elements/Anchor.tsx
@@ -13,7 +13,6 @@ export interface IAnchorProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
   /** Applies danger styling */
   isDanger?: boolean;
   /**
-   *
    * Attaches `target="_blank"` and `rel="noopener noreferrer"` to ensure
    * that an anchor that navigates to an external resource is a
    * safe [cross-origin destination link](https://web.dev/external-anchors-use-rel-noopener/)
@@ -22,6 +21,7 @@ export interface IAnchorProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
 }
 
 /**
+ * Accepts all `<a>` props
  *
  * @component
  * @name Anchor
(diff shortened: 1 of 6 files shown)


packages/buttons/src/elements/Anchor.tsx
```diff
@@ -10,12 +10,12 @@ import PropTypes from 'prop-types';
...
-  /** Apply danger styling */
+  /** Applies danger styling */
   isDanger?: boolean;
   /**
-   * Used when the anchor navigates to an external resource. Applies `target="_blank"`
-   * along with `rel="noopener noreferrer"` to ensure safe
-   * [cross-origin destination links](https://web.dev/external-anchors-use-rel-noopener/).
+   * Attaches `target="_blank"` and `rel="noopener noreferrer"` to ensure
+   * that an anchor that navigates to an external resource is a
+   * safe [cross-origin destination link](https://web.dev/external-anchors-use-rel-noopener/)
```
jzempel, line 18: This is moving in the right direction, but it still feels a bit clunky. Have you tried http://www.hemingwayapp.com/ for improving readability? It might be that this just needs to be broken up into two sentences.

packages/buttons/src/elements/ToggleButton.tsx
```diff
@@ -11,17 +11,16 @@ import Button, { IButtonProps } from './Button';
 
 export interface IToggleButtonProps extends IButtonProps {
   /**
-   * Determine if the button is pressed or not. Use `'mixed'` to indicate
-   * whether the toggle controls other elements which do not share the same
-   * value.
+   * Determines if the button is pressed. Use "mixed" to indicate that
+   * the toggle controls other elements which do not share the same value
```
jzempel, line 15: Per content [guidelines](https://garden.zendesk.com/content/punctuation#period) the second sentence should end with a period.

packages/buttons/src/elements/ToggleIconButton.tsx
```diff
@@ -11,17 +11,16 @@ import IconButton, { IIconButtonProps } from './IconButton';
 
 export interface IToggleIconButtonProps extends IIconButtonProps {
   /**
-   * Determine if the icon button is pressed or not. Use `'mixed'` to indicate
-   * whether the toggle controls other elements which do not share the same
-   * value.
+   * Determines if the button is pressed. Use "mixed" to indicate that the
+   * toggle controls other elements which do not share the same value
```
jzempel, line 15: End with a period.

jzempel (changes requested):

packages/buttons/src/elements/Anchor.tsx
Mikaelia: I did try it with this one, but never broken into two sentences. 

How about: 
Attaches `target="_blank"` and `rel="noopener noreferrer"` to an anchor that navigates to an external resource. This ensures that the anchor is a safe cross-origin destination link.

jzempel (approved):

packages/buttons/src/elements/Anchor.tsx
jzempel: Yeah, much better. Thanks.

austingreendev (approved):

jzempel:
@Mikaelia it looks like this one missed [ChevronButton](https://garden.zendesk.com/components/split-button#chevronbutton). Please check against the package root [index.ts](https://github.com/zendeskgarden/react-components/blob/main/packages/buttons/src/index.ts) as you go through these. 🙏

Mikaelia:
@jzempel I missed this one because it did not have an interface for props, where I have been seeing prop descriptions used. Is there a different way that I should be documenting files that don't have this?

jzempel:
doh, please don't mind me. All these props are inherited and will get your updated `IconButton` descriptions. 🤦 My bad.

PR #954, opened 2020-12-16
chore(chrome): update prop descriptions

Mikaelia:
## Description
Updates chrome package element prop docs

### DIFF
--- a/packages/chrome/src/elements/Chrome.tsx
+++ b/packages/chrome/src/elements/Chrome.tsx
@@ -14,7 +14,7 @@ import { ChromeContext } from '../utils/useChromeContext';
 import { StyledChrome } from '../styled';
 
 interface IChromeProps extends HTMLAttributes<HTMLDivElement> {
-  /** Applies a custom hue to chrome navigation */
+  /** Applies a custom hue to the chrome navigation */
   hue?: string;
   /** Prevents fixed positioning from being applied to the <html> element */
   isFluid?: boolean;
--- a/packages/chrome/src/elements/Chrome.tsx
+++ b/packages/chrome/src/elements/Chrome.tsx
@@ -14,9 +14,9 @@ import { ChromeContext } from '../utils/useChromeContext';
 import { StyledChrome } from '../styled';
 
 interface IChromeProps extends HTMLAttributes<HTMLDivElement> {
-  /** Apply a custom hue to chrome navigation */
+  /** Applies a custom hue to chrome navigation */
   hue?: string;
-  /** Prevent fixed positioning from being applied to the <html> element */
+  /** Prevents fixed positioning from being applied to the <html> element */
   isFluid?: boolean;
 }
 
--- a/packages/chrome/src/elements/Chrome.tsx
+++ b/packages/chrome/src/elements/Chrome.tsx
@@ -16,7 +16,7 @@ import { StyledChrome } from '../styled';
 interface IChromeProps extends HTMLAttributes<HTMLDivElement> {
   /** Applies a custom hue to the chrome navigation */
   hue?: string;
-  /** Prevents fixed positioning from being applied to the <html> element */
+  /** Prevents fixed positioning from being applied to the `<html>` element */
   isFluid?: boolean;
 }
 
--- a/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
+++ b/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
@@ -18,9 +18,9 @@ import {
 
 export interface ICollapsibleSubNavItemProps
   extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'> {
-  /** Sets the item header */
+  /** Sets the item's section header */
   header?: React.ReactNode;
-  /** Reveals the item text */
+  /** Reveals the item's section */
   isExpanded?: boolean;
   /**
    * Handles changes in the item's expansion state
--- a/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
+++ b/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
@@ -18,8 +18,15 @@ import {
 
 export interface ICollapsibleSubNavItemProps
   extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'> {
+  /** Sets the item header */
   header?: React.ReactNode;
+  /** Reveals the item text */
   isExpanded?: boolean;
+  /**
+   * Handles changes in the item's expansion state
+   *
+   * @param {boolean} isExpanded An item's expansion state
+   */
   onChange?: (isExpanded: boolean) => void;
 }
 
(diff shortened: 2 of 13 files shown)


austingreendev (approved):

packages/chrome/src/elements/Chrome.tsx
```diff
@@ -14,9 +14,9 @@ import { ChromeContext } from '../utils/useChromeContext';
 import { StyledChrome } from '../styled';
 
 interface IChromeProps extends HTMLAttributes<HTMLDivElement> {
-  /** Apply a custom hue to chrome navigation */
+  /** Applies a custom hue to the chrome navigation */
   hue?: string;
-  /** Prevent fixed positioning from being applied to the <html> element */
+  /** Prevents fixed positioning from being applied to the <html> element */
```
jzempel, line 19: `<html>` should be backtick'd.

jzempel (changes requested):

jzempel (approved):

jzempel (commented):
Oops, @Mikaelia, I realized there are a _bunch_ more top-level [element](https://github.com/zendeskgarden/react-components/tree/main/packages/chrome/src/elements)-level components to document for `chrome`. See https://garden.zendesk.com/components/chrome#api for details. Were you planning to do those here or under separate PR?

Mikaelia:
@jzempel Oops! Somehow I had it in my mind that I wasn't supposed to document SubNav. I'll get that fixed ASAP. Looking back at the code I do have a couple of questions though:

Footer -> I did not see any props to document
Header -> I see a prop `isStandalone` proptype but not the interface for it. Same with HeaderItemText for `isClipped` and SubNavItemText..are these props that I should still document?

jzempel:
Yes, there are a list of components that do not have any props. `Header` is unique in that it inherits its prop (and description) from [StyledHeader](https://github.com/zendeskgarden/react-components/blob/c46923f94061cc84cbfffab9adf2be3cdb76ec2a/packages/chrome/src/styled/header/StyledHeader.ts#L15-L18). Same for the others you mentioned. So you'll need to dip into the `styled` directory in a few cases to spruce up the JSDoc. The trick is to find the origin for all component props listed under https://garden.zendesk.com/components/chrome#api (`CollapsibleSubNavItem`, etc).

packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
```diff
@@ -18,8 +18,15 @@ import {
 
 export interface ICollapsibleSubNavItemProps
   extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'> {
+  /** Sets the item header */
   header?: React.ReactNode;
+  /** Reveals the item text */
```
jzempel, line 23: Replace ~item text~ with something like "item's content" or "sub-items" (not sure that's a term we want to use 😝 ) or, similar to [Accordion](https://zendeskgarden.github.io/react-components/storybook/?path=/docs/components-accordions-accordion--uncontrolled) "item's section" (as this is technically an accordion control).

See the demo at https://garden.zendesk.com/components/chrome#navigation-panel with "show subnav" activated for additional working details for a `CollapsibleSubNavItem`.

packages/chrome/src/styled/header/StyledBaseHeaderItem.ts
```diff
@@ -15,17 +15,11 @@ import {
 const COMPONENT_ID = 'chrome.base_header_item';
 
 export interface IStyledBaseHeaderItemProps {
-  /**
-   * Horizontally maximize a flex item in the header to take as much space as possible (i.e. breadcrumb container)
-   **/
+  /** Maximizes the width of a flex item in the header (i.e. breadcrumb container) */
```
jzempel, line 18: We should remove ~(i.e. breadcrumb container)~ as design research has determined planting a breadcrumb in the header is off-pattern.

packages/chrome/src/styled/header/StyledBaseHeaderItem.ts
```diff
@@ -15,17 +15,11 @@ import {
...
   maxX?: boolean;
-  /**
-   * Vertically maximize the height for a header item (i.e. contains a search input)
-   **/
+  /** Maximizes the height of the item (i.e. contains a search input) */
   maxY?: boolean;
-  /**
-   * Round the border radius for a header item (i.e. user icon)
-   **/
+  /** Rounds the border radius of the item (i.e. user icon) */
```
jzempel, line 22: Per guidance from Content Design and https://garden.zendesk.com/components/avatar#how-to-use-it-well, I wonder if we should swap ~user~ with "person". Or does "person icon" just sound a bit weird for dev docs?

packages/chrome/src/styled/header/StyledHeaderItemText.ts
```diff
@@ -11,9 +11,7 @@ import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-the
 const COMPONENT_ID = 'chrome.header_item_text';
 
 export interface IStyledHeaderItemTextProps {
-  /**
-   * Clip text (but leave accessible to screenreaders) for an icon-only header item
-   **/
+  /** Hides item text. Text remains accessible to screenreaders. */
```
jzempel, line 14: ~screenreaders~ -> "screen readers". https://en.wikipedia.org/wiki/Screen_reader

packages/chrome/src/styled/header/StyledLogoHeaderItem.ts
```diff
@@ -21,9 +21,7 @@ import { getNavWidth } from '../nav/StyledNav';
 const COMPONENT_ID = 'chrome.header_item';
 
 export interface IStyledLogoHeaderItemProps {
-  /**
-   * Applies product-specific color palette
-   **/
+  /** Applies a product-specific color palette */
```
jzempel, line 24: I think this could be improved with "Applies a brand color to the product logo".

For the fussy details, the given name is used to dereference a [brand color](https://garden.zendesk.com/design/color#brand-colors) from the `product` portion of the [PALETTE](https://garden.zendesk.com/components/palette#palette) object. And then that color hex is inherited by `currentColor` used within one of the two-tone "relationshape" icons seen here: https://garden.zendesk.com/design/icons/26.

jzempel (changes requested):
Thanks for picking up these leftovers. Just a few more change requests from me.

packages/chrome/src/styled/header/StyledLogoHeaderItem.ts
jzempel: TBH, I regret having built this component 😅

Mikaelia: Ahh, I see. Thanks for the explanation!

packages/chrome/src/styled/header/StyledBaseHeaderItem.ts
Mikaelia: Do we need the bit inside the parenthesis? For me I'm not sure if it adds value as it seems pretty self-explanatory

jzempel (approved):

packages/chrome/src/styled/header/StyledBaseHeaderItem.ts
jzempel: Nope, good call.

PR #955, opened 2020-12-17
feat(tables): introduce read-only styling

austingreendev:
## Description

This PR introduces a new `isReadOnly` prop to the `Table` component. This prop removes the hover/focus styling and logic from `Row` elements when provided.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/tables/src/elements/Row.spec.tsx
+++ b/packages/tables/src/elements/Row.spec.tsx
@@ -176,4 +176,35 @@ describe('Row', () => {
       expect(getByTestId('row')).toHaveStyleRule('height', '64px');
     });
   });
+
+  describe('ReadOnly', () => {
+    it('removes tabIndex while in read-only mode', () => {
+      const { getByTestId } = render(
+        <Table isReadOnly>
+          <Body>
+            <Row data-test-id="row" />
+          </Body>
+        </Table>
+      );
+
+      expect(getByTestId('row')).not.toHaveAttribute('tabindex');
+    });
+
+    it('removes :hover styling while in read-only mode', () => {
+      const { getByTestId } = render(
+        <Table isReadOnly>
+          <Body>
+            <Row data-test-id="row" />
+          </Body>
+        </Table>
+      );
+
+      expect(getByTestId('row')).toHaveStyleRule('border-bottom-color', undefined, {
+        modifier: '&:hover'
+      });
+      expect(getByTestId('row')).toHaveStyleRule('background-color', undefined, {
+        modifier: '&:hover'
+      });
+    });
+  });
 });
--- a/packages/tables/src/elements/Row.spec.tsx
+++ b/packages/tables/src/elements/Row.spec.tsx
@@ -30,6 +30,18 @@ describe('Row', () => {
     expect(getByTestId('row')).toBe(ref.current);
   });
 
+  it('adds interactive tabIndex by default', () => {
+    const { getByTestId } = render(
+      <Table>
+        <Body>
+          <Row data-test-id="row" />
+        </Body>
+      </Table>
+    );
+
+    expect(getByTestId('row')).toHaveAttribute('tabindex', '-1');
+  });
+
   it('applies focus styling', () => {
     const { getByTestId } = render(
       <Table>
@@ -188,23 +200,6 @@ describe('Row', () => {
       );
 
       expect(getByTestId('row')).not.toHaveAttribute('tabindex');
-    });
-
-    it('removes :hover styling while in read-only mode', () => {
-      const { getByTestId } = render(
-        <Table isReadOnly>
-          <Body>
-            <Row data-test-id="row" />
-          </Body>
-        </Table>
-      );
-
-      expect(getByTestId('row')).toHaveStyleRule('border-bottom-color', undefined, {
-        modifier: '&:hover'
-      });
-      expect(getByTestId('row')).toHaveStyleRule('background-color', undefined, {
-        modifier: '&:hover'
-      });
     });
   });
 });
--- a/packages/tables/src/elements/Table.tsx
+++ b/packages/tables/src/elements/Table.tsx
@@ -10,14 +10,16 @@ import PropTypes from 'prop-types';
 import { StyledTable, IStyledTableProps } from '../styled';
 import { TableContext } from '../utils/useTableContext';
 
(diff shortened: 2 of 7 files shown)


packages/tables/src/elements/Row.spec.tsx
```diff
@@ -176,4 +176,35 @@ describe('Row', () => {
...
+          </Body>
+        </Table>
+      );
+
+      expect(getByTestId('row')).toHaveStyleRule('border-bottom-color', undefined, {
+        modifier: '&:hover'
+      });
+      expect(getByTestId('row')).toHaveStyleRule('background-color', undefined, {
+        modifier: '&:hover'
+      });
```
austingreendev, line 207: @jzempel @hzhu I'm unable to get the `modifier` values here to pick up `:hover` styling 😢  It's finding `:focus` fine though.

Do you all have any ideas on getting this to work?

jzempel: I've had challenges with these, as well. I think the `toHaveStyleRule` tests work best when they directly target the `StyledRow` type components, rather than the high-order wrapper.

hzhu: I think instead of picking up `undefined`, the `.not` API can be used to ensure the border styling is not applied on hover. One suggestion is something along the lines of:

```tsx
it.only('does not show :hover styling while in read-only mode', () => {
  const { getByTestId } = render(
    <Table isReadOnly>
      <Body>
        <Row data-test-id="row" />
      </Body>
    </Table>
  );

  const row = getByTestId('row');

  userEvent.hover(row);

  expect(row).not.toHaveStyleRule(
    'border-bottom-color',
    getColor('primaryHue', 200, DEFAULT_THEME),
    {
      modifier: '&:hover'
    }
  );
});
```

packages/tables/src/elements/Row.spec.tsx
```diff
@@ -176,4 +176,35 @@ describe('Row', () => {
...
+          <Body>
+            <Row data-test-id="row" />
+          </Body>
+        </Table>
+      );
+
+      expect(getByTestId('row')).not.toHaveAttribute('tabindex');
+    });
+
+    it('removes :hover styling while in read-only mode', () => {
```
hzhu, line 193: A :hover styling is never removed in this test (e.g. it never appears because the table is read-only). What do you think about rephrasing this description to something along the lines of:

> does not show :hover styling while in read-only mode

packages/tables/src/elements/Row.spec.tsx
```diff
@@ -176,4 +176,35 @@ describe('Row', () => {
       expect(getByTestId('row')).toHaveStyleRule('height', '64px');
     });
   });
+
+  describe('ReadOnly', () => {
```
hzhu, line 180: Would it be worth it to have tests asserting that rows _are_ interactable when the read-only prop isn't applied? Since that's all guarded by logic now?

packages/tables/src/elements/Row.spec.tsx
hzhu: And the other test case would be:

```tsx
it.only('shows :hover styling while in read-only mode', () => {
  const { getByTestId } = render(
    <Table>
      <Body>
        <Row data-test-id="row" />
      </Body>
    </Table>
  );

  const row = getByTestId('row');

  userEvent.hover(row);

  expect(row).toHaveStyleRule(
    'border-bottom-color',
    getColor('primaryHue', 200, DEFAULT_THEME),
    {
      modifier: '&:hover'
    }
  );
});
```

packages/tables/src/elements/Row.spec.tsx
austingreendev: That was going to be the next step once I was able to target the styles successfully. Looks like that isn't going to be possible unless we target the `StyledRow` directly. Going to remove this so we can get a passing deployment.

packages/tables/src/elements/Row.spec.tsx
austingreendev: When I tried running something similar to this locally `toHaveStyleRule` starts to produce errors that it can't understand the identifier provided. I also tried wrapping it in a `css` block but that didn't help either.

hzhu: These were running locally for me 🤔 .

packages/tables/src/elements/Table.tsx
```diff
@@ -10,14 +10,16 @@ import PropTypes from 'prop-types';
 import { StyledTable, IStyledTableProps } from '../styled';
 import { TableContext } from '../utils/useTableContext';
 
+interface ITableProps extends IStyledTableProps, HTMLAttributes<HTMLTableElement> {
+  /** Removes interactive styling from Table Rows */
```
jzempel, line 14: Here we go with lowercase, for consistency. So ~Table Rows~ -> "table rows". The exception would be a direct reference to "component X" which we shouldn't see much (or any) of within prop descriptions.

packages/tables/src/styled/StyledRow.ts
```diff
@@ -67,10 +68,24 @@ const colorStyles = (props: IStyledRowProps & ThemeProps<DefaultTheme>) => {
...
+      }};
+      background-color: ${() => {
+        if (props.isSelected) {
+          return hoveredSelectedBackgroundColor;
+        } else if (!props.isReadOnly) {
+          return hoveredBackgroundColor;
+        }
+
+        return undefined;
+      }};
```
jzempel, line 88: Can all of this logic be pulled up above into const for `hoverBorderColor` and `hoverBackgroundColor`? It should reduce the boolean logic and make the code more readable.

jzempel (changes requested):

ginnywood (approved):

packages/tables/src/elements/Table.tsx
austingreendev: Updated to be lowercase.

packages/tables/src/styled/StyledRow.ts
austingreendev: I've moved this logic into the existing conditional. Much cleaner now.

jzempel (approved):
Brilliant. Thanks for the efficient turnaround on this.

PR #958, opened 2020-12-18
chore(forms): update prop descriptions

Mikaelia:
## Description

Updates prop descriptions for forms components

### DIFF
--- a/packages/forms/src/elements/MediaInput.tsx
+++ b/packages/forms/src/elements/MediaInput.tsx
@@ -26,9 +26,9 @@ export interface IMediaInputProps extends InputHTMLAttributes<HTMLInputElement> 
   end?: any;
   /** Applies validation state styling */
   validation?: VALIDATION;
-  /** Applies props to the wrapping fauxinput element */
+  /** Applies props to the wrapping faux-input element */
   wrapperProps?: any;
-  /** Applies a ref to the wrapping fauxinput element */
+  /** Applies a ref to the wrapping faux-input element */
   wrapperRef?: any;
   /** @ignore */
   isFocused?: boolean;
--- a/packages/forms/src/elements/MediaInput.tsx
+++ b/packages/forms/src/elements/MediaInput.tsx
@@ -26,9 +26,9 @@ export interface IMediaInputProps extends InputHTMLAttributes<HTMLInputElement> 
   end?: any;
   /** Applies validation state styling */
   validation?: VALIDATION;
-  /** Applies props to the wrapping faux-input element */
+  /** Applies props to the wrapping [FauxInput](#fauxinput) element */
   wrapperProps?: any;
-  /** Applies a ref to the wrapping faux-input element */
+  /** Applies a ref to the wrapping [FauxInput](#fauxinput) element */
   wrapperRef?: any;
   /** @ignore */
   isFocused?: boolean;
--- a/packages/forms/src/elements/MediaInput.tsx
+++ b/packages/forms/src/elements/MediaInput.tsx
@@ -14,20 +14,21 @@ import useFieldContext from '../utils/useFieldContext';
 import { VALIDATION } from '../utils/validation';
 
 export interface IMediaInputProps extends InputHTMLAttributes<HTMLInputElement> {
-  /** Apply compact styling */
+  /** Applies compact styling */
   isCompact?: boolean;
-  /** Remove borders and padding */
+  /** Removes borders and padding */
   isBare?: boolean;
-  /** Apply inset `box-shadow` styling on focus */
+  /** Applies inset `box-shadow` styling on focus */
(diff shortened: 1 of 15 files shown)


packages/forms/src/elements/MediaInput.tsx
```diff
@@ -14,20 +14,21 @@ import useFieldContext from '../utils/useFieldContext';
...
-  /** Slot for "end" icon */
+  /** Accepts an "end" icon to display */
   end?: any;
+  /** Applies validation state styling */
   validation?: VALIDATION;
-  /** Apply props to the wrapping `FauxInput` element */
+  /** Applies props to the wrapping faux-input element */
   wrapperProps?: any;
-  /** Apply ref to the wrapping `FauxInput` element */
+  /** Applies a ref to the wrapping faux-input element */
```
jzempel, line 31: In these cases, it's best to keep `FauxInput`, as we're referring directly to the other component. Another option would (maybe preferable) would be to use:

```
[FauxInput](#fauxinput)
```

As a reference to the in-page API seen here: https://garden.zendesk.com/components/input#fauxinput

packages/forms/src/elements/MultiThumbRange.tsx
```diff
@@ -28,19 +28,25 @@ import { ThemeProps, DefaultTheme } from 'styled-components';
...
-  /** Apply disabled styling */
+  /** Indicates that the element is not interactive */
   disabled?: boolean;
-  /** Handler for processing change events */
+  /**
+   * Handles change events
+   *
+   * @param {Object} updatedValues The values that have changed
+   * @param {Object} minValue The optional minimum value
+   * @param {Object} maxValue the optional maximum value
```
jzempel, line 48: Syntax nits:

```
@param {number} [updatedValues.minValue] The optional minimum value
@param {number} [updatedValues.maxValue] The optional maximum value
```

packages/forms/src/elements/tiles/Tiles.tsx
```diff
@@ -23,13 +23,13 @@ import { Icon } from './components/Icon';
 import { Label } from './components/Label';
 
 interface ITilesProps extends HTMLAttributes<HTMLDivElement> {
-  /** Value of the selected radio button */
+  /** Sets the value of the selected radio button */
   value?: string;
-  /** Callback when a radio is selected */
+  /** Handles radio selection */
   onChange?: ChangeEventHandler<HTMLInputElement>;
```
jzempel, line 29: @austingreendev I hadn't spotted this `ChangeEventHandler` before. I suspect this is the signature all Garden component `onChange` events should take. And, as documented here by @mikaelagurney, I think it is reasonable to leave the signature implicit (no extra `@param` JSDoc) since the React construct is pervasive and obvious.

Thoughts?

packages/forms/src/elements/tiles/Tiles.tsx
```diff
@@ -23,13 +23,13 @@ import { Icon } from './components/Icon';
...
-  /** The name used to reference the value of the control. */
+  /** Sets the name used to reference the value of the control */
   name: string;
-  /** Displays the tiles in their vertical arrangement */
+  /** Arranges the tiles vertically */
```
jzempel, line 32: I think there is more helpful language that can be borrowed from here: https://garden.zendesk.com/components/tiles#layout. This is more about the content within the tile, rather than the layout of multiple tiles.

jzempel (changes requested):

packages/forms/src/elements/MediaInput.tsx
Mikaelia: Will do! I don't need the full link in there?

jzempel: No, we should optimize these for `website` usage. In fact, the [tweak](https://github.com/zendeskgarden/react-components/pull/954/commits/5b813ae810c31543b8da89cafc53dc99d2cd08af) I made to your `chrome` PR is currently resulting in an unwanted external-styled link. I'll need to go back and fix this.

Screen Shot 2020-12-21 at 10 30 21 AM

Mikaelia: So in this case it should just be `[brand color](#brand-colors)` ? I can add the fix if you would like (in this or a separate PR)

jzempel: It would have to be `[brand color](/design/color#brand-colors)` in order to be valid. A fix here would be great, thanks 🙏

jzempel: For clarification, the `MediaInput` lives on the same page as the `FauxInput`, so the link can be simplified to `[FauxInput](#fauxinput)`.

jzempel (approved):

PR #959, opened 2020-12-21
chore(grid): update prop descriptions

Mikaelia:
## Description

Updates grid package prop docs

### DIFF
--- a/packages/grid/src/elements/Col.tsx
+++ b/packages/grid/src/elements/Col.tsx
@@ -19,80 +19,79 @@ import { StyledCol } from '../styled';
 import useGridContext from '../utils/useGridContext';
 
 export interface IColProps extends HTMLAttributes<HTMLDivElement> {
-  /**
-   * Determine the size, relative to the total number of `columns` in the grid,
-   * for all screen sizes
-   */
+  /** Sets the total number of grid `columns` that the col spans by default */
   size?: GRID_NUMBER;
-  /** Determine the size for extra-small screen sizes */
+  /** Defines the col size for extra-small screens.
+   *  See [breakpoints](https://garden.zendesk.com/components/theme-object#default_theme).
+   */
   xs?: BREAKPOINT;
-  /** Determine the size for small screen sizes */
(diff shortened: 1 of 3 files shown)


packages/grid/src/elements/Col.tsx
```diff
@@ -19,80 +19,79 @@ import { StyledCol } from '../styled';
...
 export interface IColProps extends HTMLAttributes<HTMLDivElement> {
-  /**
-   * Determine the size, relative to the total number of `columns` in the grid,
-   * for all screen sizes
-   */
+  /** Sets the total number of grid `columns` that the col spans on all screen sizes */
   size?: GRID_NUMBER;
-  /** Determine the size for extra-small screen sizes */
+  /** Defines the col size on extra-small screens.
+   *  See [breakpoints](https://garden.zendesk.com/components/theme-object#default_theme).
```
jzempel, line 25: Let's use the `website`-local URL here, and maybe change to `/components/theme-object#breakpoints`

packages/grid/src/elements/Row.tsx
```diff
@@ -20,49 +20,49 @@ import { StyledRow } from '../styled';
 
 export interface IRowProps extends HTMLAttributes<HTMLDivElement> {
   /**
-   * Applies the `align-items` flex container property, affecting vertical `Col`
-   * alignment, for all screen sizes
+   * Applies the `align-items` flex container property to the row.
+   * This affects vertical `Col` alignment on all screen sizes.
```
jzempel, line 24: Yes, agreed on this use of `Col` 👍

jzempel (changes requested):
For all use of `col`, I think it would be better to replace with `column`. We would say `Col` only if there was a particular reference to the component itself.

jzempel (commented):
Any thoughts on https://github.com/zendeskgarden/react-components/pull/959#pullrequestreview-556692567?

Mikaelia:
Ah yes, just updated!

jzempel (approved):

PR #962, opened 2020-12-21
chore(notifications, pagination): update prop descriptions

Mikaelia:
## Description
Updates notifications and pagination packages prop descriptions

### DIFF
--- a/packages/notifications/src/elements/Well.tsx
+++ b/packages/notifications/src/elements/Well.tsx
@@ -10,7 +10,9 @@ import PropTypes from 'prop-types';
 import { StyledWell } from '../styled';
 
 interface IWellProps {
+  /** Applies a background color to differentiate the well from the rest of the page */
   isRecessed?: boolean;
+  /** Applies a drop shadow to lift the well off of the page */
   isFloating?: boolean;
 }
 
--- a/packages/notifications/src/elements/Well.tsx
+++ b/packages/notifications/src/elements/Well.tsx
@@ -10,9 +10,9 @@ import PropTypes from 'prop-types';
 import { StyledWell } from '../styled';
 
 interface IWellProps {
-  /** Applies a background color to differentiate the well from the rest of the page */
+  /** Applies a background color */
   isRecessed?: boolean;
-  /** Applies a drop shadow to lift the well off of the page */
+  /** Applies a drop shadow */
   isFloating?: boolean;
 }
 
--- a/packages/pagination/src/elements/Pagination/Pagination.tsx
+++ b/packages/pagination/src/elements/Pagination/Pagination.tsx
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
 
 export interface IPaginationProps extends Omit<HTMLAttributes<HTMLUListElement>, 'onChange'> {
   /**
-   * The currently selected page
+   * Sets the current page
    */
   currentPage: number;
   /**
-   * The total number of pages available
+   * Defines the total number of pages
    */
   totalPages: number;
   /**
-   * The number of pages to pad the current page with when page gaps are
-   * displayed
+   * Sets the number of pages that appear between the current page and a gap indicator
    */
   pagePadding?: number;
   /**
-   * Position for the leading and trailing gap indicator, if needed based on
-   * current and total pages
+   * Positions the leading and trailing gap indicator, based on
+   * the current and total pages
    */
   pageGap?: number;
   /**
-   * @param {Any} currentPage - The currently selected page
+   * Handles page change events
+   *
+   * @param {Any} currentPage - The current page
    */
   onChange?: (currentPage: number) => void;
   /**
-   * Apply localized labels, test attributes, etc. to individual pages.
+   * Applies localized labels, test attributes, etc. to individual pages
    *
-   * @param {PAGE_TYPE} pageType - the type of the page to transform props for;
-   *  one of: `'previous'`, `'gap'`, `'page'`, `'next'`
-   * @param {Any} props - default page props to transform
+   * @param {PAGE_TYPE} pageType The type of the page accepting the props.
+   *  One of: `'previous'`, `'gap'`, `'page'`, `'next'`
+   * @param {Any} props Default page props to transform
    */
   transformPageProps?: (pageType: PAGE_TYPE, props: any) => any;
 }
--- a/packages/pagination/src/elements/Pagination/Pagination.tsx
+++ b/packages/pagination/src/elements/Pagination/Pagination.tsx
@@ -47,7 +47,7 @@ export interface IPaginationProps extends Omit<HTMLAttributes<HTMLUListElement>,
   /**
    * Applies localized labels, test attributes, etc. to individual pages
    *
-   * @param {PAGE_TYPE} pageType The type of the page accepting the props.
+   * @param {string} pageType The type of the page accepting the props.
    *  One of: `'previous'`, `'gap'`, `'page'`, `'next'`
    * @param {any} props Default page props to transform
    */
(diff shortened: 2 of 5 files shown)


Mikaelia:
Arg. Looks like somehow I forgot to branch and these updates (pagination and notifications) got combined. Would you like me to try to split them?

jzempel:
> Would you like me to try to split them?

Nah, this is fine.

packages/notifications/src/elements/Well.tsx
```diff
@@ -10,7 +10,9 @@ import PropTypes from 'prop-types';
 import { StyledWell } from '../styled';
 
 interface IWellProps {
+  /** Applies a background color to differentiate the well from the rest of the page */
   isRecessed?: boolean;
+  /** Applies a drop shadow to lift the well off of the page */
```
jzempel, line 15: I'm totally digging this. But tagging @allisonacs and @ctancinco for a consistency head-check. How design-explicit do we want to be here? Is the simpler "Applies a drop shadow" better?

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
 
 export interface IPaginationProps extends Omit<HTMLAttributes<HTMLUListElement>, 'onChange'> {
   /**
-   * The currently selected page
+   * Sets the current page
```
jzempel, line 25: Should we add zero- or one-based here?

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
...
-   * current and total pages
+   * Positions the leading and trailing gap indicator, based on
+   * the current and total pages
    */
   pageGap?: number;
   /**
-   * @param {Any} currentPage - The currently selected page
+   * Handles page change events
+   *
+   * @param {Any} currentPage - The current page
```
jzempel, line 44: ~Any~ -> "any" and no dash:

```
@param {any} currentPage The current page
```

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
...
    */
   onChange?: (currentPage: number) => void;
   /**
-   * Apply localized labels, test attributes, etc. to individual pages.
+   * Applies localized labels, test attributes, etc. to individual pages
    *
-   * @param {PAGE_TYPE} pageType - the type of the page to transform props for;
-   *  one of: `'previous'`, `'gap'`, `'page'`, `'next'`
-   * @param {Any} props - default page props to transform
+   * @param {PAGE_TYPE} pageType The type of the page accepting the props.
```
jzempel, line 50: staying true to JSDoc, the type here should be `string` (or is it `String`, I always forget).

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
...
+   *  One of: `'previous'`, `'gap'`, `'page'`, `'next'`
+   * @param {Any} props Default page props to transform
```
jzempel, line 52: Lowercase `any`

jzempel (changes requested):

packages/notifications/src/elements/Well.tsx
ctancinco: short and sweet is my vote. your edit works for me.

packages/pagination/src/elements/Pagination/Pagination.tsx
Mikaelia: I'm for it!

packages/pagination/src/elements/Pagination/Pagination.tsx
Mikaelia: Took a quick look at the docs--looks like `string`

Mikaelia: Would you like me to update the function signature as well?

Mikaelia: Hmm this actually throws off a test, too

jzempel: Only the JSDoc should change to `string`, not the signature. We want the benefit of strict typing, but documenting with [valid](https://jsdoc.app/tags-param.html) JSDoc param types (in general).

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
...
   onChange?: (currentPage: number) => void;
   /**
-   * Apply localized labels, test attributes, etc. to individual pages.
+   * Applies localized labels, test attributes, etc. to individual pages
    *
-   * @param {PAGE_TYPE} pageType - the type of the page to transform props for;
-   *  one of: `'previous'`, `'gap'`, `'page'`, `'next'`
-   * @param {Any} props - default page props to transform
+   * @param {string} pageType The type of the page accepting the props.
+   *  One of: `'previous'`, `'gap'`, `'page'`, `'next'`
```
jzempel, line 51: Oh, my bad... Should these just be non-backtick'd strings? I _think_ that's what we've used in other cases. Please correct me if I'm wrong.

`One of: "previous", "gap", "page", next"`

Mikaelia: Yes, you're right--just checked togglebuton

jzempel (approved):

packages/pagination/src/elements/Pagination/Pagination.tsx
```diff
@@ -22,33 +22,34 @@ export type PAGE_TYPE = 'next' | 'page' | 'gap' | 'previous';
...
   /**
-   * Position for the leading and trailing gap indicator, if needed based on
-   * current and total pages
+   * Positions the leading and trailing gap indicator, based on
+   * the current and total pages
    */
   pageGap?: number;
   /**
-   * @param {Any} currentPage - The currently selected page
+   * Handles page change events
```
hzhu, line 42: I'm wondering if there may be a more standardized way to describe an `onChange` handler since it is a common prop across many Garden components.

Here's one style that I think could work well for a developer audience. Something along the lines of:

| Component      | Description |
| ----------- | ----------- |
| Pagination      | A callback that is fired when a pagination page is changed       |
| Accordion   | A callback that is fired when an accordion panel is expanded or collapsed       |
| ColorPicker   | A callback that is fired when a color is selected        |

Alternatively, these could work too:

| Component      | Description |
| ----------- | ----------- |
| Pagination      | A callback that is fired when a pagination page state is changed       |
| Accordion   | A callback that is fired when an accordion panel's open state is changed        |
| ColorPicker   | A callback that is fired when a color picker state is changed        |

I like that these are a little bit more specific in their description. What do ya'll think?

jzempel: @hzhu please visit the prop documentation RFC for the intended grammar. I'm all for improving specificity, ideally within the bounds of the specified standard.

jzempel: In this case, I think "Handles..." is a solid choice for active voice present, third person, singular phrasing. How would you recommend building on that?

hzhu: I see, I think we can keep it as-is.

hzhu (approved):

PR #963, opened 2020-12-21
chore(tables): update prop descriptions

Mikaelia:
## Description

Updates tables package prop descriptions

### DIFF
--- a/packages/tables/src/styled/StyledCell.ts
+++ b/packages/tables/src/styled/StyledCell.ts
@@ -25,7 +25,7 @@ export interface IStyledCellProps {
   hasOverflow?: boolean;
   /** Adjusts the width of the cell */
   width?: string | number;
-  /** Adjusts the vertical pading of the cell */
+  /** Adjusts the vertical padding of the cell */
   size: SIZE;
 }
 
--- a/packages/tables/src/styled/StyledCell.ts
+++ b/packages/tables/src/styled/StyledCell.ts
@@ -21,12 +21,11 @@ export interface IStyledCellProps {
   isMinimum?: boolean;
   /** Applies truncated text styling */
   isTruncated?: boolean;
-  /**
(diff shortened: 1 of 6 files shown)


packages/tables/src/styled/StyledCell.ts
```diff
@@ -19,14 +19,13 @@ export interface IStyledCellProps {
...
   isTruncated?: boolean;
-  /**
-   * Styling for a cell that contains an `OverflowButton`
-   */
+  /** Applies styling for a cell that contains an overflow menu */
   hasOverflow?: boolean;
-  /** The width of the cell */
+  /** Adjusts the width of the cell */
   width?: string | number;
+  /** Adjusts the vertical pading of the cell */
```
jzempel, line 28: *padding

jzempel (changes requested):
It seems like some of the elements listed here got missed:

https://garden.zendesk.com/components/table#api

Please double-check against `yarn start:styleguidist` to ensure all props are getting coverage.

packages/tables/src/styled/StyledCell.ts
Mikaelia: Hmm I'm not sure why they are missing. I have run yarn start:styleguidist but it actually doesn't look like it is picking up any of my props in the `styled` folder..

jzempel: Yep, I just double-checked and the rendered docs are looking good, and only https://github.com/zendeskgarden/react-components/blob/67b39a7787e3f51560a8f60d84a0d6eb21192e4b/packages/tables/src/elements/SortableCell.tsx#L20-L22 is missing.

Mikaelia: Forgot to push 😅

jzempel (approved):

PR #965, opened 2020-12-21
chore(tags): update prop descriptions

Mikaelia:
## Description
Updates tags package prop descriptions

### DIFF
--- a/packages/tags/src/elements/Tag.tsx
+++ b/packages/tags/src/elements/Tag.tsx
@@ -11,16 +11,19 @@ import { StyledTag, StyledAvatar } from '../styled';
 import Close from './Close';
 
 interface ITagProps extends HTMLAttributes<HTMLDivElement> {
+  /** Adjusts font size and padding */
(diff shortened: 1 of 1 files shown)


packages/tags/src/elements/Tag.tsx
```diff
@@ -11,16 +11,19 @@ import { StyledTag, StyledAvatar } from '../styled';
...
-   * hue, but with the ability to override using any hex value.
+   * for available colors. Accepts any hex value.
    */
   hue?: string;
+  /** Applies pill styling */
   isPill?: boolean;
+  /** Applies styles to round the tag */
   isRound?: boolean;
-  /** Style using regular (non-bold) font weight */
+  /** Applies regular (non-bold) `font-weight` */
```
jzempel, line 26: It seems like you've generally opted to go with "font weight" rather than `font-weight`. Let's keep it consistent here.

jzempel (changes requested):

jzempel (approved):

PR #957, opened 2020-12-17
chore(dropdowns): update prop descriptions

Mikaelia:
## Description

Updates dropdowns package props

## Detail

Could definitely use an extra eye on `Dropdown` function props. I'm not entirely sure how to describe some of these params

PR #966, opened 2020-12-21
chore(tooltips): update prop descriptions

Mikaelia:
## Description
Updates tooltips package prop descriptions

### DIFF
--- a/packages/tooltips/src/elements/Tooltip.tsx
+++ b/packages/tooltips/src/elements/Tooltip.tsx
@@ -38,11 +38,10 @@ export interface ITooltipProps
   /** Defines the content of the tooltip */
   content: React.ReactNode;
   /**
-   * Adjusts the placement of the tooltip. Placement values differ from Popper placement values to
-   * assist with RTL layouts.
+   * Adjusts the placement of the tooltip
    **/
   placement?: GARDEN_PLACEMENT;
-  /** Passes configurations to the [Popper instance](https://popper.js.org/docs/v1/#modifiers--codeobjectcode) */
+  /** Passes configurations to the [Popper instance](https://popper.js.org/docs/v2/modifiers/) */
   popperModifiers?: Modifiers;
   /** Adjusts the padding and font size */
   size?: TOOLTIP_SIZE;
--- a/packages/tooltips/src/elements/Tooltip.tsx
+++ b/packages/tooltips/src/elements/Tooltip.tsx
@@ -45,7 +45,7 @@ export interface ITooltipProps
   popperModifiers?: Modifiers;
   /** Adjusts the padding and font size */
   size?: TOOLTIP_SIZE;
-  /** Applies either light or dark styling */
+  /** Specifies the tooltip type */
   type?: TOOLTIP_TYPE;
   /** Sets the `z-index` of the tooltip */
   zIndex?: number | string;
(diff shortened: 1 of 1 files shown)


packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -25,30 +25,37 @@ import { StyledTooltipWrapper, StyledTooltip, TOOLTIP_SIZE, TOOLTIP_TYPE } from
...
+  /** Allows the tooltip to reposition during browser resize events */
   eventsEnabled?: boolean;
+  /** Sets the ID of the tooltip */
   id?: string;
+  /** Defines the content of the tooltip */
   content: React.ReactNode;
   /**
-   * These placements differ from the default naming of Popper placements to help
+   * Adjusts the placement of the tooltip. Placement values differ from Popper placement values to
    * assist with RTL layouts.
```
jzempel, line 42: Again, I think the simplicity of your previous description is ideal:

https://github.com/zendeskgarden/react-components/pull/957/files#diff-8fc5bdeb0d77c73a3b03ce92de8df5ff9a393a3b092cc638a7700df6c8f42ec3L30-L33

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -25,30 +25,37 @@ import { StyledTooltipWrapper, StyledTooltip, TOOLTIP_SIZE, TOOLTIP_TYPE } from
...
    **/
   placement?: GARDEN_PLACEMENT;
-  /** See Popper [documentation](https://popper.js.org/docs/v1/#modifiers--codeobjectcode) for details */
+  /** Passes configurations to the [Popper instance](https://popper.js.org/docs/v1/#modifiers--codeobjectcode) */
```
jzempel, line 45: Let's stick with the updated `https://popper.js.org/docs/v2/modifiers/` URL.

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -25,30 +25,37 @@ import { StyledTooltipWrapper, StyledTooltip, TOOLTIP_SIZE, TOOLTIP_TYPE } from
...
   popperModifiers?: Modifiers;
+  /** Adjusts the padding and font size */
   size?: TOOLTIP_SIZE;
+  /** Applies either light or dark styling */
```
jzempel, line 49: This feels a wee bit redundant based on what we see here: https://garden.zendesk.com/components/tooltip#tooltip

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -25,30 +25,37 @@ import { StyledTooltipWrapper, StyledTooltip, TOOLTIP_SIZE, TOOLTIP_TYPE } from
...
   type?: TOOLTIP_TYPE;
+  /** Sets the `z-index` of the tooltip */
   zIndex?: number | string;
-  /** Determine visibility on initial render */
+  /** Displays the tooltip on initial render */
   isInitialVisible?: boolean;
-  /** Control visibility state of the Tooltip */
+  /** Displays the tooltip */
   isVisible?: boolean;
   children: React.ReactElement;
```
jzempel, line 57: Good call not documenting. Can this be removed from the interface without prompting a TS error? Generally, we don't define default props.

jzempel (changes requested):

packages/tooltips/src/elements/Tooltip.tsx
Mikaelia: Looks like it breaks a test

packages/tooltips/src/elements/Tooltip.tsx
Mikaelia: Would something like "Specifies the tooltip type" be any better?

jzempel: Yeah, I think so.

jzempel (approved):

PR #967, opened 2020-12-21
chore(typography): update prop descriptions

Mikaelia:
## Description
Updates typography package prop descriptions

### DIFF
--- a/packages/typography/src/elements/Ellipsis.tsx
+++ b/packages/typography/src/elements/Ellipsis.tsx
@@ -10,11 +10,9 @@ import PropTypes from 'prop-types';
 import { StyledEllipsis } from '../styled';
 
 interface IEllipsisProps extends HTMLAttributes<HTMLSpanElement> {
-  /**
-   * Optional override for the auto-generated `title` attribute
-   */
+  /** Overrides the auto-generated `title` attribute */
   title?: string;
-  /** Any valid element for the styled component */
+  /** Updates the element tag */
   tag?: any;
 }
 
--- a/packages/typography/src/elements/Ellipsis.tsx
+++ b/packages/typography/src/elements/Ellipsis.tsx
@@ -12,7 +12,7 @@ import { StyledEllipsis } from '../styled';
 interface IEllipsisProps extends HTMLAttributes<HTMLSpanElement> {
   /** Overrides the auto-generated `title` attribute */
   title?: string;
-  /** Updates the element tag */
+  /** Updates the element's HTML tag */
   tag?: any;
 }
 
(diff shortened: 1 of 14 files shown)


packages/typography/src/elements/Ellipsis.tsx
```diff
@@ -10,11 +10,9 @@ import PropTypes from 'prop-types';
 import { StyledEllipsis } from '../styled';
 
 interface IEllipsisProps extends HTMLAttributes<HTMLSpanElement> {
-  /**
-   * Optional override for the auto-generated `title` attribute
-   */
+  /** Overrides the auto-generated `title` attribute */
   title?: string;
-  /** Any valid element for the styled component */
+  /** Updates the element tag */
```
jzempel, line 15: ~tag~ -> "HTML tag"? Or is "element's HTML tag" too much? ditto for similar props below...

packages/typography/src/elements/Span.tsx
```diff
@@ -10,16 +10,16 @@ import PropTypes from 'prop-types';
 import { StyledFont, StyledIcon } from '../styled';
 
 interface ISpanProps extends HTMLAttributes<HTMLSpanElement> {
-  /** Any valid DOM element for the styled component */
+  /** Updates the element tag */
   tag?: any;
-  /** Render bold font; `undefined` = inherited weight, `true` = bolded weight, `false` = regular weight */
+  /** Applies bold font style */
```
jzempel, line 15: I think it might be important to incorporate the use of `undefined` (or default) here.

Something like: `Applies bold font style. Font weight is inherited by default.`

packages/typography/src/elements/Span.tsx
```diff
@@ -10,16 +10,16 @@ import PropTypes from 'prop-types';
...
   isBold?: boolean;
-  /** Render monospace font */
+  /** Renders with monospace font */
   isMonospace?: boolean;
   /**
-   * Apply a span color – typically constrained to a
+   * Applies a font color. Use
    * [palette](https://zendeskgarden.github.io/react-components/theming/#palette)
```
jzempel, line 21: please update to `[PALETTE](/components/palette#palette)`

jzempel (changes requested):
"marker style" – awesome. TIL https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#Usage_notes

packages/typography/src/elements/Ellipsis.tsx
Mikaelia: I think "element's HTML tag" works!

packages/typography/src/elements/Span.tsx
Mikaelia: Agreed

jzempel (approved):

hzhu (approved):

PR #986, opened 2021-01-15
chore(tooltips): migrate tooltip example to storybook

austingreendev:
## Description

This PR migrates the existing styleguidist examples for `react-tooltips` to Storybook.

## Checklist

      designer as a reviewer)


packages/tooltips/README.md
```diff
@@ -39,7 +39,7 @@ their parent [ThemeProvider](https://zendeskgarden.github.io/react-components/th
 
 #### English (LTR) Placements
```
jzempel, line 40: The benefit of these LTR/RTL sections are questionable (and not easily digestible). All Garden components respond to RTL. The callout here feels a bit gratuitous and could probably just be removed.

packages/tooltips/stories/2-Tooltip.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+  },
+  size: {
+    name: 'Size',
+    control: {
+      type: 'select',
+      options: ['default', 'small', 'medium', 'large', 'extra-large']
+    }
+  },
+  forceVisibility: {
+    name: 'Foce Visibility'
```
jzempel, line 166: *Force. Also [nit] use sentence case.

packages/tooltips/stories/2-Tooltip.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+        'end-top',
+        'end-bottom',
+        'start',
+        'start-top',
+        'start-bottom'
+      ]
+    }
+  },
+  hasArrow: {
+    name: 'Has Arrow'
```
jzempel, line 156: [nit] use sentence case.

packages/tooltips/stories/2-Tooltip.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+    name: 'Size',
+    control: {
+      type: 'select',
+      options: ['default', 'small', 'medium', 'large', 'extra-large']
+    }
+  },
+  forceVisibility: {
+    name: 'Foce Visibility'
+  },
+  delayMS: {
```
jzempel, line 168: If controls can be dynamically disabled, then `delayMS` should be disabled when `forceVisibility` is true. https://storybook.js.org/docs/react/essentials/controls#disable-controls-for-specific-properties

jzempel (changes requested):

packages/tooltips/stories/2-Tooltip.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <>
+                <Title>Light Tooltip</Title>
+                <Paragraph>{retrieveTooltipContent(size, 'light')}</Paragraph>
+              </>
+            }
+            hasArrow={hasArrow}
+            placement={placement}
+            size={size === 'default' ? undefined : size}
+            type="light"
+            isVisible={forceVisibility ? true : undefined}
```
hzhu, line 109: Would this line be equivalent to `isVisible={forceVisibility}`?

hzhu (approved):
LGTM after the other comments are addressed.

packages/tooltips/README.md
austingreendev: Agreed, just wasn't sure if we wanted to remove that much content. I've removed both sections!

packages/tooltips/stories/2-Tooltip.stories.tsx
austingreendev: Updated.

packages/tooltips/stories/2-Tooltip.stories.tsx
austingreendev: Updated.

packages/tooltips/stories/2-Tooltip.stories.tsx
austingreendev: I don't think we have access to the current story state in the `argTypes`. It's static like `propTypes`.

Another option is a warning message within the story based on those states, but that might be heavy-handed.

packages/tooltips/stories/2-Tooltip.stories.tsx
austingreendev: Since this is a controlled value, `isVisible={false}` would not allow the Tooltips to be shown even when hovered. `undefined` lets the user interact with the mouse in its default state.

packages/tooltips/stories/2-Tooltip.stories.tsx
jzempel: Yeah, no need to get fancy with a warning.

jzempel (approved):

PR #990, opened 2021-01-20
fix(dropdowns): allow menu usage in table elements

austingreendev:
## Description

We have found that when `Dropdowns` (mostly `Trigger`) are used within a `Table` component we see additional scroll space and scrolling issues when dropdowns are interacted.

https://codesandbox.io/s/brave-cray-7yzwt?file=/src/App.js

## Detail

During the investigation, I found the culprit was our hidden `<input />` that is included for a11y and Downshift reasons. The `position: absolute` value works as intended in standard flows, but placing it within a `td` or `th` element resets this positioning.

To maintain styling across standard page layouts and table usages I have moved this to `position: fixed`.

I looked into different approaches of apply `position: relative` to the table cell and content, but all solutions required a wrapping `<div>` as well as complications for positioning the menu within it.

I've also updated our "Scrollable" table example to include overflow menus to help validate this change.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 86987,
-    "minified": 54069,
-    "gzipped": 11060
+    "bundled": 86984,
+    "minified": 54066,
+    "gzipped": 11063
   },
   "index.esm.js": {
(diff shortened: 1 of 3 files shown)


jzempel (approved):
TIL

hzhu (approved):

PR #994, opened 2021-01-24
feat(chrome): add SkipNav component

### LINKED ISSUE #314: Add "Skip Navigation" to `Chrome`
## Expectations

Accessible navigation should provide a "Skip Navigation" link for users who use keyboard or screen readers to navigate ([WebAIM](https://webaim.org/techniques/skipnav/)), so the `Chrome` package should provide this, either by default or as a component.

## Suggestions

The implementation could be similar to the `SkipNav` component from Reach UI which is fairly simple ([code](https://github.com/reach/reach-ui/blob/master/packages/skip-nav/src/index.js)), but could point to `<Main />`.

### Fine Print

* Component: `@zendeskgarden/react-chrome`
* Browsers: all

jzempel:
## Description

Provides a new `SkipNav` component which allows a keyboard user to bypass `Nav` and `Header` and jump to a targeted ID.

## Detail

Closes #314

## Checklist

      designer as a reviewer)


packages/chrome/src/elements/Chrome.tsx
```diff
@@ -38,10 +39,12 @@ const Chrome = React.forwardRef<HTMLDivElement, IChromeProps>(({ hue, isFluid, .
 
     return false;
   }, [hue, theme]);
+  const uid = useUID();
```
austingreendev, line 42: Would the functionality provided by [useUIDSeed()](https://github.com/thearnica/react-uid#hooks-168) be a good fit for here?

Also, with this ID being added to the query params of the page when selected, should we look into using a non-unique ID? Since there is only a single `SkipNav` available per-page this could be hard-coded to something more human-readable, like `#main-content`.

jzempel: > could be hard-coded to something more human-readable, like `#main-content`
we need to steer clear of any existing IDs used throughout product; something like `#main-content` is likely already in use.

> Since there is only a single `SkipNav` available per-page
I was considering some of our own pages like the [website](https://garden.zendesk.com/components/chrome) or [storybook](https://zendeskgarden.github.io/react-components/storybook/?path=/docs/components-chrome--default) where there are multiples. But I could be convinced that this accommodation is unnecessary, and those particular pages can specify `Content` IDs as workarounds.

> Would the functionality provided by useUIDSeed() be a good fit for here?
🤷 what makes `useUIDSeed` preferable over `useUID` here? This case doesn't involve generation of multiples.

Either way, I can be convinced that the inclusion of `react-uid` is overkill and we can do just fine without it.

jzempel: But, you're right, the query param (and knock-on analytics, etc) is gross. Hopefully folks would be encouraged to set their own `Content` ID.

jzempel: Sorry for talking to myself on this thread, but I could also be convinced to just go with `#main-content` and let folks adjust to their needs as desired.

austingreendev: Another option would be to make the target ID a required prop. Most of this needs to be thought throw anyway to make sure the main content is the "correct" content to focus on. There could be nested areas that could theoretically receive focus.

Maybe something like:
```
<SkipNav targetId="main-content" /> // required targetId prop

<Content id="main-content">...</Content> //optional prop, spread onto standard element
```

This would also make it easier for products with non-standard `Chrome` implementations to use the `SkipNav`

jzempel: Yeah, I think this simplifies the implementation substantially. Will update...

austingreendev (approved):
LGTM

hzhu (approved):

allisonacs (approved):

VilhelmNielsen:
😍 🎉

PR #999, opened 2021-01-27
feat(colorpicker): introduce color picker package

hzhu:
## Description

This PR introduces the `react-colorpicker` package. It exports two components: 

1) `ColorPicker`
Screen Shot 2021-01-27 at 12 48 36 PM

2) `ColorDialog`
Screen Shot 2021-01-27 at 12 48 25 PM

### Component API

The `ColorPicker` accepts a `color` prop to determine the current color state. The component is uncontrolled and the state is managed internally (via `reducer.ts`).

The color picker is driven by various color models (rgb/a, hsv, hsla, etc). The controlled `color` prop is used to initialize a color picker state. The color picker state is managed by a single `useReducer` function call. The logic is separated from the UI and the reducer logic lives in a `reducer.ts` file.

### Testing

The `ColorPicker` and `ColorDialog` are tested by simulating user events and asserting against the UI. These tests aim to follow the integration structure as described by `react-testing-library`. Unit tests are written for utility functions such as color model conversion.

### Cross-browser support

The color picker is supported in modern Chrome, Firefox, Safari, Edge, and IE11. Styling native input ranges were quite difficult. And to get parity amongst all browsers, some extra CSS and IE-specific CSS was needed. I tried my best to clean this up, but please let me know it can be cleaned up more so.

Screen Shot 2021-01-27 at 11 33 32 AM

## Checklist

      designer as a reviewer)


demo/index.html
```diff
@@ -110,7 +110,10 @@ <h1 class="c-main__title">React Components</h1>
               <a class="u-fg-inherit" href="storybook/?path=/story/components-chrome-readme--page">Chrome</a>
             </div>
             <div class="u-mb-sm">
-              <a class="u-fg-inherit" href="storybook/?path=/story/components-datepickers-readme--page">Datepickers</a>
+              <a class="u-fg-inherit" href="storybook/?path=/story/components-colorpicker-readme--page">Colorpicker</a>
```
austingreendev, line 113: This makes the first color larger by 2 component entries. Should we rebalance and move `Dropdowns` to the top of the next column?

packages/colorpicker/README.md
```diff
@@ -0,0 +1,29 @@
...
+npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
+```
+
+## Usage
+
+```jsx
+import { ThemeProvider } from '@zendeskgarden/react-theming';
+import { ColorPicker } from '@zendeskgarden/react-colorpicker';
+
+const [color, setColor] = React.useState('#17494D');
```
austingreendev, line 21: Should we include a wrapping `Example` function around this to make it a valid example? Take a look at `Tabs/README.md`

packages/colorpicker/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@zendeskgarden/react-colorpicker",
+  "description": "Components related to colorpicker in the Garden Design System",
```
austingreendev, line 3:
```suggestion
  "description": "Components related to colorpickers in the Garden Design System",
```

packages/colorpicker/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "version": "8.29.3",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "../../utils/scripts/build.sh",
+    "build:demo": "../../utils/scripts/build-demo.sh",
```
austingreendev, line 19: This isn't needed anymore with the move to Storybook. Was this included in the package template process?

packages/colorpicker/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@zendeskgarden/react-forms": "^8.29.3",
+    "@zendeskgarden/react-modals": "^8.29.3",
+    "polished": "^4.0.0",
+    "react-merge-refs": "^1.1.0"
+  },
+  "peerDependencies": {
+    "@zendeskgarden/react-theming": "^8.1.0",
+    "prop-types": "^15.6.1",
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
```
austingreendev, line 35:
```suggestion
    "react": ">=16.8.0",
    "react-dom": ">=16.8.0",
```

austingreendev: This allows React v17 as a valid peer dependency.

packages/colorpicker/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "styled-components": "^4.2.0 || ^5.0.0"
+  },
+  "devDependencies": {
+    "@types/lodash.throttle": "4.1.6",
+    "@zendeskgarden/react-theming": "^8.0.0",
+    "@zendeskgarden/svg-icons": "6.27.0",
+    "lodash.throttle": "4.1.1"
```
austingreendev, line 42: If this is used within any non-storybook code we need to include it as a `dependency` to ensure that it's included for consumers.

packages/colorpicker/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            }
+          }}
+        >
+          <StyledDialogPreview backgroundColor={color} />
+          <StyledIcon isRotated={referenceElement}>
+            <Chevron />
+          </StyledIcon>
+        </StyledButton>
+      </StyledFauxInput>
+
```
austingreendev, line 64: [nit] extra space

austingreendev (commented):
Just did a quick run-through of the package structure, but everything is looking great so far!

I did notice a small bug relating to mobile drag in the saturation block. The values are updated correctly, but the visual thumb isn't keeping up.


https://user-images.githubusercontent.com/4030377/106181432-70a4e780-6152-11eb-9423-e96bb679e715.mov

packages/colorpicker/package.json
jzempel: * color pickers

packages/colorpicker/package.json
jzempel: Please retrofit the `.template` as we find these bits.

hzhu:
> I did notice a small bug relating to mobile drag in the saturation block. The values are updated correctly, but the visual thumb isn't keeping up.

@austingreendev great catch. I've removed the touch handlers because I think this is a better way to handle saturation/lightness selection on a touch device. Alternatively, we could hook into touch events, but I think we would need to consider disabling the global scroll events to prevent scrolling when dragging the saturation thumb. Could you try out the mobile experience now and let me know what you think?

packages/colorpicker/package.json
hzhu: Moved to deps. Thanks!

demo/index.html
hzhu: Updated

packages/colorpicker/README.md
hzhu: Ah yes, I don't want a floating `useState` in the example. It's actually not necessary, so I've removed it and simplified the example code.

packages/colorpicker/package.json
hzhu: Updated 👍🏽

packages/colorpicker/package.json
hzhu: Updated `.template` 👍🏽

packages/colorpicker/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    return (
+      <StyledColorPicker ref={ref} {...props}>
+        <Saturation
+          hue={state.hue}
+          saturation={state.saturation}
+          lightness={state.lightness}
+          onChange={(hsv: IHSVColor) => {
+            dispatch({
+              type: 'saturation change',
```
jzempel, line 70: Use reducer constants throughout.

packages/colorpicker/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  ({ color, labels = {}, onChange, ...props }, ref) => {
+    const [state, dispatch] = useReducer(reducer, getInitialState(color));
+
+    useEffect(() => {
+      onChange && onChange(state);
+    }, [state, onChange]);
+
+    return (
+      <StyledColorPicker ref={ref} {...props}>
+        <Saturation
```
jzempel, line 64: Can we use @steue's specified `ColorWell` terminology (throughout) rather than `Saturation`?

packages/colorpicker/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  blue?: string;
+  alpha?: string;
+}
+
+export interface IColorPickerProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'color' | 'onChange'> {
+  /** A hex string, RGB string, RGB object, or color picker state that represents the current color */
+  color: string | IRGBColor | IColorPickerState;
+  /** A callback that is fired when an the color picker state is changed */
+  onChange?: (state: IColorPickerState) => void;
```
jzempel, line 46: Please re-phrase, using RFC-defined grammar. Consult all existing website props for details/ideas. The `onChange` function must be fully JSDoc'd.

packages/colorpicker/src/elements/ColorPicker/Saturation.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  return (
+    <StyledSaturationWrapper>
+      <StyledSaturation hue={hue} ref={container} role="presentation" onMouseDown={handleMouseDown}>
+        <StyledSaturationWhite>
+          <StyledSaturationBlack />
+          <StyledSaturationPointer top={topPosition} left={leftPosition}>
+            <StyledSaturationCircle />
+          </StyledSaturationPointer>
```
jzempel, line 78: Did you explore consolidating this into one `div` element and factor the circle center color choice accordingly?

packages/colorpicker/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  StyledButton,
+  StyledFauxInput,
+  StyledDialogPreview,
+  StyledTooltipModal,
+  StyledTooltipBody
+} from '../../styled';
+
+export interface IColorDialogProps extends IColorPickerProps {
+  /** A callback function that is called when the color dialog is closed */
+  onClose?: (selectedColor: string | IRGBColor | IColorPickerState) => void;
```
jzempel, line 26: Please re-phrase per specified grammar RFC. All functions must be fully JSDoc'd. Consult the `website` for info.

packages/colorpicker/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  /** Positions the placement of the color dialog */
```
jzempel, line 27: Per other website descriptions, let's go with "Adjusts the placement of the color dialog".

packages/colorpicker/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  placement?: GARDEN_PLACEMENT;
+}
+
+/**
+ * @extends HTMLAttributes<HTMLButtonElement>
+ */
+export const ColorDialog = forwardRef<
```
jzempel, line 34: I think the `ColorDialog` API will need to be fundamentally refactored to off flexibility similar to [Select](https://garden.zendesk.com/components/select).

jzempel (changes requested):
- This package should be named `colorpickers` (plural) in keeping with most other `react-components` package naming.
- Is the utils/conversions.ts file duplicating color conversions we get from `polished`?
- How do either of these components work in the context of a form?
- How can I invoke the `ColorDialog` with an alternate trigger (i.e. `IconButton`)?

packages/colorpicker/src/styled/ColorPicker/StyledSaturationCircle.ts
```diff
@@ -0,0 +1,36 @@
...
+  /* stylelint-disable */
+  transform: translate(
+    -${props => props.theme.space.base * 2}px,
+    -${props => props.theme.space.base * 2}px
+  );
+  /* stylelint-enable */
+  border-radius: 50%;
+  /* stylelint-disable */
+  box-shadow: 0 0 0 1.5px ${props => props.theme.colors.background},
+    inset 0 0 1px 1px rgba(0, 0, 0, 0.3), 0 0 1px 2px rgba(0, 0, 0, 0.4);
```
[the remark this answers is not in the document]
hzhu: Updated to use `getColor` and match design specs.

packages/colorpicker/src/elements/ColorDialog/index.tsx
hzhu: The `ColorDialog` has an API the covers the design specs. Are you suggesting an API that allows consumers to customize the dialog trigger?

packages/colorpicker/stories/examples/ColorPicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+    red: 'R',
+    green: 'G',
+    blue: 'B',
+    alpha: 'A'
+  }
+};
+
+Default.argTypes = {
+  color: { control: 'disable' },
+  labels: { control: 'object' }
```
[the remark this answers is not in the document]
hzhu: Each label can be driven by an individual text control to improve UX.

packages/colorpicker/stories/examples/ColorDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+    blue: 'B',
+    alpha: 'A'
+  }
+};
+
+Default.argTypes = {
+  color: { control: { disable: true } },
+  labels: { control: 'object' },
+  placement: {
+    labels: { control: 'object' },
```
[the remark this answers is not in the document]
hzhu: Each label can be driven by an individual text control to improve UX.

packages/colorpicker/src/styled/ColorPicker/StyledSaturationWrapper.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const COMPONENT_ID = 'colorpicker.saturation.wrapper';
+
+export const StyledSaturationWrapper = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  position: relative;
+  margin-bottom: ${props => props.theme.space.base * 2}px;
+  padding-bottom: 75%;
```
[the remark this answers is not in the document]
hzhu: It isn't corresponding with the spec due to oversight. Updated to match spec.

packages/colorpicker/src/styled/ColorPicker/StyledSaturationCircle.ts
```diff
@@ -0,0 +1,36 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpicker.saturation.circle';
+
+export const StyledSaturationCircle = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  /* stylelint-disable */
```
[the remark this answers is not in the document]
hzhu: `stylelint-disable-next-line` doesn't seem to work for multi-line CSS. The multi-line disable/enable doesn't seem to allow for calling out rules.

packages/colorpicker/src/styled/ColorPicker/StyledLabel.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import styled from 'styled-components';
+import { Label } from '@zendeskgarden/react-forms';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpicker.label';
+
+export const StyledLabel = styled(Label).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
```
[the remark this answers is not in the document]
hzhu: Updated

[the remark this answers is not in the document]
hzhu: Thanks, updated.

packages/colorpicker/src/styled/ColorPicker/StyledHue.ts
```diff
@@ -0,0 +1,146 @@
...
+  `;
+};
+
+export const colorStyles = (props: ThemeProps<DefaultTheme>) => {
+  const SHADE = 600;
+  const thumbBackgroundColor = getColor('white', SHADE, props.theme);
+  const thumbBorderColor = thumbBackgroundColor;
+  const thumbBoxShadow = props.theme.shadows.lg(
+    math(`${props.theme.space.base} * 1px`),
+    math(`${props.theme.space.base} * 2px`),
```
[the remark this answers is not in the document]
hzhu: Thanks, updated.

packages/colorpicker/src/styled/ColorDialog/StyledTooltipModal.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TooltipModal } from '@zendeskgarden/react-modals';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colordialog.tooltipmodal';
+
+export const StyledTooltipModal = styled(TooltipModal).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  width: 354px;
```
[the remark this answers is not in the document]
hzhu: Yes, this should. I've updated it to reflect that.

packages/colorpicker/src/elements/ColorPicker/index.tsx
hzhu: Yeah, updated to use `ColorWell` terminology.

packages/colorpicker/src/styled/ColorPicker/StyledAlpha.ts
```diff
@@ -0,0 +1,66 @@
...
+    background: none;
+  }
+
+  &::-ms-fill-lower {
+    background: none;
+  }
+
+  &::-webkit-slider-thumb {
+    margin-top: -5px;
+  }
```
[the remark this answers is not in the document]
hzhu: This was specifically the get the slider thumbs to align with the track on Chrome. It was the last bit of CSS needed to achieve parity across all the browsers. I wasn't able to remove it the first time around, but I'll give it another look and see if there's a better way to handle the styling.

packages/colorpicker/src/elements/ColorPicker/Saturation.tsx
hzhu: Oops, these should be a single styled component.

packages/colorpicker/stories/examples/ColorDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const [color, setColor] = useState<string | IRGBColor>({
+    red: 23,
+    green: 73,
+    blue: 77,
+    alpha: 100
+  });
+
+  return (
+    <div style={{ display: 'flex', justifyContent: 'center' }}>
+      <ColorDialog color={color} labels={labels} onClose={setColor} placement={placement} />
```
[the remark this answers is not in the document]
hzhu: Yes, that's a great idea. I'll add another story to demonstrate that.

packages/colorpicker/src/elements/ColorDialog/index.tsx
hzhu: > How can I invoke the ColorDialog with an alternate trigger (i.e. IconButton)?

Thanks for bringing this up because this was a missing feature. The API has been modified to accept any (valid) element via `props.children` as a trigger. I've created a Storybook example to demonstrate the usage with an `IconButton`.

packages/colorpicker/src/styled/ColorPicker/StyledSaturationWhite.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpicker.saturation.white';
+
+export const StyledSaturationWhite = styled.div.attrs({
```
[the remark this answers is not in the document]
hzhu: I ran into linear-gradient issues last week while attempting to combine them into a single styled component. Will give it another go.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const ColorWell: React.FC<IColorWellProps> = ({ hue, saturation, lightness, onChange }) => {
+  const { rtl } = useContext(ThemeContext);
+  const container = useRef<HTMLDivElement>(null);
+  const hsv = hsl2hsv(hue, saturation, lightness);
+  const [position, setPosition] = useState<{ s: number; v: number }>({
+    s: rtl ? 100 - hsv.s : hsv.s,
+    v: hsv.v
+  });
```
austingreendev, line 35: [nit] not sure if this is complex enough to warrant a [lazy initial state syntax](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state), but it's up to you.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+      setPosition(saturationPosition);
+      onChange && onChange(nextHsv, e);
+    }
+  }, 50);
+
+  const topPosition = 100 - position.v;
+  const leftPosition = position.s;
+
+  const handleChange = (e: any) => throttledChange(e);
```
austingreendev, line 50: Is `throttledChange` ever called without this event? You could just provided `throttledChange` rather than `handleChange`

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const handleMouseUp = () => {
+    throttledChange.cancel();
+    window.removeEventListener('mousemove', handleChange);
+    window.removeEventListener('mouseup', handleMouseUp);
```
austingreendev, line 55: If we're attaching global events we need a useEffect to remove these in the event the component is unmounted before the mouseUp event occurs.

austingreendev: You could also have a single `isMousedDown` state with an effect that applies/removes events based on that value.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * @extends HTMLAttributes<HTMLDivElement>
+ */
+export const ColorPicker = forwardRef<HTMLDivElement, IColorPickerProps>(
+  ({ color, labels = {}, onChange, ...props }, ref) => {
+    const [state, dispatch] = useReducer(reducer, getInitialState(color));
+
+    useEffect(() => {
+      onChange && onChange(state);
+    }, [state, onChange]);
```
austingreendev, line 73: Is this dependency array accurate? If someone were to provide a new instance of `onChange` it would immediately get called with the current state.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+          saturation={state.saturation}
+          lightness={state.lightness}
+          onChange={(hsv: IHSVColor) => {
+            dispatch({
+              type: SATURATION_CHANGE,
+              payload: hsv
+            });
+          }}
+        />
+
```
austingreendev, line 88: [nit] extra space

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+                }}
+              />
+              <StyledCheckered />
+              <StyledAlphaGradient
+                rgb={{ red: state.red, green: state.green, blue: state.blue, alpha: state.alpha }}
+              />
+            </StyledAlphaField>
+          </StyledSliders>
+        </StyledSliderGroup>
+
```
austingreendev, line 123: [nit] extra space

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -0,0 +1,298 @@
...
+
+export type ColorPickerActionTypes =
+  | ISaturationChange
+  | IHueSliderChange
+  | IAlphaSliderChange
+  | IHexChange
+  | IRedChange
+  | IGreenChange
+  | IBlueChange
+  | IAlphaChange;
```
austingreendev, line 103: There are many ways to type these reducer actions, but let's try and stick with the format provided in [react-datepickers](https://github.com/zendeskgarden/react-components/blob/main/packages/datepickers/src/elements/Datepicker/utils/datepicker-reducer.ts#L85-L93)

This will cut down on a lot of the TS boilerplate.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -0,0 +1,298 @@
...
+
+export type ReducerType = (
```
austingreendev, line 105: Does React provide any generics for this type? It seems to mimic some of their dispatcher actions.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -0,0 +1,298 @@
...
+      blue,
+      redInput: red,
+      greenInput: green,
+      blueInput: blue,
+      alpha: 100,
+      alphaInput: 100,
+      hex: initialColor
+    };
+  }
+  const { red, green, blue, alpha = 100 } = initialColor;
```
austingreendev, line 130: [nit] add a space above

packages/colorpickers/src/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export * from './elements/ColorPicker';
+export * from './elements/ColorDialog';
+export * from './elements/ColorPicker/reducer';
```
austingreendev, line 10: What reducer methods are we providing to consumers? Is this intended?

packages/colorpickers/src/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+export * from './elements/ColorPicker';
```
austingreendev, line 8: Should we do named exports to match all other packages? Wild-card exports can affect the ability to tree-shake effectively (especially with `sideEffects: false` provided)

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colordialog.button';
```
austingreendev, line 11: Should all of these styled elements be nested under `colorpickers.*` like other packages?

packages/colorpickers/src/styled/ColorDialog/StyledDialogPreview.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { IRGBColor } from '../../elements/ColorPicker/reducer';
```
austingreendev, line 10: Would it be useful to have a single `types.ts` file for these shared definitions? These values are used in many areas/props besides the reducer.

packages/colorpickers/src/styled/ColorDialog/StyledDialogPreview.ts
```diff
@@ -0,0 +1,47 @@
...
+      props.theme,
+      0.25
+    )}`};
+  border-radius: ${props => props.theme.borderRadii.sm};
+  background: ${props =>
+    typeof props.backgroundColor === 'string'
+      ? props.backgroundColor
+      : `rgba(${props.backgroundColor.red}, ${props.backgroundColor.green}, ${
+          props.backgroundColor.blue
+        }, ${props.backgroundColor.alpha ? props.backgroundColor.alpha / 100 : 0})`};
```
austingreendev, line 38: This is difficult for me to understand as a nested ternary. Could this be refactored to a full function (still inline if you like)?

packages/colorpickers/src/styled/ColorPicker/StyledAlphaField.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { Field } from '@zendeskgarden/react-forms';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpicker.alpha_field';
```
austingreendev, line 12: With the package rename to `colorpickers` should all of the COMPONENT_ID's be renamed as well?

packages/colorpickers/src/styled/ColorPicker/StyledColorWell.ts
```diff
@@ -0,0 +1,63 @@
...
+  bottom: 0;
+  left: 0;
+  background: ${props => `hsl(${props.hue}, 100%, 50%)`};
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledColorWellGradient.defaultProps = {
+  theme: DEFAULT_THEME
+};
```
austingreendev, line 59: [nit] move these default props to the same section as the styled element.

packages/colorpickers/styleguide.config.js
```diff
@@ -0,0 +1,35 @@
+/**
```
austingreendev, line 1: This file shouldn't be needed if we are using storybook for the documentation. You can remove this from the `.template` file if it's still there.

austingreendev (commented):
Looking better! Here is another round of 👀 .

packages/colorpicker/src/styled/ColorPicker/StyledSliderGroup.ts
```diff
@@ -0,0 +1,26 @@
...
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpicker.flex';
+
+export const StyledSliderGroup = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  justify-content: space-between;
```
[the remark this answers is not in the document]
hzhu: Updated, I've made some adjustments to the CSS and measured that each (hue & alpha) slider measures at `12px` each. The latter with a `top: 8px` from the former. This should match the `32px` spec.

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
hzhu: Yeah, updated.

hzhu:
> This package should be named colorpickers (plural) in keeping with most other react-components package naming.

Updated 👍🏽 .

> Is the utils/conversions.ts file duplicating color conversions we get from polished?

The majority of these conversions involve HSV, which does not seem to be supported by `polished`. There are two of the functions that can leverage `polished`: `rgbToHsl` and `hsl2rgb`. However, I  noticed that `polished` is returning a slightly off RGB for the `hsl2rgb` coversion. I think these should be kept so that Garden can have fine-grain control over the color conversion output.

> How do either of these components work in the context of a form?

These components can complement form elements and a Storybook example has been added to demonstrate this.

> How can I invoke the ColorDialog with an alternate trigger (i.e. IconButton)?

The API has been updated slightly to allow for using an alternative trigger and a Storybook example has been added to demonstrate this.

m-lai:
@hzhu Not sure how much we can customize but:

1. The chevron icon for the ColorDialog should be `grey-600` by default and `grey-700` on focus (right now it looks black?)
2. Are we able to retain focus state styling on the ColorDialog when clicked? Right now [it flashes on/off](https://share.getcloudapp.com/bLugrxZX) but it's not like how our Garden components behave (retains focus styling when element is clicked open and it goes away when mouse moves away)
3. From an interaction perspective I feel like the color in the ColorDialog should update when a user is choosing between colors in the ColorPicker while it's open, but right now it only updates after the ColorPicker is closed. Not sure if this can be updated?

hzhu:
Thanks for the review @m-lai . The chevron colors don’t currently sync up to UI states. @steue let’s sync up and iron that out 👍🏽 .

> Are we able to retain focus state styling on the ColorDialog when clicked? Right now it flashes on/off but it's not like how our Garden components behave (retains focus styling when element is clicked open and it goes away when mouse moves away)

Since the color dialog is a button (not a form input), the focus indicator moves from the button to the active element (dialog or colorwell). In Garden’s case, the active element is the dialog. However, if the color well supported keyboard navigation (it currently doesn't), then the focus indicator should be around the color well. This interaction is similar to that of the [dialog example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/datepicker-dialog.html#ex_label) outlined on the W3. 

Garden's color dialog does not have a visual indicator when it is active. This could cause confusion and is a fair discussion topic.

> From an interaction perspective I feel like the color in the ColorDialog should update when a user is choosing between colors in the ColorPicker while it's open, but right now it only updates after the ColorPicker is closed. Not sure if this can be updated?

The two preview boxes (dialog and picker) don’t change at the same time because the color dialog preview is intended to show the final color selection — whereas the color picker preview is intended to show intermediary selections. The APIs are named to reflect the interaction: the color dialog has an `onClose` prop and the color picker has an `onChange` prop. I presumed this was the intended interaction as it follows other open-source color dialog patterns. @steue is this the type of interaction you had in mind?

m-lai:
Cool, I was coming into this cold so thanks for clarifying how it all works!

> The two preview boxes (dialog and picker) don’t change at the same time because the color dialog preview is intended to show the final color selection — whereas the color picker preview is intended to show intermediary selections. The APIs are named to reflect the interaction: the color dialog has an `onClose` prop and the color picker has an `onChange` prop. I presumed this was the intended interaction as it follows other open-source color dialog patterns. @steue is this the type of interaction you had in mind?

I shall leave this to Steph re: intention. Just was thinking most of the color pickers I've used have had the dialog update the colors in real time but it might not be what Steph had mind. Cheers!

jzempel:
You're instincts are correct, @m-lai. For the right user experience, the dialog does need to update the selected color in real time – similar to Sketch.

packages/colorpicker/stories/examples/ColorPicker.tsx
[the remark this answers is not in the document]
hzhu: I went ahead and made this adjustment locally, but wanted to confirm if this is how the labels should be presented. These aren't props, so won't have a description:

Screen Shot 2021-02-02 at 9 38 06 AM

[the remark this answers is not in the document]
jzempel: I don't think its that big of a deal here, but in other cases we've just used natural language as a slight differentiator:
- Alpha slider label
- Hue slider label
- Hex label
- Red label
- Green label
- Blue label
- Alpha label

ymmv

packages/colorpicker/src/styled/ColorPicker/StyledLabel.ts
```diff
@@ -0,0 +1,25 @@
...
+})`
+  font-weight: ${props => props.theme.fontWeights.regular};
```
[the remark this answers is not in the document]
hzhu: I added styling for ellipsis overflow. Let me know if this addressed the problem:

Screen Shot 2021-02-02 at 9 26 38 AM

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const throttledChange = throttle(e => {
+    if (container.current) {
+      const nextHsv = calculateNextHsv(e, hsv, container.current, rtl);
+      const saturationPosition = getSaturationPosition(e, container.current);
+
+      setPosition(saturationPosition);
+      onChange && onChange(nextHsv, e);
+    }
+  }, 50);
```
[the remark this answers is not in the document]
hzhu: I wasn't able to re-create the unexpected calls, could you outline the steps to re-create the unexpected calls so I could investigate further?

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
hzhu: In what scenario would `ColorWell` unmount prior to the `mouseup` event? I'm unable to think of such a scenario. Would there be bigger issues than lingering event listeners in such a scenario?

packages/colorpickers/src/styled/ColorDialog/StyledDialogPreview.ts
hzhu: A new `types.ts` has been created for shared definitions. I've removed the `exports` that aren't used and aren't intended for consumer usage.

packages/colorpickers/src/index.ts
hzhu: Only two types definitions. Originally doing this to get rid of some HMR warnings. But updated to use named exports for specific values.

packages/colorpickers/src/index.ts
hzhu: Yeah, updated 👍🏽 .

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
hzhu: Updated and removed `handleChange`. Thanks!

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
hzhu: I'll leave as-is for now until the initial state gets expensive to calculate. At that point, this could be moved into a function and leverage lazy initial state.

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [referenceElement, setReferenceElement] = useState<HTMLButtonElement | null>();
+  const mergedRef = mergeRefs([ref, buttonRef]);
+
+  const onClick = () => {
+    if (referenceElement) {
+      setReferenceElement(null);
+    } else {
+      setReferenceElement(buttonRef.current);
+    }
+  };
```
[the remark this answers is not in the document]
hzhu: It's stateful but I didn't see the benefit to wrapping it in a `useCallback`.

packages/colorpickers/src/elements/ColorPicker/index.tsx
hzhu: It should be accurate. If the `onChange` was called with the current state (not the updated state) then the color picker should not update. 

However, it does seem to be updating. I don't think there is a problem with this `useEffect` but happy to investigate further provided more detail.

packages/colorpickers/src/styled/ColorDialog/StyledDialogPreview.ts
hzhu: I think having it in a separate function would be nice. Updated.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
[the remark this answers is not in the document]
austingreendev: With the throttle being at 50ms this might not be something that is easy to reproduce. I would still lean toward using `useRef` or `useCallback` to memorize this as mentioned in the stackoverflow thread.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
austingreendev: It would be rare but is still a possibility. Unmounts can be triggered by most things outside of our control. 

In general, the best practice is to remove global event listeners within a useEffect, right? [MultiThumbRange example](https://github.com/zendeskgarden/react-components/blob/main/packages/forms/src/elements/MultiThumbRange.tsx#L94-L102)

packages/colorpickers/src/elements/ColorPicker/index.tsx
austingreendev: I think the `state` dependency is correct, I'm mostly wondering about the `onChange`. With the default `Object.is` comparison in `useEffect` changing that prop to a new function would trigger an `onChange(currentState)` to be immediately called. Is that the expected API?

packages/colorpickers/src/elements/ColorPicker/reducer.ts
hzhu: > Does React provide any generics for this type?

Not to my knowledge. How can I find out? The `useReducer` is correctly typed here. Is there an alternative way to typing out the `userReducer` function?

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
hzhu: Upon further analysis, the position of the `ColorWell` thumb can be derived from props alone. I'm going to simplify this file and remove local component state.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
hzhu: The `MultiThumbRange` code aligns with how I think about registering and cleaning up listeners in a `useEffect`. I think it's best practice to remove event listeners registered within a `useEffect` in its cleanup function. 

However, these event listeners aren't registered inside a `useEffect`. It seems fair to register the listeners on `mousedown` and remove them on the corresponding `mouseup` event.

packages/colorpickers/src/elements/ColorPicker/index.tsx
hzhu: > With the default Object.is comparison in useEffect changing that prop to a new function would trigger an onChange(currentState) to be immediately called. Is that the expected API?

I think this is the expected API. By the time the `useEffect` is called, it's called with the (expected) updated state. I've also verified that the `useEffect` is called only once per state update.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -22,37 +22,52 @@ interface IColorWellProps {
...
+      if (container.current) {
+        const nextHsv = calculateNextHsv(e, hsv, container.current, rtl);
 
-      onChange && onChange(nextHsv, e);
-    }
-  }, 50);
+        onChange && onChange(nextHsv, e);
+      }
+    }, 50);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
austingreendev, line 42: Is this to remove the need for the `onChange` dependency? Why is this necessary?

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -22,37 +22,52 @@ interface IColorWellProps {
...
   };
 
+  useEffect(() => {
+    return () => {
+      throttledChange.cancel();
+      window.removeEventListener('mousemove', handleMouseMove);
+      window.removeEventListener('mouseup', handleMouseUp);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [throttledChange]);
```
austingreendev, line 66: Is there a reason that we don't include the `handleMouseMove` and `handleMouseUp` in the dependency array? I think the warnings here are valid. On each render these callbacks would receive new references and not be removed correctly (except in the mouse up logic).

Some suggestions:
- Remove all of the rule-of-hooks exceptions
- Since `throttledChange` is now memorized we should migrate the mouse handlers to `useCallback`

steue:
@m-lai @jzempel thank you both for feedback! replying here for posterity. 

1. re: chevrons, yes they should change in shade between default and hover / active states.
2. re: focus, as henry mentioned, focus should be brought to the hex value when dialog is opened.
3. re: selected color updating. yes, that is the intended behavior! as a user is updating any values within the picker, the dialog should update in real time. this is different from other interaction patterns where the user needs to [confirm the selection within the dialog](https://share.getcloudapp.com/yAu6YDlG).

henry and i have synced on these items and will be addressed in the next commits or have been addressed already. ☺️

steue:
i noticed a bug(?) when clearing out any of the rgba inputs and then changing any of the other values. the cleared input stays blank, even though it looks like the previous valid value is still maintained. is it possible to resurface the previous valid value when the other values are changed?

https://user-images.githubusercontent.com/39511661/107328372-77413200-6a63-11eb-84bb-c552b5f87851.mp4

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [isMouseDown, setIsMouseDown] = React.useState(false);
+
+  // State for thumb position when change come from mouse activity on the color well
+  const [x, setX] = useState(0);
+  const [y, setY] = useState(0);
+  const { topFromMouse, leftFromMouse } = getThumbPosition(x, y, rtl, container);
+
+  // State for thumb position when change come from saturation and lightness props without mouse activity
+  const [topPosition, setTopPosition] = useState(0);
+  const [leftPosition, setLeftPosition] = useState(0);
```
austingreendev, line 45: This state seems like a duplication of X and Y from above. Is there a reason that saturation and lightness aren't able to manipulate X and Y directly when changed?

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+  saturation: number;
+  lightness: number;
+  onChange?: (hsv: IHSVColor, event: MouseEvent) => void;
+}
+
+export const ColorWell: React.FC<IColorWellProps> = ({ hue, saturation, lightness, onChange }) => {
+  const theme = useContext(ThemeContext);
+  const environment = useDocument(theme);
+  const { rtl } = useContext(ThemeContext);
+  const container = useRef<HTMLDivElement>(null);
```
austingreendev, line 33: This should be named `containerRef` to follow naming conventions. This also allows the rules-of-hooks to understand dependencies better.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { topFromMouse, leftFromMouse } = getThumbPosition(x, y, rtl, container);
+
+  // State for thumb position when change come from saturation and lightness props without mouse activity
+  const [topPosition, setTopPosition] = useState(0);
+  const [leftPosition, setLeftPosition] = useState(0);
+
+  useEffect(() => {
+    setTopPosition(100 - hsv.v);
+    setLeftPosition(rtl ? 100 - hsv.s : hsv.s);
+  }, [hsv.s, hsv.v, rtl]);
```
austingreendev, line 50: Should the `hsv` calculation be memoized and its entire object be used as a dependency?

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export const ColorWell: React.FC<IColorWellProps> = ({ hue, saturation, lightness, onChange }) => {
+  const theme = useContext(ThemeContext);
+  const environment = useDocument(theme);
+  const { rtl } = useContext(ThemeContext);
+  const container = useRef<HTMLDivElement>(null);
+  const hsv = hslToHsv(hue, saturation, lightness);
+  const mouseActiveRef = useRef(false);
+  const [isMouseDown, setIsMouseDown] = React.useState(false);
```
austingreendev, line 36: We don't reference with `React.*` for other hooks. Let's import `useState` directly.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [leftPosition, setLeftPosition] = useState(0);
+
+  useEffect(() => {
+    setTopPosition(100 - hsv.v);
+    setLeftPosition(rtl ? 100 - hsv.s : hsv.s);
+  }, [hsv.s, hsv.v, rtl]);
+
+  useEffect(() => {
+    if (environment) {
+      const originalUserSelect = environment.body.style.userSelect;
```
austingreendev, line 54: I don't think I understand this logic correctly. If `isMouseDown=false` then `environment.body.style.userSelect` is applied to itself.

This assignment makes more sense if it's referenced within the effect cleanup.

packages/colorpickers/src/elements/ColorPicker/ColorWell.tsx
```diff
@@ -0,0 +1,137 @@
...
+        environment.body.style.webkitUserSelect = originalUserSelect;
+        (environment.body.style as any).msUserSelect = originalUserSelect;
+      }
+    }
+
+    return () => {
+      if (environment) {
+        environment.body.style.userSelect = 'auto';
+        environment.body.style.webkitUserSelect = 'auto';
+        environment.body.style['msUserSelect' as any] = 'auto';
```
austingreendev, line 71: I think this will break consuming applications. Anytime the ColorWell is used this would modify the body style of the application and not return it to the previous state.

Why are we removing the ability to select text globally and not just scoped to this component?

packages/colorpickers/src/utils/types.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface IColor {
+  hex: string;
+  hue: number;
+  saturation: number;
+  lightness: number;
+  red: number;
+  green: number;
+  blue: number;
+  alpha: number;
+}
```
austingreendev, line 17: [nit] missing space

packages/colorpickers/src/utils/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  s: number;
+  l: number;
+  a?: number;
+}
+
+export interface IRGBColor {
+  red: number;
+  green: number;
+  blue: number;
+  alpha?: number;
```
austingreendev, line 35: All other color types use abbreviations. Is there a need for us to spell `r,g,b,a` explicitly?

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,473 @@
...
+export const ColorPicker = forwardRef<HTMLDivElement, IColorPickerProps>(
+  ({ color, defaultColor, labels = {}, autofocus, onChange, ...props }, ref) => {
+    const isControlled = color !== null && color !== undefined;
+    const [uncontrolledState, dispatch] = useReducer(reducer, toState(defaultColor));
+
+    useEffect(() => {
+      if (isControlled === false) {
+        onChange && onChange(uncontrolledState);
+      }
+    }, [uncontrolledState, onChange, isControlled]);
```
austingreendev, line 82: Could you go into the reasoning behind having both a controlled vs uncontrolled state manager here?

It looks like the state objects below replicate what is already being accomplished in the reducer.

austingreendev (changes requested):

hzhu:
Closing this pull request in favor of opening a fresh one. We will get a fresh comment thread and it will be easier to review.

PR #1011, opened 2021-02-05
fix(deps): ensure all packages include prop-types and lodash correctly

austingreendev:
## Description

While auditing our dependencies I found 2 issues across our common libraries: `prop-types` and `lodash.debounce`

## `prop-types`

We currently list `prop-types` as a peer dependency. This makes sense since it's similar to `react` and `react-dom`, but after reading their documentation. [They suggest always including it as a dependency with a specific open-range of `^15.5.7`](https://github.com/facebook/prop-types#how-to-depend-on-this-package)

I have also removed all references to `prop-types` in our README and updated our package template.

## `lodash.debounce`

There were some packages that erroneously listed `lodash.debounce` as a devDependency. Those have been updated.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -92,7 +92,6 @@ dependencies include:
   - `polished`
 - Peer dependencies
   - `@zendeskgarden/react-theming`
-  - `prop-types`
   - `react`
   - `react-dom`
   - `styled-components`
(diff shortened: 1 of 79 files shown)


jzempel (approved):
🙌

PR #1012, opened 2021-02-05
feat(typography): add line highlighting to CodeBlock component

jzempel:
## Description

- adds a new `highlightLines` prop for specifying an array of lines to highlight
- while restructuring the component, I discovered the a11y need for [scrollable region focus](https://dequeuniversity.com/rules/axe/4.1/scrollable-region-focusable) if the code content exceeds the boundaries of the code block container; a `tabindex` is now applied accordingly

### Details

@steue in the situations where the `CodeBlock` allows keyboard focus, I've applied our standard `box-shadow` treatment. Please double-check that this maps to design intent. :pray:

The `ScrollableRegionFocusable` logic needs to be pulled out into a new `react-containers` hook.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/typography/src/elements/CodeBlock.tsx
+++ b/packages/typography/src/elements/CodeBlock.tsx
@@ -5,9 +5,15 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { HTMLAttributes } from 'react';
+import React, { HTMLAttributes, useCallback, useEffect, useRef, useState } from 'react';
 import Highlight, { Language, Prism } from 'prism-react-renderer';
-import { StyledCodeBlock, StyledCodeBlockLine, StyledCodeBlockToken } from '../styled';
+import debounce from 'lodash.debounce';
+import {
+  StyledCodeBlock,
+  StyledCodeBlockContainer,
+  StyledCodeBlockLine,
+  StyledCodeBlockToken
+} from '../styled';
 
 export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
   /** Selects the language used by the Prism tokenizer */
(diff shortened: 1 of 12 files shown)


packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -18,13 +24,41 @@ export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
...
  */
 export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
-  ({ children, isLight, isNumbered, language, size, ...other }, ref) => {
+  (
+    { children, containerProps, highlightLines, isLight, isNumbered, language, size, ...other },
+    ref
+  ) => {
+    const [containerTabIndex, setContainerTabIndex] = useState(-1);
+    const containerRef = useRef<HTMLDivElement>(null);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
austingreendev, line 43: Sometimes the eslint `rules-of-hooks` can get confused when not using an immediate inline function with `useCallback`. Since you're returning a reference, `useMemo` might be more applicable here. (we ran into this with the ColorPicker PR)

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -18,13 +24,41 @@ export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
...
+    const updateContainerTabIndex = useCallback(
+      debounce(() => {
+        if (containerRef.current) {
+          const codeBlock = containerRef.current.children[0];
```
austingreendev, line 47: Would [firstChild](https://developer.mozilla.org/en-US/docs/Web/API/Node/firstChild) work in this case?

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -36,32 +70,44 @@ export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
       _size = 'lg';
     }
 
+    useEffect(() => {
+      addEventListener('resize', updateContainerTabIndex);
+      updateContainerTabIndex();
+
+      return () => {
+        removeEventListener('resize', updateContainerTabIndex);
```
austingreendev, line 78: Do we also need to cancel any pending denounces here as well? [debounce.cancel()](https://lodash.com/docs/4.17.15#debounce)

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -18,13 +24,41 @@ export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
...
+          const containerWidth = containerRef.current.offsetWidth;
+
+          if (codeBlockWidth > containerWidth || codeBlockHeight > containerHeight) {
+            setContainerTabIndex(0);
+          } else {
+            setContainerTabIndex(-1);
+          }
+        }
+      }, 100),
+      [containerRef, setContainerTabIndex]
```
austingreendev, line 60: Both of these shouldn't be required dependencies as their references don't change. Do you still get these eslint requirements when using `useMemo`?

packages/typography/src/elements/CodeBlock.tsx
jzempel: Gives me a Node vs the desired Element which is dereferenced for width/height.

packages/typography/src/elements/CodeBlock.tsx
jzempel: After giving this a closer look, I'm not convinced there's much here that benefits from memoization. Please take a look at the updates and let me know what you think.

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -18,13 +24,37 @@ export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
...
  */
 export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
-  ({ children, isLight, isNumbered, language, size, ...other }, ref) => {
+  (
+    { children, containerProps, highlightLines, isLight, isNumbered, language, size, ...other },
+    ref
+  ) => {
+    const [containerTabIndex, setContainerTabIndex] = useState(-1);
+    const containerRef = useRef<HTMLDivElement>(null);
+    const updateContainerTabIndex = debounce(() => {
```
austingreendev, line 43: With this no longer being memoized, would a re-render (not including props that are dependencies within the useEffect) remove the correct listener?

Even with no dependencies I think memoizing this is necessary. This would create a new debounce reference on every render.

steue (approved):
lgtm! thanks for doing this jz!

austingreendev (approved):
Thanks for updating the debounce logic.

PR #1013, opened 2021-02-06
fix(accordions): Stop stepper's counter incrementing twice in React.StrictMode

### LINKED ISSUE #849: Stepper not rendering properly in React.StrictMode
## Expectations
Stepper's label will start from 1 and increment by 1 as shown below:
 
image

## Reality
Stepper's label starts from 2 and incremented by 2.

image

## Steps to Reproduce

1. Create an app using Create React App
2. Install `@zendeskgarden/react-theming` and `@zendeskgarden/react-accordions`
3. Use the example of `Stepper` from https://zendeskgarden.github.io/react-components/accordions/ 

## Example 
https://codesandbox.io/s/zendesk-garden-stepper-w5ujw

MarkSFrancis:
## Description

In [React.StrictMode](https://reactjs.org/docs/strict-mode.html), components are rendered twice. This causes the `Label` in the `Stepper` to go up in twos (e.g. step 2, 4, 6  instead of 1, 2, 3), which is undesired behaviour.

> Strict mode can’t automatically detect side effects for you, but it can help you spot them by making them a little more deterministic. This is done by intentionally double-invoking the following functions:
> 
> * Class component constructor, render, and shouldComponentUpdate methods
> * Class component static getDerivedStateFromProps method
> * Function component bodies
> * State updater functions (the first argument to setState)
> * Functions passed to useState, useMemo, or useReducer

## Detail

This change uses useEffect to increment/ decrement the value as the component is unmounted/ remounted, fixing the issue of double-rendering components. 

Tests have been added which prove that the change is backwards compatible, but also fixes the issue. These tests will not pass in the older version.

Closes #849

## Checklist

Not sure which of the checklist apply - please can someone confirm?

      designer as a reviewer)~
- [ ] ~:globe_with_meridians: demo is up-to-date (`yarn start`)~

### DIFF
--- a/packages/accordions/src/elements/stepper/components/Step.spec.tsx
+++ b/packages/accordions/src/elements/stepper/components/Step.spec.tsx
@@ -48,7 +48,7 @@ describe('Step', () => {
   });
 
   it('does not change step number when running outside strict mode', async () => {
-    const { findByText } = render(
+    const { findAllByText } = render(
       <Stepper>
         <Stepper.Step>
           <Stepper.Label />
@@ -59,12 +59,14 @@ describe('Step', () => {
       </Stepper>
     );
 
-    expect(await findByText('1')).toBeInTheDocument();
-    expect(await findByText('2')).toBeInTheDocument();
+    const elements = await findAllByText(/1|2/u);
+
+    expect(elements[0]).toContainHTML('1');
+    expect(elements[1]).toContainHTML('2');
   });
 
   it('does not change step number when running in strict mode', async () => {
-    const { findByText } = render(
+    const { findAllByText } = render(
       <React.StrictMode>
         <Stepper>
           <Stepper.Step>
@@ -77,7 +79,9 @@ describe('Step', () => {
       </React.StrictMode>
     );
 
-    expect(await findByText('1')).toBeTruthy();
-    expect(await findByText('2')).toBeTruthy();
+    const elements = await findAllByText(/1|2/u);
+
+    expect(elements[0]).toContainHTML('1');
+    expect(elements[1]).toContainHTML('2');
   });
 });
--- a/packages/accordions/src/elements/stepper/components/Step.spec.tsx
+++ b/packages/accordions/src/elements/stepper/components/Step.spec.tsx
@@ -46,4 +46,38 @@ describe('Step', () => {
 
     expect(queryAllByTestId('step-line')).toHaveLength(0);
   });
+
+  it('does not change step number when running outside strict mode', async () => {
+    const { findByText } = render(
+      <Stepper>
+        <Stepper.Step>
+          <Stepper.Label />
+        </Stepper.Step>
+        <Stepper.Step>
+          <Stepper.Label />
+        </Stepper.Step>
+      </Stepper>
+    );
+
+    expect(await findByText('1')).toBeInTheDocument();
+    expect(await findByText('2')).toBeInTheDocument();
+  });
+
+  it('does not change step number when running in strict mode', async () => {
+    const { findByText } = render(
+      <React.StrictMode>
+        <Stepper>
+          <Stepper.Step>
+            <Stepper.Label />
+          </Stepper.Step>
+          <Stepper.Step>
+            <Stepper.Label />
+          </Stepper.Step>
+        </Stepper>
+      </React.StrictMode>
+    );
+
+    expect(await findByText('1')).toBeTruthy();
+    expect(await findByText('2')).toBeTruthy();
+  });
 });
--- a/packages/accordions/src/elements/stepper/components/Step.tsx
+++ b/packages/accordions/src/elements/stepper/components/Step.tsx
@@ -5,19 +5,42 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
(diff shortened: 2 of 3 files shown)


packages/accordions/src/elements/stepper/components/Step.tsx
```diff
@@ -5,19 +5,42 @@
...
-  const { currentIndexRef, isHorizontal } = useStepperContext();
-  const stepIndexRef = useRef(currentIndexRef.current++);
-  const stepContextValue = { currentStepIndex: stepIndexRef.current };
+  const [currentStepIndex, context] = useStep();
 
   return (
-    <StepContext.Provider value={stepContextValue}>
-      <StyledStep ref={ref} isHorizontal={isHorizontal} {...props}>
-        {isHorizontal && <StyledLine data-test-id="step-line" />}
+    <StepContext.Provider value={{ currentStepIndex }}>
```
jzempel, line 41: @hzhu in the (near) future, I'd like for us to find our way to a pure CSS solution to this problem. Reference: https://github.com/zendeskgarden/react-components/blob/ceb91ac0ee5f38192bb66cf3f999bc3031aaa985/packages/typography/src/styled/StyledCodeBlockLine.ts#L24-L25

packages/accordions/src/elements/stepper/components/Step.spec.tsx
```diff
@@ -46,4 +46,38 @@ describe('Step', () => {
 
     expect(queryAllByTestId('step-line')).toHaveLength(0);
   });
+
+  it('does not change step number when running outside strict mode', async () => {
```
austingreendev, line 50: I think these tests are accurate, but should we also be asserting that the steps are in the correct order as well?

packages/accordions/src/elements/stepper/components/Step.tsx
```diff
@@ -5,19 +5,34 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { forwardRef, useRef, LiHTMLAttributes } from 'react';
+import React, { forwardRef, LiHTMLAttributes, useEffect, useState } from 'react';
 import { StyledStep, StyledLine } from '../../../styled';
 import { StepContext, useStepperContext } from '../../../utils';
 
+const useStep = () => {
```
austingreendev, line 12: Even though this abstraction is helpful; if we aren't using this hook multiple places it might make more sense to have the logic live within `<Step />` directly. This also would require returning the remaining `context` values as they aren't involved in the step calculation.

packages/accordions/src/elements/stepper/components/Step.tsx
```diff
@@ -5,19 +5,34 @@
...
+  const { currentIndexRef, ...context } = useStepperContext();
+  const [index, setIndex] = useState(currentIndexRef.current);
+
+  useEffect(() => {
+    setIndex(currentIndexRef.current);
+    currentIndexRef.current++;
+    const currentIndex = currentIndexRef;
+
+    return () => {
+      currentIndex.current--;
```
austingreendev, line 22: Is this decrement effect cleanup necessary? The `currentIndexRef` is set to zero whenever the parent `<Stepper />` is rendered.

packages/accordions/src/elements/stepper/components/Step.tsx
austingreendev: The CSS solution would help with the visible step name, but we also have to track it for the `activeIndex` styling in controlled modes.

packages/accordions/src/elements/stepper/components/Step.tsx
MarkSFrancis: Sometimes this is re-rendered when the parent component is not re-rendered. This is what StrictMode does. It'll render these child components twice without rendering `<Stepper />`

packages/accordions/src/elements/stepper/components/Step.tsx
MarkSFrancis: Happy to change this

packages/accordions/src/elements/stepper/components/Step.tsx
austingreendev: 👍  Thanks for the explanation

packages/accordions/src/elements/stepper/components/Step.tsx
austingreendev: Let's bring this logic back into the component. Overall the logic looks good to me.

packages/accordions/src/elements/stepper/components/Step.spec.tsx
MarkSFrancis: I'm not sure what the best way to do this is, so I've used `findAllByText`, and then asserted the order of elements in the array

austingreendev: LGTM 👍

austingreendev (approved):
Looks good to me! Thanks for taking the time to tweak this @MarkSFrancis

hzhu (approved):
Thanks @MarkSFrancis! Once the snapshots are updated and the build is passing, this PR can be merged.

austingreendev:
Thanks for the help with this @MarkSFrancis. This will get published during our weekly release next week.

PR #1016, opened 2021-02-11
feat: introduce new state-management example

austingreendev:
## Description

@hzhu and @jzempel here is my first crack at another potential state management pattern for ColorPickers. This implementation is a fully controlled/uncontrolled pattern that should be roughly an apples-to-apples comparison to the existing branch.

This updated pattern allows us to treat invalid input with more flexibility, so I've updated the logic to align with the [Fluent-UI ColorPicker](https://developer.microsoft.com/en-us/fluentui#/controls/web/colorpicker) and [react-color](https://casesandberg.github.io/react-color/). These changes required some test updates. I've commented out a few tests just to get this out the door, but they're unrelated to the state management logic.

## Detail

The most difficult detail in ColorPicker, is that a single input change can affect other style values. Both of the libraries above maintain a "temporary" color within local state that stores these values separate from the inputs. I've used this same pattern in the implementation.

This temporary color also allows us to handle input errors more aggressively. To match the other libraries it now resets the input values based on the current controlled (or temporary) color.

**DO NOT MERGE. TEMPORARY DRAFT.**


packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -72,73 +70,99 @@ export interface IColorPickerProps
...
+      }
+
+      previousStateColorRef.current = state.color;
+    }, [color, onChange, state.color]);
 
     useEffect(() => {
-      if (isControlled === false) {
-        onChange && onChange(uncontrolledState);
+      if (!areColorsEqual(controlledColor, previousComputedColorRef.current)) {
+        dispatch({ type: 'RESET_COLOR', payload: controlledColor });
```
austingreendev, line 106: Whenever a new controlled value is applied we reset the temporary color and the associated inputs.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -72,73 +70,99 @@ export interface IColorPickerProps
...
-        } else {
-          dispatch({
-            type: 'SATURATION_CHANGE',
-            payload: hsv
-          });
-        }
-      },
-      [isControlled, onChange, state.alpha]
-    );
+    const handleHueChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
```
austingreendev, line 119: All color computation logic now lives within the `reducer`. For both controlled and uncontrolled usages.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -72,73 +70,99 @@ export interface IColorPickerProps
...
+      dispatch({ type: 'RESET_COLOR', payload: controlledColor });
+    }, [controlledColor]);
 
     return (
       <StyledColorPicker ref={ref} {...props}>
         <ColorWell
-          hue={state.hue}
-          saturation={state.saturation}
-          lightness={state.lightness}
+          hue={controlledColor.hue}
```
austingreendev, line 154: All of these value pass-throughs could probably be simplified with a unified `colo={IColor}` prop, but that is unrelated to state management.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -214,59 +206,11 @@ export const ColorPicker = forwardRef<HTMLDivElement, IColorPickerProps>(
...
-                          blue: rgb.blue,
-                          alpha: state.alpha
-                        });
-                    } else {
-                      dispatch({ type: 'HEX_CHANGE', payload: hexInputString });
-                    }
-                  }
-                }
-              }}
+              onBlur={handleBlur}
```
austingreendev, line 212: We now reset values based on the computed color whenever a blur event occurs. This cleans up _valid_ number inputs like `e` and other non-int notations.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -18,12 +19,57 @@ type ColorPickerActionTypes =
...
+interface IColorPickerState {
+  color: IColor;
+  hexInput: string;
+  redInput: string;
+  blueInput: string;
+  greenInput: string;
+  alphaInput: string;
+}
+
+export function convertStringToColor(colorString: string): IColor | undefined {
```
austingreendev, line 34: Extracted from `getInitialState` as this is used in many areas. Could probably be moved to the color conversion utility, but left it here for this review.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -18,12 +19,57 @@ type ColorPickerActionTypes =
...
+  const colorA = typeof a === 'string' ? convertStringToColor(a) : a;
+  const colorB = typeof b === 'string' ? convertStringToColor(b) : b;
 
-export function toState(color?: string | IColor) {
-  const white = {
+  if (colorA === undefined || colorB === undefined) {
+    return false;
+  }
+
+  return isEqual(colorA, colorB);
```
austingreendev, line 68: `Object.is()` equivalency doesn't cut it within our effects. We have to perform a shallow comparison against these values.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -133,73 +172,136 @@ export const reducer: ReducerType = (state, action) => {
...
-      return state;
+      return {
+        ...state,
+        hexInput: action.payload,
+        color
+      };
     }
     case 'RED_CHANGE': {
-      const red = Number(action.payload);
+      let red = parseInt(action.payload, 10);
```
austingreendev, line 182: `parseInt` is a more specific type of number parsing that will help remove invalid notation.

packages/colorpickers/src/elements/ColorPicker/reducer.ts
```diff
@@ -133,73 +172,136 @@ export const reducer: ReducerType = (state, action) => {
...
+      }
 
       return {
         ...state,
-        hex,
-        red: action.payload === '' ? state.red : red,
-        hue: hsl.h,
-        lightness: hsl.l,
-        saturation: hsl.s
+        redInput: action.payload,
```
austingreendev, line 209: Input can be updated regardless of whether the temporary color has changed. This matches the 3rd party libraries referenced in the description.

jzempel:
These changes and suggestions appear to have helpful merit.

hzhu (approved):
LGTM! Thanks, @austingreendev for the suggestions. These have been incorporated in the new color picker pull request.

PR #1018, opened 2021-02-12
feat(colorpickers): introduce color pickers package

hzhu:
## Description

This PR introduces the `react-colorpickers` package. It offers a fresh start and irons out many of the suggestions from #999.

## Detail

Please see the corresponding Abstract assets and API RFC doc for details.

## Checklist

      designer as a reviewer)


packages/colorpickers/README.md
```diff
@@ -0,0 +1,28 @@
...
+This package includes components related to colorpickers in the
+[Garden Design System](https://zendeskgarden.github.io/).
+
+## Installation
+
+```sh
+npm install @zendeskgarden/react-colorpickers
+
+# Peer Dependencies - Also Required
+npm install react react-dom prop-types styled-components @zendeskgarden/react-theming
```
austingreendev, line 12: [nit] prop-types is no longer a required peer dependency

packages/colorpickers/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "polished": "^4.0.0",
+    "react-merge-refs": "^1.1.0"
+  },
+  "peerDependencies": {
+    "@zendeskgarden/react-theming": "^8.1.0",
+    "prop-types": "^15.6.1",
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
+    "styled-components": "^4.2.0 || ^5.0.0"
+  },
```
austingreendev, line 38: We have updated the peer version ranges for `react` and `react-dom`. Let's update it here as well.

We have also moved `prop-types` to be a direct dependency.

austingreendev (changes requested):

austingreendev (commented):
LGTM on the dev side!

packages/colorpickers/package.json
hzhu: Updated, thanks!

nyckie (approved):

steue:
looks great so far! wondering if this is a bug? the alpha thumb is still blue. 😅

Image 2021-02-17 at 11 11 08 AM

steue:
> looks great so far! wondering if this is a bug? the alpha thumb is still blue. 😅
> 
> Image 2021-02-17 at 11 11 08 AM

nvm false alarm. after reloading a 5th time it didn't surface

steue:
@hzhu it looks and behaves as intended! just one final nit:

the faux input / color dialog's spacing looks off by 1 or 2 pixels. after inspection and the `StyledButton` looks to be spaced with em instead of pixels, not sure if this is important, but after changing the padding to 12px, it looked properly spaced. also, the `StyledDialogPreview` and chevron off center vertically, but after changing the margin to -1px, both fell back into place. i'm not sure if these are appropriate code changes, but the visual result is what it should look like.

Image 2021-02-17 at 11 42 55 AM

packages/colorpickers/src/styled/ColorPicker/StyledColorWellThumb.ts
```diff
@@ -0,0 +1,46 @@
...
+  border: 2px solid ${props => props.theme.colors.background};
+  cursor: default;
+  /* stylelint-disable */
+  transform: translate(
+    -${props => props.theme.space.base * 2.5}px,
+    -${props => props.theme.space.base * 2.5}px
+  );
+  /* stylelint-enable */
+  border-radius: 50%;
+  box-shadow: 0 4px 8px 0 ${props => getColor('neutralHue', 800, props.theme, 0.24)};
```
jzempel, line 37: Express `4px 8px` via theming values.

packages/colorpickers/src/styled/ColorPicker/StyledColorWell.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colorpicker.colorwell';
```
jzempel, line 11: It's problematic that the COMPONENT_ID is being used for two components – preventing future theme targeting.

packages/colorpickers/src/styled/ColorDialog/StyledFauxInput.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const COMPONENT_ID = 'colorpickers.colordialog.fauxinput';
+
+export const StyledFauxInput = styled(FauxInput).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  padding: 0;
+  width: ${props => props.theme.space.base * 17}px;
+  height: 40px;
```
jzempel, line 20: Express via `theme` values.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+            />
+          </StyledRGBAField>
+        </StyledInputGroup>
+      </StyledColorPicker>
+    );
+  }
+);
+
+ColorPicker.defaultProps = {
+  defaultColor: '#ffffff'
```
jzempel, line 259: Prefer shorthand `#fff` as this is the value that the ColorPicker renders for white.

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  StyledDialogPreview,
+  StyledTooltipModal,
+  StyledTooltipBody
+} from '../../styled';
+
+export interface IColorDialogProps extends IColorPickerProps {
+  /**
+   * Handles close actions. Can be triggered from the backdrop.
+   *
+   * @param {Object} color a color picker state
```
jzempel, line 37: Capital "A" here.

(reference https://garden.zendesk.com/components/autocomplete#dropdown)

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {
+  StyledIcon,
+  StyledButton,
+  StyledFauxInput,
+  StyledDialogPreview,
+  StyledTooltipModal,
+  StyledTooltipBody
+} from '../../styled';
+
+export interface IColorDialogProps extends IColorPickerProps {
```
jzempel, line 33: This component is missing the following menu-style props: `eventsEnabled`, `hasArrow`, `isAnimated` (true by default), `popperModifiers`, and `zIndex`.

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    return (
+      <>
+        {children ? (
+          cloneElement(Children.only(children as ReactElement), {
+            onClick,
+            ref: mergedRef
+          })
+        ) : (
+          <StyledFauxInput tabIndex={-1} disabled={disabled}>
+            <StyledButton disabled={disabled} ref={mergedRef} onClick={onClick} {...props}>
```
jzempel, line 83: This construct is problematic. This component needs to function as a Garden `<button>`. Right now, it behaves as some kind of hybrid between a `button` and a `select`. Functionally, it should behave just like https://garden.zendesk.com/components/menu#default – handling mouse and keyboard focus accordingly.

The `StyledButton` should receive all of the intended CSS to get the backround & borders right per designs. I would suggest that `StyledFauxInput` is not needed.

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+              <Button.EndIcon>
+                <StyledIcon isRotated={referenceElement}>
+                  <Chevron />
+                </StyledIcon>
+              </Button.EndIcon>
+            </StyledButton>
+          </StyledFauxInput>
+        )}
+        <StyledTooltipModal
+          hasArrow={false}
```
jzempel, line 94: Receive this value (and others) via a prop.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Handles color picker changes
+   *
+   * @param {Object} color a color picker state
+   */
+  onChange?: (color: IColor) => void;
+  /** Replaces the default labels within the color picker */
+  labels?: IColorPickerLabels;
+  /** Autofocuses the hex input element */
+  autofocus?: boolean;
```
jzempel, line 65: Let's go over some of this wording offline. We need to remain structurally consistent with existing website copy.

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
```diff
@@ -0,0 +1,61 @@
...
+      color: ${activeColor};
+    }
+
+    &:disabled {
+      color: ${disabledForegroundColor};
+    }
+  `;
+};
+
+export const StyledButton = styled(Button).attrs({
```
jzempel, line 40: In addition to the items @steue mentioned, there seem to be some styling problems with this component. 

1. Chevron-side padding overflows the button
2. The component isn't rendering right in IE11

Screen Shot 2021-02-17 at 5 26 01 PM

jzempel (changes requested):

packages/colorpickers/src/styled/ColorPicker/StyledHexField.ts
```diff
@@ -0,0 +1,28 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { Field } from '@zendeskgarden/react-forms';
+
+const COMPONENT_ID = 'colorpickers.colorpicker.hex_field';
+
+export const StyledHexField = styled(Field).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
```
jzempel, line 17: Please apply a `'spellcheck': false` attribute to this field. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck

jzempel (changes requested):

packages/colorpickers/src/styled/ColorPicker/StyledAlphaField.ts
```diff
@@ -0,0 +1,42 @@
...
+    height: ${props => props.theme.space.base * 3}px;
+  }
+
+  /* stylelint-disable-next-line */
+  @media screen and (-ms-high-contrast: active), screen and (-ms-high-contrast: none) {
+    input {
+      top: -10px; /* [1] */
+      height: ${props => props.theme.space.base * 8}px; /* [1] */
+    }
+  }
```
jzempel, line 35: Shouldn't these styles be directly on `StyledAlpha`. Also, it appears that `height` is an unnecessary override.

jzempel:
@allisonacs @steue you'll find that dragging the mouse outside of the `ColorDialog` inadvertently dismisses the color picker. 
We know this is an issue with the Tooltip Modal and needs to be addressed under a separate PR. Please help by providing design opinion on whether or not this is a showstopper for shipping a first version of the color picker. 🙏

steue:
> @allisonacs @steue you'll find that dragging the mouse outside of the `ColorDialog` inadvertently dismisses the color picker.
> We know this is an issue with the Tooltip Modal and needs to be addressed under a separate PR. Please help by providing design opinion on whether or not this is a showstopper for shipping a first version of the color picker. 🙏

when would we be able to address the tooltip modal bug? if we squish the bug quickly after release, then this seems fine. i don't want to prolong the release, nor would i feel comfortable shipping a buggy experience.

packages/colorpickers/src/utils/saturation.ts
```diff
@@ -0,0 +1,87 @@
...
+  const { saturation, bright } = getSaturationLightness(container, e.pageX, e.pageY, rtl);
+
+  return {
+    h: hue,
+    s: saturation,
+    v: bright
+  };
+}
+
+export const getThumbPosition = (
```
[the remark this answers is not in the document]
hzhu: Interesting bugs, the first bug is a result of the hue slider having an incorrect max hue of `359`. It should be `360`. The color values (hex, rgb) should be calculated correctly now.

The second bug seems to be related to Storybook rather than the color picker. It can be replicated for other color pickers such as [Fluent UI color picker](https://fluent-color-spike.netlify.app/?path=/docs/components-colorpicker--default) and [react-color](https://color-preview.netlify.app/?path=/docs/components-colorpicker-colorpicker--default). I'll investigate if there's something that can be done at the Storybook level to address this.

packages/colorpickers/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "name": "@zendeskgarden/react-colorpickers",
+  "description": "Components related to color pickers in the Garden Design System",
+  "license": "Apache-2.0",
+  "author": "Zendesk Garden <garden@zendesk.com>",
+  "homepage": "https://garden.zendesk.com/",
+  "repository": "https://github.com/zendeskgarden/react-components",
+  "bugs": {
+    "url": "https://github.com/zendeskgarden/react-components/issues"
+  },
+  "version": "8.29.3",
```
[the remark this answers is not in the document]
hzhu: Updated

packages/colorpickers/stories/examples/ColorPicker/Uncontrolled.tsx
```diff
@@ -0,0 +1,68 @@
...
+  hueSlider: 'Hue slider',
+  hex: 'Hex',
+  red: 'R',
+  green: 'G',
+  blue: 'B',
+  alpha: 'A'
+};
+
+Uncontrolled.argTypes = {
+  color: { control: 'disable' },
```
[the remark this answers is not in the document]
hzhu: Updated

packages/colorpickers/README.md
```diff
@@ -0,0 +1,27 @@
+# @zendeskgarden/react-colorpickers [![npm version](https://flat.badgen.net/npm/v/@zendeskgarden/react-colorpickers)](https://www.npmjs.com/package/@zendeskgarden/react-colorpickers)
+
+This package includes components related to colorpickers in the
```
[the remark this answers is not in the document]
hzhu: Updated

README.md
```diff
@@ -73,6 +74,12 @@ to install.
 [chrome size link]: https://bundlephobia.com/result?p=@zendeskgarden/react-chrome
 [chrome dependency status]: https://flat.badgen.net/david/dep/zendeskgarden/react-components/packages/chrome
 [chrome dependency link]: https://david-dm.org/zendeskgarden/react-components?path=packages/chrome
+[colorpicker npm version]: https://flat.badgen.net/npm/v/@zendeskgarden/react-colorpickers
+[colorpicker npm link]: https://www.npmjs.com/package/@zendeskgarden/react-colorpickers
+[colorpicker size bundle]: https://flat.badgen.net/bundlephobia/minzip/@zendeskgarden/react-colorpickers
+[colorpicker size link]: https://bundlephobia.com/result?p=@zendeskgarden/react-colorpickers
+[colorpicker dependency status]: https://flat.badgen.net/david/dep/zendeskgarden/react-components/packages/colorpicker
+[colorpicker dependency link]: https://david-dm.org/zendeskgarden/react-components?path=packages/colorpicker
```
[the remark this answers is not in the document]
hzhu: Updated

README.md
```diff
@@ -28,6 +28,7 @@ to install.
 | [`@zendeskgarden/react-breadcrumbs`](packages/breadcrumbs)     | [![npm version][breadcrumbs npm version]][breadcrumbs npm link]     | [![Bundle Size][breadcrumbs size bundle]][breadcrumbs size link]     | [![Dependency Status][breadcrumbs dependency status]][breadcrumbs dependency link]     |
 | [`@zendeskgarden/react-buttons`](packages/buttons)             | [![npm version][buttons npm version]][buttons npm link]             | [![Bundle Size][buttons size bundle]][buttons size link]             | [![Dependency Status][buttons dependency status]][buttons dependency link]             |
 | [`@zendeskgarden/react-chrome`](packages/chrome)               | [![npm version][chrome npm version]][chrome npm link]               | [![Bundle Size][chrome size bundle]][chrome size link]               | [![Dependency Status][chrome dependency status]][chrome dependency link]               |
+| [`@zendeskgarden/react-colorpickers`](packages/colorpicker)    | [![npm version][colorpickers npm version]][colorpickers npm link]   | [![Bundle Size][colorpickers size bundle]][colorpickers size link]   | [![Dependency Status][colorpickers dependency status]][colorpickers dependency link]   |
```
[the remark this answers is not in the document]
hzhu: Updated

packages/colorpickers/stories/examples/ColorPicker/Controlled.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Button onClick={() => setColor('#CE9FB7')}>Control to #CE9FB7</Button>
+      </div>
+      <div style={{ display: 'flex', justifyContent: 'center' }}>
+        <ColorPicker color={color} onChange={setColor} labels={labels} />
+      </div>
+    </>
+  );
+};
+
+Controlled.args = {
```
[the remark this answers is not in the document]
hzhu: I've removed the `labels` control which is a change required in `Uncontrolled.tsx`.

packages/colorpickers/stories/examples/ColorPicker/Uncontrolled.tsx
```diff
@@ -0,0 +1,68 @@
...
+  hueSlider: {
+    control: 'text',
+    name: 'Hue slider label',
+    description: 'A label for the hue slider'
+  },
+  hex: { control: 'text', name: 'Hex input label', description: 'A label for the hex input' },
+  red: { control: 'text', name: 'Red input label', description: 'A label for the red input' },
+  green: { control: 'text', name: 'Green input label', description: 'A label for the green input' },
+  blue: { control: 'text', name: 'Blue input label', description: 'A label for the blue input' },
+  alpha: { control: 'text', name: 'Alpha input label', description: 'A label for the alpha input' }
```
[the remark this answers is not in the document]
hzhu: I think `argTypes` aren't defined in `Controlled` because only the main story (top of page) has controls. Without these `argType` values, the controls would not have a description text.

packages/colorpickers/stories/examples/ColorDialog/WithIconButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <IconButton style={{ color: iconColor }} aria-label="leaf">
+          <LeafIcon />
+        </IconButton>
+      </ColorDialog>
+    </div>
+  );
+};
+
+WithIconButton.args = {
+  placement: 'bottom',
```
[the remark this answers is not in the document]
hzhu: It looks like it's because only the "main" story can have controls. In this case the "main" story is `Uncontrolled`.

packages/colorpickers/src/styled/ColorPicker/StyledAlphaField.ts
hzhu: I don't think the styles can be on `StyledAlpha` since they target the `input`. I was able to remove the `height` for non-IE browsers, but IE requires it so that the thumb shadow styles don't get cut off.

jzempel: I'm confused. Isn't `StyledAlpha` the `<input type="range">`?

hzhu: Oops, it is. I've moved over all the input styles to `StyledAlpha` and `StyledHue` accordingly 👍🏽 .

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+  green?: string;
+  blue?: string;
+  alpha?: string;
+}
+
+export interface IColorPickerProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'color' | 'onChange'> {
+  /** A default color for a uncontrolled color picker. Can be a valid color string or IColor state */
+  defaultColor?: string | IColor;
+  /** A color value for a controlled color picker. Can be a valid color string or IColor state */
```
jzempel, line 56: *Sets the color for a controlled color picker

~. Can be a valid color string or IColor state~ is unnecessary as the info is apparent from the prop sheet data.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+  color?: string | IColor;
+  /**
+   * Handles color picker changes
+   *
+   * @param {Object} color A color picker state
+   */
+  onChange?: (color: IColor) => void;
+  /** Replaces the default labels within the color picker */
```
jzempel, line 64: *Replaces label text within the color picker

Is there an effective way to help docgen pick up the interface structure here? Showing an interface that the consumer doesn't have access to in the docs isn't super helpful.

packages/colorpickers/src/elements/ColorPicker/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+  labels?: IColorPickerLabels;
+  /** Autofocuses the hex input element */
+  autofocus?: boolean;
```
jzempel, line 67: Should be set to `true` in `ColorPicker.defaultProps`.

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
```diff
@@ -0,0 +1,63 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { ThemeProps, DefaultTheme } from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { Button } from '@zendeskgarden/react-buttons';
+
+const COMPONENT_ID = 'colorpickers.colordialog_button';
+
+const colorStyles = (props: ThemeProps<DefaultTheme>) => {
```
jzempel, line 14: This button was changed from @steue's designs. It should default to a `neutralHue` `border-color` (the remaining hover/active/focus stateful colors work as expected) in order to fit in well with surrounding form fields. The `:disabled` `border-color` also needs correct styling.

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
```diff
@@ -0,0 +1,63 @@
...
+  return `
+    color: ${baseColor};
+
+    &:hover {
+      color: ${hoverColor};
+    }
+
+    &:active,
+    &[aria-pressed='true'],
+    &[aria-pressed='mixed'] {
```
jzempel, line 30: These are not necessary as this is not a toggle button.

packages/colorpickers/src/styled/ColorDialog/StyledButton.ts
```diff
@@ -0,0 +1,63 @@
...
+  width: ${props => props.theme.space.base * 17}px;
+  height: ${props => props.theme.space.base * 10}px;
+
+  &:hover {
+    background-color: ${props => props.theme.colors.background};
+  }
+
+  &:disabled {
+    background-color: ${props => getColor('neutralHue', 100, props.theme)};
+  }
```
jzempel, line 54: Simplify by moving these into the `colorStyles` function – consolidate all styling related to color.

packages/colorpickers/src/styled/ColorDialog/StyledTooltipBody.ts
```diff
@@ -0,0 +1,25 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { TooltipModal } from '@zendeskgarden/react-modals';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colordialog_tooltipmodal_body';
+
+export const StyledTooltipBody = styled((TooltipModal.Body as unknown) as 'div').attrs({
```
jzempel, line 14: can this be simplified to `styled(TooltipModal.Body as 'div')` by using `as: 'div'` in the attrs?

packages/colorpickers/src/styled/ColorPicker/StyledHue.ts
```diff
@@ -0,0 +1,169 @@
...
+      );
+      /* stylelint-enable */
+      height: ${props.theme.space.base * 3}px;
+  `)}
+
+  ${trackLowerStyles(`opacity: 0;`)}
+
+  &::-webkit-slider-thumb {
+    margin-top: -${props => props.theme.space.base / 2}px;
+  }
```
jzempel, line 154: why is this not being set using `thumbStyles`? If it's a hack, it should be commented.

packages/colorpickers/src/styled/ColorPicker/StyledAlpha.ts
```diff
@@ -0,0 +1,57 @@
...
+  position: absolute;
+  z-index: 2;
+  /* stylelint-disable-next-line declaration-no-important */
+  margin-top: 0 !important;
+  border-radius: 0;
+  height: ${props => props.theme.space.base * 3}px;
+
+  &::-webkit-slider-thumb {
+    margin-top: -5px;
+  }
```
jzempel, line 32: why is this not being set using `thumbStyles`? If it's a hack, it should be commented.

packages/colorpickers/src/styled/ColorPicker/StyledAlpha.ts
```diff
@@ -0,0 +1,57 @@
...
+const COMPONENT_ID = 'colorpickers.colorpicker_alpha';
+
+/**
+ * 1. Adjust spacing in IE11 to match other browsers
+ * 2. Provides height for the input so that the thumb shadow styles are not cut off in IE11.
+ */
+export const StyledAlpha = styled((Range as unknown) as 'input').attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
```
jzempel, line 22: ditto re: simplifying the TS here by using `as: 'input'` attrs

packages/colorpickers/src/styled/ColorPicker/StyledHue.ts
```diff
@@ -0,0 +1,169 @@
...
+  `;
+};
+
+/**
+ * 1. Provides height for the input so that the thumb shadow styles are not cut off in IE11.
+ */
+export const StyledHue = styled((Range as unknown) as 'input').attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
```
jzempel, line 122: ditto re: TS

packages/colorpickers/src/styled/ColorPicker/StyledAlpha.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { Range } from '@zendeskgarden/react-forms';
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { colorStyles, thumbStyles, trackStyles, trackLowerStyles } from './StyledHue';
```
jzempel, line 11: Should shared styling be abstracted into a `StyledRange` base class instead? Even with this, there seems to be additional crossover that would be cleaner and more maintainable to abstract and subsequently wrapped by both `StyledAlpha` and `StyledHue`.

packages/colorpickers/src/styled/ColorPicker/StyledAlphaField.ts
```diff
@@ -0,0 +1,25 @@
...
+import styled from 'styled-components';
+import { Field } from '@zendeskgarden/react-forms';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colorpicker_alpha_field';
+
+export const StyledAlphaField = styled((Field as unknown) as 'div').attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
```
jzempel, line 17: ditto re: simplified TS.

packages/colorpickers/src/styled/ColorPicker/StyledHexField.ts
```diff
@@ -0,0 +1,33 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { Field } from '@zendeskgarden/react-forms';
+
+const COMPONENT_ID = 'colorpickers.colorpicker_hex_field';
+
+export const StyledHexField = styled((Field as unknown) as 'div').attrs({
```
jzempel, line 14: simplified TS?

packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
```diff
@@ -0,0 +1,25 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { Input } from '@zendeskgarden/react-forms';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colorpicker_input';
+
+export const StyledInput = styled(Input as any).attrs({
```
jzempel, line 14: why not `as 'input'`?

packages/colorpickers/src/styled/ColorPicker/StyledLabel.ts
```diff
@@ -0,0 +1,27 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { Label } from '@zendeskgarden/react-forms';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colorpicker_label';
+
+export const StyledLabel = styled(Label as any).attrs({
```
jzempel, line 14: why not `as 'label'`?

packages/colorpickers/src/styled/ColorPicker/StyledRGBAField.ts
```diff
@@ -0,0 +1,28 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { Field } from '@zendeskgarden/react-forms';
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'colorpickers.colorpicker_rgb_field';
+
+export const StyledRGBAField = styled((Field as unknown) as 'div').attrs({
```
jzempel, line 14: simplified TS?

packages/colorpickers/stories/examples/ColorDialog/WithIconButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+  alphaSlider: 'Alpha slider',
+  hueSlider: 'Hue slider',
+  hex: 'Hex',
+  red: 'R',
+  green: 'G',
+  blue: 'B',
+  alpha: 'A',
+  disabled: false,
+  hasArrow: false,
```
jzempel, line 79: `hasArrow` should be true by default on this story.

packages/colorpickers/stories/examples/ColorDialog/WithIconButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+        hasArrow={hasArrow}
+        placement={placement}
+        isAnimated={isAnimated}
+        onClick={() => setIsPressed(true)}
+        onClose={nextColor => {
+          setIsPressed(false);
+          setSelectedColor(nextColor);
+        }}
+      >
+        <ToggleIconButton
```
jzempel, line 56: Humble apologies, I got this one wrong. I double-checked with design, and this should be a regular (no-toggle) `IconButton`. Toggles are reserved for actually toggling functionality like "bold" or "italic".

packages/colorpickers/stories/examples/ColorDialog/WithIconButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+          aria-label="leaf"
+          disabled={disabled}
+          isPressed={isPressed}
+          style={{ color: disabled ? undefined : iconColor }}
+        >
+          <LeafIcon />
```
jzempel, line 62: The new palette icon that shipped today with `svg-icons` v6.29.0 would be preferable here.

packages/colorpickers/stories/examples/ColorDialog/WithFormInput.tsx
```diff
@@ -0,0 +1,143 @@
...
+}) => {
+  const labels = { alphaSlider, hueSlider, hex, red, green, blue, alpha };
+  const [input, setInput] = useState('#17494d');
+  const [color, setColor] = useState<string | IColor>('rgba(23, 73, 77, 100)');
+
+  return (
+    <StyledField>
+      <Label>Favorite color</Label>
+      <StyledContainer>
+        <InputGroup>
```
jzempel, line 61: I'm glad we pushed this boundary by digging into the system more deeply. We find that border radius styling is leaking into the RGBA fields. Those fields will need specificity overrides to prevent this from happening.

Screen Shot 2021-02-23 at 11 40 49 AM

Also, check with @steue, but the default placement on this example should _probably_ be `bottom-end`.

jzempel: Also note that the alpha value is showing up as `1000` in this example.

packages/colorpickers/stories/examples/ColorDialog/Controlled.tsx
```diff
@@ -0,0 +1,104 @@
...
+  green,
+  blue,
+  alpha,
+  placement,
+  disabled,
+  hasArrow,
+  isAnimated
+}) => {
+  const labels = { alphaSlider, hueSlider, hex, red, green, blue, alpha };
+  const [color, setColor] = useState<string | IColor>('rgba(22,73,77,100)');
```
jzempel, line 32: The default alpha is showing up as `1000` in this example.

packages/colorpickers/stories/examples/ColorDialog/WithIconButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+  blue,
+  alpha,
+  disabled,
+  hasArrow,
+  isAnimated
+}) => {
+  const labels = { alphaSlider, hueSlider, hex, red, green, blue, alpha };
+  const [isPressed, setIsPressed] = useState(false);
+  const [color, setColor] = useState<string | IColor>('rgba(23, 73, 77, 100)');
+  const [selectedColor, setSelectedColor] = useState<string | IColor>('rgba(23, 73, 77, 100)');
```
jzempel, line 35: The default alpha is showing up as `1000` in this example.

jzempel (changes requested):

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <Button.EndIcon isRotated={referenceElement != null}>
+              <Chevron />
+            </Button.EndIcon>
+          </StyledButton>
+        )}
+        <StyledTooltipModal
+          hasArrow={hasArrow}
+          popperModifiers={popperModifiers}
+          zIndex={zIndex}
+          isAnimated={isAnimated}
```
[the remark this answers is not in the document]
jzempel: Also, reorientation of the dialog is not working as expected on window scroll. See https://60353946748b1d07920d870e--zendeskgarden-react-components.netlify.app/storybook/?path=/docs/components-colorpickers-colordialog--uncontrolled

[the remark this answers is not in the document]
jzempel: Might just be that the stories need the appropriate `popperModifiers` added? See https://garden.zendesk.com/components/tooltip-modal#placement

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+    ref
+  ) => {
+    const theme = useContext(ThemeContext);
+    const isControlled = color !== null && color !== undefined;
+    const buttonRef = useRef<HTMLButtonElement>(null);
+    const colorPickerRef = useRef<HTMLDivElement>(null);
+    const mergedRef = mergeRefs([ref, buttonRef]);
+    const [referenceElement, setReferenceElement] = useState<HTMLButtonElement | null>();
+    const [uncontrolledColor, setUncontrolledColor] = useState<string | IColor>(
+      defaultColor || '#ffffff'
```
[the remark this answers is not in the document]
hzhu: I've typed out the value to include `undefined` values for `defaultColor`. This way `#fff` doesn't need to be hard coded here and `defaultColor` is managed like `ColorPicker`. Updated 👍🏽 .

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+  onClose: PropTypes.func,
+  onChange: PropTypes.func,
+  labels: PropTypes.object,
+  color: PropTypes.oneOfType<any>([PropTypes.object, PropTypes.string]),
+  defaultColor: PropTypes.oneOfType<any>([PropTypes.object, PropTypes.string])
+};
+
+ColorDialog.defaultProps = {
+  hasArrow: false
+};
```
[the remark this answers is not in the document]
hzhu: Removed `hasArrow` default. Added `placement`, `isAnimated`, and `zIndex`.

packages/colorpickers/src/elements/ColorPicker/index.tsx
hzhu: Updated the description copy. Inlining the `labels` type results in a detailed type:

Screen Shot 2021-02-24 at 9 17 04 AM

Alternatively, there could be a link in the description that links directly to the line of code where `IColorPickerLabels` is defined.

jzempel: I think a link is likely to fall out-of-date over time. Better to provide the information inline.

packages/colorpickers/src/styled/ColorPicker/StyledAlpha.ts
hzhu: This is intended to fix a TypeScript compilation issue so modifying the DOM tag wouldn't fix the compilation issue.

It seems like there are multiple versions of `@types/react` when a package uses `react-modals`. The `react-transitiongroup` also depends on `@types/react` and is duplicating it. Austin and I attempted to use `yarn workspaces` to hoist this dependency but ran into issues around the build process. It looks like type-casting is a suitable workaround for now.

packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
hzhu: I think it's because `input` does not have attributes such as `isCompact`, `isBare`, etc. 

`Input as React.ForwardRefExoticComponent<IInputProps>` compiles locally, but in CI runs into the issue regarding multiple versions of `@types/react`.

jzempel: @austingreendev thots on an improved TS construct or running `npx yarn-deduplicate`? Seems like a lockfile might be less than pristine.

packages/colorpickers/src/styled/ColorPicker/StyledHue.ts
hzhu: I've added a comment `"Adjusts the spacing to align thumb with track on Chrome, Safari, and Edge"`.

jzempel: In that case, this feels a bit like a structural problem. Where _all_ browsers should receive generalized `thumbStyles` and then the overrides for IE should appear below. Or am I missing something here?

packages/colorpickers/src/styled/ColorPicker/StyledAlpha.ts
hzhu: Refactored shared styling to `/common/StyledRange` 👍🏽

packages/colorpickers/src/elements/ColorDialog/index.tsx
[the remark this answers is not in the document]
hzhu: Added close after fade out transition and spacing around examples to demo reorientation.

packages/colorpickers/stories/examples/ColorDialog/WithFormInput.tsx
hzhu: Added overrides to prevent those UI issues and works across all browsers except IE11.

Screen Shot 2021-02-26 at 10 41 41 AM

packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
austingreendev: `yarn-deduplicate` wasn't able to help us here as we aren't hoisting dependencies. There are 2 different versions of `@types/react` and even if they're the same version I think we would have these inference issues.

@hzhu did you try removing and re-adding `@types/react-transition-group` in `react-modals`?

packages/colorpickers/src/elements/ColorDialog/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+      let timeout: number;
+
+      if (isOpen) {
+        setReferenceElement(buttonRef.current);
+      } else if (isAnimated) {
+        // Match the duration of the menu fade out transition.
+        timeout = window.setTimeout(() => setReferenceElement(null), 200);
+      } else {
+        setReferenceElement(null);
+      }
```
jzempel, line 111: @hzhu I believe this should live on the `TooltipModal` so that it is fixed for all tooltip modal instances.

hzhu: Do you think that would belong in a new bug fix pull request for `react-modals`? If so, I can remove it here and open a separate PR.

jzempel: I'm fine with that being a separate PR. We'll just want to be sure the fix works here as expected.

packages/colorpickers/stories/examples/ColorDialog/WithFormInput.tsx
hzhu: Update: I've pushed a commit that fixes this in IE11 by refactoring the styling and using a `span` within the `botton` element.

packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
hzhu: Yes, I've tried removing and re-adding `@types/react-transition-group` in `react-modals`. It'd be great to take another look at this one more time, but if it requires more dedicated time I think this issue could be addressed outside the context of this PR.

jzempel (approved):

austingreendev (approved):

hzhu:
I noticed two cross-browser related issues:

* IE11: when the “Set to #CE9A”” button is pressed, the color dialog preview square does not reflect the new controlled color
* Firefox: the alpha slider’s gradient doesn’t seem to take up the full height of the input element

jzempel:
thanks @hzhu. I think I've got the cross-browser bugs swatted with the recent batch of commits.

steue (approved):
lgtm on the design side. looking forward for the modal fix. tysvm @hzhu @jzempel!!

m-lai:
Not sure whether this is on purpose, but in the Custom Trigger demo, the [delay for the little pointer popping up](https://share.getcloudapp.com/nOulrJvQ) from the menu feels a bit odd. Should it be the same part of the menu and appear at the same time?

The Input in the Form Input demo displays 8 digit values instead of 6 for the hex code, is this a bug? (It looks like "99" is added to the end of any hex code). I can also add any text, not just hex code values. https://share.getcloudapp.com/mXu14zxb

jzempel:
@m-lai the arrow animation delay is precisely the same we see with [`Menu`](https://garden.zendesk.com/components/menu#arrow) and  [`TooltipModal`](https://zendeskgarden.github.io/react-components/storybook/?path=/story/components-modals-tooltipmodal--default) with `isAnimated` set to `true`. I am seeing a 1px discrepancy on the tooltip modal arrow, but that is a long-standing bug that we should address outside of this PR.

There are many CSS hex colors that are valid outside of 6 characters, including 8 and 4 character values. This story demo attempts to show that the range of possibilities is working as expected. But it is not attempting to moderate invalid values. This would be an exercise left to the consumer (or Garden website) to get sorted to perfection. See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Syntax_2 for details.

m-lai (approved):

m-lai:
> @m-lai the arrow animation delay is precisely the same we see with [`Menu`](https://garden.zendesk.com/components/menu#arrow) and [`TooltipModal`](https://zendeskgarden.github.io/react-components/storybook/?path=/story/components-modals-tooltipmodal--default) with `isAnimated` set to `true`. I am seeing a 1px discrepancy on the tooltip modal arrow, but that is a long-standing bug that we should address outside of this PR.
> 
> There are many CSS hex colors that are valid outside of 6 characters, including 8 and 4 character values. This story demo attempts to show that the range of possibilities is working as expected. But it is not attempting to moderate invalid values. This would be an exercise left to the consumer (or Garden website) to get sorted to perfection. See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Syntax_2 for details.

Cool, thanks for the explainer.

PR #1021, opened 2021-02-18
feat(notifications): introduce toast pattern and utilities

austingreendev:
## Description

This PR introduces the `Toast` pattern and related utilities. 

### Components and API

Toasts are implemented with a `ToastProvider` component and a `useToast()` hook. The toast implementation has no concept around the content it's provided. The examples below show our `Notification` component, but consumers can render any element they want.

### Animation

The toast animation is `400ms` in duration and transitions the `opacity`, `translateY`, and `max-height`. We modify the `translateY` to provide directionality for bottom vs top placements. We also modify the transition delay based on enter/exit states.

This is a rough estimate of the spec'd animation so please provide any tweaks that you would be looking for.

## Checklist

      designer as a reviewer)


m-lai (changes requested):
Hey @austingreendev , one thing I noticed for now that's different to specs - we'd want to reverse the Show Animation for top and bottom variations (we want what's happening on  for top toasts on Show to happen for bottom toasts, and vice versa). Hide is fine!

jzempel:
Early thoughts... not quite sure how to describe it, but I think something will need to happen when dismissable notifications stack up quickly. Something like added microseconds in order to make the fade-swipe animations work (or a set distance for animated dismissal)? Otherwise, notification 1 dismisses at height = 1x, notification 2 at height = 2x, notification 3 at height = 3x, etc. And the system doesn't feel as intentional as it should.

Besides that, early demo look and feel is 😍

m-lai:
> Early thoughts... not quite sure how to describe it, but I think something will need to happen when dismissable notifications stack up quickly. Something like added microseconds in order to make the fade-swipe animations work (or a set distance for animated dismissal)? Otherwise, notification 1 dismisses at height = 1x, notification 2 at height = 2x, notification 3 at height = 3x, etc. And the system doesn't feel as intentional as it should.
> 
> Besides that, early demo look and feel is 😍

Yeah agreed, what JZ said. I think a small adjustment is to have all subsequent notifications on dismissal & fade out aim for the same height notification 1 aims for, for lack of better words

jon-kerwin:
Hey team, adding some thoughts in here as well. Please feel free to give me direct feedback on the best way to deliver this info. Overall though, these are some wonderfully simple toasts. I love the thinking around stacking and dismissible/non-dismissible toasts, and the animations are on their way.

+1 to others' comments on stacked dismissible animations. I notice that the 3rd and 4th toast fade a bit too quickly in relation to the 1st and 2nd. Needs some tweaking to feel like each toast rests properly and has time to be viewed.

Another thing I noticed is that in the "Top" placement the toast displays behind the dropdown.
Screen Shot 2021-02-24 at 12 08 39 PM

austingreendev:
@jzempel @m-lai @jon-kerwin I've tweaked the animation some more after sitting down with Mandy. Here are some of the highlights:

- Tweaked enter vs exit timing for opacity. We now use `ease-in` only on specific values and have changed the delays to feel more deliberate.
- The "stacked" notification issue should feel a little better now.
  - Rather than adding/removing stacked items from the DOM we now always render the notification.
  - We visually hide the notifications until they are within the bounds provided by `limit`
  - Once it is within the limit we _only_ transition the opacity. This avoids any double-swoops with the `translateY`
  - This will also help with A11Y since a screen-readers should be notified immediately due to `aria-live`

@jzempel @hzhu TIL of the `hideVisually()` utility within polished https://polished.js.org/docs/#hidevisually

austingreendev:
https://user-images.githubusercontent.com/4030377/109204344-151e4780-775a-11eb-9a20-40e7c00de254.mov

jzempel:
> TIL of the hideVisually() utility within polished https://polished.js.org/docs/#hidevisually

yeah, the [implementation](https://github.com/styled-components/polished/blob/main/src/mixins/hideVisually.js) matches [ours](https://github.com/zendeskgarden/css-components/blob/2e3d86d9a69a335fb3dec698be65b6e9301c299e/packages/utilities/src/visibility.css#L13-L27) almost exactly and tailwind's [implementation](https://tailwindcss.com/docs/screen-readers) exactly.

packages/notifications/src/elements/toaster/ToastSlot.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return index >= limit;
+    },
+    [limit, placement, toasts.length]
+  );
+
+  return (
+    <StyledTransitionGroup
+      key={placement}
+      placement={placement}
+      zIndex={zIndex}
```
hzhu, line 71: I noticed a warning about passing `zIndex` to the native dom element in the console while running the branch locally:

Screen Shot 2021-02-26 at 11 31 55 AM

austingreendev: These components might be a good first candidate for [transient style props](https://styled-components.com/docs/api#transient-props)

m-lai:
>  Screen.Recording.2021-02-25.at.11.02.04.AM.mov

This is feeling pretty good to me now. @allisonacs @nyckie @steue Would love to get your thoughts on the entry/exit animation and speed for single & stacked notifications (also when you dismiss them).

jzempel:
Would these benefit from a subtle bounce similar to our menus, selects, datepickers, etc? The entry feels like it is lacking some life. Probably not so hyper as dropbox paper, but something with a bit more spring? The exit feels better than the entry for me.

m-lai:
> Would these benefit from a subtle bounce similar to our menus, selects, datepickers, etc? The entry feels like it is lacking some life. Probably not so hyper as dropbox paper, but something with a bit more spring? The exit feels better than the entry for me.

That's a good call. @austingreendev and I focused on tweaking `ease-in` mostly. Are there already similar values we have for existing components so we can match the spring-y feel?

allisonacs:
Played around this afternoon, and it is feeling really strong, but I have a few nits for timing. 

- The fade out timing feels too fast. I'm not sure exactly how much the time currently is but I think we probably need at least another half second on there. 
- The timing for dismissal on stacked notifications is _way_ too fast. If a user has 3 notifications all coming in around the same time, they can't read them all in the time allotted. Maybe we either need to reset the countdown when a notification reaches a new position, or we need to at least add time when reaching a new position. 
- Agree that we need bounce in here… I would have thought we would have been pulling the effects from our other components.
- I noticed that if I have some text highlighted, but then move my mouse away from the notification, the notification still autodismisses. I think that's a little counterintuitive, I would have thought that since I focused the browser on an element in the notification that it would stay on screen. 

I like how hovering anywhere in the notification stack pauses dismissal for all the notifications, that's a nice touch. 

@austingreendev I took a look at these a bit during my 1:1 with JZ, and he pointed me at react-toastify. They have a visual indication of the dismissal timing. JZ floated the idea that maybe we can add that in for visual debugging… I think that would be really helpful as we try to dial all this in.

austingreendev:
> The fade out timing feels too fast. I'm not sure exactly how much the time currently is but I think we probably need at least another half second on there.

We were trying to keep enter/exit animations equal to avoid issues with notifications that stack past the defined limit. With different enter/exit animations that timing could overlap. It looks like we can [customize appear vs enter timing](The fade out timing feels too fast. I'm not sure exactly how much the time currently is but I think we probably need at least another half second on there.) which could help. Will look into this.

> The timing for dismissal on stacked notifications is way too fast. If a user has 3 notifications all coming in around the same time, they can't read them all in the time allotted. Maybe we either need to reset the countdown when a notification reaches a new position, or we need to at least add time when reaching a new position.

Could this be achieved with a longer default timeout? The existing toast implementations don't provide this reset functionality. FWIW in real-world scenarios there will realistically only be one or two notifications at a time. 

> I noticed that if I have some text highlighted, but then move my mouse away from the notification, the notification still autodismisses.

Notifications are not focusable or interactive by default. This means we don't have a focus state to key off of. The removal timer is based off of mouse state only. Would it make sense to disable user-selection for all elements within the toast placement?

The current API doesn't provide time information to the rendered notifications. It would be a significant refactor to supply that state to the elements even for visual debugging. Unless we plan on adding the feature in the future I would advise that we not provide remaining time as state to avoid re-renders. Since `react-toastify` provides the UI as well as placement they are able to enforce memoization more efficiently than we are able to.

packages/notifications/src/elements/toaster/ToastSlot.tsx
austingreendev: I've updated these props to use the `$*` transient naming. We will still see this issues until we upgrade to v5, but it will be resolved at that time.

austingreendev:
@allisonacs @jzempel @jon-kerwin Mandy and I sat down and worked on introducing a similar cubic-bezier animation ("bounce") to the toast. Here is what we came up with:

- Increase the fade-in/translate timing and delays for the _exit_ animation to slow it down
- Added a bounce transition that aligns with our relatively placed menus/dropdowns to the _enter_ animation only
  - With our exit animation having both a transform and opacity change, adding the bounce can create some weird visuals as the opacity goes to 0.

Let us know what you think.

allisonacs:
This is feeling much improved. Couple thoughts…

- Take a look at this screencast: https://share.getcloudapp.com/YEuXwRBK (audio on, I speak during it). I do think a longer default time onscreen would be helpful, although I am not sure that will fix the thing I'm concerned about. You said that the realistic scenario is 1-2 notifications on screen at any given time. I agree that's probably the p95 or higher. But regardless, even with two notifications, I don't think they're on screen long enough to reasonably get through both of them in the time allotted—this is made worse because the last notification moves in the middle of you trying to read it. 
- I think the bounce might be a bit too pronounced, especially compared to the bounce we see in menus and modals. Agree that it should be bigger than those examples bc the UI is bigger, and it's dropped farther but it feels like the physics aren't the same. I've asked Tori from our animation team to take a look. 
- You said the notifications aren't interactive… what if there is an interactive element in the notification? like a link? the button to close the notifcation works as expected, including with a keyboard. 
- That said, I don't think we should prevent people from selecting the text. Folks use it to help themselves read.

allisonacs:
I don't think I reviewed the most recent deploy btw

allisonacs:
Tori's feedback on the revised animations:

> Hiii, I just looked at these and I think they look really nice. The overshoot on them (how they bounce down and back up) isn't really something we do in our brand animation very often, but it makes sense for this use case. These are messages that you want people to see and the extra pop will help grab their attention.

So from a pure animation perspective, I think these are all good now.

m-lai:
@austingreendev All good to go after discussion with Allison and Tori's feedback above!

packages/notifications/src/elements/toaster/Toast.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }, [pauseTimers, remainingTime]);
+
+  useEffect(() => {
+    let timeout: any;
+
+    if (!isNaN(remainingTime)) {
+      startTimeRef.current = Date.now();
+
+      timeout = setTimeout(() => {
```
hzhu, line 56: I learned recently that using `window.setTimeout` allows TypeScript to infer the timeout id (`number`). I think it could enable the removal of the `any` type on line 51.

```tsx
let timeout: number | undefined = undefined;

timeout = window.setTimeout(callback, delay);
```

packages/notifications/src/elements/toaster/ToastSlot.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+  }, [environment, handleVisibilityChange]);
+
+  const handleMouseEnter = useCallback(() => {
+    setPauseTimers(true);
+  }, []);
+
+  const handleMouseLeave = useCallback(() => {
+    setPauseTimers(false);
+  }, []);
```
hzhu, line 54: Could you clarify the purpose of memorizing these functions? What are the consequences of not memoizing? Say with something along the lines of:

```jsx
onMouseEnter={() =>  setPauseTimers(true)}
```

hzhu (approved):
LGTM 🚀

packages/notifications/src/elements/toaster/ToastSlot.tsx
austingreendev: I think without memoization this would re-render the `StyledTransitionGroup` component every time? (emphasis on _I think_ 😄 )

packages/notifications/package.json
```diff
@@ -21,7 +21,10 @@
   "sideEffects": false,
   "types": "dist/typings/index.d.ts",
   "dependencies": {
-    "prop-types": "^15.5.7"
+    "nanoid": "^3.1.20",
```
jzempel, line 24: fancy 💃

packages/notifications/src/elements/toaster/Toast.tsx
jzempel: Aren't there SSR concerns for dereferencing `window`?

packages/notifications/stories/5-Toasts.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <Notification style={{ width: 275 }} type={type}>
+            <Title>{type.charAt(0).toUpperCase() + type.slice(1)} notification</Title>
+            <Paragraph>
+              Turnip greens yarrow ricebean cauliflower sea lettuce kohlrabi amaranth water
+            </Paragraph>
+            <Close aria-label="Close Alert" onClick={() => close()} />
+          </Notification>
+        ),
+        {
+          autoDismiss: type === 'error' ? false : 5000,
```
jzempel, line 53: Can this `5000` value be expressed as a story control?

packages/notifications/src/elements/toaster/ToastProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+} from './reducer';
+import { ToastSlot } from './ToastSlot';
+
+export const ToastContext = createContext<
+  { state: IToasterState; dispatch: Dispatch<ToasterReducerAction> } | undefined
+>(undefined);
+
+export interface IToastProviderProps {
+  /**
+   * Limit the number of visible toasts
```
jzempel, line 26: [nit] Change "Limit" to "Limits" for alignment with the JSDoc RFC

packages/notifications/src/elements/toaster/ToastProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+      {toastsByPlacement('bottom')}
+      {toastsByPlacement('bottom-end')}
+    </ToastContext.Provider>
+  );
+};
+
+ToastProvider.displayName = 'ToastProvider';
+
+ToastProvider.defaultProps = {
+  limit: 4
```
jzempel, line 71: How can the limit be made unlimited? I could definitely be wrong, but it feels like `undefined` should be the default.

packages/notifications/src/elements/toaster/styled.ts
```diff
@@ -0,0 +1,158 @@
...
+import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
+import { hideVisually } from 'polished';
+import { TransitionGroup } from 'react-transition-group';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { ToastPlacement } from './reducer';
+
+export const TRANSITION_CLASS = 'garden-toast-transition';
+
+export const StyledFadeInTransition = styled.div<{ isHidden: boolean; placement: ToastPlacement }>`
+  transition: opacity 400ms ease-in 300ms;
```
jzempel, line 17: It would be good to factor `400ms` into a constant and then update all following values based on this constant. This way, we clearly understand which timing depends upon another.

jzempel (changes requested):
As this is essentially non-discoverable, it would be good to add a minimal example to the README.

Other than the minor mentions, looking 🏆

packages/notifications/src/elements/toaster/Toast.tsx
hzhu: Yeah, that's a good point. SSR support slipped my mind. ⛔ `window` reference.

packages/notifications/stories/5-Toasts.stories.tsx
austingreendev: I've added this as a knob to the example. I'm still disabling the `autoDismiss` for the error examples though, if that's ok?

packages/notifications/src/elements/toaster/ToastProvider.tsx
austingreendev: @m-lai originally raised concerns about readability with more than 4 items which is what led us to the default. "unlimited" seems a unique use-case. If that was something that needed to occur the user could provide [Number.POSITIVE_INFINITY](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/POSITIVE_INFINITY).

packages/notifications/src/elements/toaster/styled.ts
austingreendev: 👍 Updated to a shared constant.

packages/notifications/stories/5-Toasts.stories.tsx
jzempel: This makes sense to me.

packages/notifications/src/elements/toaster/ToastProvider.tsx
jzempel: Got it – makes sense.

jzempel (approved):

PR #1022, opened 2021-02-26
fix(modals): match tooltipmodal close with fade out transition

hzhu:
## Description

From the color picker work it was determined that the `TooltipModal` needed to [handle the close fade-out](https://github.com/zendeskgarden/react-components/pull/1018#discussion_r581296253) animation properly. 

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/elements/TooltipModal/TooltipModal.tsx
+++ b/packages/modals/src/elements/TooltipModal/TooltipModal.tsx
@@ -122,6 +122,21 @@ export const TooltipModal = React.forwardRef<HTMLDivElement, ITooltipModalProps>
     const previousReferenceElementRef = useRef<HTMLElement | null>();
     const modalRef = useCombinedRefs(ref);
     const [popperElement, setPopperElement] = useState<HTMLDivElement | null>();
+    const [isOpen, setIsOpen] = useState(false);
+    let closeTimeout: number | undefined = undefined;
+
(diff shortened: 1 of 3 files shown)


packages/modals/src/elements/TooltipModal/TooltipModal.tsx
```diff
@@ -130,7 +145,7 @@ export const TooltipModal = React.forwardRef<HTMLDivElement, ITooltipModalProps>
       getModalProps
     } = useModal({
       id,
-      onClose,
+      onClose: handleClose,
```
hzhu, line 148: For `TooltipModals` that are animated, the `onClose` should invoke asynchronously — after the animation completes. This will prevent UI that is determined by the state that's changed in the `onClose` from being out of sync.

jzempel (commented):
@hzhu are you seeing the fade out effect? I'm not using the demo with `isAnimated` true. Are you sure you have the right setup?

hzhu:
@jzempel I figured it was subtle and it was there, but maybe it isn't. Is the fade out effect visible here with `isAnimated` toggled on:
https://zendeskgarden.github.io/react-components/storybook/?path=/docs/components-modals-tooltipmodal--default?

jzempel:
No it's not. Temporarily crank up the animation timing along with your useEffect hook to prove to yourself that you've got the right fix.

PR #1024, opened 2021-03-04
fix(modals): fadeout tooltip modal

hzhu:
### Description

This PR fades out the tooltip modal using `react-transition-group`.  

### Detail

Tooltip modals with the `isAnimated` prop hook into `react-transition-group`'s exit transition state and transitions the opacity to `0`. The outcome is that the fade-out transition is achieved without any API changes for the end consumers. There are edge cases where dynamic modal content could change immediately `onClose`, but this can be handled by consumers.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/styled/StyledTooltipModalBackdrop.ts
+++ b/packages/modals/src/styled/StyledTooltipModalBackdrop.ts
@@ -28,6 +28,11 @@ export const StyledTooltipModalBackdrop = styled.div.attrs({
   font-family: ${props => props.theme.fonts.system};
   direction: ${props => props.theme.rtl && 'rtl'};
 
+  &.garden-tooltip-modal-transition-exit-active div {
+    transition: opacity 200ms;
+    opacity: 0;
+  }
+
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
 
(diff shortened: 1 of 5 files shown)


packages/modals/src/styled/StyledTooltipModalBackdrop.ts
```diff
@@ -28,6 +28,11 @@ export const StyledTooltipModalBackdrop = styled.div.attrs({
   font-family: ${props => props.theme.fonts.system};
   direction: ${props => props.theme.rtl && 'rtl'};
 
+  &.garden-tooltip-modal-transition-exit-active div {
+    transition: opacity 200ms;
+    opacity: 0;
+  }
```
hzhu, line 34: This PR only involves adding the "fade-out" transition. The "fade-in" remains the same as before which leverages CSS animation from `menuStyles`.

jzempel (commented):
thanks for this, @hzhu. Please add `isAnimated: true` to the `TooltipModal.defaultProps`. @allisonacs confirms that this is the desired default for this component and matches the rest of Garden.

austingreendev (approved):
LGTM!

jzempel (approved):

PR #1051, opened 2021-03-22
chore(deps): upgrade styled-components to v5

austingreendev:
## Description

This PR upgrades `styled-components` to v5 and upgrades our related tooling and tests.

## Detail

@jzempel there are some unique warnings about our `StyledThumb` implementation in `Range` that I'm not able to decipher. Can you take a look in the CircleCI jest logs?

## Checklist

      designer as a reviewer)


jzempel:
The jest warning appear to be connected to https://github.com/zendeskgarden/react-components/blob/8fb1326568572622a18ca47c0777bd4c2be07008/packages/colorpickers/src/styled/Colorpicker/StyledAlphaRange.ts#L34-L35. However, this construct is absolutely intentional and necessary to get the `Range` inheritance styled as expected. Refactoring would be beyond the scope of this PR. Is there a way to get the console warnings dialed back?

jzempel (approved):

jzempel (changes requested):
The background on the `Colorpicker` alpha slider is whacked.

jzempel:
Ugh, from what I'm seeing, it looks as if `styled-components` is no longer honoring the `backgroundSize` followed by `background`. When I remove the line 34 shown above, I'm able to locally reproduce what we see in the production staging deploy.

It could be that `Range` will need to receive an internal (i.e. `/** @ignore */`) prop that allows us to knock-out background sizing while it is being extended by these `Colorpicker` components. TBH, it would also allow the accompanying logic to be knocked-out.

https://github.com/zendeskgarden/react-components/blob/8fb1326568572622a18ca47c0777bd4c2be07008/packages/forms/src/elements/Range.tsx#L22-L55

austingreendev:
@jzempel for the colorpicker checkered background logic to work I think we need to split the `checkeredBackground` utility into the equivalent shorthand CSS properties. I've been trying this, but I'm not able to match the original styling you intended. Could you give `StyledAlphaRange` implementation a look?

jzempel:
@austingreendev I think this is going to be a lot more straightforward to apply a prop to the underlying `Range` that knocks out the `backgroundSize` styling. Splitting up the `checkeredBackground` logic isn't going to result in code we want to maintain going forward.

jzempel (approved):

PR #1060, opened 2021-03-26
chore(forms, colorpickers): provide mechanism for removing Range lower track

jzempel:
## Description

Allow `Colorpicker` to remove competing `background` CSS from underlying `Range` components.

## Detail

@austingreendev this should clear the way for #1051.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/forms/src/elements/Range.tsx
+++ b/packages/forms/src/elements/Range.tsx
@@ -10,11 +10,16 @@ import { composeEventHandlers, useCombinedRefs } from '@zendeskgarden/container-
 import useFieldContext from '../utils/useFieldContext';
 import { StyledRangeInput } from '../styled';
 
+export interface IRangeProps extends InputHTMLAttributes<HTMLInputElement> {
+  /** @ignore */
+  hasLowerTrack?: boolean;
+}
+
 /**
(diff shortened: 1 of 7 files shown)


packages/forms/src/elements/Range.tsx
```diff
@@ -10,11 +10,16 @@ import { composeEventHandlers, useCombinedRefs } from '@zendeskgarden/container-
 import useFieldContext from '../utils/useFieldContext';
 import { StyledRangeInput } from '../styled';
 
+export interface IRangeProps extends InputHTMLAttributes<HTMLInputElement> {
```
jzempel, line 13: I'm kind of thinking I don't need this `export` since the prop is internal-only. Thoughts?

austingreendev (approved):

packages/forms/src/elements/Range.tsx
austingreendev: LGTM!

austingreendev: Could we use the `__NAME` scheme to denote an internal prop?

hzhu (approved):

PR #1071, opened 2021-04-01
fix(colorpickers): allow props to be applied to color dialog trigger

hzhu:
## Description

This change allows consumers to pass props to the color dialog trigger element. Useful for passing `aria-label` and other props to the trigger element.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/colorpickers/src/elements/ColorpickerDialog/index.tsx
+++ b/packages/colorpickers/src/elements/ColorpickerDialog/index.tsx
@@ -63,7 +63,7 @@ export interface IColorpickerDialogProps extends IColorpickerProps {
   /**
    * Passes HTML attributes to the color dialog button element
    */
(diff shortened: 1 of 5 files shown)


packages/colorpickers/src/elements/ColorpickerDialog/index.tsx
```diff
@@ -60,6 +60,10 @@ export interface IColorpickerDialogProps extends IColorpickerProps {
    * Applies inset `box-shadow` styling on focus
    */
   focusInset?: boolean;
+  /**
+   * Passes HTML attributes to the color dialog button element
+   */
+  triggerProps?: HTMLAttributes<HTMLButtonElement>;
```
jzempel, line 66: [nit] I'm leaning toward naming this `buttonProps` as these props are _not_ applied when a generic child trigger is in use.

jzempel (approved):

PR #1072, opened 2021-04-01
fix(notifications): use IE11 compatible react-uid

hzhu:
## Description

It seems like `nanoid` is causing Storybook to be in a perpetual loading state in Internet Explorer 11. A perpetual loading state can be reproduced locally and in production in IE11.

## Detail

This PR replaces `nanoid` usages with `react-uid` so that Storybook can be loaded on IE11. 

## Checklist

      designer as a reviewer)~


jzempel (approved):
ha, the bundle size actually reduced :tada:

PR #1076, opened 2021-04-09
feat(forms): add new radiogroup component

hzhu:
## Description

This PR introduces a `RadioGroup` component that encourages consumers to properly group radio buttons.

## Checklist

      designer as a reviewer)


packages/forms/stories/8-Radio.stories.tsx
```diff
@@ -30,24 +31,27 @@ export const Default: Story<ICheckboxStoryProps & InputHTMLAttributes<HTMLInputE
...
-            <Radio disabled={disabled} name="example">
-              <Label hidden={isHidden} isRegular={isRegular}>
-                Radio
-              </Label>
-              {showHint && <Hint>Hint</Hint>}
-              {showMessage && <Message validation={validation}>Message</Message>}
-            </Radio>
-          </Field>
+          <MD isBold>Radio group label</MD>
+          <RadioGroup aria-label="radio group label">
```
hzhu, line 35: Alternatively, this could be structured as the following:

```jsx
<RadioGroup aria-labelledby="id_1">
  <MD isBold id="id_1">Radio group label</MD>
  {/* radio buttons */}
<RadioGroup>
```

jzempel: Why not `<RadioGroup label="Radio group label">` and then use that for both the display and `aria-label`?

packages/forms/stories/8-Radio.stories.tsx
```diff
@@ -30,24 +31,27 @@ export const Default: Story<ICheckboxStoryProps & InputHTMLAttributes<HTMLInputE
...
+            <Field>
+              <Radio disabled={disabled} name="example">
+                <Label hidden={isHidden} isRegular={isRegular}>
```
jzempel, line 38: `Label`s that participate in a `RadioGroup` should use context to reverse the font-weight and use regular weight by default. In other words, the only way to get them back to semibold would be to set `isRegular={false}`.

packages/forms/stories/8-Radio.stories.tsx
```diff
@@ -30,24 +31,27 @@ export const Default: Story<ICheckboxStoryProps & InputHTMLAttributes<HTMLInputE
...
-          </Field>
-          <Field className="u-mt-xxs">
-            <Radio disabled={disabled} name="example">
-              <Label hidden={isHidden} isRegular={isRegular}>
-                Radio
-              </Label>
-              {showHint && <Hint>Hint</Hint>}
-              {showMessage && <Message validation={validation}>Message</Message>}
-            </Radio>
-          </Field>
```
jzempel, line 50: We should leave this story as is and add a new one for `RadioGroup`.

jzempel:
@allisonacs this PR presents the opportunity for design to determine spacing:
- between the group heading and the radios
- in between the radios themselves
- in between radios with/without hint text

As we look at the current example under https://garden.zendesk.com/components/radio#validation, it could be that the current layout (which has no spacing applied) would be considered too tight and needs a bit of air.

packages/forms/stories/8-Radio.stories.tsx
hzhu: No particular reasons not to do that. Before I make the changes I want to understand the suggetion.

Would the API look this this?

```jsx
<RadioGroup label="Choose a plant lifecycle">
  <Field>
    <Radio name="validation-example" value="annual">
      <Label>Annual</Label>
    </Radio>
  </Field>
  <Field>
    <Radio name="validation-example" value="perennial">
      <Label isRegular>Perennial</Label>
    </Radio>
  </Field>
</RadioGroup>
```

And the DOM output look like this?

```html
<div role="group" aria-label="Choose a growth type">
  <div data-garden-id="typography.font">Choose a growth type</div>
  <div>
     <input type="radio">
     <label>Annual</label>
  </div>
  <div>
     <input type="radio">
     <label>Perennial</label>
  </div>
</div>
```

packages/forms/stories/8-Radio.stories.tsx
hzhu: Updated and added some tests for this.

jzempel (changes requested):
@hzhu, I think there is a missed opportunity to follow the [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1) by using canonical DOM elements (`fieldset`). There are other items I want to dig into (API, styling, and demo tweaks), but this is the first item we should discuss and address.

packages/forms/stories/8-Radio.stories.tsx
hzhu: Marking as resolved due to change in implementation.

packages/forms/src/styled/common/StyledLegend.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const StyledLegend = styled.legend.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledLegend>`
+  margin-bottom: ${props =>
+    props.isCompact ? props.theme.space.base : props.theme.space.base * 2}px;
+  ${props => props.isHidden && hideVisually()}
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
```
jzempel, line 27: This should be (in abbreviated format):

```js
/**
 * 1. legend element reset.
 * /
StyledLegend = styled(StyledLabel).attrs({
  as: 'legend',
  ...
})`
  display: block;
  margin-bottom: ${props => props.theme.space.base * (props.isCompact ? 1 : 2)}px;
  padding: 0; /* [1] */

`;
```

Picks up all the missing styles, resets appropriately, and applies `hidden` as expected.

packages/forms/src/styled/common/SyledLegend.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render, screen } from 'garden-test-utils';
+import { StyledLegend } from './StyledLegend';
+
+describe('StyledLegend', () => {
```
jzempel, line 12: Only use this to test what is unique to `StyledLegend` apart from `StyledLabel`. i.e. `display: block;`, compact styling, or rendering as a `'LEGEND'` element.

packages/forms/src/styled/common/SyledLegend.spec.tsx
```diff
@@ -0,0 +1,37 @@
+/**
```
jzempel, line 1: File is misspelled. `StyledLegend.spec.tsx`.

packages/forms/src/styled/common/StyledLegend.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import styled from 'styled-components';
+import { hideVisually } from 'polished';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'forms.fieldset_legend';
+
+interface IStyledLegend {
+  isCompact?: boolean;
+  isHidden?: boolean;
```
jzempel, line 16: Remove. The correct usage is via the `hidden` attribute, as supported by all other `Label` components. See https://garden.zendesk.com/components/radio#hidden-label for details.

packages/forms/src/styled/common/StyledFieldset.ts
```diff
@@ -0,0 +1,29 @@
...
+export const StyledFieldset = styled.fieldset.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  margin: 0; /* [1] */
+  border: 0; /* [1] */
+  padding: 0; /* [1] */
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
```
jzempel, line 25: This should just be...

```js
export const StyledFieldset = styled(StyledField).attrs({
  as: 'fieldset',
})`
`;
```

It would also be consistent to include a `StyledFieldset.spec.tsx` that simply tests for a `'FIELDSET'` element rendering.

packages/forms/src/elements/common/Legend.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef, HTMLAttributes } from 'react';
+import { StyledLegend } from '../../styled';
+import useFieldsetContext from '../../utils/useFieldsetContext';
+
+export interface ILegendProps extends HTMLAttributes<HTMLLegendElement> {
+  isHidden?: boolean;
```
jzempel, line 13: Remove. The consistent option is to use the `hidden` HTML attribute.

packages/forms/src/elements/common/Fieldset.tsx
```diff
@@ -0,0 +1,51 @@
...
+  PropsWithoutRef,
+  FieldsetHTMLAttributes,
+  ForwardRefExoticComponent
+} from 'react';
+import { Legend } from './Legend';
+import { StyledFieldset } from '../../styled';
+import { FieldsetContext } from '../../utils/useFieldsetContext';
+
+export interface IFieldsetProps extends FieldsetHTMLAttributes<HTMLFieldSetElement> {
+  isCompact?: boolean;
```
jzempel, line 21: Missing documentation. Recommend "Applies compact styling" per https://garden.zendesk.com/components/input#input.

packages/forms/src/elements/common/Fieldset.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+interface IStaticFieldsetExport<T, P>
+  extends ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>> {
+  Legend: typeof Legend;
+}
+
+/* eslint-disable react/display-name */
+/**
+ * @extends HTMLAttributes<HTMLFieldSetElement>
```
jzempel, line 31: Should be `FieldsetHTMLAttributes<HTMLFieldSetElement>`

jzempel (changes requested):

packages/forms/src/styled/radio/StyledRadioLabel.ts
```diff
@@ -40,6 +40,8 @@ export const StyledRadioLabel = styled(StyledLabel).attrs({
   position: relative;
   cursor: pointer;
   user-select: none;
+  font-weight: ${props =>
+    props.isRegular === false ? props.theme.fontWeights.semibold : props.theme.fontWeights.regular};
```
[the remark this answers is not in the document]
hzhu: I had misunderstood https://github.com/zendeskgarden/react-components/pull/1076#discussion_r610846183. This has been reverted since it exists in `StyledLabel`.

packages/forms/stories/story-types.ts
```diff
@@ -22,6 +22,15 @@ export interface ICheckboxStoryProps {
   validation?: VALIDATION;
 }
 
+export interface IFieldsetStoryProps {
+  isRegular: boolean;
```
jzempel, line 26: Per https://github.com/zendeskgarden/react-components/pull/1076#discussion_r621486362, remove `isRegular`.

Per previous conversations, the `Fieldset` context should be leveraged so that the `Label` for any child element changes the default of `isRegular={true}`. In other words radios, checkboxes, and toggles below a `Legend` must display with non-bold text, by default. This means a user could override those elements with `isRegular={false}` to obtain bold text, but in general this is undesired from a design POV.

jzempel (changes requested):

packages/forms/stories/story-types.ts
hzhu: I pushed up a [new commit](https://github.com/zendeskgarden/react-components/pull/1076/commits/05d3573d02cdeca033d95a7d128d4c01792a1485) which allows labels within a `Fieldset` to render regular label styling by default. It can be overwritten with `isRegular={false}`.

packages/forms/src/elements/common/Legend.tsx
```diff
@@ -0,0 +1,23 @@
...
+import useFieldsetContext from '../../utils/useFieldsetContext';
+
+/**
+ * @extends HTMLAttributes<HTMLLegendElement>
+ */
+export const Legend = forwardRef<HTMLLegendElement, HTMLAttributes<HTMLLegendElement>>(
+  (props, ref) => {
+    const fieldsetContext = useFieldsetContext();
+
+    return <StyledLegend {...props} {...fieldsetContext} ref={ref} />;
```
jzempel, line 19: Should this be refined? With the recent change `isRegular` is being passed as an invalid prop to the `StyledLegend`.

packages/forms/stories/15-Fieldset.stories.tsx
```diff
@@ -0,0 +1,157 @@
...
+  showHint,
+  showMessage
+}) => {
+  return (
+    <Grid>
+      <Row>
+        <Col lg={4} offsetLg={4} md={6} offsetMd={3}>
+          <Fieldset isCompact={isCompact} disabled={disabled}>
+            <Fieldset.Legend hidden={isHidden}>Toggle growth types</Fieldset.Legend>
+            <Field style={{ marginBottom: '4px' }}>
```
jzempel, line 131: Please remove this margin customization. We want the designers to review the default layout.

jzempel (changes requested):

packages/forms/src/elements/common/Legend.tsx
hzhu: I think this should be okay. It looks like `StyledLegend` only consumes `isCompact`. Or should it also have an `isRegular` prop that matches the characteristics of `Label`?

jzempel: We don't want regular weight functionality applied to the `Legend`. Perhaps it would be more clear to destructure `isCompact` out for clarity. YMMV.

jzempel (approved):
Ok @allisonacs. This is now ready for a design review pass. 🙏

jzempel:
@allisonacs, just a reminder that I'm expecting to take the follow-on task for handling the desired `Message` spacing for radios, checkboxes, and toggles.

allisonacs (changes requested):
This is looking pretty good. Couple of things…

**Hint text**
Right now, we are only providing the ability to add hint text to individual form elements. See screenshot. I would expect that if we have a fieldset that we would want to include the hint under the primary label, a la text inputs. 

image

**Spacing in general**

Looks like an additional .8px is being added to the form element + label group. 

image

This is especially apparent on toggles. Since the toggles are 20px tall, I’d expect them to actually butt up on each other, but that’s not happening. With the .8 px being added to this group, there’s effectively an extra pixel on screen. 

image

**Toggle spacing**
Spacing looks good on radio and checkbox fieldsets, but when we get to toggle groups, things feel too tight (our toggles are 20px tall vs. 16px for the radio/checkboxes). See screenshot. 

image

So I’d like to address this. I think there are two options.

1. Special spacing just for toggles. We’d add 4px between each toggle element, which would make the toggles _feel_ equally spaced as the radios/checkboxes. But it does mess with vertical rhythm a bit (I’d expect a fieldset with 3 items to be the same size).
2. Add spacing for all fieldsets: checkboxes, radios, toggles. 

Option 2 is the most elegant. So let’s go with this:
- In all default fieldsets, add 8px between each form element
- In all compact fieldsets, add 4px between each form element

image

jzempel:
Great feedback @allisonacs. Thank you.

> Right now, we are only providing the ability to add hint text to individual form elements.

@hzhu let's just knock the control out of the stories

> Looks like an additional .8px is being added to the form element + label group.

@allisonacs this is out of scope for this PR as it currently affects all `StyledField` block display wrappers that surround labels, inputs, hints, and messages (it's been this way since forever). Block/line-height treatment for these types of DOM structures vary wildly between browsers and from one patch version to the next. (Chrome has a very bad track record here.) If we commit to pixel-chasing this one, let's do it under a separate effort.

> but when we get to toggle groups, things feel too tight

yeah, we expected you to see this and help us address it.

> Add spacing for all fieldsets: checkboxes, radios, toggles.

👍 @hzhu, I'd expect you can use CSS-only constructs to achieve what is being desired here. Please connect with me to go over the approach.

@allisonacs so is the spacing between `Toggle` components actually reduced a bit in order to make them match the others?

jzempel:
> If we commit to pixel-chasing this one, let's do it under a separate effort.

Actually, I lied. The fix here is pretty straightforward. We just need to add `font-size: 0;` to `StyledField`.

allisonacs:
> @allisonacs so is the spacing between `Toggle` components actually reduced a bit in order to make them match the others?

No. There should be 8px between each form element when stacked vertically in the default size, and 4px vertical between each in the compact setting. The screen shot I attached in the original comment shows how I expect it to look.

packages/forms/src/elements/common/Fieldset.tsx
```diff
@@ -43,7 +43,7 @@ export const Fieldset = forwardRef<HTMLFieldSetElement, IFieldsetProps>((props,
 
   return (
     <FieldsetContext.Provider value={fieldsetContext}>
-      <StyledFieldset {...props} ref={ref} />
+      <StyledFieldset {...props} ref={ref} isCompact={props.isCompact} />
```
jzempel, line 46: Should isCompact be destructured out of `{ isCompact, ...props }` in the function signature so that it isn't applied to the `StyledFieldset` twice? Or is the `{...props}` sufficient on this line without `isCompact`?

hzhu: Oops, yeah this is not needed since props are already passed to `StyledFieldset`.

allisonacs:
@hzhu @jzempel I understand we’ve knocked out the hint against individual form elements (in the example), but is it possible to have hint text on the fieldset label level?

jzempel:
> but is it possible to have hint text on the fieldset label level?

Yes, it is possible to drop the `<Hint>` component below the `<Fieldset.Legend>`. However, the styling setup will need some additional attention, as the current layout looks like this:

Screen Shot 2021-05-07 at 8 02 56 AM

Should we plan to tackle this as a PR follow-on?

jzempel:
p.s. @allisonacs if this is the desired layout, then it is a two-line change and would be trivial to get into this PR...

**Default**
Screen Shot 2021-05-07 at 8 42 11 AM

**Compact**
Screen Shot 2021-05-07 at 8 42 46 AM

jzempel (commented):
@hzhu this is the diff for supporting `<Hint>`. And, arguably, improved CSS over what we landed on for spacing.

Screen Shot 2021-05-07 at 9 17 11 AM

hzhu:
Thanks for the feedback! @allisonacs I've pushed a commit to add the hint text under the legend.

allisonacs:
The hint is looking good, thanks for the fix. One last change request:

In the default layout, the validation message should have a top margin of 8px. It should be 4px in the compact layout. This will make it consistent with all other validation states (input, etc.)

jzempel:
@allisonacs just a reminder that your `<Message>` concern is planned as a follow-up bug fix, rather than adding it as part of this feature. See https://github.com/zendeskgarden/react-components/pull/1076#issuecomment-832084974

PR #1081, opened 2021-05-04
feat(buttons,colorpickers,dropdowns,forms): add support for neutral button styling

jzempel:
## Description

Adds `isNeutral` prop with associated styling to all `Button` variants. The styling is designed to mesh well when combined with `Input` components as seen with the [color picker dialog](https://garden.zendesk.com/components/color-picker#form-input) and other usages.

## Detail

There was a list of knock-on effects – tweaks for consistency and bug fixes – resulting from the `isNeutral` addition:
- chore: all media icons (not just chevrons) respond to stateful actions (`:hover`, `:focus`, `:active`) across the component set – `Button`, `MediaInput`, `Autocomplete`, `Multiselect`, `Select`, and `Combobox`
- fix: focus `z-index` handling for both `ButtonGroup` and `InputGroup` components
- fix: apply missing `disabled` styling to `ColorpickerDialog`

**Note to reviewers**

Hit the following storybook components hard by testing the long list of control combinations...
- `Button`
- `ButtonGroup`
- `IconButton`
- `SplitButton`
- `ToggleButton`
- `ToggleIconButton`
- `Autocomplete`
- `Multiselect`
- `Select`
- `Combobox`
- `MediaInput`
- `InputGroup`

## Checklist

      designer as a reviewer)


packages/buttons/src/styled/StyledButton.ts
```diff
@@ -124,18 +127,25 @@ const colorStyles = (
...
-      color: ${baseColor};
+      color: ${foregroundColor};
 
       &:hover {
-        border-color: ${!props.isBasic && hoverColor};
-        background-color: ${rgba(baseColor as string, 0.08)};
-        color: ${hoverColor};
+        border-color: ${!props.isBasic && hoverBorderColor};
+        background-color: ${!props.isNeutral && rgba(baseColor as string, 0.08)};
+        color: ${hoverForegroundColor};
```
hzhu, line 144: I noticed that a button group (when toggled to use basic buttons) has a background color on hover and active:



It looks like the same does not happen for when the buttons are neutral. Is this expected?

jzempel: Yes, this is a storybook snafu. The only way out of it is to clear the URL parameters and refresh the page. This does not happen with normal usage.

jzempel: Yes, this is a storybook snafu, which seems to be crumbling with certain combos. The solution is to clear URL parameters and refresh the page. This does not seem to be happening with normal usage.

hzhu (approved):

allisonacs (changes requested):
- neutral plus danger button: there is no visible hover effect (the outline stays red)
- neutral plus basic button: no visible hover effect
- default button group - looks like it is missing its outline - same issue on neutral version: https://share.getcloudapp.com/z8u6RGYD'
- split buttons are missing their outlines in the default state - 
image
- Input group - I’d argue we should change the default input group to neutral
- The lack of background hover color is somewhat concerning to me. On the one hand, I know we want this to mimic the input design as much as possible. But in thinking of cases where someone wants to use a neutral button out in the wild (or even as part of the input group) it’s not super awesome. I wonder what we could do here. (maybe nothing, since we won’t be recommending standalone usage?) What do you think @steue @jzempel?

steue:
> * The lack of background hover color is somewhat concerning to me. On the one hand, I know we want this to mimic the input design as much as possible. But in thinking of cases where someone wants to use a neutral button out in the wild (or even as part of the input group) it’s not super awesome. I wonder what we could do here. (maybe nothing, since we won’t be recommending standalone usage?) What do you think @steue @jzempel?

Great point. The intention of the design was to work mostly in tandem with our input groups, since this styled button and input fields both rely on the border to communicate interactivity. This approach of `isNeutral` prop brings up the conflict of using both `isBasic` and `isNeutral` at the same time. We have a similar situation with the form inputs `isBare` prop. 

I think we should recommend as a best practice to use `isNeutral` in conjunction with input groups. I haven't seen any use for `isNeutral` treatment for stand alone buttons

steue:
Maybe it's a bug with storybook and not the component itself, but when using `isNeutral` for Chevron button, the color of both the button and chevron does not "neutralize" to grey.

jzempel:
@allisonacs...

> neutral plus danger button: there is no visible hover effect (the outline stays red)

fixed.

> neutral plus basic button: no visible hover effect

yeah, not great. I did the best based on the design info I had. But I'd +1 @steue on "I think we should recommend as a best practice to use isNeutral in conjunction with input groups."

> default button group - looks like it is missing its outline - same issue on neutral version

blerg. See https://github.com/zendeskgarden/react-components/pull/1081#discussion_r626035629. The initial storybook render gets this right. See:
- https://609187db7cda550ac0b7701b--zendeskgarden-react-components.netlify.app/?path=/story/components-buttons-buttongroup--default
- https://609187db7cda550ac0b7701b--zendeskgarden-react-components.netlify.app/?path=/story/components-buttons-buttongroup--default&args=isNeutral:true
Once `isPrimary` is toggled, the controls need to be reset and the page refreshed in order to get storybook back to a stable state.

> Input group - I’d argue we should change the default input group to neutral

Ah, missed that. Thanks. Fixed the control default.

> The lack of background hover color is somewhat concerning to me. ... (...even as part of the input group)

Yeah, it feels a bit off to me, and it has unexpected impacts for various system combinations – which we try to avoid. I'm good to go with a final recommendation between you and @steue. But it might help to know the `Button` would benefit from a fairly substantial code clean-up if we can convince ourselves to retain the button hover background color.

> when using `isNeutral` for Chevron button, the color of both the button and chevron does not "neutralize" to grey.

@steue Oops, I forgot to connect the control to the story. Thanks. Fixed now.

jzempel:
@steue @allisonacs when the deployment completes, the updated hover color background is ready for inspection.

Screen Shot 2021-05-06 at 4 06 23 PM

allisonacs (approved):

steue (approved):
thanks jz!

PR #1083, opened 2021-05-05
fix(dropdowns): handle combobox controlled open

jzempel:
## Description

The original intent of this PR was to add `isOpen` controls for debugging storybook dropdown demos. However, it brought to light a missing `useEffect` on `Combobox` for focusing the input when `isOpen` is set to true.

### DIFF
--- a/packages/dropdowns/stories/examples/Menu/Default.tsx
+++ b/packages/dropdowns/stories/examples/Menu/Default.tsx
@@ -30,6 +30,7 @@ interface IStoryProps {
   isAnimated: boolean;
   isCompact: boolean;
   isDisabled: boolean;
+  isOpen: boolean;
 }
 
 export const Default: Story<IStoryProps> = ({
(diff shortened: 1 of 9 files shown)


packages/dropdowns/stories/examples/Menu/Default.tsx
```diff
@@ -30,21 +30,23 @@ interface IStoryProps {
...
+  isDisabled,
+  isOpen
 }) => {
   return (
     <Grid>
       <Row style={{ minHeight: 200 }}>
         <Col textAlign="center">
           {/* eslint-disable-next-line no-alert */}
-          <Dropdown onSelect={item => alert(item)}>
+          <Dropdown onSelect={item => alert(item)} isOpen={isOpen || undefined}>
```
hzhu, line 49: What do you think about making this example a fully controlled component? I think it would get rid of the warnings in the console. Some [other](https://github.com/reach/reach-ui/blob/develop/packages/accordion/src/index.tsx#L77) [libraries](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMInput.js#L39) also log a warning for switching usages during the component life cycle.

screenshot

jzempel: ack, yeah I'll get these updated as you've suggested.

hzhu (approved):

packages/dropdowns/stories/examples/Menu/Default.tsx
jzempel: leaving as-is for now per offline conversation.

PR #1089, opened 2021-05-18
[VEG-397] feat(Dropdowns): Add data-test-id to Dropdown component

token-cjg:
## Description

In order to select Dropdowns easily during Cypress testing, it would be useful to be able to assign a unique data-test-id to each and every Dropdown on a page in order to reduce difficulty in testing changes.

In particular, while working on [this change](https://github.com/zendesk/admin_center_framework_webhooks/pull/106) I noticed that the Garden Dropdown component does not seem to support this at present, so the intent of this change is to alter said component so it does.

## References
* Cypress regression test PR that this change will support (after a bump to react-components in the relevant project, of course): https://github.com/zendesk/admin_center_framework_webhooks/pull/106
* Discussion in #garden: https://zendesk.slack.com/archives/C0AANB3HS/p1596224114239400

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
@@ -12,7 +12,7 @@ import { Dropdown, Trigger, Menu, Item, NextItem, PreviousItem } from '../..';
 import { IDropdownProps } from './Dropdown';
 
 const ExampleDropdown = (props: IDropdownProps) => (
-  <Dropdown {...props}>
+  <Dropdown dataTestId="dropdown" {...props}>
     <Trigger>
       <button data-test-id="trigger">Trigger</button>
     </Trigger>
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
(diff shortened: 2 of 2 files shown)


packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -16,7 +16,7 @@ import { DropdownContext, DROPDOWN_TYPE } from '../../utils/useDropdownContext';
 
 export const REMOVE_ITEM_STATE_TYPE = 'REMOVE_ITEM';
 
-export interface IDropdownProps {
+export interface IDropdownProps extends HTMLAttributes<HTMLSelectElement> {
```
jzempel, line 19: This is definitely not a DOM element – the addition of `extends` here is incorrect.

packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
```diff
@@ -12,7 +12,7 @@ import { Dropdown, Trigger, Menu, Item, NextItem, PreviousItem } from '../..';
 import { IDropdownProps } from './Dropdown';
 
 const ExampleDropdown = (props: IDropdownProps) => (
-  <Dropdown {...props}>
+  <Dropdown data-test-id="dropdown" {...props}>
```
jzempel, line 15: As a rule, we do not provide external test IDs – they are stripped by the build config. This is left as an exercise for the consumer in their own component test setup.

jzempel (changes requested):
Apologies but Garden won't be able to accept this PR.

Issue #1090, opened 2021-05-19
Group nav items collide with collapsed group


RomainEndelin:
## Expectations

NavItems and SubNavItems should line-up with no collision.

![Screenshot 2021-03-30 at 16 25 42](https://user-images.githubusercontent.com/1875908/118740475-5290fd80-b88f-11eb-9310-6dcc73354623.jpg)

## Reality

The list of items is refreshed after rendering but the `maxHeight` isn’t updated. This turns into a collision with the item below.

![Screenshot 2021-03-30 at 16 16 32](https://user-images.githubusercontent.com/1875908/118740401-207f9b80-b88f-11eb-81c5-8bd91bd8d426.jpg)

After a bit of digging, it seems to be caused by `useEffect` not working well with `useRef` ([here](https://github.com/zendeskgarden/react-components/blob/main/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx#L56-L60)), and React docs offer an [alternative solution](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) with `useCallback`.

## Steps to Reproduce

- Create a `CollapsibleSubNavItem`, containing an empty set of `SubNavItem`
- Add a `SubNavItem` below the `CollapsibleSubNavItem`
- Dynamically add some `SubNavItem` elements inside this `CollapsibleSubNavItem` (through a `useEffect`)

https://codesandbox.io/s/flamboyant-cerf-dedgq

### Proposed fix

```diff
index c0dff4d7d6..72c09b3f13 100644
--- a/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
+++ b/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
-import React, { useEffect, useRef, useState, ButtonHTMLAttributes } from 'react';
+import React, { useState, ButtonHTMLAttributes, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { useAccordion } from '@zendeskgarden/container-accordion';
 import { getControlledValue } from '@zendeskgarden/container-utilities';
@@ -35,7 +35,6 @@ export interface ICollapsibleSubNavItemProps
  */
 export const CollapsibleSubNavItem = React.forwardRef<HTMLDivElement, ICollapsibleSubNavItemProps>(
[...]

PR #1117, opened 2021-05-24
fix(chrome): recalculate height of subnavs when children change

hzhu:
## Description

The `CollapsibleSubNavItem` is not correctly recalculating its `maxHeight` when it's children updates.

## Detail

I've updated `CollapsibleSubNavItem` to recalculate its `maxHeight` when children changes. Also, introduced a new `itemCount` Storybook knob to help visualize the fix. Fixes https://github.com/zendeskgarden/react-components/issues/1090.

# Screens

**BEFORE:**

**AFTER:**

## Checklist

### DIFF
--- a/packages/chrome/.size-snapshot.json
+++ b/packages/chrome/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 58058,
-    "minified": 44720,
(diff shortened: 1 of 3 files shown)


jzempel (approved):
nice.

RomainEndelin:
Thanks!

PR #1120, opened 2021-06-08
feat(accordions): create new timeline component

hzhu:
## Description

This pull request introduces a new `Timeline` component as described by internal RFC and Figma assets.

## Detail

The `Timeline` component is used to render a list of chronological events. The `Timeline` component can be composed with other UIs to form more complex user experiences. For (unofficial) example, consumers can wrap [timeline items in tooltips](https://timeline-composition.netlify.app/iframe.html?id=components-accordions-timeline--default&args=&viewMode=story). Or they can place accordions, tabs, tables, etc. in timeline items. These higher level implementations are feature specific and is left as an exercise for the consumer at this time.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/accordions/src/elements/timeline/components/Item.tsx
+++ b/packages/accordions/src/elements/timeline/components/Item.tsx
@@ -1,0 +1,52 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, {
+  useMemo,
+  forwardRef,
+  Children,
+  ReactNode,
+  ReactElement,
+  HTMLAttributes
+} from 'react';
+import { Timeline } from '../Timeline';
+import { StyledTimelineItem } from '../../../styled';
+import { TimelineItemContext, useTimelineContext } from '../../../utils';
+
+export interface IItem extends HTMLAttributes<HTMLLIElement> {
+  /** Defines the icon rendered in place of the dot */
+  icon?: ReactNode;
+  /** Provides surface color for an SVG icon placed on a non-white background */
+  surfaceColor?: string;
+}
+
+export const Item = forwardRef<HTMLLIElement, IItem>(({ icon, surfaceColor, ...props }, ref) => {
+  const value = useMemo(() => ({ icon, surfaceColor }), [icon, surfaceColor]);
+  const { isAlternate } = useTimelineContext();
+  let hasOppositeContent = false;
+
+  Children.forEach(props.children, child => {
+    if (child) {
+      if ((child as ReactElement).type === Timeline.OppositeContent) {
+        hasOppositeContent = true;
+      }
+    }
+  });
+
+  return (
+    <TimelineItemContext.Provider value={value}>
+      <StyledTimelineItem
+        ref={ref}
+        isAlternate={isAlternate}
+        hasOppositeContent={hasOppositeContent}
+        {...props}
+      />
+    </TimelineItemContext.Provider>
+  );
+});
+
+Item.displayName = 'Item';
--- a/packages/accordions/src/styled/timeline/StyledItem.ts
+++ b/packages/accordions/src/styled/timeline/StyledItem.ts
@@ -1,0 +1,63 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { css } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledSeparator } from './StyledSeparator';
+import { StyledTimelineContent } from './StyledContent';
+import { StyledOppositeContent } from './StyledOppositeContent';
+
+const COMPONENT_ID = 'timeline.item';
+
+interface IStyledTimelineItem {
+  surfaceColor?: string;
+  isAlternate?: boolean;
+  hasOppositeContent?: boolean;
+}
+
+export const StyledTimelineItem = styled.li.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledTimelineItem>`
+  display: flex;
+  position: relative;
+
+  &:last-of-type ${StyledSeparator}::after {
+    display: none;
+  }
+
+  ${props =>
+    !props.hasOppositeContent &&
+    props.isAlternate &&
+    css`
+      &:before {
+        flex: 1;
+        content: '';
+        padding: ${props.theme.space.base * 4}px;
+      }
+    `}
+
+  ${props =>
+    props.isAlternate &&
+    css`
+      &:nth-child(even) {
+        flex-direction: row-reverse;
+        ${StyledOppositeContent} {
+          text-align: ${props.theme.rtl ? 'right' : 'left'};
+        }
+        ${StyledTimelineContent} {
+          text-align: ${props.theme.rtl ? 'left' : 'right'};
+        }
+      }
+    `}
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledTimelineItem.defaultProps = {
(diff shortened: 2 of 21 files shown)


packages/accordions/src/elements/timeline/components/Item.tsx
```diff
@@ -0,0 +1,52 @@
...
+  Children,
+  ReactNode,
+  ReactElement,
+  HTMLAttributes
+} from 'react';
+import { Timeline } from '../Timeline';
+import { StyledTimelineItem } from '../../../styled';
+import { TimelineItemContext, useTimelineContext } from '../../../utils';
+
+export interface IItem extends HTMLAttributes<HTMLLIElement> {
```
JamesSingleton, line 20: Just curious if `IItem` is meant to just be `Item`?

hzhu: Garden adopted the convention to prefix interface names with the letter "I".

JamesSingleton: Oh interesting, good to know 👍🏼

packages/accordions/src/elements/timeline/components/Item.tsx
```diff
@@ -0,0 +1,52 @@
...
+  surfaceColor?: string;
+}
+
+export const Item = forwardRef<HTMLLIElement, IItem>(({ icon, surfaceColor, ...props }, ref) => {
+  const value = useMemo(() => ({ icon, surfaceColor }), [icon, surfaceColor]);
+  const { isAlternate } = useTimelineContext();
+  let hasOppositeContent = false;
+
+  Children.forEach(props.children, child => {
+    if (child) {
```
ryanseddon, line 33: This is probably a redundant, if the component has no children then the [forEach short circuits and returns null](https://github.com/facebook/react/blob/aecb3b6d114e8fafddf6982133737198e8ea7cb3/packages/react/src/ReactChildren.js#L229-L231) so the callback is never entered.

packages/accordions/src/styled/timeline/StyledItem.ts
```diff
@@ -0,0 +1,63 @@
...
+import styled, { css } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { StyledSeparator } from './StyledSeparator';
+import { StyledTimelineContent } from './StyledContent';
+import { StyledOppositeContent } from './StyledOppositeContent';
+
+const COMPONENT_ID = 'timeline.item';
+
+interface IStyledTimelineItem {
+  surfaceColor?: string;
```
ryanseddon, line 17: Is this right name for this prop would something more obvious like `iconBackgroundColor` be more obvious? Feel free to ignore, I just haven't seen the `surface` wording used anywhere else in garden.

packages/accordions/src/styled/timeline/StyledOppositeContent.ts
```diff
@@ -0,0 +1,26 @@
...
+const COMPONENT_ID = 'timeline.opposite.content';
+
+export const StyledOppositeContent = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  flex: 1;
+  margin-right: auto;
+  padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 4}px`};
+  text-align: ${props => (props.theme.rtl ? 'left' : 'right')};
```
ryanseddon, line 20: Is IE11 support holding us back from using CSS logical properties?

packages/accordions/src/styled/timeline/StyledSeparator.ts
```diff
@@ -0,0 +1,54 @@
...
+    position: relative;
+    z-index: 2;
+    background: ${props => props.surfaceColor || props.theme.colors.background};
+    padding: ${props => props.theme.space.base}px 0;
+    color: ${props => getColor('neutralHue', 600, props.theme)};
+  }
+
+  &::after {
+    position: absolute;
+    left: calc(50% - 0.5px);
```
ryanseddon, line 42: Looks like you're getting subpixel issues here with the use of `.5px` looking at the deployed storybook the line is slightly off center. In fact if you remove the calc and just do 50% it doesn't actually move so the browser is ignoring the subpixel shift all together.

To actually make this truly center you'd need to make the circle an uneven number width e.g. 13px then you can do `left: (50% - 1px)` to have true center, I can't access the design assets so not sure if they're doing subpixel alignment or not.

ryanseddon (changes requested):
Looking good just some feedback and general questions

packages/accordions/src/styled/timeline/StyledTimeline.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'timeline.timeline';
+
+/**
+ * 1. <ol> reset.
+ */
+export const StyledTimeline = styled.ol.attrs({
```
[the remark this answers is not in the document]
hzhu: I had originally used a `<ul>` and @jzempel suggested the `<ol>`. I think it makes sense because a timeline is chronological (thus ordered). The `<ol>` seems more suitable from a semantic perspective.

packages/accordions/src/elements/timeline/components/Item.tsx
hzhu: This `if` check is intended to prevent runtime errors when users pass `undefined`, `false`, or `null` as children since there isn't a `type` property on these children.

packages/accordions/src/styled/timeline/StyledOppositeContent.ts
hzhu: I'll take a look at CSS logical properties — looks interesting. Garden components do need to support IE11 for the time being.

Normally, text alignment uses the default alignment and auto adjusts to RTL / LTR accordingly (via `direction` CSS property). In this particular case, timeline items can alternate, so explicit adjustment is needed.

packages/accordions/src/styled/timeline/StyledSeparator.ts
hzhu: I'm not seeing where the subpixel issues are. It looks like applying `left: 50%;` would cause a centering issue:

Currently, looks centered:
Screen Shot 2021-06-10 at 11 22 14 AM

With `left: 50%;` it no longer looks centered:
Screen Shot 2021-06-10 at 11 21 47 AM

hzhu: I think Garden needs to support centering for all sized icons. Garden's `svg-icons` (all even sizing) are used here. But these icons can be replaced with any valid React node including images, emojis, etc.

With an uneven size of `13px` and `left: calc(50% - 0.5)` it looks centered:
Screen Shot 2021-06-10 at 11 38 33 AM

With an uneven size of `13px` and `left: calc(50% - 1)` it looks not centered:
Screen Shot 2021-06-10 at 11 37 27 AM

hzhu: I noticed that the screenshots I posted in the pull request description is out-of-date. I think the Netlify deploy (click "View Deployment") should be up-to-date and have a centered line.

hzhu: I am noticing pixel issues here — but only sometimes — I think it can occur when the viewport width changes. Taking a deeper look and re-trying your suggestions.

hzhu: > make the circle an uneven number width e.g. 13px then you can do left: (50% - 1px) to have true center

I went with this solution 🙏🏽 . @steue had mention `13px` might be too large so went with `11px`.

packages/accordions/src/styled/timeline/StyledTimeline.ts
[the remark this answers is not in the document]
JamesSingleton: I agree that an `<ol>` here makes sense

JamesSingleton (approved):

ryanseddon (approved):

bethanybeech:
Looks good!

bethanybeech (approved):

PR #1126, opened 2021-06-29
feat(forms): create file list component

hzhu:
## Description

This pull request introduces a new `FileList` and `File` component in the `@zendeskgarden/forms` package. The component can be used to render a list of files, and can be used alongside the `FileUpload` component to build a complete file uploading experience.

Please see the accompanying (internal) file list RFC and Figma assets.

## Detail

The `FileList` component shows a list of (uploaded) files. Consumers may use it with a `Progress` component from Garden's `@zendeskgarden/loaders` package. The `FileUpload` story demonstrates this.

## Demo

Explore the demo using the bottom-most "View Deployment" link.  The file upload story can be compared to the current production story [here](https://zendeskgarden.github.io/react-components/?path=/story/components-forms-fileupload--default). The `FileList` and `File` components are demonstrated in two stories: 1) FileUpload 2) FileList.

1) `FileUpload` story demonstrates the usage of `FileUpload` with the `FileList`
2) `FileList` story demonstrates the usage of `FileList` alone

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/elements/file-list/utils.tsx
+++ b/packages/forms/src/elements/file-list/utils.tsx
@@ -1,0 +1,34 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import FilePdfStroke from '@zendeskgarden/svg-icons/src/16/file-pdf-stroke.svg';
+import FileZipStroke from '@zendeskgarden/svg-icons/src/16/file-zip-stroke.svg';
+import FileImageStroke from '@zendeskgarden/svg-icons/src/16/file-image-stroke.svg';
+import FileDocumentStroke from '@zendeskgarden/svg-icons/src/16/file-document-stroke.svg';
+import FileSpreadsheetStroke from '@zendeskgarden/svg-icons/src/16/file-spreadsheet-stroke.svg';
+import FilePresentationStroke from '@zendeskgarden/svg-icons/src/16/file-presentation-stroke.svg';
+
+export type FILE_TYPE = 'pdf' | 'zip' | 'image' | 'document' | 'spreadsheet' | 'presentation';
+
+export const ARRAY_FILE_TYPE: FILE_TYPE[] = [
+  'pdf',
+  'zip',
+  'image',
+  'document',
+  'spreadsheet',
+  'presentation'
+];
+
+export const fileIcons: Record<FILE_TYPE, React.ReactNode> = {
+  pdf: <FilePdfStroke />,
+  zip: <FileZipStroke />,
+  image: <FileImageStroke />,
+  document: <FileDocumentStroke />,
+  spreadsheet: <FileSpreadsheetStroke />,
+  presentation: <FilePresentationStroke />
+};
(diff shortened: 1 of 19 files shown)


allisonacs:
@hzhu there are a bunch of visual design issues I’m seeing. 

image

- The icon is vertically off center
- The text is #000000 (all black, not a garden color typically)
- The text is 16px (not a Garden type size)

Can you double check what’s going on there?

packages/forms/src/elements/file-list/utils.tsx
```diff
@@ -0,0 +1,34 @@
...
+export type FILE_TYPE = 'pdf' | 'zip' | 'image' | 'document' | 'spreadsheet' | 'presentation';
+
+export const ARRAY_FILE_TYPE: FILE_TYPE[] = [
+  'pdf',
+  'zip',
+  'image',
+  'document',
+  'spreadsheet',
+  'presentation'
+];
```
ryanseddon, line 25: Is it possible to use an enum here over repeating the strings twice? Fell free to ignore if this complicates stuff.

packages/forms/src/styled/file-list/StyledClose.ts
```diff
@@ -0,0 +1,65 @@
...
+ * 1. Reset for <button> element.
+ * 2. Remove dotted outline from Firefox on focus.
+ */
+export const StyledClose = styled.button.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<any>`
+  display: block;
+  transition: background-color 0.1s ease-in-out, color 0.25s ease-in-out;
+  z-index: 1;
```
ryanseddon, line 28: There's no `position` applied here so I don't think the `z-index` is necessary

ryanseddon (approved):
Not a lot of feedback lgtm

packages/forms/src/elements/file-list/utils.tsx
hzhu: Yeah it's possible. I've refactored this to use `enum` instead of `type`.  Looks like a `type` can be derived from an `enum` using `keyof typeof`. The `enum` is also flexible to be used as an object (albeit with a few caveats).

### Before

```typescript
export type FILE_TYPE = 'pdf' | 'zip' | 'image' | 'document' | 'spreadsheet' | 'presentation';

export const ARRAY_FILE_TYPE: FILE_TYPE[] = [
  'pdf',
  'zip',
  'image',
  'document',
  'spreadsheet',
  'presentation'
];
```

### After

```typescript
export enum FileType {
  pdf = 'pdf',
  zip = 'zip',
  image = 'image',
  document = 'document',
  spreadsheet = 'spreadsheet',
  presentation = 'presentation'
}

export type FILE_TYPE = keyof typeof FileType;

export const ARRAY_FILE_TYPE: FILE_TYPE[] = [...Object.values(FileType)];
```

https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

packages/forms/src/styled/file-list/StyledClose.ts
hzhu: You're right — after examining the DOM, this element is not a positioned element. Which is why I'm confused at why the `z-index` affects the element 🤔 .

With no `z-index`
Screen Shot 2021-07-06 at 11 20 56 AM

With `z-index: 1;`
Screen Shot 2021-07-06 at 11 21 01 AM

bethanybeech (approved):

packages/forms/src/styled/file-list/StyledClose.ts
jzempel: Please see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context#the_stacking_context for the scenarios in which a stacking context is formed.

PR #1127, opened 2021-06-30
fix(deps): upgrade polished dep due to security vulnerability

JamesSingleton:
https://app.snyk.io/vuln/npm:polished

                                                                                                                      

## Description

Upgrades the `polished` dependency to address https://app.snyk.io/vuln/npm:polished

## Checklist

      designer as a reviewer)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -13047,11 +13047,11 @@ pnp-webpack-plugin@1.6.4:
     ts-pnp "^1.1.6"
 
 polished@^4.0.5:
-  version "4.1.1"
-  resolved "https://registry.yarnpkg.com/polished/-/polished-4.1.1.tgz#40442cc973348e466f2918cdf647531bb6c29bfb"
-  integrity sha512-4MZTrfPMPRLD7ac8b+2JZxei58zw6N1hFkdBDERif5Tlj19y3vPoPusrLG+mJIlPTGnUlKw3+yWz0BazvMx1vg==
+  version "4.1.3"
+  resolved "https://registry.yarnpkg.com/polished/-/polished-4.1.3.tgz#7a3abf2972364e7d97770b827eec9a9e64002cfc"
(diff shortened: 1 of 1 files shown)


hzhu:
Looks good. What was the steps (commands) to make this upgrade? I want to verify it locally on my machine.

JamesSingleton:
> Looks good. What was the steps (commands) to make this upgrade? I want to verify it locally on my machine.

@hzhu, unfortunately, `yarn` does not give an easy way to upgrade sub dependencies like `npm`. Basically what I had to do was `rm yarn.lock && yarn install`. Copy the bit for `polished` from the `yarn.lock` file and then `git checkout .` and replace the old `polished` in the `yarn.lock` with the new. Doing anything like `yarn upgrade` or `yarn upgrade polished@latest` didn't work as doing `yarn upgrade polished@latest` is like doing `yarn add polished@latest`. There is a GitHub [Issue](https://github.com/yarnpkg/yarn/issues/2394) in yarn that talks about this and it not getting native support in yarn.

JamesSingleton:
@hzhu actually might have to decline this PR as it looks like I might have to update `@babel/runtime` as well... But let me know.

hzhu:
It looks like polished isn't a shared dependency at the root mono repo level, and the dependency found in the root `yarn.lock` is a result of an indirect dependency. Notice that polished isn't found in the root `package.json`. Also, this project strives to not directly modify the `yarn.lock` manually.

There are sub-packages that use polished, and those may be the packages that you're looking to upgrade? E.g. `react-forms`, `react-colorpickers`.

JamesSingleton:
@hzhu the issue looks to be `@zendeskgarden/react-theming` as it requires `polished@^4.0.0`. However, when installed into new deps it will install whatever is the latest at the time.

hzhu:
@JamesSingleton 

Looks like `react-theming` is one of the sub-packages that use polished. You could try `yarn upgrade polished` which should update the corresponding sub-package `yarn.lock` file. 

Since polished isn't a cross-package dependency (e.g not defined at the top-level `package.json`), I think the top level `yarn.lock` should remain unchanged. It should update accordingly as cross-package dependencies update to later versions of polished. These top-level dependencies are managed by Renovate.

hzhu:
@JamesSingleton any updates on this one?

PR #1158, opened 2021-07-22
fix(dropdowns): add missing data-garden- attributes to menu item icon elements

### LINKED ISSUE #1128: Missing data-garden-id on Menu Item Icon
https://github.com/zendeskgarden/react-components/blob/d4fd7bc1cd62ce5107f8464b00fbec197d96d673/packages/dropdowns/src/elements/Menu/Items/Item.tsx#L95

Is it possible to add a `data-garden-id` to the Icon for a selected menu item so it can be targeted by css selector outside?? In previous garden versions where these were inserted via a pseduo selector they could easily be found, but in the new setup without a `data-garden-id` on the `StyledItemIcon` its quite a pain.

jzempel:
## Description

Add missing `data-garden-[id/version]` attributes to `StyledItemIcon` components.

## Detail

Closes #1128

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/dropdowns/stories/examples/Menu/Advanced.tsx
+++ b/packages/dropdowns/stories/examples/Menu/Advanced.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { useState } from 'react';
+import React from 'react';
 import { Story } from '@storybook/react';
 import {
   Dropdown,
@@ -46,6 +46,7 @@ interface IStoryProps {
   isAnimated: boolean;
   isCompact: boolean;
   isDisabled: boolean;
+  isOpen: boolean;
 }
 
 export const Advanced: Story<IStoryProps> = ({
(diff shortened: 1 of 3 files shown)


packages/dropdowns/stories/examples/Menu/Advanced.tsx
```diff
@@ -46,29 +46,22 @@ interface IStoryProps {
...
         <Col textAlign="center">
-          <Dropdown
-            isOpen={isOpen}
-            onStateChange={changes => {
-              if (Object.prototype.hasOwnProperty.call(changes, 'isOpen')) {
-                setIsOpen(changes.isOpen === true);
-              }
-            }}
-          >
+          <Dropdown isOpen={isOpen || undefined}>
```
jzempel, line 64: I realize this treatment flips the component between controlled and uncontrolled (and results in a warning). But that the point – the control demonstrates both uses. More importantly, we have the ability to observe and use dev tools to debug the menu items which is imperative for the storybook demo.

hzhu: I think a more suitable structure would be to demonstrate one of each usage per story — similar to the [accordion](https://zendeskgarden.github.io/react-components/?path=/story/components-accordions-accordion--uncontrolled) stories. Probably not the right time for re-structuring, but just wanted to throw the idea out there.

hzhu (approved):

packages/dropdowns/stories/examples/Menu/Advanced.tsx
jzempel: Perhaps. This directly copies the `Default` dropdowns story. So we may want to address both.

PR #1159, opened 2021-07-28
chore(dropdowns,tooltips): Upgraded react-popper to 2.2.x

carvallegro:
## Description

Updated `react-popper` and `@popper/core` for the tooltips and dropdowns packages to align with the version used in modals.

This was creating a compatibility issue when using those three packages as only `v1.3.4` would be used.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/package.json
+++ b/packages/dropdowns/package.json
@@ -21,13 +21,14 @@
   "sideEffects": false,
   "types": "dist/typings/index.d.ts",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jzempel:
@carvallegro thanks for the PR. Unfortunately, Garden is not prepared to make this change as it results in a major breaking change to the API (namely, the `popperModifiers` prop). Look for this in an eventual v9.

PR #1160, opened 2021-08-02
fix(accordions): timeline view styling

jzempel:
## Description

This PR addresses a series of view styling issues with the `Timeline` component for improved theming alignment. See inline comments below for change rationale. Notably, odd-sized icons increased from 11 to 13px. Also add missing `surfaceColor` control to the storybook demo.

## Checklist

      designer as a reviewer)


packages/accordions/src/elements/timeline/components/Content.tsx
```diff
@@ -6,16 +6,17 @@
...
 
 export const Content = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>((props, ref) => {
   const { icon, surfaceColor } = useTimelineItemContext();
 
   return (
     <>
-      <StyledSeparator surfaceColor={surfaceColor}>
-        {icon || <StyledCircleStrokeIcon />}
+      <StyledSeparator>
+        <StyledItemIcon surfaceColor={surfaceColor}>{icon || <CircleIcon />}</StyledItemIcon>
```
jzempel, line 19: All icons should receive the same styling treatment, whether custom or not. And `surfaceColor` should be targeted to the component for which it directly applies.

packages/accordions/src/styled/timeline/StyledItemIcon.ts
```diff
@@ -0,0 +1,42 @@
...
+).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledItemIcon>`
+  z-index: 1;
+  box-sizing: content-box;
+  background-color: ${props => props.surfaceColor || props.theme.colors.background};
+  padding: ${props => props.theme.space.base}px 0;
+  width: ${props => math(`${props.theme.iconSizes.sm} + 1`)}; /* [1] */
+  height: ${props => math(`${props.theme.iconSizes.sm} + 1`)}; /* [1] */
```
jzempel, line 34: ☝️ Size should be expressed as a function of the `iconSizes` theming values.

packages/accordions/src/styled/timeline/StyledOppositeContent.ts
```diff
@@ -15,7 +15,6 @@ export const StyledOppositeContent = styled.div.attrs({
   'data-garden-version': PACKAGE_VERSION
 })`
   flex: 1;
-  margin-right: auto;
```
jzempel, line 18: ☝️ @hzhu do you know what this was intended for? If needed, it would have to be tuned for LTR.

packages/accordions/src/styled/timeline/StyledSeparator.ts
```diff
@@ -10,43 +10,23 @@ import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden
...
-  }
-
-  svg,
-  img {
-    position: relative;
-    z-index: 2;
-    background: ${props => props.surfaceColor || props.theme.colors.background};
-    padding: ${props => props.theme.space.base}px 0;
-    color: ${props => getColor('neutralHue', 600, props.theme)};
-  }
```
jzempel, line 40: ☝️ When possible, always prefer to add CSS directly against the styled component for which it applies, rather than relying on descendent selectors which are a bit more brittle to DOM structural changes.

packages/accordions/src/styled/timeline/StyledSeparator.ts
```diff
@@ -10,43 +10,23 @@ import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden
...
-
   &::after {
     position: absolute;
-    z-index: 1;
```
jzempel, line 44: ☝️ @hzhu not sure what this `z-index` was needed for?

packages/accordions/src/styled/timeline/StyledTimeline.ts
```diff
@@ -8,7 +8,7 @@
 import styled from 'styled-components';
 import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
-const COMPONENT_ID = 'timeline.timeline';
+const COMPONENT_ID = 'timeline';
```
jzempel, line 11: ☝️ Given that this component hasn't hit the website, it seemed safe to replace this non-standard naming scheme.

packages/accordions/stories/4-Timeline.stories.tsx
```diff
@@ -15,26 +15,24 @@ import EmailIcon from '@zendeskgarden/svg-icons/src/12/email-stroke.svg';
 import CartIcon from '@zendeskgarden/svg-icons/src/12/shopping-cart-stroke.svg';
 import ClipboardIcon from '@zendeskgarden/svg-icons/src/12/clipboard-blank-stroke.svg';
 
-const iconSize = { width: 11, height: 11 };
```
jzempel, line 18: ☝️ Direct attribute manipulation like this in storybook is a solid clue that the underlying component should be enhanced to cover expected cases.

packages/accordions/src/styled/timeline/StyledOppositeContent.ts
hzhu: I think this might have been accidentally left from previous iterations, and can be safely removed.

packages/accordions/src/styled/timeline/StyledSeparator.ts
hzhu: I think this element needs to be positioned on top of the icon. Since the icon's been updated to have a `z-index: 1` this should be safe to remove. I should've styled it like so to begin with.

hzhu (approved):

PR #1165, opened 2021-08-23
Add option to hide Alpha Modifiers for ColorPicker Component

maladeveloper:
__feat(ColorPicker):add option to hide alpha modifiers__

## Description
Add an option for the developer to hide the Alpha modification in the ColorPicker Component by passing in relavant prop. Motivation for this exists because the modifying the alpha value changes the hex string from 6 -> 8 elements, which, cannot be handled by some APIs.

| Before         |     After   (when prop hideAlphaModifiers passed as true)  |
| -------------- | :-----------: |
|image|image|

## Checklist

      designer as a reviewer)
- [x] :memo: tested in Chrome, Firefox, Safari, Edge, and ❌  IE11

## TODO
- Need approval from Garden Designer. @zendesk-garden  
- Need to update StoryBook such that the demo works with new prop added. 
- Need to test out on IE11 browser.
- Need to find out whether this needs to be analyzed via [axe].
- Need to find out how to render with Bedrock CSS.

### DIFF
--- a/packages/colorpickers/src/elements/Colorpicker/index.spec.tsx
+++ b/packages/colorpickers/src/elements/Colorpicker/index.spec.tsx
@@ -59,6 +59,15 @@ describe('Colorpicker', () => {
     expect(alphaInput.value).toBe('100');
   });
 
+  it('does not render alpha modifiers when passed with relavant prop value', () => {
+    render(<Colorpicker hideAlphaModifiers={true} />);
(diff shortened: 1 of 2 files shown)


jzempel:
Hello @maladeveloper. Thanks for bringing this for Garden's consideration. We actually have a designed task in our backlog to make this change. As it turns out, the design is going to require a bit more styling attention than just knocking out the elements. But please look for this update coming from the Garden team soon :wink:

PR #1173, opened 2021-09-03
chore(storybook): add browser env for disabling addon-docs

jzempel:
## Description

Adds a `BROWSER` environment variable check, that can be used to [disable Storybook docs-addon](https://storybook.js.org/docs/react/configure/babel) for localhost BrowserStack testing against IE11. Command usage shown below.

## Detail

```sh
BROWSER=ie11 yarn start
```

## Checklist

      designer as a reviewer)~

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -9,10 +9,11 @@ const path = require('path');
 const webpack = require('webpack');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const TS_CONFIG_PATH = path.resolve(__dirname, '../tsconfig.json');
+const docs = process.env.BROWSER ? process.env.BROWSER.toUpperCase() !== 'IE11' : true;
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


.storybook/main.js
```diff
@@ -9,10 +9,11 @@ const path = require('path');
 const webpack = require('webpack');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const TS_CONFIG_PATH = path.resolve(__dirname, '../tsconfig.json');
+const docs = process.env.BROWSER ? process.env.BROWSER.toUpperCase() !== 'IE11' : true;
```
jzempel, line 12: I'm open to feedback on this env var name/value. What makes more sense?

Francois-Esquire: `BROWSER` makes sense for me. Is it meant to be used like `BROWSER=ie11 yarn start`?

jzempel: yep, as described ☝️ 😉

Francois-Esquire (approved):

PR #1175, opened 2021-09-09
chore(scripts): enhance deploy with added bandwidth checking

jzempel:
## Description

Verify available Netlify bandwidth prior to staging deploy.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "jsx": "preserve",
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
     "moduleResolution": "node",
     "baseUrl": "./",
     "paths": {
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "jsx": "preserve",
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
```
jzempel, line 8: 79cfcfb commit addresses previous failure. Please consider https://www.typescriptlang.org/tsconfig#skipLibCheck during review.

hzhu: Which previous failures is this flag addressing? Is it the build failures where multiple copies of `@types/react` is the cause?

jzempel: Please note the commit I referenced with linked error info. This change resolves for `@types/download` duplicates scattered throughout various scripting tools. Note that the recommended `skipLibCheck` setting does not skip type checking for component dependency references. I think it is not only safe, but likely a preferred change for Garden.

hzhu (approved):

Issue #1167, opened 2021-08-24
A11y: Autocomplete's value isn't ready by VoiceOver


kgoggin:
## Expectations

When using the `Autocomplete` component with VoiceOver enabled, the field's selected value ought to be read by the screen reader. 

## Reality

The screen reader doesn't read the current value of the field.

![Screen Shot 2021-08-16 at 1 49 06 PM](https://user-images.githubusercontent.com/4549617/130667841-d2e1b99c-6eb2-4bfb-9006-2abf49a8a9d8.png)

## Steps to Reproduce

1. In the Garden storybook, open the Autocomplete story.
2. Turn on VoiceOver in Mac OS (Cmd-F5)
3. Tab to the field in the story. Notice that the screen reader doesn't read the value (default is Asparagus).

## Initial Investigation Notes

I took some time to try to understand what was causing this. I'll say up front that this is my second week at Zendesk and so I'm 100% new to Garden and its code 😆.

After inspecting the rendered output, I noticed that the underlying `<input />` that's being rendered didn't have any `value` set on it. The input props being [passed in from the prop getter](https://github.com/zendeskgarden/react-components/blob/main/packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx#L115) seem to always have `value` set to `""`, even after selecting a value. In fact, even when the Storybook component initially renders and `selectedItem` is set to "Asparagus", the input's value is `""` - this is because the input's state is being controlled externally via the story itself, but even after selecting a new value the input's value gets reset to `""`.

I wanted to see if having a value set would fix the screen reader issue, so I stopped passing in the `inputValue` to the example and let Downshift take over managing that state. I found that, on the initial render, the underlying `<input />` had its value set to "Asparagus", and the screen reader correctly read it out! 🎉 . But, if I changed the value, the `inputValue` being tracked in the internal state again got reset to `""`, and it broke again.

It looks like there's some logic that was added in #747 to handle explicitly clearing out the input field in certain circumstances. I think this is likely what's resetting that value in the internal state and leaving the DOM element without a value. Since I'm not familiar with the desired behavior there I wasn't comfortable trying to modify it, but it seems to me that if that could be tweaked to maintain the input's value that would effectively fix this.

[...]

PR #1176, opened 2021-09-09
fix(dropdowns): Ensure the input value is preserved and announced by screen readers

smritimadan:
## Description

Addresses a bug where the `value` of the `input` was cleared once the user blurred out of it, resulting in the screen reader not announcing the value even if there was a selection made and was visible to sighted users.

## Detail

This PR rolls back part of the changes in [this PR](https://github.com/zendeskgarden/react-components/pull/747) that caused an a11y regression described in [this issue](https://github.com/zendeskgarden/react-components/issues/1167). We no longer clear out the input's value and let the user clear it out themselves if they so choose. This ensures that the `input` always accurately reflects the selected value when not in use and the result of the user's input when in use.

This means the autocomplete will have a slightly different (albeit arguably more desirable) behavior from what exists today. The user will need to manually clear out the value of the input if needed. This is in line with ARIA ([example here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html#ex2_label)).

Closes https://github.com/zendeskgarden/react-components/issues/1167

## Checklist

      designer as a reviewer)
- [ ] :~guardsman: includes new unit tests~

### DIFF
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
@@ -241,7 +241,7 @@ describe('Dropdown', () => {
       container.querySelector('input')!.readOnly = false;
       userEvent.type(container.querySelector('input')!, 't');
 
-      expect(onInputValueChangeSpy.mock.calls[1][0]).toBe('t');
+      expect(onInputValueChangeSpy.mock.calls[0][0]).toBe('t');
     });
   });
 });
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -197,9 +197,6 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
           onStateChange && onStateChange(changes, stateAndHelpers);
         }}
         stateReducer={(_state, changes) => {
-          /**
-           * Set inputValue to an empty string during any event that updates the inputValue
-           */
           switch (changes.type) {
             case Downshift.stateChangeTypes.changeInput:
               if (changes.inputValue === '' && dropdownType === 'combobox') {
@@ -207,14 +204,6 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
               }
 
               return changes;
-            case Downshift.stateChangeTypes.controlledPropUpdatedSelectedItem:
-            case Downshift.stateChangeTypes.mouseUp:
-            case Downshift.stateChangeTypes.keyDownSpaceButton:
-            case Downshift.stateChangeTypes.blurButton:
-              return {
-                ...changes,
-                inputValue: ''
-              };
             case Downshift.stateChangeTypes.keyDownEnter:
             case Downshift.stateChangeTypes.clickItem:
             case REMOVE_ITEM_STATE_TYPE as any:
--- a/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
@@ -157,8 +157,23 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
         highlightedIndex={highlightedIndex}
         selectedItem={selectedItem || null} // Ensures that selectedItem never becomes controlled internally by Downshift
         inputValue={inputValue}
-        onInputValueChange={onInputValueChange}
+        onInputValueChange={(inputVal, stateAndHelpers) => {
+          if (onInputValueChange) {
+            if (stateAndHelpers.isOpen) {
+              onInputValueChange(inputVal, stateAndHelpers);
+            } else if (dropdownType === 'multiselect') {
+              onInputValueChange('', stateAndHelpers);
+            }
+          }
+        }}
         onStateChange={(changes, stateAndHelpers) => {
+          if (
+            dropdownType === 'autocomplete' &&
+            changes.isOpen === false &&
+            !changes.selectedItem
+          ) {
+            onSelect && onSelect(selectedItem, stateAndHelpers);
+          }
           if (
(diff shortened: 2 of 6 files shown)


packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -196,8 +188,8 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
               onSelect && onSelect(changes.selectedItem, stateAndHelpers);
             }
 
-            // Reset input value when item is selected
-            if (dropdownType !== 'combobox') {
+            // Reset input value when item is selected for multiselect - to clear input for next use
+            if (dropdownType === 'multiselect') {
```
smritimadan, line 192: Kept the same behavior for `Multiselect` until it can be addressed separately for its different concerns, likely needing a different a11y approach.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -219,7 +211,6 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
             case Downshift.stateChangeTypes.mouseUp:
             case Downshift.stateChangeTypes.keyDownSpaceButton:
             case Downshift.stateChangeTypes.blurButton:
```
smritimadan, line 213: I was on the fence about clearing the `input` when the button is blurred, so chose to keep parity with v7.

jzempel:
@smritimadan looks like the PR is missing an updated `.size-snapshot`. Can you add the commit so we get a staging deployment? I'm also adding @allisonacs to the reviewer list here.

smritimadan:
> @smritimadan looks like the PR is missing an updated `.size-snapshot`. Can you add the commit so we get a staging deployment? I'm also adding @allisonacs to the reviewer list here.

Just did!

jzempel (changes requested):
The changes appear to affect `Combobox` – that dropdown component needs to be able to persist the value that the user has entered. See https://zendeskgarden.github.io/react-components/?path=/story/components-dropdowns-combobox--default for comparison.

packages/dropdowns/stories/examples/Autocomplete/Default.tsx
```diff
@@ -66,7 +66,7 @@ export const Default: Story<IStoryProps> = ({
   showStartIcon
 }) => {
   const [selectedItem, setSelectedItem] = useState(options[0]);
-  const [inputValue, setInputValue] = useState('');
+  const [inputValue, setInputValue] = useState(options[0]);
```
smritimadan, line 69: Update our example to ensure the initial selection (and subsequent selections, see line 98) that a sighted user would see always match the `<input>`s `value` that a screen reader would announce.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -206,7 +213,7 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
         }}
         stateReducer={(_state, changes) => {
           /**
-           * Set inputValue to an empty string during any event that updates the inputValue
+           * Close the menu when the combobox input is cleared by the user
```
smritimadan, line 216: Retained this behavior from before, just added the comment for posterity.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -215,19 +222,6 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
...
-            case Downshift.stateChangeTypes.blurButton:
-            case Downshift.stateChangeTypes.blurInput:
-              return {
-                ...changes,
-                inputValue: ''
-              };
-            case Downshift.stateChangeTypes.keyDownEnter:
-            case Downshift.stateChangeTypes.clickItem:
-            case REMOVE_ITEM_STATE_TYPE as any:
-              return { ...changes, inputValue: '', isOpen: false };
```
smritimadan, line 230: Avoid all unneeded `stateReducer` logic because we are moving towards the approach where the user manually clears out the input when they want, instead of us doing it for them.

packages/dropdowns/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -161,12 +161,19 @@ const Dropdown: React.FunctionComponent<IDropdownProps & ThemeProps<DefaultTheme
...
           }
         }}
         onStateChange={(changes, stateAndHelpers) => {
+          if (
+            dropdownType === 'autocomplete' &&
+            changes.isOpen === false &&
+            !changes.selectedItem
+          ) {
+            onSelect && onSelect(selectedItem, stateAndHelpers);
+          }
```
smritimadan, line 176: Ensure we go back to the original selection in the event the user removes part of the input but doesn't commit to the change - by blurring or by selecting the same option again (which wouldn't trigger a selection on its own as that won't register as a "change").

packages/dropdowns/src/elements/Dropdown/Dropdown.spec.tsx
```diff
@@ -241,7 +241,7 @@ describe('Dropdown', () => {
       container.querySelector('input')!.readOnly = false;
       userEvent.type(container.querySelector('input')!, 't');
 
-      expect(onInputValueChangeSpy.mock.calls[1][0]).toBe('t');
+      expect(onInputValueChangeSpy.mock.calls[0][0]).toBe('t');
```
smritimadan, line 244: Since we no longer clear the input, this test makes a single mock fn call now. Previously, it was making 2 calls before, first with "" and second with the typed character, which was less than ideal anyway.

jzempel (approved):
lgtm. @allisonacs how does the behavior feel to you?

allisonacs (approved):
Looks good! Thanks Smriti!

hzhu (approved):

PR #1207, opened 2021-09-15
fix(datepickers): remove unused Month onChange prop

jzempel:
## Description

The `onChange` prop provided via context in the `DatepickerRange` component is not being used.

## Detail

Removes the unused prop identified (and subsequently ignored) in the #1189 renovate.

## Checklist

N/A

### DIFF
--- a/packages/datepickers/src/elements/DatepickerRange/utils/useDatepickerRangeContext.ts
+++ b/packages/datepickers/src/elements/DatepickerRange/utils/useDatepickerRangeContext.ts
@@ -17,7 +17,6 @@ export interface IDatepickerRangeContext {
   maxValue?: Date;
   startValue?: Date;
   endValue?: Date;
-  onChange?: (values: { startValue?: Date; endValue?: Date }) => void;
   startInputRef: MutableRefObject<HTMLInputElement | undefined>;
   endInputRef: MutableRefObject<HTMLInputElement | undefined>;
 }
--- a/packages/datepickers/.size-snapshot.json
+++ b/packages/datepickers/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 133388,
-    "minified": 75449,
-    "gzipped": 16823
+    "bundled": 133291,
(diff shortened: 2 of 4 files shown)


packages/datepickers/src/elements/DatepickerRange/utils/useDatepickerRangeContext.ts
```diff
@@ -17,7 +17,6 @@ export interface IDatepickerRangeContext {
   maxValue?: Date;
   startValue?: Date;
   endValue?: Date;
-  onChange?: (values: { startValue?: Date; endValue?: Date }) => void;
```
hzhu, line 20: Since `onChange` is removed from the context, should it also be removed from being [passed down](https://github.com/zendeskgarden/react-components/blob/main/packages/datepickers/src/elements/DatepickerRange/DatepickerRange.tsx#L146) the `<DatepickerRangeContext.Provider>`?

jzempel: Yeah. Might need to go in the opposite direction with this one and actually apply [documented](https://github.com/zendeskgarden/react-components/blob/77499ab31d87fb7a012c8c6746e64e41b2591b08/packages/datepickers/src/elements/DatepickerRange/DatepickerRange.tsx#L45-L52) `onChange` to the intended target. 🕵️

mtomcal:
So this is being removed from Context as the `onChange` prop is handled in [`Start.tsx`](https://github.com/zendeskgarden/react-components/blob/b8072d56e62a1ad272f171653484490f9fc45c6f/packages/datepickers/src/elements/DatepickerRange/components/Start.tsx#L15) via the Reducer state changes?

jzempel:
@mtomcal per https://github.com/zendeskgarden/react-components/pull/1207#discussion_r709353007 I'm not actually sure removal is the right move here. Just as likely that the implementation missed applying `onChange` to the proper underlying component(s).

jzempel:
Closing for now. Will re-open after further consideration.

PR #1213, opened 2021-09-21
refactor(components): replace utility useCombinedRefs with mergeRefs

Francois-Esquire:
## Description

Removes all usage of (to be) deprecated method `useCombinedRefs` in favor of [`mergeRefs` from `react-merge-refs`](https://github.com/gregberge/react-merge-refs).

## Detail

This is a refactor that takes all `useCombinedRefs` usage and replace it with `mergeRefs` without any behavior or API change. There were two scenarios, one being to merge forwardRefs with an internal ref per component and the other usage was combining multiple APIs exposing different refs that needed to be merged into the `ref={...}` prop.

## Checklist

      designer as a reviewer)


mtomcal (approved):
`mergeRefs` most certainly reads better and seems more inline with React best practices. 🚀

Francois-Esquire:
> I'm curious why it was necessary to default `useRef` calls to `null` rather than the previous `undefined`?

@jzempel there was no particular reason why `useRef` was given `null` other than to follow the projects convention when using `useRef` with elements (used in [Modal](https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/Modal.tsx#L105), [DrawerModal](https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/DrawerModal/DrawerModal.tsx#L92), [Dot Loader](https://github.com/zendeskgarden/react-components/blob/main/packages/loaders/src/elements/Dots.tsx#L49-L51), etc). Do you think we should omit the default `null`?

packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx
```diff
@@ -33,15 +34,15 @@ interface IAutocompleteProps extends HTMLAttributes<HTMLDivElement> {
 }
 
 const Autocomplete = React.forwardRef<HTMLDivElement, IAutocompleteProps>(
-  ({ children, inputRef: controlledInputRef, start, ...props }, ref) => {
+  ({ children, inputRef: controlledInputRef = null, start, ...props }, ref) => {
```
hzhu, line 37: Could you clarify why the addition of `null` default value is needed here?

packages/accordions/src/elements/accordion/components/Panel.tsx
```diff
@@ -5,15 +5,15 @@
...
 import debounce from 'lodash.debounce';
-import { useCombinedRefs } from '@zendeskgarden/container-utilities';
+import mergeRefs from 'react-merge-refs';
 import { useAccordionContext, useSectionContext } from '../../../utils';
 import { StyledPanel, StyledInnerPanel } from '../../../styled';
 
 export const Panel = forwardRef<HTMLElement, HTMLAttributes<HTMLElement>>((props, ref) => {
   const { isCompact, isBare, isAnimated, getPanelProps, expandedSections } = useAccordionContext();
-  const panelRef = useCombinedRefs<HTMLElement>(ref);
+  const panelRef = useRef<HTMLElement | null>(null);
```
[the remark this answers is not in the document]
Francois-Esquire: Agreed, we can compound this as part of a bigger effort and would help us keep usage consistency.

packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx
Francois-Esquire: The reason for the `null` default for the input ref is for [`react-merge-refs` and to avoid passing in an `undefined` ref](https://github.com/gregberge/react-merge-refs/blob/main/src/index.tsx#L10).

jzempel: @Francois-Esquire note that the loose `!=` equality check in `react-merge-refs` should work without needing to modify this code.

jzempel (changes requested):
Can we revert the `null` additions in this PR so that it is just focused on `mergeRefs` replacements? We can discuss `null` vs `undefined` preferences separately.

jzempel (approved):

packages/accordions/src/elements/accordion/components/Panel.tsx
```diff
@@ -5,15 +5,17 @@
...
-import React, { useCallback, forwardRef, HTMLAttributes } from 'react';
+import React, { useCallback, useRef, forwardRef, HTMLAttributes } from 'react';
 import debounce from 'lodash.debounce';
-import { useCombinedRefs } from '@zendeskgarden/container-utilities';
-import { useAccordionContext, useSectionContext } from '../../../utils';
+import mergeRefs from 'react-merge-refs';
+import { useAccordionContext, useSectionContext, IAccordionContext } from '../../../utils';
 import { StyledPanel, StyledInnerPanel } from '../../../styled';
 
+type PanelProps = IAccordionContext | { isBare?: boolean; isExpanded?: boolean };
```
hzhu, line 14: Not a blocker, but I think the TypeScript compiler will pick up the `isBare` type from the `IAccordionContext` interface.

hzhu, line 14: It seems like `isBare` is typed on `IAccordionContext`. Does the TypeScript compiler compile without `isBare` explicitly typed here?

packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx
hzhu: Could an alternative be to do a conditional check against `null` prior to using `mergeRefs`? I think that would satisfy the TypeScript compiler without changing the function signature.

Not a blocker, though, as `controlledInputRef` doesn't seem to be used elsewhere for now.

hzhu (approved):

packages/accordions/src/elements/accordion/components/Panel.tsx
Francois-Esquire: > It seems like `isBare` is typed on `IAccordionContext`. Does the TypeScript compiler compile without `isBare` explicitly typed here?

@hzhu you are absolutely right, it may have been my IDE linter didn't like `isBare` at the time. Everything seems fine without it as you pointed out. Going to remove.

hzhu (approved):

packages/dropdowns/src/elements/Autocomplete/Autocomplete.tsx
Francois-Esquire: @jzempel good point! Technically would be fine if it was `undefined` which was my initial concern. @hzhu I did return the fn signature however I wasn't able to completely remove the `null` from the `mergeRefs` call.

PR #1215, opened 2021-09-23
fix(types): export IAvatarProps and ILabelProps

mtomcal:
## Description

Some developers need access to `ILabelProps` and `IAvatarProps` in order to extend them.
* exported `ILabelProps`
* exported `IAvatarProps`

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/index.ts
+++ b/packages/dropdowns/src/index.ts
@@ -6,21 +6,32 @@
  */
 
 export { default as Dropdown } from './elements/Dropdown/Dropdown';
+export type { IDropdownProps } from './elements/Dropdown/Dropdown';
 export { default as Trigger } from './elements/Trigger/Trigger';
+export type { ITriggerProps } from './elements/Trigger/Trigger';
 export { default as Autocomplete } from './elements/Autocomplete/Autocomplete';
+export type { IAutocompleteProps } from './elements/Autocomplete/Autocomplete';
 export { default as Combobox } from './elements/Combobox/Combobox';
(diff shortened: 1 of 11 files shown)


packages/dropdowns/src/index.ts
```diff
@@ -28,5 +28,5 @@ export { MediaItem } from './elements/Menu/Items/MediaItem';
 export { NextItem } from './elements/Menu/Items/NextItem';
 export { PreviousItem } from './elements/Menu/Items/PreviousItem';
 export type { GARDEN_PLACEMENT, POPPER_PLACEMENT } from './utils/garden-placements';
-
+export type { ILabelProps } from './elements/Fields/Label';
```
mtomcal, line 31: Do we need to export all interfaces from these subcomponents in Dropdown retroactively? 🤔

saescapa:
+1! We are Typescriptfying some components that are wrappers for some of the Dropdown components and it would help a lot if we had these!

packages/dropdowns/src/index.ts
jzempel: All applicable element interfaces should be exported. See https://github.com/zendeskgarden/react-components/blob/main/packages/buttons/src/index.ts

jzempel (approved):

Francois-Esquire (approved):

hzhu (approved):

mtomcal:
Added all available interfaces to the export for dropdown. For other components, we can open up new PRs for that. At the present, this will go out in next week's release on Wed 29th.

Francois-Esquire (approved):

PR #1216, opened 2021-09-27
feat(typography): add diff language styling to `CodeBlock`

jzempel:
## Description

Prism.js, the underlying tokenizer for `CodeBlock`, supports `diff` language tokenization. This PR updates `CodeBlock` with first class styling support for:

```tsx
<CodeBlock language="diff">
@@ -1,3 +1,9 @@
-This paragraph contains
+This is an important
 part of the document
</CodeBlock>
```

## Detail

Screen Shot 2021-09-28 at 10 15 48 AM

## Checklist

      designer as a reviewer)


packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -9,14 +9,16 @@ import React, { HTMLAttributes, useMemo, useRef } from 'react';
...
+  SIZE,
   StyledCodeBlock,
   StyledCodeBlockContainer,
   StyledCodeBlockLine,
   StyledCodeBlockToken
 } from '../styled';
 
 export interface ICodeBlockProps extends HTMLAttributes<HTMLPreElement> {
-  /** Selects the language used by the Prism tokenizer */
+  /** Selects the language used by the [Prism](https://prismjs.com/) tokenizer */
```
mtomcal, line 21: ❤️ PrismJS 👍

mtomcal:
Not sure if this applies to this effort but Accessibility contrast color is flagged on Staging for CodeBlock:
Screen Shot 2021-09-27 at 10 00 14 AM

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -60,26 +68,52 @@ export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
         <Highlight Prism={Prism} code={code ? code.trim() : ''} language={language || 'tsx'}>
           {({ className, tokens, getLineProps, getTokenProps }) => (
             <StyledCodeBlock className={className} ref={ref} isLight={isLight} {...other}>
```
steue, line 70: think this might be a minor bug, but when switching from default to `isLight` back to default, the FG text sticks and doesn't change back.

[i've linked a gif screencap here to demonstrate what's happening](https://p-a39a79.f0.n0.cdn.getcloudapp.com/items/Qwu5ZNB6/df2312c1-104f-4fb4-b989-c450ca07c8d3.gif)

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -41,14 +43,20 @@ export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
     const containerRef = useRef<HTMLDivElement>(null);
```
mtomcal, line 43: If we have decided on `undefined` do we want to do preventative maintenance and remove `null`? Don't think this would be a breaking change.

steue:
> Not sure if this applies to this effort but Accessibility contrast color is flagged on Staging for CodeBlock:
> Screen Shot 2021-09-27 at 10 00 14 AM
> Screen Shot 2021-09-27 at 10 03 19 AM

i think i was getting that same visual bug? when toggling from default to `isLight` back to default.

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -41,14 +43,20 @@ export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
...
+        _size = 'sm';
+        break;
+
+      case 'medium':
+        _size = 'md';
+        break;
+
+      case 'large':
+        _size = 'lg';
+        break;
```
exelarios, line 59: Just for my own curiosity, is there a reason a switch statement is being used instead of a dictionary lookup?

mtomcal:
@steue Yeah definitely could be given Accessibility was flagged

packages/typography/src/elements/CodeBlock.tsx
jzempel: @steue I meant to note in the PR description, this is a CSS ordering problem specific to Storybook. Set the controls you want and then refresh the browser in order to see the correct styling.

jzempel:
@mtomcal https://github.com/zendeskgarden/react-components/pull/1216#discussion_r716903666

packages/typography/src/elements/CodeBlock.tsx
jzempel: Removing breaks typescript for `useScrollRegion`.

```
Type 'MutableRefObject<HTMLDivElement | undefined>' is not assignable to type 'RefObject<HTMLElement>'.
  Types of property 'current' are incompatible.
    Type 'HTMLDivElement | undefined' is not assignable to type 'HTMLElement | null'.
```

packages/typography/src/elements/CodeBlock.tsx
```diff
@@ -60,26 +68,52 @@ export const CodeBlock = React.forwardRef<HTMLPreElement, ICodeBlockProps>(
...
+                      } else if (token.types.includes('coord')) {
+                        retVal = 'hunk';
+                      } else if (token.types.includes('diff')) {
+                        retVal = 'change';
+                      }
+                    }
+                  }
+
+                  return retVal;
+                };
```
hzhu, line 92: Could `getDiff` be moved into the component function scope and accept a `line` argument so that `getDiff` does not have to be defined for each `line` in the map function scope?

hzhu (approved):

packages/typography/src/elements/CodeBlock.tsx
jzempel: There might be, but it isn't a good one 😅. Updated 😉.

packages/typography/src/elements/CodeBlock.tsx
jzempel: Hmm, I had it that way, but didn't have access to prism-react-renderer's `Token[]` interface. Here the scoped interface is implicit. I'll take a closer look and see what I can come up with.

hzhu: I saw that the `Token` type wasn't exported. It may have to defined it in the file (which seemed to work). It's a small type, but maybe the juice isn't worth the squeeze. 

Just a thought, not a blocker, so I'll resolve this.

mtomcal (approved):
LGTM. Reviewed for storybook, voiceover, and code style

smritimadan (approved):
Looks good!

steue (approved):
lgtm! thanks for working on these with me JZ!!!

Francois-Esquire (approved):

PR #1230, opened 2021-10-28
fix(colorpickers): retain previous indices for uncontrolled color swatch

hzhu:
## Description

The uncontrolled color swatch navigation does not work correctly after the first mount.

## Detail

The updated grid indices from `ColorSwatch` component (`onChange`) are not stored in the `ColorSwatchDialog` component. As a result, the focus indices are reset to `0, 0` after the first mount and subsequent navigations may not work correctly.

## Steps to reproduce 🐞 

1) Navigate to the uncontrolled [Storybook example](https://zendeskgarden.github.io/react-components/?path=/story/components-colorpickers-colorswatchdialog--uncontrolled)
2) Open the color swatch dialog 
3) Use arrow keys to select `Red-400`
4) Press `ESC` to close the dialog
5) Press `Enter` to open the dialog
6) Press left arrow key, and notice that the focus does not move ❌ 

## Screens

**Before**:

**After**: 

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/colorpickers/.size-snapshot.json
+++ b/packages/colorpickers/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 67310,
-    "minified": 42995,
-    "gzipped": 9956
+    "bundled": 67428,
+    "minified": 43005,
+    "gzipped": 9964
   },
   "index.esm.js": {
-    "bundled": 63100,
-    "minified": 39437,
-    "gzipped": 9680,
+    "bundled": 63218,
+    "minified": 39447,
+    "gzipped": 9690,
     "treeshaked": {
       "rollup": {
-        "code": 32894,
+        "code": 32906,
         "import_statements": 960
       },
       "webpack": {
-        "code": 36534
+        "code": 36545
       }
     }
   }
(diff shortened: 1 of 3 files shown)


jzempel (approved):

exelarios (commented):
There's a bug when the dialog isn't active it still reads key input, which causes the focus to jump to a different cell.
demo

hzhu:
@exelarios I'm unable to visualize that bug given the description and the fast screen. Is that an existing bug? Or a result of something that was changed in this PR? If it's an existing bug, I can create a new task to address it.

What are the steps to reproduce that bug?

exelarios:
@hzhu 
I just tried to reproduce it, but it seems I mistaken it was a bug since I thought the focus would set to where the checkmark cell is once the dialog is toggled, but it's actually retaining the position where it lasted focus and I thought it jumped to a random position. 😅

jzempel:
@exelarios @hzhu I'm seeing the bug mentioned under https://github.com/zendeskgarden/react-components/pull/1230#pullrequestreview-795790026. It should be addressed either here or under a follow-on.

hzhu:
@exelarios @jzempel 

Looks like the recording is describing a bug where the focused indices are persisted when the dialog is closed. This is normally okay, but when the focused indices are different than the selected indices, the next time the dialog is opened focus navigation starts at the focused indices (incorrect).

I pushed up a new commit which should address that issue by setting the internal uncontrolled indices to the selected indices when the dialog is closed.

jzempel (approved):
Nice. Great catch @exelarios

PR #1231, opened 2021-10-29
feat(chrome): create sheet components

Francois-Esquire:
## Description

Adds `Sheet` and subcomponents that make up the parts of the `Sheet`.

## Detail

The `Sheet` is a complementary component that will be part of the `Chrome` suite of components.

Screen Shot 2021-10-29 at 5 15 42 PM

It supports RTL and placement awareness:

Screen Shot 2021-10-29 at 5 15 28 PM

It can be animated when opened and closed as an optional prop:

sheet-open-close

## Checklist

      designer as a reviewer)


packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+  it('uses placement prop to apply smart border to the correct side', () => {
+    const { getByRole } = render(<Example isOpen placement="start" />);
+
+    expect(getByRole('complementary')).toHaveStyleRule(
+      'border-right',
+      `1px solid ${DEFAULT_THEME.palette.grey[300]}`
+    );
+  });
+
+  it('renders correctly in rtl mode', () => {
```
hzhu, line 98: Style assertions are usually colocated within the `styled` directory. For example, here's a [style assertion](https://github.com/zendeskgarden/react-components/blob/main/packages/accordions/src/styled/accordion/StyledRotateIcon.spec.tsx#L49) on RTL in the accordions package that's colocated in the `style` directory.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <Sheet.Body>body</Sheet.Body>
+          <Sheet.Footer data-test-id="footer">
+            <Sheet.FooterItem>
+              <button>button1</button>
+            </Sheet.FooterItem>
+            <Sheet.FooterItem>
+              <button>button2</button>
+            </Sheet.FooterItem>
+          </Sheet.Footer>
+          <Sheet.Close data-test-id="close-btn" />
```
hzhu, line 39: I think these test IDs can be removed in favor of [priority querying](https://testing-library.com/docs/queries/about/#priority). For elements like `Sheet.Close`, simulation of `tab` + `enter` key events can be used to make assertions against trigger behavior.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+import React, { useRef, useState, forwardRef } from 'react';
+import userEvent from '@testing-library/user-event';
+import { render } from 'garden-test-utils';
+import { ThemeProvider, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+import { Sheet, ISheetProps } from './Sheet';
+
+describe('Sheet', () => {
+  const Example = forwardRef<HTMLElement, ISheetProps>((props: ISheetProps, ref) => {
+    const [isOpen, setIsOpen] = useState(() => props.isOpen);
```
hzhu, line 17: Could you clarify the purpose of using lazy initial state?

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const { getByRole } = render(<Example isOpen placement="start" />);
+
+    expect(getByRole('complementary')).toHaveStyleRule(
+      'border-right',
+      `1px solid ${DEFAULT_THEME.palette.grey[300]}`
+    );
+  });
+
+  it('renders correctly in rtl mode', () => {
+    const { getByTestId } = render(
```
hzhu, line 99: Feel free to use the `renderRtl` [utility](https://github.com/zendeskgarden/react-components/blob/main/utils/test/garden-test-utils.tsx#L47) which is also exported from `garden-test-utils`.

packages/chrome/src/elements/sheet/components/Close.tsx
```diff
@@ -0,0 +1,26 @@
...
+import XStrokeIcon from '@zendeskgarden/svg-icons/src/16/x-stroke.svg';
+import { IconButton } from '@zendeskgarden/react-buttons';
+
+import { StyledSheetCloseButtonContainer } from '../../../styled';
+
+export const SheetCloseButton = forwardRef<HTMLElement, HTMLAttributes<HTMLElement>>(
+  (props, ref) => {
+    return (
+      <StyledSheetCloseButtonContainer>
+        <IconButton aria-expanded aria-label="Close Sheet" ref={ref} {...(props as any)}>
```
hzhu, line 18: Will hard coding `aria-label` in English be problematic for i18n?

hzhu (changes requested):

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,138 @@
...
+export const Sheet = React.forwardRef<HTMLElement, ISheetProps>(
+  ({ id, isOpen, isAnimated, focusOnMount, restoreFocus, children, ...props }, ref) => {
+    const sheetRef = useRef<HTMLElement>(null);
+
+    const seed = useUIDSeed();
+    const [idPrefix] = useState<string>(id || seed(`sheet_${PACKAGE_VERSION}`));
+    const titleId = `${idPrefix}--title`;
+    const descriptionId = `${idPrefix}--description`;
+
+    const context = useMemo(() => ({ idPrefix }), [idPrefix]);
```
mtomcal, line 73: `sheetContext`? I was confused at first #nitpick

packages/chrome/src/styled/sheet/StyledSheetBody.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const COMPONENT_ID = 'chrome.sheet_body';
+
+export const StyledSheetBody = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  flex: 1;
+  overflow-y: auto;
+  padding: 20px;
```
mtomcal, line 19: Any particular reason not to use `theme` driven padding configuration? (e.g. [`StyledPanel.ts`](https://github.com/zendeskgarden/react-components/blob/main/packages/accordions/src/styled/accordion/StyledPanel.ts))

packages/chrome/src/styled/sheet/StyledSheetCloseButtonContainer.ts
```diff
@@ -0,0 +1,34 @@
...
+    ${placement}: ${props.theme.space.base * 1}px;
+  `;
+};
+
+export const StyledSheetCloseButtonContainer = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  display: block;
+  position: absolute;
```
mtomcal, line 27: Any reasoning to not use `display: flex` for positioning?

packages/chrome/src/styled/sheet/StyledSheetFooterItem.ts
```diff
@@ -0,0 +1,24 @@
...
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet_footer_item';
+
+export const StyledSheetFooterItem = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  margin: auto 10px;
```
mtomcal, line 17: Similar https://github.com/zendeskgarden/react-components/pull/1231/files?authenticity_token=ocFDB1GmyeZM0qKO%2B%2B%2BaueEsdBPb4UU%2F%2FUUGDkPrbzizvqjKB%2FQzJUVHh0bjsIMD1%2FkviL9azLOv18Ws9Ws0vA%3D%3D&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#r741539973

packages/chrome/src/styled/sheet/StyledSheetHeader.ts
```diff
@@ -0,0 +1,39 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  display: flex;
+  flex-flow: ${props => (props.theme.rtl ? 'row-reverse' : 'row')} wrap;
+  align-content: space-between;
+  align-items: center;
+  border-bottom: 1px solid ${props => props.theme.palette.grey[300]};
+  padding: 20px;
+  min-height: 40px;
```
mtomcal, line 23: [Similar](https://github.com/zendeskgarden/react-components/pull/1231/files?authenticity_token=ocFDB1GmyeZM0qKO%2B%2B%2BaueEsdBPb4UU%2F%2FUUGDkPrbzizvqjKB%2FQzJUVHh0bjsIMD1%2FkviL9azLOv18Ws9Ws0vA%3D%3D&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#r741539973)

packages/chrome/stories/examples/Default.tsx
```diff
@@ -81,9 +87,11 @@ export const Default: Story<IDefaultStoryProps> = ({
   isExpanded,
   showSubnav,
   showSidebar,
+  showSheet,
   hueColor,
   itemCount
 }) => {
+  const [_, updateArgs] = useArgs();
```
mtomcal, line 94: What does `_` mean? Im confused.

mtomcal:
Deployment of storybook would be helpful. 🥺

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Good call on this, it isn't necessary. Will update

packages/chrome/src/elements/sheet/components/Close.tsx
Francois-Esquire: It could be problematic if it is not documented. Is there a standard Garden way of handling internationalization scenarios like this?

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Good to know, I'll place the appropriate tests there.

packages/chrome/stories/examples/Default.tsx
Francois-Esquire: It doesn't have a meaning really, it was meant to be a placeholder and treating the array like parameters (which isn't necessary). Will update to `[, updateArgs]`

packages/chrome/src/styled/sheet/StyledSheetHeader.ts
Francois-Esquire: Like the `StyledBody` it was specific to the `Sheet` design spec.

packages/chrome/src/styled/sheet/StyledSheetFooterItem.ts
Francois-Esquire: According to the `Sheet` design spec and like `StyledBody`, `FooterItem`s have a 10px margin. Do you think this should be a scalar value based on another value?

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: Yes, the `Sheet` context binds the `Sheet` with it's counterparts for accessibility purposes. The `idPrefix` is used for `aria-describedby` and `aria-labelledby` on the `aside` element that point to the `Sheet.Title` and `Sheet.Description`.

mtomcal: Rename to `sheetContext` for clarity?

packages/chrome/src/styled/sheet/StyledSheetCloseButtonContainer.ts
Francois-Esquire: I think either would work - just realized a `div` has a default `display: block`. Any particular reason to use `flex` in this instance?

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: Sure, will update if that makes it clearer.

packages/chrome/src/elements/sheet/components/Close.tsx
hzhu: Looks like the other `Close` buttons in Garden pass through the attribute so that consumers can pass in arbitrary strings. E.g. `<Sheet.Close aria-label={translation} />`. I think the hard-coded `aria-label` can be removed since it prevents users from passing in custom strings, particularly for i18n.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Agreed, I've updated and removed the test ids.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Updated and moved the tests to the `styled/` directory.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: TIL, updated tests to use this!

packages/chrome/src/styled/sheet/StyledSheetBody.ts
Francois-Esquire: Updated to utilize theme space base, from our discussion offline.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,81 @@
...
+  flex-direction: column;
+  order: 1;
+
+  transition: width 0.25s ease-in-out;
+  will-change: width;
+
+  width: 380px;
+  min-height: 100%;
+
+  & > * {
```
hzhu, line 45: It looks like this selector applies a transition (`transition: opacity 0.5s ease-in-out)` to the children elements of the sheet. Is there a reason why header, body, footer, and the close button need that transition?

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const descriptionId = `${idPrefix}--description`;
+
+    const sheetContext = useMemo(() => ({ idPrefix }), [idPrefix]);
+
+    useFocusableMount({ targetRef: sheetRef, isMounted: isOpen, focusOnMount, restoreFocus });
+
+    if (isAnimated) {
+      return (
+        <SheetContext.Provider value={sheetContext as ISheetContext}>
+          <CSSTransition in={isOpen} unmountOnExit timeout={250} classNames="side-sheet-transition">
```
hzhu, line 80: When the sheet is closing, the transition feels like it is collapsing both start and the end of the sheet, in other words, it's transitioning from both horizontal sides into the center. Is this because the `width` is being transitioned?

I think it should only transition from start to end (LTR) or end to start (RTL). Similar to the drawer modal.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,81 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION,
+  tabIndex: -1
+})<IStyledSheetProps>`
+  display: flex;
+  position: relative;
+  flex-direction: column;
+  order: 1;
+
+  transition: width 0.25s ease-in-out;
```
hzhu, line 39: I believe [`transform`](https://csstriggers.com/transform) can be used in this situation instead of [`width`](https://csstriggers.com/width).  The `transform` property does not trigger geometry changes or painting and can be carried out by the compositor thread with the help of the GPU.

exelarios: I didn't know there was a difference between `transform` and `transition`. Thanks for pointing that out, we'll definitely look into it and change it.

packages/chrome/stories/examples/Sheet.tsx
```diff
@@ -0,0 +1,156 @@
...
+};
+
+SheetStory.args = {
+  rtl: false,
+  isOpen: true,
+  isAnimated: true,
+  focusOnMount: false,
+  restoreFocus: false,
+  isCompact: false,
+  placement: 'end'
```
hzhu, line 123: I don't notice any difference in the UI/UX when the `placement` knob is toggled. How is the UI/UX different when the `placement` toggle is switched to `start` (vs `end`)?

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <StyledSheet
+              id={idPrefix}
+              aria-labelledby={titleId}
+              aria-describedby={descriptionId}
+              ref={mergeRefs([sheetRef, ref])}
+              {...(props as any)}
```
hzhu, line 86: The TypeScript compiler is throwing an error here because the type definition for `StyledSheet` indicates `placement` as a required prop.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const titleId = `${idPrefix}--title`;
+    const descriptionId = `${idPrefix}--description`;
+
+    const sheetContext = useMemo(() => ({ idPrefix }), [idPrefix]);
+
+    useFocusableMount({ targetRef: sheetRef, isMounted: isOpen, focusOnMount, restoreFocus });
+
+    if (isAnimated) {
+      return (
+        <SheetContext.Provider value={sheetContext as ISheetContext}>
```
hzhu, line 79: I think the type for `sheetContext` can be inferred by TypeScript so the type casting can be removed.

packages/chrome/src/elements/sheet/components/Body.tsx
```diff
@@ -0,0 +1,17 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef, HTMLAttributes } from 'react';
+import { StyledSheetBody } from '../../../styled';
+
+export const SheetBody = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
+    return <StyledSheetBody tabIndex={0} ref={ref} {...(props as any)} />;
```
hzhu, line 13: It seems like TypeScript compiles without this `as any` type cast. These type casts might be left over from before the `@types/react` was reconciled and can be removed.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,81 @@
...
+  will-change: width;
```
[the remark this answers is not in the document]
exelarios: The property was intended to be a performance improvement for animations. Mike and I weren't aware it was intended to be used as a last resort. We'll remove it from the `StyledSheet` since there aren't any performance problems, to begin with. Thanks!

packages/chrome/src/elements/sheet/components/Body.tsx
hzhu: Could you clarify why the body element can receive keyboard focus?

packages/chrome/src/elements/sheet/components/Close.tsx
hzhu: According to ARIA the usage for `aria-expanded`: 

> Indicates whether the element, or another grouping element it controls, is currently expanded or collapsed.

My interpretation is that `aria-expanded` is applied to the trigger that controls the sheet — not a close trigger within the sheet. For example, if there was a element that toggles the sheet:

```jsx
<button aria-expanded={isOpen}>Toggle Sheet</button>

<Sheet />
```

packages/chrome/src/styled/sheet/StyledSheet.ts
exelarios: It's there so when the container collapses or expands the contents from the sub-components don't appear squished.

Transition applied to the children:
demo1

Without Transition applied to children:
demo1

packages/chrome/stories/examples/Sheet.tsx
exelarios: The `placement` prop is to allow the consumer to explicitly state where the `Sheet` will be positioned, therefore the borders will react accordingly to prevent double bordering. 

When the `Sheet` is positioned on the `right`, therefore placement will be `end`
image

When the `Sheet` is positioned on the `left`, therefore placement will be `start`
image

Francois-Esquire: It shifts over the border to expect the `Sheet` (which is very subtle) to be placed on that respective placement.

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: Good call, will update this.

packages/chrome/src/elements/sheet/Sheet.tsx
exelarios: Nice catch! Thanks!

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  it('does not animate a transition when isAnimated is false', () => {
+    const { getByRole, getByText } = render(<Example isAnimated={false} />);
+
+    userEvent.click(getByText('Toggle Sheet'));
+
+    expect(getByRole('complementary')).not.toHaveAttribute('class', 'side-sheet-transition');
+  });
+
+  it('contains a11y bindings to label and describe the sheet', () => {
```
hzhu, line 74: Asserting against attributes (typically) side-steps `testing-library`'s [guiding principles](https://testing-library.com/docs/guiding-principles/) by asserting against implementation details. The user does not care if the title has a specific ID — they care if their browser is communicating that the title is mapped to the correct element.

The following would more align with `testing-library`'s guiding principles.

```jsx
it('has an accessible title and description', () => {
  render(<Example isOpen />);

  const sheet = screen.getByRole('complementary');

  // asserts the title is associated with the correct element
  expect(screen.getByLabelText('title')).toBe(sheet);

  // asserts the sheet has the correct description
  expect(sheet).toHaveDescription('description');
});
```

_Note 1: there is legacy Garden code that asserts against attributes directly that should be cleaned up_
_Note 2: may want to use modern `toHaveAccessibleDescription` with type casting_

packages/chrome/src/elements/sheet/components/Body.tsx
Francois-Esquire: The reason for the `tabIndex=0` is to allow keyboard users to use arrow keys to scroll through the text when it overflows. Is there a better way of doing it?

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const { getByRole, getByText } = render(<Example ref={ref} />);
+
+    expect(() => getByRole('complementary')).toThrow();
+
+    userEvent.click(getByText('Toggle Sheet'));
+
+    expect(getByRole('complementary')).toBe(ref.current);
+  });
+
+  it('does not animate a transition when isAnimated is false', () => {
```
hzhu, line 66: What do you think about rephrasing this description because animations can't be tested in jsdom. Maybe something like `does not apply the animation class` would be more accurate?

Also, should the inverse test also exist? `does apply the animation class`?

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: The reason the `Sheet` is collapsing to the center is due to the story layout being `'center'`. As a complementary element alongside the main content, resizing the width doesn't have a direction, unless I am missing something here.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByRole, getByText } = render(<Example ref={ref} />);
+
+    userEvent.click(getByText('Toggle Sheet'));
+
+    expect(getByRole('complementary')).toBe(ref.current);
+  });
+
+  it('is only mounted when isOpen is true', () => {
```
hzhu, line 55: Not a blocker, but most tests (testing from user perspective) should have declarative descriptions. Maybe some variant of `opens and closes sheet`.

Note: there are times for imperative descriptions — when Garden tests implementation details. Library authors occasionally have to test implementation details. For those tests, imperative descriptions make more sense.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByRole, getByText } = render(<Example ref={ref} />);
+
+    expect(() => getByRole('complementary')).toThrow();
+
+    userEvent.click(getByText('Toggle Sheet'));
+
+    expect(getByRole('complementary')).toBe(ref.current);
```
hzhu, line 63: The [`toBeInTheDocument`](https://github.com/testing-library/jest-dom#tobeinthedocument) matcher seems more appropriate than checking if an error occurred and if the sheet is the `ref`.

packages/chrome/src/elements/sheet/components/Close.spec.tsx
```diff
@@ -0,0 +1,21 @@
...
+import React from 'react';
+import { render } from 'garden-test-utils';
+import { SheetCloseButton as Close } from './Close';
+
+describe('Sheet.Close', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { getByTestId } = render(<Close data-test-id="close-btn" ref={ref} />);
+
+    const btn = getByTestId('close-btn');
```
hzhu, line 17: Alternatively, can use `getByRole('button')` as that query is [prioritized](https://testing-library.com/docs/queries/about/#priority) over data test IDs.

packages/chrome/src/elements/sheet/components/Close.tsx
Francois-Esquire: I do recall that the [Tag close button](https://github.com/zendeskgarden/react-components/blob/main/packages/tags/src/styled/StyledClose.ts#L16) did use an `aria-close` attribute which inclined me to add it. The other reason was to have a default value for `aria-close`. Wouldn't spreading the props after `aria-close` has been defined allow it to be overridden?

Good call on `aria-expanded` I will update this.

packages/chrome/src/elements/sheet/components/Title.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react';
+import { render } from 'garden-test-utils';
+import { SheetTitle as Title } from './Title';
+
+describe('Sheet.Title', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLDivElement>();
+    const { container } = render(<Title ref={ref} />);
+
+    expect(container.firstChild).toBe(ref.current);
```
hzhu, line 17: Not a blocker, but can be written to align more with `testing-library` guiding principles:

```tsx
render(<Title ref={ref}>Zendesk</Title>);

expect(screen.getByText('Zendesk')).toBe(ref.current);
```

packages/chrome/src/styled/sheet/StyledSheet.ts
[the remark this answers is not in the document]
Francois-Esquire: Good call on this, will remove.

packages/chrome/src/styled/sheet/StyledSheet.ts
Francois-Esquire: This makes sense to use `transform` instead of `width`, will look into this!

packages/chrome/src/elements/sheet/Sheet.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Sheet, ISheetProps } from './Sheet';
+
+describe('Sheet', () => {
+  const Example = forwardRef<HTMLElement, ISheetProps>((props: ISheetProps, ref) => {
+    const [isOpen, setIsOpen] = useState(props.isOpen);
+    const buttonRef = useRef<HTMLButtonElement>(null);
+
+    return (
+      <>
+        <button ref={buttonRef} onClick={() => setIsOpen(!isOpen)}>
```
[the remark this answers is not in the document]
Francois-Esquire: Yes, good catch! Will remove.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: This makes sense, will update.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Noted! I think I follow, could you give an example of what you mean by imperative tests?

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: That makes sense, I will update. I agree having a test for the inverse to make sure we are animating.

packages/chrome/src/elements/sheet/Sheet.spec.tsx
Francois-Esquire: Thanks this is insightful! I've updated the test and need to review to see what I can do about other tests to best align with this.

packages/chrome/src/elements/sheet/components/Close.tsx
Francois-Esquire: Removed `aria-expanded`!

packages/chrome/src/elements/sheet/Sheet.spec.tsx
hzhu: Yeah for sure! I meant "imperative description" and not "imperative tests". Here's an [example](https://github.com/zendeskgarden/react-containers/blob/main/packages/grid/src/GridContainer.spec.tsx#L411) of testing implementation details. Imperative descriptions are written from the perspective of the developer (not end user) and typically contain programming phrases such as `isOpen`, `true`, `onChange`, "invokes", "calls", etc.

packages/chrome/src/elements/sheet/components/Body.tsx
hzhu: I believe the idiomatic way to achieve that interaction is to set a negative tabindex (`-1`) and programmatically focus on the element when the sheet is opened. Here's an [example](https://yj5ik.csb.app/) of that interaction from a Material UI implementation.

jzempel (changes requested):
I think my biggest concern here is around animation and how the DOM structure around that is being managed. Let's connect and go over some alternatives. Might be that the RFC needs to be revisited 🤷 

The other bits are more about retaining connection and consistency with the remainder of the Garden codebase and should be relatively straightforward to knock-out.

Thanks for the good work here 🙌

packages/chrome/src/styled/sheet/StyledSheet.ts
hzhu: @Francois-Esquire @exelarios 

On second thought, I think transform doesn't move other elements adjacent to it. I think `width` is probably correct here. I'm going to resolve this comment. I think the focus is to address [this comment](https://github.com/zendeskgarden/react-components/pull/1231#discussion_r744257990) instead.

packages/chrome/src/styled/sheet/StyledSheetCloseButtonContainer.ts
mtomcal: No hard requirements here. Flex has more flexibility (no pun intended) for dynamic layouts and RTL capabilities with `flex-direction: row-reverse` and could potentially scale nicely with whatever `theme` variables might be used.

packages/chrome/package.json
```diff
@@ -24,8 +24,13 @@
     "@zendeskgarden/container-accordion": "^1.0.3",
     "@zendeskgarden/container-keyboardfocus": "^0.4.7",
```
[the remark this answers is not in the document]
exelarios: The `accordion` is being used by `CollapsibleSubNavItem`, so we only removed `keyboardfocus` since it's not being used anywhere.

These containers were not additional to `Sheet` and were already here.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,138 @@
...
+  /** Restores focus to the previous element after the Sheet is closed **/
+  restoreFocus?: boolean;
+  /** Determines the position of the Sheet **/
+  placement?: 'start' | 'end';
+}
+
+/**
+ * @extends HTMLAttributes<HTMLElement>
+ */
+// eslint-disable-next-line react/display-name
```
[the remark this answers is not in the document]
jzempel: How does setting https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md#ignoretranspilername-default-false to `true` affect linting?

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,79 @@
...
+  &.side-sheet-transition-enter {
+    width: 0;
+
+    & > * {
+      opacity: 0;
+    }
+  }
+
+  &.side-sheet-transition-enter-active {
+    width: 380px;
```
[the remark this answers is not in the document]
Francois-Esquire: We've updated the dimension to be DRY. I see what you mean, however I am concerned about the width being conveniently modified by consumers if both `width` and `max-width` are at play. The way I see it is if `<Sheet style={{ width: '500px' }} />` wouldn't need anything additional to set the dimension to the desired value.

packages/chrome/src/elements/sheet/components/Close.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { StyledSheetCloseButtonContainer } from '../../../styled';
+
+export const SheetCloseButton = forwardRef<HTMLButtonElement, HTMLAttributes<HTMLButtonElement>>(
+  (props, ref) => {
+    return (
+      <StyledSheetCloseButtonContainer>
+        <IconButton aria-label="Close Sheet" ref={ref} {...props}>
+          <XStrokeIcon />
+        </IconButton>
```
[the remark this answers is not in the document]
Francois-Esquire: I'm curious, what would be the reason not to use `IconButton`? We noticed [ColorSwatchDialogue](https://github.com/zendeskgarden/react-components/blob/main/packages/colorpickers/src/elements/ColorSwatchDialog/index.tsx#L20) re-uses the `react-buttons` package and we'd like to take advantage of the re-use, unless there are reasons not to.

There is the trade-off of bundling the `IconButton` in multiple places for consumer applications, however I think we can remedy that using `react-buttons` as a peer dependency?

packages/chrome/src/styled/sheet/StyledSheetHeader.ts
```diff
@@ -0,0 +1,39 @@
...
+  align-content: space-between;
+  align-items: center;
+  border-bottom: 1px solid ${props => props.theme.palette.grey[300]};
+  padding: ${props => props.theme.space.base * 5}px;
+  min-height: ${props => props.theme.space.base * 10}px;
+  text-align: ${props => (props.theme.rtl ? 'right' : 'left')};
+
+  & > * {
+    width: 100%;
+  }
```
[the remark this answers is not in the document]
Francois-Esquire: Agreed, it has been removed.

packages/chrome/package.json
```diff
@@ -24,8 +24,13 @@
...
     "@zendeskgarden/container-utilities": "^0.6.0",
+    "@zendeskgarden/react-buttons": "^8.44.1",
+    "@zendeskgarden/react-typography": "^8.44.1",
```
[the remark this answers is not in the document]
Francois-Esquire: Yes, the reason we added these dependencies was to include them in the `Sheet` - we did remove `@zendeskgarden/react-typography` in favor of `StyledSheetTitle` and `StyledSheetDescription`.

packages/chrome/src/elements/sheet/components/Close.tsx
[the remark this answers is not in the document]
Francois-Esquire: We've removed the button in favor of `StyledSheetClose` button.

packages/chrome/package.json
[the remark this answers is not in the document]
Francois-Esquire: Both dependencies have been removed.

packages/chrome/src/elements/sheet/components/Body.tsx
```diff
@@ -0,0 +1,17 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef, HTMLAttributes } from 'react';
+import { StyledSheetBody } from '../../../styled';
+
+export const SheetBody = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
+    return <StyledSheetBody tabIndex={0} ref={ref} {...props} />;
```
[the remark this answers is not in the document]
Francois-Esquire: Per our conversation offline, we've removed the tab index.

packages/chrome/src/elements/sheet/components/Body.tsx
Francois-Esquire: Agree with what you are saying. We've removed the `tabIndex` for the time being!

packages/chrome/src/styled/sheet/StyledSheet.ts
Francois-Esquire: We've updated this and removed that selector.

packages/chrome/src/styled/sheet/StyledSheetCloseButtonContainer.ts
Francois-Esquire: Agreed! I think for the button, it's only going to have a single child (SVG icon) and can get away with the default display for that reason I believe. It's also positioned as `absolute` against the Sheet layout.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,79 @@
...
+  flex-direction: column;
+  order: 1;
+
+  transition: width 0.25s ease-in-out;
+
+  width: 380px;
+  min-height: 100%;
+
+  & > * {
+    transition: opacity 0.5s ease-in-out;
```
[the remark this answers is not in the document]
Francois-Esquire: We've spoken offline about this and I think we're in the ballpark with `StyledSheetWrapper` addition.

packages/chrome/src/styled/sheet/StyledSheet.ts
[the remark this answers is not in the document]
Francois-Esquire: I think the new addition of `size` should cater to consumer needs of sizing the `Sheet` to the desired dimension.

packages/chrome/src/elements/sheet/Sheet.tsx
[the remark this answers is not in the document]
exelarios: I added `"react/display-name:": ["off", {"ignoreTranspilerName": true}]` within the rules for `eslintrc.json`, it still errors out. 🤷‍♂️

packages/chrome/src/utils/useFocusableMount.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  useLayoutEffect(() => {
+    if (isMounted && focusOnMount && targetRef.current) {
+      triggerRef.current = activeElement();
+      wasOpenRef.current = true;
+      targetRef.current.focus();
+    }
+  }, [wasOpenRef, isMounted, focusOnMount, targetRef]);
+
+  useLayoutEffect(() => {
```
hzhu, line 35: Is there a particular reason why `useLayoutEffect` is used instead of `useEffect`?

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,152 @@
...
+    const sheetRef = useRef<HTMLElement>(null);
+
+    const seed = useUIDSeed();
+    const [idPrefix] = useState<string>(id || seed(`sheet_${PACKAGE_VERSION}`));
+    const titleId = `${idPrefix}--title`;
+    const descriptionId = `${idPrefix}--description`;
+
+    const sheetContext = useMemo(() => ({ idPrefix }), [idPrefix]);
+
+    useFocusableMount({ targetRef: sheetRef, isMounted: isOpen, focusOnMount, restoreFocus });
```
hzhu, line 77: What was the conclusion regarding sighted keyboard only users and scrolling body text?

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,152 @@
...
+  /** Determines whether animation for opening and closing the Sheet is used **/
+  isAnimated?: boolean;
+  /** Focuses on the Sheet when isOpen is true and mounted **/
+  focusOnMount?: boolean;
+  /** Restores focus to the previous element after the Sheet is closed **/
+  restoreFocus?: boolean;
+  /** Determines the position of the Sheet **/
+  placement?: 'start' | 'end';
+  /** Sets the width in pixels, based on the placement of the Sheet */
+  size?: number;
```
jzempel, line 59: The `size` prop should be typed as a string to accept all valid CSS dimensions. See [Menu maxHeight](https://garden.zendesk.com/components/menu#menu) for comparison. Also, the prop is not optional (but receives a default below).

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,152 @@
...
+  id: PropTypes.string,
+  isOpen: PropTypes.bool,
+  isAnimated: PropTypes.bool,
+  focusOnMount: PropTypes.bool,
+  restoreFocus: PropTypes.bool,
+  placement: PropTypes.oneOf(['start', 'end']),
+  size: PropTypes.number
+};
+
+Sheet.defaultProps = {
```
jzempel, line 146: `size` should default to `380px` here and all other constants and conditional logic removed from styled components.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { css, ThemeProps, DefaultTheme } from 'styled-components';
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet';
+
+export const SHEET_SIZE = 380;
```
jzempel, line 13: Remove. We can expect the `Sheet` element will provide a valid size.

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,73 @@
...
+export const StyledSheet = styled.aside.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSheetProps>`
+  display: flex;
+  position: relative;
+  order: 1;
+  ${props =>
+    css`
+      transition: ${props.isAnimated ? 'width 0.5s ease-in-out, border 0.5s ease-in-out' : 'none'};
```
jzempel, line 57: The `border` should not be animated. Either it is visible on `isOpen` or transparent. Also, some UAs are not able to properly animate from `transparent`. Also, this speed feels very sluggish for Garden. Compare with DrawerModal that animates at 0.25s.

jzempel: This construct with extra `css` is not necessary. Use:

```
transition: ${props => props.isAnimated && 'width 0.25s ease-in-out'};
```

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,73 @@
...
+  ${props => sheetBorderStyle(props)};
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledSheet.defaultProps = {
+  theme: DEFAULT_THEME,
+  placement: 'end',
+  isOpen: false,
+  isAnimated: true,
+  size: SHEET_SIZE
```
jzempel, line 72: Not needed. These should all have specified defaults coming from the `Sheet` element.

packages/chrome/src/utils/useFocusableMount.ts
```diff
@@ -0,0 +1,41 @@
...
+  targetRef
+}: IUseFocusableMountProp) {
+  const wasOpenRef = useRef<boolean>(false);
+  const triggerRef = useRef<HTMLElement | Element | null>();
+
+  useLayoutEffect(() => {
+    if (isMounted && focusOnMount && targetRef.current) {
+      triggerRef.current = activeElement();
+      wasOpenRef.current = true;
+      targetRef.current.focus();
```
jzempel, line 31: The `StyledSheet` is not handling focus styling properly. It should have

```css
&:focus {
  outline: none;
}
```

See `StyledDrawerModal` for details.

packages/chrome/src/styled/sheet/StyledSheetWrapper.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled, { ThemeProps, DefaultTheme } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+import { SHEET_SIZE } from './StyledSheet';
```
jzempel, line 11: Remove.

packages/chrome/src/styled/sheet/StyledSheetWrapper.ts
```diff
@@ -0,0 +1,51 @@
...
+  &.side-sheet-transition-exit-active {
+    opacity: 0;
+  }
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledSheetWrapper.defaultProps = {
+  theme: DEFAULT_THEME,
+  size: SHEET_SIZE
```
jzempel, line 50: Not necessary. We can expect a valid `size` from the `Sheet` element.

packages/chrome/src/styled/sheet/StyledSheetWrapper.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const StyledSheetWrapper = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSheetWrapperProps & ThemeProps<DefaultTheme>>`
+  display: flex;
+  position: relative;
+  flex-direction: column;
+  transition: opacity 0.5s ease-in-out;
```
jzempel, line 26: I _really_ think this transition needs to be removed. It does not jive with the Garden system overall. In the case of modals, the UX is presenting (what appears to the user) _new_ content. As such, a fade transition makes sense there.

In the case of this and `Accordion` etc. the UX is presenting (again, what appears to be) _existing_ content. The UI should appear as if the information was there all along. Not as if it is somehow being summoned into the view.

packages/chrome/src/styled/sheet/StyledSheetFooter.ts
```diff
@@ -0,0 +1,35 @@
...
+export const StyledSheetFooter = styled.footer.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSheetFooterProps & ThemeProps<DefaultTheme>>`
+  display: flex;
+  flex-flow: row wrap;
+  align-items: center;
+  justify-content: ${props => (props.isCompact ? 'center' : 'flex-end')};
+  border-top: ${props => `${props.theme.borders.sm} ${getColor('neutralHue', 300, props.theme)}}`};
+
```
jzempel, line 27: [nit] why this extra whitespace?

packages/chrome/src/styled/sheet/StyledSheetDescription.ts
```diff
@@ -0,0 +1,31 @@
...
+import styled, { css, ThemeProps, DefaultTheme } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { math } from 'polished';
+
+const COMPONENT_ID = 'chrome.sheet_description';
+
+const fontStyles = (props: ThemeProps<DefaultTheme>) => {
+  return css`
+    line-height: ${math(`${props.theme.lineHeights.sm} - 1px`)};
+    font-size: ${math(`${props.theme.fontSizes.sm} - 1px`)};
```
jzempel, line 17: These values are off-system (and off design). Use the [getLineHeight](https://garden.zendesk.com/components/utilities#getlineheight) function for unitless line height. Also, the design-specified color is missing.

Please reference https://github.com/zendeskgarden/react-components/blob/main/packages/tags/src/styled/StyledTag.ts for an example of expected code format.

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,67 @@
...
+      color: ${activeColor};
+    }
+  `;
+};
+
+export const StyledSheetClose = styled.button.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  transition: background-color 0.25s ease-in-out, color 0.25s ease-in-out;
```
jzempel, line 52: Missing `box-shadow` transition. See https://github.com/zendeskgarden/react-components/blob/06565680b7fedc832637b711302c5b3f8e6eacbf/packages/modals/src/styled/StyledClose.ts#L55-L58

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,67 @@
...
+  border: 0;
+  border-radius: 50%;
+  padding: 0;
+  width: 40px;
+  height: 40px;
+  line-height: 0;
+
+  ${props => placementStyles(props)};
```
jzempel, line 60: [nit] not entirely clear this needed to diverge from https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledClose.ts

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,67 @@
...
+export const StyledSheetClose = styled.button.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  transition: background-color 0.25s ease-in-out, color 0.25s ease-in-out;
+  border: 0;
+  border-radius: 50%;
+  padding: 0;
+  width: 40px;
+  height: 40px;
```
jzempel, line 57: Need to be dimensioned on-theme. See https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledDrawerModalClose.ts

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,67 @@
...
+  line-height: 0;
```
jzempel, line 58: Why was this `line-height` added?

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    &:active,
+    &[aria-pressed='true'],
+    &[aria-pressed='mixed'] {
+      background-color: ${rgba(baseColor as string, 0.2)};
+      color: ${activeColor};
+    }
+  `;
+};
```
jzempel, line 46: This should basically be pulling from https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledClose.ts.

Please re-visit this styled component.

jzempel (changes requested):
In general, the only `defaultProps` needed on styled-components will be `theme: DEFAULT_THEME`. Please re-check the code for unnecessary specified default props.

packages/chrome/src/elements/sheet/components/Description.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import React, { forwardRef, HTMLAttributes } from 'react';
+
+import { StyledSheetDescription } from '../../../styled';
+import { useSheetContext } from '../../../utils/useSheetContext';
+
+export const SheetDescription = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
+  ({ id, ...props }, ref) => {
+    const { idPrefix } = useSheetContext();
+    const descriptionId = `${idPrefix}--description`;
```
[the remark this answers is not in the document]
exelarios: Do you mean have `descriptionId` and `titleId` within context?

The `uid` is generated through a hook, so I can't set the `uid` upon the context creation, so it needs to be passed through `Sheet` in order for it to be set as the context's value.

packages/chrome/src/elements/sheet/Sheet.tsx
jzempel: The consideration is out-of-scope for the initial round. If this behavior is critical (it very well may be), then Garden will take a step back and re-evaluate a11y for potential scroll containers (`DrawerModal`, `TabPanel`, `Accordion.Panel`, modal `Body`, `Stepper.Content`, etc.). Otherwise, the expectation is for the consumer to implement the [useScrollRegion](https://github.com/zendeskgarden/react-containers/tree/main/packages/scrollregion) hook.

packages/chrome/src/elements/sheet/components/Description.tsx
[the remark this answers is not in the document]
jzempel: Yes adding `descriptionId` and `titleId` to context. The variable needing to be redefined here in order to match the definition in Sheet.tsx is not DRY.

packages/chrome/src/utils/useFocusableMount.ts
Francois-Esquire: Initially `useLayoutEffect` was used to prevent this code block from running when server side rendering. [After re-reading the documention](https://reactjs.org/docs/hooks-reference.html#uselayouteffect), it looks like that is not the case, and is ran during SSR. Will update this to `useEffect` instead. Good call out!

packages/chrome/src/elements/sheet/components/Description.tsx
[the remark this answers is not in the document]
Francois-Esquire: Agreed, will update this.

packages/chrome/src/utils/useFocusableMount.ts
Francois-Esquire: The `Sheet` differs from the modality of the `DrawerModal`, do you think they should be treated the same in regards to focus? Would it be confusing for users without an indication (without screen readers)? This is a broad question to ask, however my instinct tells me to leave this to the user-agent, or be more explicit with it.

jzempel: Already covered in the RFC, the `Sheet` will offer `focusOnMount` to allow the consumer to control whether or not the `Sheet` will receive focus. In the case where a consumer determines that it should, we should not be seeing the browser's default outline.

packages/chrome/src/styled/sheet/StyledSheetDescription.ts
Francois-Esquire: Good catch on size being off design, will update!

packages/chrome/src/styled/sheet/StyledSheetWrapper.ts
Francois-Esquire: Agreed, will remove this.

packages/chrome/src/styled/sheet/StyledSheetClose.ts
Francois-Esquire: The `line-height` did center the icon used within the button when it was set to `0`. I think `display: flex` can accomplish this as well. will update.

packages/chrome/src/styled/sheet/StyledSheet.ts
Francois-Esquire: Will update the speed of the transition and will remove border transition.

packages/chrome/src/utils/useFocusableMount.ts
Francois-Esquire: After doing some experimentation with this, I noticed flickering when the `Sheet` animation was running within the Chrome suite. Do you think focusing on the element would be better ran after the component did update? Any reason comes to mind for not using `useLayoutEffect`?

packages/chrome/src/styled/sheet/StyledSheetCloseButtonContainer.ts
Francois-Esquire: Good call on this, after reverting to a custom button, `flex` layout was the right choice here!

packages/chrome/src/utils/useFocusableMount.ts
Francois-Esquire: Fair point, will update.

packages/chrome/src/utils/useFocusableMount.ts
Francois-Esquire: From our conversation, we decided to stick to `useEffect`. Updated!

packages/chrome/src/styled/sheet/StyledSheetTitle.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export const StyledSheetTitle = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  margin-bottom: ${props => props.theme.space.base / 4}px;
+
+  ${props => colorStyles(props)};
+  ${props => fontStyles(props)};
```
jzempel, line 41: Ditto. These functions feel a tad bit overdone and don't necessarily jive with [API rules](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules).

packages/chrome/src/utils/useSheetContext.ts
```diff
@@ -0,0 +1,22 @@
...
+import { createContext, useContext } from 'react';
+
+export interface ISheetContext {
+  titleId: string;
+  descriptionId: string;
+}
+
+export const SheetContext = createContext<ISheetContext>({
+  titleId: 'sheet--title',
+  descriptionId: 'sheet--description'
```
jzempel, line 17: Is there a reason these are being initialized here? These would almost never be valid ID values.

jzempel (changes requested):
Do we know why all the non-modals `yarn.lock` files (especially the root) are being updated in this PR? Something about that feels off. Or have there been changes to `main` that require this?

packages/chrome/src/styled/sheet/StyledSheetFooter.ts
```diff
@@ -0,0 +1,34 @@
...
+import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet_footer';
+
+export interface IStyledSheetFooterProps {
+  /** Sets the SheetFooter padding to half the standard and centers the elements  */
+  isCompact?: boolean;
+}
+
+export const StyledSheetFooter = styled.footer.attrs({
```
jzempel, line 18: We're seeing unexpected `isCompact` font-size styling (a 16px Anchor link in Storybook) within this component. However, it could be that the whole `Footer` or maybe the entire `Sheet` should be set to a font size of `theme.fontSizes.md` (?) and the inheritance will cascade as-needed to this child. There are advantages that need to be considered for where to set the font size.

jzempel (changes requested):

packages/chrome/src/styled/sheet/StyledSheetFooterItem.ts
```diff
@@ -0,0 +1,24 @@
...
+import styled, { ThemeProps, DefaultTheme } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet_footer_item';
+
+export const StyledSheetFooterItem = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  margin: auto ${props => props.theme.space.base * 2.5}px;
```
jzempel, line 17: I just compared to the spec'd designs, and this margining is off – leaving the end item incorrectly positioned with respect to the padding. Please compare with https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/styled/StyledFooterItem.ts which applies the correct margins.

jzempel (changes requested):

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,71 @@
...
+      box-shadow: ${props.theme.shadows.md(
+        getColor(backgroundColor, 600, props.theme, 0.35) as string
+      )};
+    }
+
+    &:active {
+      /* prettier-ignore */
+      transition:
+        background-color 0.1s ease-in-out,
+        color 0.1s ease-in-out;
```
mtomcal, line 36: `props.isAnimated` check?

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -0,0 +1,71 @@
...
+  position: absolute;
+  top: ${props => props.theme.space.base * 2.5}px;
+  ${props => (props.theme.rtl ? 'left' : 'right')}: ${props => `${props.theme.space.base * 2}px`};
+  align-items: center;
+  justify-content: center;
+  /* prettier-ignore */
+  transition:
+    box-shadow 0.1s ease-in-out,
+    background-color 0.25s ease-in-out,
+    color 0.25s ease-in-out;
```
mtomcal, line 57: https://github.com/zendeskgarden/react-components/pull/1231/files/1e703fe3f145fc29870a2d8e7423a1eae5e7a912..107764f6598bdea5f2748380b2577e054fc99de5#r751674928

packages/chrome/src/styled/sheet/StyledSheetClose.ts
jzempel: Fair question. Garden's `isAnimated` prop is used to control potentially expensive macro animations. These micro (button hover, etc) animations happen as a matter of course and intended by design.

packages/chrome/src/utils/useSheetContext.ts
Francois-Esquire: Good call on this, will remove.

packages/chrome/src/elements/sheet/components/Footer.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+describe('Sheet.Footer', () => {
+  it('passes ref to underlying DOM element', () => {
+    const ref = React.createRef<HTMLElement>();
+    const { container } = render(<Footer ref={ref} />);
+
+    expect(container.firstChild).toBe(ref.current);
+  });
+
+  it('is styled correctly when isCompact', () => {
```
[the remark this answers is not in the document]
Francois-Esquire: Good catch on this.

packages/chrome/src/styled/sheet/StyledSheetFooter.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+describe('StyledSheetFooter', () => {
+  it('renders correctly when isCompact is true', () => {
+    const { getByText } = render(<StyledSheetFooter isCompact>footer</StyledSheetFooter>);
+
+    const footer = getByText('footer');
+
+    expect(footer).toHaveStyleRule('justify-content', 'center');
+    expect(footer).toHaveStyleRule('padding', `${DEFAULT_THEME.space.base * 2.5}px`);
+  });
```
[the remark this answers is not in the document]
Francois-Esquire: That makes sense, will update.

packages/chrome/src/styled/sheet/StyledSheet.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+  it('renders correctly in rtl mode', () => {
+    const { getByRole } = renderRtl(<StyledSheet placement="end" />);
+
+    expect(getByRole('complementary')).toHaveStyleRule('border-right', '1px solid transparent');
+  });
+
+  it('renders correctly when placement is set', () => {
+    const { getByRole } = render(<StyledSheet placement="start" />);
+
+    expect(getByRole('complementary')).toHaveStyleRule('border-right', '1px solid transparent');
```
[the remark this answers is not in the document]
Francois-Esquire: Good call, this was a miss when we moved the logic to the styled component, will update.

packages/chrome/src/styled/sheet/StyledSheetFooter.ts
Francois-Esquire: Updated, added the font-size to the `StyledSheet`.

packages/chrome/src/styled/sheet/StyledSheetFooterItem.ts
Francois-Esquire: Good call on this, updated!

mtomcal:
Saw some warnings on the latest deploy:
Screen Shot 2021-11-18 at 11 25 24 AM

hzhu:
It looks like when placement is set to `start`, the sheet opens from the `end` position. Should the sheet open from the `start` position when placement is set to `start`?

jzempel:
@mtomcal 

> Saw some warnings on the latest deploy

re: `<svg role="presentation">` it was determined [a while ago](https://github.com/zendeskgarden/react-components/blob/080e5bd6c8f8f4f787453c0445d1c7b7637a0ef6/.svgo.yml#L5) that this attribute needed to be added to SVGs. It would take me a while to remember why.

re: scrollable region see https://github.com/zendeskgarden/react-components/pull/1231#discussion_r750297206

packages/chrome/src/styled/sheet/StyledSheetTitle.ts
```diff
@@ -0,0 +1,32 @@
...
+  DEFAULT_THEME
+} from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet_title';
+
+export const StyledSheetTitle = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  margin-bottom: ${props => props.theme.space.base / 4}px;
```
jzempel, line 22: This smacks of "pixel nudging" – typically an indication that the design has a bug. Garden never relies on margin nudging like this to match designs. The inherent line height is the only thing that should be managing this space for us.

jzempel (changes requested):
With the removal of this final bit along with addressing @hzhu's start/end issue, I think we're there. Woot.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -0,0 +1,125 @@
...
+  Header: typeof SheetHeader;
+  Body: typeof SheetBody;
+  Footer: typeof SheetFooter;
+  FooterItem: typeof SheetFooterItem;
+  Close: typeof SheetCloseButton;
+}
+
+export interface ISheetProps extends HTMLAttributes<HTMLElement> {
+  /** Sets the root ID. A unique ID is created if none is provided. **/
+  id?: string;
```
jzempel, line 46: Please note that `id`, which naturally exists on the extended element, is never documented throughout the Garden codebase. Reference container `useUIDSeed` usages for Tabs, Accordion, form Field, Modal, Tooltip. This can be removed.

jzempel (changes requested):

packages/chrome/src/styled/sheet/StyledSheetTitle.ts
Francois-Esquire: Understood, will update.

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: Makes sense, I will remove this.

Francois-Esquire:
> Saw some warnings on the latest deploy: Screen Shot 2021-11-18 at 11 25 24 AM



> It looks like when placement is set to `start`, the sheet opens from the `end` position. Should the sheet open from the `start` position when placement is set to `start`?
> 
> 

Since we don't have control over which side the sheet is going to be placed, we leave it up to the consumer to make that decision with the `placement` prop.

hzhu:
> Since we don't have control over which side the sheet is going to be placed, we leave it up to the consumer to make that decision with the placement prop.

Should the consumer be able to experience a working Storybook demo of `Sheet` interaction in LTR+ a `placement="start"` though? How would a consumer know how to accomplish that working version in product development?

Francois-Esquire:
> Should the consumer be able to experience a working Storybook demo of Sheet interaction in LTR+ a placement="start" though? How would a consumer know how to accomplish that working version in product development?

@hzhu you make a solid point here. After adding to the example and experiencing the animation, it was very inconsistent between the placements. Thanks for catching this! 🙌

packages/chrome/src/styled/sheet/StyledSheetTitle.ts
```diff
@@ -0,0 +1,31 @@
...
+} from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'chrome.sheet_title';
+
+export const StyledSheetTitle = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  line-height: ${props => getLineHeight(props.theme.space.base * 5, props.theme.fontSizes.md)};
+  color: ${props => getColor('neutralHue', 800, props.theme)};
```
jzempel, line 23: This should be `color: ${props.theme.colors.foreground};`

jzempel (changes requested):

packages/chrome/src/styled/sheet/StyledSheet.ts
```diff
@@ -0,0 +1,66 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSheetProps>`
+  display: flex;
+  order: 1;
+  transition: ${props => props.isAnimated && 'width 250ms ease-in-out'};
+  background-color: ${props => props.theme.colors.background};
+  width: ${props => (props.isOpen ? props.size : '0px')};
+  min-height: 100%;
+  max-height: 100%;
```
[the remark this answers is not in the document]
Francois-Esquire: Good call, I've updated.

packages/chrome/src/styled/sheet/StyledSheetTitle.ts
Francois-Esquire: Updated!

jzempel (approved):
🎉 thanks a mil @exelarios & @Francois-Esquire.

mtomcal (approved):
Super slick. Playing with Sheet in the Storybook is very nice and polished.

hzhu (approved):

jon-kerwin (approved):
Design reviewed.

PR #1233, opened 2021-11-02
fix(spinner): fix placeholder proportions

mtomcal:
## Description

We found a bug where Spinner Placeholder does not match the dimensions the loaded Spinner. By default Placeholder has a 1:0.9 aspect ratio using `em`s. However Spinner has a 1:1 aspect ratio. Also, the placeholder currently has `display: inline` which collapses the placeholder dimensions relative to the Spinner itself. The display mode `display: block` better fits this placeholder for a block based SVG spinner element. 

## Detail

**Note:** For the screenshots below, I removed the `display: flex` from the Storybook as it masks the bug and matches more closely the [CodeSandbox that reproduces the problem](https://codesandbox.io/s/elated-dust-e8ypl). 

### Placeholder

Screen Shot 2021-11-02 at 12 10 44 PM

### Spinner

Screen Shot 2021-11-02 at 12 10 59 PM

### Placeholder

Screen Shot 2021-11-02 at 12 14 45 PM

### Spinner

Screen Shot 2021-11-02 at 12 12 22 PM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/loaders/src/styled/StyledLoadingPlaceholder.ts
+++ b/packages/loaders/src/styled/StyledLoadingPlaceholder.ts
@@ -15,7 +15,7 @@ export const StyledLoadingPlaceholder = styled.div.attrs({
   'data-garden-version': PACKAGE_VERSION,
   role: 'progressbar'
 })<{ fontSize: string | number; width?: string; height?: string }>`
-  display: block;
+  display: inline-block;
   width: ${props => props.width || '1em'};
   height: ${props => props.height || '0.9em'};
   font-size: ${props => props.fontSize};
--- a/packages/loaders/src/styled/StyledLoadingPlaceholder.ts
+++ b/packages/loaders/src/styled/StyledLoadingPlaceholder.ts
@@ -14,10 +14,10 @@ export const StyledLoadingPlaceholder = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
   role: 'progressbar'
-})<{ fontSize: string | number }>`
-  display: inline;
-  width: 1em;
-  height: 0.9em;
+})<{ fontSize: string | number; width?: string; height?: string }>`
+  display: block;
+  width: ${props => props.width || '1em'};
+  height: ${props => props.height || '0.9em'};
   font-size: ${props => props.fontSize};
 
   ${props => retrieveComponentStyles(COMPONENT_ID, props)}
--- a/packages/loaders/src/styled/StyledSVG.ts
+++ b/packages/loaders/src/styled/StyledSVG.ts
@@ -14,6 +14,8 @@ interface IStyledSVGProps {
   fontSize?: string | number;
   width: number | string;
   height: number | string;
+  containerWidth?: string;
+  containerHeight?: string;
 }
 
 export const StyledSVG = styled.svg.attrs<IStyledSVGProps>(props => ({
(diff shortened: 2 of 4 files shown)


packages/loaders/src/styled/StyledLoadingPlaceholder.ts
```diff
@@ -14,10 +14,10 @@ export const StyledLoadingPlaceholder = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
   role: 'progressbar'
-})<{ fontSize: string | number }>`
-  display: inline;
-  width: 1em;
-  height: 0.9em;
+})<{ fontSize: string | number; width?: string; height?: string }>`
+  display: block;
```
jzempel, line 18: Are we sure about this? The default display for `<svg/>` (the element that replaces this) is `inline`. The placeholder should mimic that element as closely as possible.

We know for sure these loaders are not meant to take up the full block by default. I would think `inline-block` is the most we could do here.

jzempel: I also noticed `font-size` is bleeding through as a prop to the SVG. That shouldn't be happening, although not exactly the focus of this PR.

mtomcal: Screen Shot 2021-11-02 at 12 45 02 PM

DevTools says the SVG element is display block. Am I missing something?

FWIW Inline block also achieves the same results in placeholder we are looking for.

jzempel: Hmm, the `display` default for SVGs here – https://zendeskgarden.github.io/svg-icons/?path=/story/12-px--icons – is `inline` 🤔. Also, stripped away https://codepen.io/chriscoyier/pen/gpcDi is `inline`. I can't quite make sense of why the component storybook shows block. But we know that altering all `svg` to `display: block` in the `svg-icons` example would be rather undesirable. I think the same applies here.

If `inline-block` solves for you, then great. Just be sure to [mind the gap](https://dev.to/christiankaindl/the-strange-img-gap-in-html) (not exactly sure if this applies here).

packages/loaders/src/styled/StyledSVG.ts
```diff
@@ -14,6 +14,8 @@ interface IStyledSVGProps {
   fontSize?: string | number;
   width: number | string;
   height: number | string;
+  containerWidth?: string;
+  containerHeight?: string;
```
Francois-Esquire, line 18: Out of curiosity, is there any specific reason for using `containerWidth` instead of `width`?

mtomcal: There's a viewport surrounding the SVG dimensions that serves as a frame. 

[Width and height power Viewbox](https://github.com/zendeskgarden/react-components/pull/1233/files#diff-e79c04ddf2586ab7abb3c833519c98a1224b0e5b12edb2582a65ede9532f10ddR27)

In my opinion if we didn't mind breaking changes I would refactor `width` to `viewBoxWidth` and `height` to `viewBoxHeight`. Then `containerWidth|Height` would be just `width` and `height`.

jzempel: These are `styled` component adaptations. We use whatever utilitarian naming serves the purpose here (as long as the props don't leak into unwanted HTML attributes). These components are not exported nor subject to a breaking change discussion.

packages/loaders/src/styled/StyledLoadingPlaceholder.ts
jzempel, line 18: If I'm following, the best we can do here is `inline-block`. Current PR's `block` display would introduce unexpected layout issues.

jzempel (changes requested):

packages/loaders/src/styled/StyledLoadingPlaceholder.ts
mtomcal: Fixed in https://github.com/zendeskgarden/react-components/pull/1233/commits/895ee404126283a828e8af47c4c615101dd36f85

packages/loaders/src/styled/StyledSVG.ts
mtomcal: Im going to leave this alone as I tried prop name refactor quickly however it caused more problems than the scope of this effort.

jzempel (approved):

hzhu (approved):

PR #1242, opened 2021-11-10
feat(accordions): add default expanded sections prop

hzhu:
## Description

This pull request allows an uncontrolled component to be configured with an initial state.

## Checklist

### DIFF
--- a/packages/accordions/stories/examples/Uncontrolled.tsx
+++ b/packages/accordions/stories/examples/Uncontrolled.tsx
@@ -109,6 +109,12 @@ Uncontrolled.args = {
   isBare: false
 };
 
+Uncontrolled.argTypes = {
+  defaultExpandedSections: {
+    control: { disable: true }
+  }
+};
+
 Uncontrolled.parameters = {
   docs: {
     description: {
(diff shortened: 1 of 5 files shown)


packages/accordions/stories/examples/Uncontrolled.tsx
```diff
@@ -109,6 +109,12 @@ Uncontrolled.args = {
   isBare: false
 };
 
+Uncontrolled.argTypes = {
+  defaultExpandedSections: {
+    control: { disable: true }
```
jzempel, line 114: Can we open this control up to allow the demo to be affected by it?

hzhu: Unfortunately, the demo would not be affected by the knob because the knob value used as initial state for a `useState` in `Accordion`. The result is that the demo would render correctly for the initial `defaultExpandedSections`, but would not show updates for subsequently different `defaultExpandedSections`. I think adding a `defaultValue` knob to the `Input` would have the same issues.

I think if the demo forced a re-render of the `Accordion` each time the `defaultExpandedSections` changed, then the demo would reflect the knob. But this requires some one-off code to live in the story.

jzempel: got it.

jzempel (approved):

PR #1264, opened 2021-12-15
feat(forms): hide semantic Radio, Checkbox, and Toggle inputs more inclusively

jgorfine-zendesk:
## Description

This PR aims to hide the inputs that are part of the Checkbox, Radio, and Toggle components in a more inclusive way. This technique comes from Sara Soueidan via her blog post, “Inclusively Hiding & Styling Checkboxes and Radio Buttons:” [Hiding the checkboxes inclusively](https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/#hiding-the-checkboxes-inclusively).

## Detail

As Sara articulates in her blog post, there are assistive technologies — such as TalkBack, on Android phones — that enable low-vision and blind users to explore the web by touch. As far as checkboxes and radio buttons go, this means that the actual, semantic HTML inputs need to be the same size (i.e., occupy the same space) as the decorative inputs, so that someone exploring by touch can feel them and interact with them. 

### Before

The inputs were smaller than their decorative counterparts:

Before: less inclusive radio button

Before: less inclusive checkbox

Before: less inclusive toggle

### After (this PR)

The inputs are the same size as their decorative counterparts. They're positioned on top of the decorative inputs, and made completely transparent:

After: more inclusive radio button

After: more inclusive checkbox

After: more inclusive toggle

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/styled/radio/StyledRadioInput.ts
+++ b/packages/forms/src/styled/radio/StyledRadioInput.ts
@@ -139,6 +139,7 @@ export const StyledRadioInput = styled.input.attrs({
       box-shadow .1s ease-in-out,
       background-color .25s ease-in-out,
       color .25s ease-in-out;
+    z-index: -1;
     border: ${props => props.theme.borders.sm};
     border-radius: 50%;
     background-repeat: no-repeat;
@@ -148,6 +149,7 @@ export const StyledRadioInput = styled.input.attrs({
 
   & ~ ${StyledRadioLabel} > svg {
     position: absolute;
+    z-index: -1;
   }
 
   ${props => sizeStyles(props)};
--- a/packages/forms/src/styled/radio/StyledRadioInput.ts
+++ b/packages/forms/src/styled/radio/StyledRadioInput.ts
@@ -92,6 +92,10 @@ const sizeStyles = (props: IStyledRadioInputProps & ThemeProps<DefaultTheme>) =>
   const marginTop = `${props.theme.space.base * (props.isCompact ? 1 : 2)}px`;
 
   return css`
+    top: ${top};
+    width: ${size};
+    height: ${size};
+
     & ~ ${StyledRadioLabel}::before {
       top: ${top};
       background-size: ${props.theme.iconSizes.sm};
@@ -124,7 +128,7 @@ export const StyledRadioInput = styled.input.attrs({
 })<IStyledRadioInputProps>`
   /* hide <input> but retain accessiblity */
   position: absolute;
-  clip: rect(1px, 1px, 1px, 1px);
+  opacity: 0;
 
   & ~ ${StyledRadioLabel}::before {
     position: absolute;
--- a/packages/forms/src/styled/radio/StyledRadioInput.ts
+++ b/packages/forms/src/styled/radio/StyledRadioInput.ts
@@ -129,6 +129,7 @@ export const StyledRadioInput = styled.input.attrs({
   /* hide <input> but retain accessiblity */
   position: absolute;
   opacity: 0;
+  margin: 0;
 
   & ~ ${StyledRadioLabel}::before {
     position: absolute;
--- a/packages/forms/src/styled/radio/StyledRadioInput.ts
+++ b/packages/forms/src/styled/radio/StyledRadioInput.ts
@@ -140,7 +140,6 @@ export const StyledRadioInput = styled.input.attrs({
       box-shadow .1s ease-in-out,
       background-color .25s ease-in-out,
       color .25s ease-in-out;
-    z-index: -1;
     border: ${props => props.theme.borders.sm};
     border-radius: 50%;
     background-repeat: no-repeat;
@@ -150,7 +149,6 @@ export const StyledRadioInput = styled.input.attrs({
 
   & ~ ${StyledRadioLabel} > svg {
     position: absolute;
-    z-index: -1;
   }
 
   ${props => sizeStyles(props)};
--- a/packages/forms/.size-snapshot.json
+++ b/packages/forms/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 153102,
-    "minified": 105524,
-    "gzipped": 18387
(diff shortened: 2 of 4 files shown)


jgorfine-zendesk:
[Update Garden checkbox and radio button styles to be more inclusive](https://trello.com/c/9wBWAdb9/38-update-garden-checkbox-and-radio-button-styles-to-be-more-inclusive)

jzempel:
Thanks @jgorfine-zendesk. Please note that `Toggle` is part of this family as well. It would be great if you can branch internally vs. fork. Then we'll get your PR running through CI and SHA deployments which makes review more straightforward.

jgorfine-zendesk:
Hi @jzempel — I'll be sure to take a look at the Toggle and include it in this effort. 🙂 

> It would be great if you can branch internally vs. fork
Would you mind taking another look? I could I have sworn I did it this way already, but maybe I'm missing something.

jzempel:
> Would you mind taking another look? I could I have sworn I did it this way already, but maybe I'm missing something.

oic, yeah it looks good.

Francois-Esquire:
@jgorfine-zendesk I like this approach from Sara Soueidan and appreciate the PR. I am a little confused, were we expecting to see `Checkbox` included as well? I think this is ready to review tho, would love to see this PR/branch deployed - maybe we can take it out of draft?

jgorfine-zendesk:
> @jgorfine-zendesk I like this approach from Sara Soueidan and appreciate the PR. I am a little confused, were we expecting to see `Checkbox` included as well? I think this is ready to review tho, would love to see this PR/branch deployed - maybe we can take it out of draft?

Hi @Francois-Esquire — the work that's currently in the draft should cover both the Radio and Checkbook component; however, it does not include the Toggle, and @jzempel requested that it does. 

At this point, we could either:
1) Move forward with the Radio and Checkbox, and then I can come back to the Toggle sometime in the next couple of weeks
2) Keep waiting until the Toggle has also been addressed, and then merge in this PR. 

Would it be helpful if I left a follow up comment here articulating why I'm blocked (technically speaking) on the Toggle work?

jzempel:
Couple clarifying points here...

1. Given the nature of inheritance between these `Radio`, `Checkbox`, and `Toggle` really need to be addressed as a unit of work
2. Draft PRs will deploy, but it looks like the build sees @jgorfine-zendesk as [Unauthorized](https://app.circleci.com/pipelines/github/zendeskgarden/react-components?branch=jgorfine%2Fstyled-checkboxes-and-radios-inclusively). Jen, please authenticate with that link. At that point, CircleCI should be aware of your write permissions and allow future deploys to go through

packages/forms/src/styled/radio/StyledRadioInput.ts
```diff
@@ -135,6 +140,7 @@ export const StyledRadioInput = styled.input.attrs({
       box-shadow .1s ease-in-out,
       background-color .25s ease-in-out,
       color .25s ease-in-out;
+    z-index: -1;
```
jzempel, line 143: Without a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) reset, these visual artifacts are susceptible to being hidden by an intervening stacked layer (i.e. absolute `div`). I think there are a couple of things that could be done here:
- Add `z-index: 0;` to the `StyledRadioLabel`
-or-
- Update the `Radio`, `Checkbox`, `Toggle` element components to render the styled input _after_ the `{children}` and remove the `z-index: -1;` here

I'd generally be in favor of the latter over trying to manipulate stacking contexts – which always feels a bit messy.  But only if it can be done with minimal CSS updates. If we find ourselves updating a bunch of CSS to make it work, then I'd advise sticking with the former.

The problem can be seen with Bedrock enabled in Storybook: https://61e9d60fd1320011fdcba22c--zendeskgarden-react-components.netlify.app/?path=/story/packages-forms-fieldset--checkbox&globals=bedrock:enabled

packages/forms/src/styled/radio/StyledRadioInput.ts
```diff
@@ -144,6 +150,7 @@ export const StyledRadioInput = styled.input.attrs({
 
   & ~ ${StyledRadioLabel} > svg {
     position: absolute;
+    z-index: -1;
```
jzempel, line 153: Ditto :wink:

jzempel (changes requested):

packages/forms/src/styled/radio/StyledRadioInput.ts
jzempel: Actually, the more I think about it, I think there are a bunch of sibling selectors that break if the `{children}`/input are reordered. So I'd recommend sticking with the `z-index: 0;` added to `StyledRadioLabel`.

jgorfine-zendesk: @jzempel — thanks for showing me the Bedrock thing! (I didn't know about it before.)

> I'd generally be in favor of the latter over trying to manipulate stacking contexts – which always feels a bit messy.
I totally agree with you here. In this particular case, I think part of the challenge comes from the faux input (the `label::before`) being positioned absolutely, as well as the `input`. I can't help but wonder if styling whole component with Flexbox instead might help alleviate this. 🤔 

RE the `z-index`: I just tried it out again locally — both with and without Bedrock — and it looks like we might not need to hard-set the `z-index` after all? I just pushed up a commit that removes it altogether. Would you mind taking another look?

jzempel (approved):
@jgorfine-zendesk thanks for this contribution, representing an inclusive uplift to an often used class of Garden components. With one more @zendeskgarden/maintainers review, we'll get this merged and published ASAP.

Francois-Esquire (approved):
PR looks great! Thank you @jgorfine-zendesk 🙌

hzhu (approved):

jgorfine-zendesk:
My pleasure, @jzempel! Thank you for the opportunity. 😄

PR #1265, opened 2021-12-28
chore(demo): refactor storybook

jzempel:
## Description

This PR represents a pretty massive overhaul of the Storybook component demos. You might find that reading through the new [demo documentation](https://github.com/zendeskgarden/react-components/blob/c31bae4face4e57cfe40c3f03d2c80827a0c09f4/docs/demo.md) is the best way to get acquainted with the refactored code.

## Detail

Highlights include:

- a substantial uptick in Story controls with categorized differentiation between main element props, subcomponents, and "Story" controls
- moved stories to package `/demo` folders to better align with other Garden repos
- standardized MDX-to-TSX Story configuration
- added Figma designs (all currently available) to story tabs
- removal of outmoded documentation (the website is Garden's documentation source of truth)

(@smritimadan this recent [staging deployment](https://61bab387d9e13be77d0a27ca--zendeskgarden-react-components.netlify.app/) can be used, as needed, for ongoing API doc recovery efforts.)

## Checklist

      designer as a reviewer)~


.storybook/manager.js
```diff
@@ -10,6 +10,7 @@ import { addons } from '@storybook/addons';
 import { DEFAULT_THEME } from '../packages/theming/src';
 
 addons.setConfig({
+  panelPosition: 'right',
```
jzempel, line 13: This default layout makes better UX sense for component-controls testing.

.storybook/preview.js
```diff
@@ -17,7 +17,8 @@ export const parameters = {
     grid: { disable: true }
   },
   controls: {
-    hideNoControlsWarning: true
+    hideNoControlsWarning: true,
+    sort: 'alpha'
```
jzempel, line 21: No more `01-FileName.stories.mdx` 😉

.svgo.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
             focusable: false
           },
           {
-            role: 'presentation'
+            'aria-hidden': true
```
jzempel, line 11: This clears out SVG errors seen in Storybook's a11y tab and applies the correct accessibility semantics for visually decorative Garden icons. See https://css-tricks.com/accessible-svgs/#example-2-standalone-icon-decorative for details.

package.json
```diff
@@ -121,6 +121,7 @@
     "rollup-plugin-license": "2.6.0",
     "rollup-plugin-size-snapshot": "0.12.0",
     "rollup-plugin-typescript2": "0.31.1",
+    "storybook-addon-designs": "6.2.0",
```
jzempel, line 124: This package connects Figma to Storybook. https://help.figma.com/hc/en-us/articles/360045003494-Storybook-and-Figma

packages/accordions/src/index.ts
```diff
@@ -6,7 +6,10 @@
  */
 
 export { Stepper } from './elements/stepper/Stepper';
+export type { IStepperProps } from './elements/stepper/Stepper';
```
jzempel, line 9: There were a number of missing prop interface exports. This PR recovered many (probably not all) so they could be reused within `Story` modules.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -124,10 +124,7 @@ Sheet.propTypes = {
 };
 
 Sheet.defaultProps = {
-  isOpen: false,
   isAnimated: true,
-  focusOnMount: false,
-  restoreFocus: false,
```
jzempel, line 130: Note that Garden does not specify defaults for `undefined` booleans.

packages/dropdowns/src/elements/Multiselect/Multiselect.tsx
```diff
@@ -432,10 +444,3 @@ Multiselect.defaultProps = {
 };
 
 Multiselect.displayName = 'Multiselect';
-
-/**
- * @extends HTMLAttributes<HTMLDivElement>
- */
-export default withTheme(Multiselect) as React.FunctionComponent<
```
jzempel, line 439: All `withTheme` exports were failing to properly render Storybook prop sheets. This PR swaps out the preferred `styled-components` [ThemeContext](https://styled-components.com/docs/advanced#via-usecontext-react-hook).

utils/build/Intl.d.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+declare namespace Intl {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  interface DateTimeFormatOptions extends Intl.DateTimeFormatOptions {
+    dateStyle?: 'full' | 'long' | 'medium' | 'short';
```
jzempel: Resolves [missing types](https://github.com/microsoft/TypeScript/issues/35865).

docs/demo.md
```diff
@@ -0,0 +1,218 @@
...
+  is ordered first.
+- Subsequent component story files are named `componentName.stories.mdx`
+- Use Garden's standard prop naming for story args (i.e. isXxx/hasXxx for
+  boolean args), but use `argTypes` to rename with subcomponent notation. For
+  example:
+- All `Story` TSX files are placed under a `demo/stories` directory. Additionally:
+  - Move all boilerplate data constants (i.e. control data for a component's
+    `children`) into `demo/stories/data.ts`
+  - Capture all types and interfaces specific to story data under
+    `demo/stories/types.ts`
```
Francois-Esquire, line 190: I was not expecting the info on `data.ts` and `types.ts` to be this far down in the document (or under `## Naming`) 😅  IMO, this feels like it belongs under `## Conventions` before diving into the Story conventions.. maybe under `### File structure` (not the best name). This is excellent practice for our code hygiene and great to have!

docs/demo.md
```diff
@@ -0,0 +1,218 @@
...
+- If a package has patterns (see section below), tuck all pattern demo code
+  under a `demo/~patterns` directory. Prefix with a tilde to ensure these
+  stories are ordered last.
```
Francois-Esquire, line 193: I would have liked to see this sooner in the document as well, possibly under the same `### File structure` section as mentioned in the previous comment.

docs/demo.md
```diff
@@ -0,0 +1,218 @@
...
+
+A pattern is a demo story that highlights a component's ability to work together
+with other Garden components or in conjunction with external libraries.
+Oftentimes, a pattern ends up violating the best practice of duplicating example
+documentation that belongs on the website. However, due to component development
+sequencing or a need to persist a visual test, a pattern can be a helpful tool
+for (temporarily) demonstrating component layout or behavior that supercedes the
+isolation of the component itself.
+
+Follow [naming](#naming) conventions to keep patterns collected in a
```
Francois-Esquire, line 214: Deviating thought here: would `Patterns/{{pattern name}}` be a good counterpart to `Packages/{{package name}}/{{component name}}`? Having a nested `[patterns]` subdirectory in Storybook makes it hard to find, and feels slightly out of place IMO.

docs/demo.md
```diff
@@ -0,0 +1,218 @@
...
+
+Take time to ensure Story `args` and `argTypes` are defined well. Often you will
+find that the majority of demo development time is spent on `args` and
+`argTypes` and the `Story` itself essentially falls out of the goal to
+demonstrate component flexibility in isolation.
+
+## Naming
+
+- Follow the existing Storybook naming hierarcy:
+  `Packages/{{PackageName}}/{{ComponentName}}`
```
Francois-Esquire, line 178: What is the idea behind `Packages` instead of `Component`?

Francois-Esquire (commented):
Monumental effort @jzempel ! I really like the direction this is heading and what it will offer to the development experience.. excited for this!

Francois-Esquire (changes requested):
It looks like a few stories didn't make it out the same when checking locally..

Grid:
Screen Shot 2022-01-03 at 10 30 23 AM

Chrome:
Screen Shot 2022-01-03 at 10 33 54 AM

.storybook/preview.js
jzempel: Oops, you're right @Francois-Esquire. I misstated the purpose of this sort, which is to ensure Controls are sorted alphabetically.

.storybook/babel.config.js
```diff
@@ -7,7 +7,7 @@
 
 const path = require('path');
 const { readdirSync } = require('fs');
-const babel = require('@storybook/core-common/dist/cjs/utils/babel');
+const babel = require('@storybook/core-common');
```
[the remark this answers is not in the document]
jzempel: Yeah, I agree. I left this to reduce the change set, but I'll update as you suggest.

.storybook/preview.js
jzempel: I'd rather not use `storySort`, preferring rather to have the MDX files sort naturally in alphabetical order.

package.json
```diff
@@ -5,21 +5,21 @@
...
-    "lint:css": "stylelint '{packages,utils,stories}/**/*.{js,ts,tsx}' '!packages/**/*.spec.{js,ts,tsx}' '!packages/**/dist/**'",
-    "lint:js": "eslint packages/*/src/ utils/ .storybook/ --ext js,ts,tsx --max-warnings 0",
-    "lint:md": "markdownlint README.md docs/*.md stories/**/*.md packages/*/examples/*.md packages/*/src/**/*.md packages/*/README.md",
+    "lint:css": "stylelint '{packages,utils}/**/*.{js,ts,tsx}' '!packages/**/*.spec.{js,ts,tsx}' '!packages/**/dist/**'",
+    "lint:js": "eslint packages/*/src/ packages/*/demo/ utils/ .storybook/ --ext js,ts,tsx --max-warnings 0",
+    "lint:md": "markdownlint README.md packages/*/README.md packages/*/demo/**/*.mdx",
     "new": "utils/scripts/new.js",
     "prepare": "yarn build",
-    "start": "start-storybook -s ./utils/storybook -p 6006",
+    "start": "start-storybook --no-version-updates -p 6006",
```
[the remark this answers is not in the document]
jzempel: Yes (run `yarn start-storybook --help` to see options). It suppresses the update check to reduce the nag noise. Renovate does this for us.

.storybook/main.js
```diff
@@ -6,13 +6,18 @@
...
-const externalConfig = require('../.svgo.config.js');
+const svgoConfig = require('../.svgo.config.js');
 const ESLintPlugin = require('eslint-webpack-plugin');
 const docs = process.env.BROWSER ? process.env.BROWSER.toUpperCase() !== 'IE11' : true;
 
 module.exports = {
-  stories: ['../packages/*/stories/**/*.stories.@(js|jsx|ts|tsx|mdx)'],
-  addons: [{ name: '@storybook/addon-essentials', options: { docs } }, '@storybook/addon-a11y'],
+  stories: ['../packages/*/demo/**/*.stories.@(js|jsx|ts|tsx|mdx)'],
+  staticDirs: ['../utils/storybook'],
```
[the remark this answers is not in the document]
jzempel: I agree with you and actually thought about doing what you suggest, but backed off to reduce the changes. But given the complexity, this PR is probably a reasonable place to make this change.

docs/demo.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Do this
+
+- Expose controls for all essentials – inherent props, conditional subcomponent
+  renders, child content
+- Allow natural component layout – full width for block, partial width for inline
+- Use the Garden `Grid` to manage layout for components that respond to
+  placement values (start, end, top, bottom, etc)
+- Rely on `useArgs` from Storybook's client API in order to keep component state
+  connected with it's associated Storybook control
```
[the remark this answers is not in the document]
jzempel: This is `true` (ha, see what I did there). But this is one where I prefer to go with the grain of Storybook and see that boolean props are `undefined` as expected. Taking this approach lowers the devX bar (I don't have to chase down pre-defining all the component boolean `args`) and it helped us find inconsistent prop default definitions (https://github.com/zendeskgarden/react-components/pull/1265#discussion_r775973782).

docs/demo.md
jzempel: Per https://github.com/zendeskgarden/react-components/pull/1265#discussion_r778096091, the naming reflects the basic structure of `react-components`. Also, when we find ourselves setting `{{PackageName}}` as the second level, basic taxonomy science should tell us that the root level is "Packages" (plural).

jzempel: If Garden ever decides to deviate the basic package structure, then this naming will need to change. For now, it clearly tells us which components are part of which packages.

docs/demo.md
jzempel: Good call – update on the way...

jzempel:
> It looks like a few stories didn't make it out the same

The stories are definitely not the same, nor intended to be. The updates fall in line with the new `demo.md` documentation, pulling out `styled-components` that leave false impressions of default component renderings, initially rendering with prop defaults (while providing 100% control), and ensuring that the website will be the true source of documented component usage truth.

packages/colorpickers/demo/colorPicker.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+# API
+
+<ArgsTable />
+
+# Demo
+
+## Uncontrolled
+
+<Canvas>
+  <Story name="Uncontrolled" argTypes={{ color: { control: false } }}>
```
hzhu, line 27: I noticed that setting `defaultColor` for the uncontrolled demo causes an error because Storybook passes `{}`. Should this knob be disabled (or maybe have a default set)?

Screen Shot 2022-01-04 at 2 21 10 PM

jzempel: Yeah. I ended up leaving this in place, because if you enter a valid value such as `"#888"`, switch to another story, and then come back to the uncontrolled color picker – the `defaultValue` sticks as expected.

jzempel: This could present an opportunity to make the component a bit more robust 🤷 rather than taking away a control that manifests a component error.

packages/dropdowns/demo/menu.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+    }}
+  >
+    {args => {
+      const updateArgs = useArgs()[1];
+      const handleSelect = selectedItems => updateArgs({ selectedItems });
+      const handleStateChange = changes =>
+        changes.hasOwnProperty('isOpen') &&
+        changes.type !== Downshift.stateChangeTypes.blurInput /*account for Storybook control*/ &&
+        updateArgs({ isOpen: changes.isOpen });
+      return <TriggerStory {...args} onSelect={handleSelect} onStateChange={handleStateChange} />;
```
hzhu, line 95: It looks like there is an issue with the chevron when the nested menu is expanded.

Screen Shot 2022-01-04 at 2 45 54 PM

jzempel: Yes. This is just one of a number of issues where a finessed storybook was hiding component cases that can happen IRL. In this case a `NextItem` that is set as the `Dropdown` `selectedItem` styles with a checkmark. The component styling should be preventing this from happening. I am filing a list of follow-on bugs to be addressed after this PR.

packages/forms/demo/fileUpload.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+    }}
+    parameters={{
+      design: {
+        allowFullscreen: true,
+        type: 'figma',
+        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=103%3A20268'
+      }
+    }}
+  >
+    {args => <FileUploadStory {...args} />}
```
hzhu, line 51: Anyone else having issues triggering the file upload UI? I'm clicking the dropzone area but nothing seems to happen.

hzhu (approved):
This is a huge improvement. I love the new structure! Thanks @jzempel!

packages/forms/demo/fileUpload.stories.mdx
jzempel: The `FileUpload` demo is meant to only show that component in isolation with no amendments or alterations. To see the component in action (combined with `react-dropzone`) go to the `Packages/Forms/[patterns]` section.

See https://github.com/zendeskgarden/react-components/blob/c31bae4face4e57cfe40c3f03d2c80827a0c09f4/docs/demo.md

> The development goal for demos is to render components in isolation with the least amount of structural scaffolding together with the greatest amount of control for content and flexibility.

Also take special note of the "Patterns" section at the bottom of that documentation.

docs/demo.md
Francois-Esquire: That makes sense, thanks for explaining - yeah let's keep this in mind and see if we can come up with something that could work better (if any).

docs/demo.md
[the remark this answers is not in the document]
Francois-Esquire: Hmm, I disagree with this to some degree, since a boolean for example is a primitive data type that would default to false in most cases, and can mitigate inconsistencies with types. Definitely not a blocker.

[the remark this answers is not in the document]
jzempel: I agree that Garden could've opted for `false` defaults, and I could even agree that might be ideal. But the decision to go with `undefined` booleans throughout the component API long pre-dates the making of this PR.

exelarios:
I noticed with `Typography/paragraph` it displays `MD` tags, not entirely sure where that's coming from?
image

jzempel:
> I noticed with Typography/paragraph it displays MD tags, not entirely sure where that's coming from?

Similar to the [website](https://garden.zendesk.com/components/paragraph#size) this is indicating that `MD` typography is meant to be used with `medium` sized `Paragraph` and so on. From the perspective of Garden design, there is no other valid combination (although, technically, a `large`-spaced Paragraph could use `SM` typography).

mtomcal:
There seems to be styling divergence between old Grid and new Grid docs:

Old
Screen Shot 2022-01-06 at 2 24 54 PM
New
Screen Shot 2022-01-06 at 2 24 44 PM

Was this intended?

jzempel:
> Was this intended?

@mtomcal yes, this Grid demo is rendering as intended. Demos should render the default state initially and then offer controls (in this case, `debug`) to change, well, basically just about everything.

See previous https://github.com/zendeskgarden/react-components/pull/1265#issuecomment-1004917162 re: similar observation.

mtomcal:
> > Was this intended?
> 
> @mtomcal yes, this Grid demo is rendering as intended. Demos should render the default state initially and then offer controls (in this case, `debug`) to change, well, basically just about everything.
> 
> See previous [#1265 (comment)](https://github.com/zendeskgarden/react-components/pull/1265#issuecomment-1004917162) re: similar observation.

When I see this, I ask myself, "what am I looking at?". I am not sure when I see this. Any chance this could be documented in the story itself about the presence of the "debug" toggle?

jzempel:
> Any chance this could be documented in the story itself about the presence of the "debug" toggle?

My preference is to keep the component demos essentially doc-free. Given that the Grid demo in particular has caused repeated concern, how would you feel about making an exception for the "default state" rule and allowing `debug=true` in this one, isolated case?

mtomcal:
It might be easiest to show the "Docs" tab first then the "Canvas" tab which documents the "debug" toggle front and center. This gives me context and information about this more abstract demo.

Screen Shot 2022-01-06 at 3 11 08 PM

Another storybook from the Azure team shows the "Docs" first then the "Preview" (similar to "Demo") tab. 
https://azure.github.io/communication-ui-library/?path=/docs/ui-components-gridlayout--grid-layout

This might also help document the case mentioned here as well: https://github.com/zendeskgarden/react-components/pull/1265#issuecomment-1006030348

mtomcal:
Aside from documenting how to use the demo, this is an amazing update and epic reorganization of the docs. Im looking forward to using this for Splitter component. 👏

jzempel:
> It might be easiest to show the "Docs" tab first

Garden determined early on that the Canvas was most valuable to the greater audience (design + dev). I'm not keen to reverse that decision here.

mtomcal:
@jzempel What would you suggest on messaging "how to use this abstract demo" to the developer?

jzempel:
> What would you suggest on messaging...

@mtomcal sorry, I'm not tracking with you. We're not super interested in documenting "how to use this" as it is primarily a tool for Garden visual testing and the how-to boils down to familiarity with Storybook 6.x. The Garden team is expected to be familiar with Storybook 6.x.

mtomcal:
@jzempel Let's take this offline then and follow up there. Thanks!

packages/dropdowns/demo/stories/DropdownFieldStory.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const DropdownFieldStory: Story<IArgs> = ({
+  dropdownProps,
+  label = 'Label',
+  isLabelRegular,
+  isLabelHidden,
+  hasHint = true,
+  hint = 'Hint',
+  hasMessage = true,
+  message = 'Message',
+  validation,
```
Francois-Esquire, line 47: I couldn't tell where this was defined [in the story](https://github.com/zendeskgarden/react-components/blob/82561ad10fb06cedacae3d26b3426f764d8d741c/packages/dropdowns/demo/combobox.stories.mdx). Can you point me to where it can be found?

packages/dropdowns/src/elements/Menu/Menu.tsx
```diff
@@ -108,13 +109,15 @@ const Menu: React.FunctionComponent<IMenuProps & ThemeProps<DefaultTheme>> = pro
...
+  const themeContext = useContext(ThemeContext);
+
   // Reset Downshift refs on every render
   itemIndexRef.current = 0;
   nextItemsHashRef.current = {};
   previousIndexRef.current = undefined;
   itemSearchRegistry.current = [];
 
-  const popperPlacement = isRtl(props)
+  const popperPlacement = themeContext.rtl
```
Francois-Esquire, line 120: Do you remember if there are any more instances of `isRtl` we should modernize? We could track the rest for a future date if there are any left.

Francois-Esquire (approved):
LGTM 🙌  thanks for this huge effort @jzempel

packages/dropdowns/src/elements/Menu/Menu.tsx
jzempel: I _think_ we got em' all 😅. My "element" component typedef work will tell us for certain. And we'll follow-on deprecate `isRtl` for removal in v9.

packages/dropdowns/demo/stories/DropdownFieldStory.tsx
jzempel: In dropdowns (and fields) `validation` is a concept that spans both the input and associated `Message` component. In the case you referenced, it applies styling to both the `Combobox` and associated `Field` > `Message` component. The same control can be seen with dropdowns `Autocomplete`, `Multiselect`, `Select` and similar "Forms" components.

We don't have solid design evidence that validation should apply (as it currently does) disparate from input & message. But the current implementation allows for cases we don't know about.

jzempel: p.s. The MDX story does not (should not) need to expose props as args which are inherent in the component identified in `<Meta>`. Which, in this case, is `Combobox` and includes `validation`, `focusInset`, `isBare`, `isCompact`, etc.

PR #1272, opened 2022-01-20
fix(chrome): add padding to sheet header when sheet close button is p…

Francois-Esquire:
## Description

Adds padding to the `Sheet.Header` when the `Sheet.Close` is rendered. This prevents the `Sheet.Close` button from overlapping & obscuring the text in the header (`Sheet.Title` & `Sheet.Description`).

## Detail

This PR adds to `Sheet` context and lets the containing component tree know when the `Sheet.Close` button is used. The `Sheet.Header` adds additional padding when the `Sheet.Close` button is used.

Example with overflowing text obstructed by `Sheet.Close`:

Screen Shot 2022-01-20 at 4 42 00 PM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/chrome/demo/stories/SheetStory.tsx
+++ b/packages/chrome/demo/stories/SheetStory.tsx
@@ -23,14 +23,18 @@ interface ISheetComponentProps extends ISheetProps {
   footerItems: IFooterItem[];
   hasHeader: boolean;
   isCompact: boolean;
+  hasTitle?: boolean;
   title: string;
+  hasDescription?: boolean;
   description: string;
 }
 
 export const SheetComponent = ({
   hasHeader,
   title,
+  hasTitle = !!title,
   description,
+  hasDescription = !!description,
   hasBody,
   body,
   hasFooter,
@@ -43,8 +47,8 @@ export const SheetComponent = ({
   <Sheet {...props}>
     {hasHeader && (
       <Sheet.Header>
-        <Sheet.Title>{title}</Sheet.Title>
-        <Sheet.Description>{description}</Sheet.Description>
+        {hasTitle && <Sheet.Title>{title}</Sheet.Title>}
+        {hasDescription && <Sheet.Description>{description}</Sheet.Description>}
       </Sheet.Header>
     )}
     {hasBody ? <Sheet.Body>{body}</Sheet.Body> : body}
--- a/packages/chrome/src/elements/sheet/Sheet.tsx
+++ b/packages/chrome/src/elements/sheet/Sheet.tsx
@@ -19,7 +19,7 @@ import { useUIDSeed } from 'react-uid';
 import mergeRefs from 'react-merge-refs';
 
 import { StyledSheet, StyledSheetWrapper } from '../../styled';
-import { SheetContext } from '../../utils/useSheetContext';
+import { ISheetContext, SheetContext } from '../../utils/useSheetContext';
 import { useFocusableMount } from '../../utils/useFocusableMount';
 
 import { SheetTitle } from './components/Title';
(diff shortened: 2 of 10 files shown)


packages/chrome/demo/stories/SheetStory.tsx
```diff
@@ -23,14 +23,18 @@ interface ISheetComponentProps extends ISheetProps {
   footerItems: IFooterItem[];
   hasHeader: boolean;
   isCompact: boolean;
+  hasTitle?: boolean;
   title: string;
+  hasDescription?: boolean;
```
jzempel, line 28: I think you can guarantee that Storybook is going to give you a `boolean` and not `undefined`.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -67,16 +67,27 @@ export const Sheet = React.forwardRef<HTMLElement, ISheetProps>(
...
+        }
+      }),
+      [titleId, descriptionId, isCloseButtonPresent]
+    );
 
     useFocusableMount({ targetRef: sheetRef, isMounted: isOpen, focusOnMount, restoreFocus });
 
     return (
-      <SheetContext.Provider value={sheetContext}>
+      <SheetContext.Provider value={sheetContext as ISheetContext}>
```
jzempel, line 90: is there a means for typing `useMemo` to indicate the return type and avoid `as`?

packages/chrome/src/styled/sheet/StyledSheetHeader.ts
```diff
@@ -10,13 +10,22 @@ import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden
...
-})<ThemeProps<DefaultTheme>>`
+})<IStyledSheetHeaderProps & ThemeProps<DefaultTheme>>`
   border-bottom: ${props =>
     `${props.theme.borders.sm} ${getColor('neutralHue', 300, props.theme)}}`};
   padding: ${props => props.theme.space.base * 5}px;
+  ${props =>
+    props.isCloseButtonPresent &&
+    // the padding size accounts for 40px (10 bse units) size of the button,
+    // 8px additional padding and 8px padding for the button position from a given side.
+    `padding-${props.theme.rtl ? 'left' : 'right'}: ${props.theme.space.base * 14}px;`}
```
jzempel, line 28: Feels like this could be derived from an exported `width` on `StyledSheetClose` plus the intended padding.

jzempel (commented):
Thanks for pushing the Story to its limits and adding `StyledSheetHeader` testing. A couple of follow-on considerations:
- how does the context callback construct compare to or improve upon an existing construct like [useFieldContext](https://github.com/zendeskgarden/react-components/blob/main/packages/forms/src/utils/useFieldContext.ts)
- what consideration was given to pushing this solution deeper into `DrawerModal` and `Modal`?

james-sann (commented):
Hey Mike,

This looks great from a design perspective and adheres to the 8px gap between header content and the close button 👏

I can't remember where we landed on if I remove the header and just use the Sheet `body`? If I do this currently with the Storybook example in this PR, the design breaks. If there's nothing we can do technically to resolve this, all good. Just wanted to get clarification 🙌

jzempel:
> I can't remember where we landed on if I remove the header and just use the Sheet `body`?

If I remember right, my recollection is that we determined the implementation would be in unspecified design territory at this point, and subsequently responsible for applying fallback styling.

packages/chrome/demo/stories/SheetStory.tsx
Francois-Esquire: The reason it was optional is that this story is re-used by the `Chrome` story and those two controls are not present there.

packages/chrome/src/elements/sheet/Sheet.tsx
Francois-Esquire: Ahh good call, let me take a go at that.

packages/chrome/demo/stories/SheetStory.tsx
jzempel: gotcha. Thanks for the clarification.

packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -67,16 +67,27 @@ export const Sheet = React.forwardRef<HTMLElement, ISheetProps>(
     const sheetRef = useRef<HTMLElement>(null);
 
     const seed = useUIDSeed();
+    const [isCloseButtonPresent, setCloseButtonPresent] = useState<boolean>(false);
```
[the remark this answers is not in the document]
Francois-Esquire: I think because the `Sheet.Close` button is optional was why the decision was made for padding to be conditional, otherwise it makes sense to do so.Another challenge that makes this nonstandard is that the `Sheet.Close` doesn't nest within the `Sheet.Header` container, which makes it tricky.

jzempel (approved):
Approved, although I have some open questions in https://github.com/zendeskgarden/react-components/pull/1272#pullrequestreview-860001743

hzhu (approved):

Francois-Esquire:
> Thanks for pushing the Story to its limits and adding `StyledSheetHeader` testing. A couple of follow-on considerations:
> 
> * how does the context callback construct compare to or improve upon an existing construct like [useFieldContext](https://github.com/zendeskgarden/react-components/blob/main/packages/forms/src/utils/useFieldContext.ts)
> * what consideration was given to pushing this solution deeper into `DrawerModal` and `Modal`?

@jzempel
I think the `useFieldContext` compares to the current implementation found here, nothing immediate strikes me as needing improvement.
For `DrawerModal` & `Modal` we should track and evaluate if we can ship the changes for v8 or later. Making a note to add to track this.

PR #1297, opened 2022-02-18
chore(tooltips): refactor types and element interface

jzempel:
## Description

An improved approach for type/interface management. The arrangement in this PR will fix the `type` obfuscation seen in this doc-generated API: https://garden.zendesk.com/components/tooltip#tooltip

## Detail

Following this PR and internal RFC review, we'll add a new `docs/typescript.md` documenting the standard approach and start to pick off the non-standard definitions one package at a time.

## Checklist

      designer as a reviewer)~


packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -14,46 +14,9 @@ import { useTooltip } from '@zendeskgarden/container-tooltip';
...
-  /** Displays the tooltip */
-  isVisible?: boolean;
-  /** @ignore ReactNode override */
-  children: React.ReactElement;
-  /** Defines the ref key used to position the tooltip */
-  refKey?: string;
-}
+import { getPopperPlacement, getRtlPopperPlacement } from '../utils/gardenPlacements';
+import { StyledTooltipWrapper, StyledTooltip } from '../styled';
+import { ITooltipProps, PLACEMENT, SIZE, TYPE } from '../utils/types';
```
jzempel, line 19: In this PR's approach, the public interface pushes _out_ rather than pulling the interface _up_ from styled view elements. I believe this is the better angle, as the public interface is focal for typed prop data and documentation. And any styled interfaces should pull from the public interface as-needed.

jzempel: For example, with the public interface pushed _out_, a styled component's interface could look something like the following without creating a circular code reference:

```
import { ITooltipProps } from '...';

interface IStyledTooltipProps {
  reusedProp: ITooltipProps['prop'];
}
```

packages/tooltips/src/elements/Tooltip.tsx
```diff
@@ -194,24 +157,10 @@ Tooltip.propTypes = {
...
-    'start',
-    'start-top',
-    'start-bottom'
-  ]),
+  placement: PropTypes.oneOf(PLACEMENT),
   popperModifiers: PropTypes.any,
-  size: PropTypes.oneOf(['small', 'medium', 'large', 'extra-large']),
-  type: PropTypes.oneOf(['light', 'dark']),
+  size: PropTypes.oneOf(SIZE),
+  type: PropTypes.oneOf(TYPE),
```
jzempel, line 163: Even though TypeScript helps enforce the values, all `propTypes` throughout the Garden codebase should be reusing ARRAY constants rather than maintaining separate lists.

packages/tooltips/src/utils/gardenPlacements.spec.ts
```diff
@@ -33,17 +28,17 @@ describe('Garden Placement Utilities', () => {
         'start-bottom': 'left-end'
       };
 
-      Object.keys(GARDEN_PLACEMENTS).forEach(gardenPlacement => {
-        expect(getPopperPlacement(gardenPlacement as GARDEN_PLACEMENT)).toBe(
-          GARDEN_PLACEMENTS[gardenPlacement as GARDEN_PLACEMENT]
-        );
+      const gardenPlacements = Object.keys(GARDEN_PLACEMENTS) as GardenPlacement[];
```
jzempel, line 31: Prevent multiple `as` type assertions whenever possible. See https://stackoverflow.com/questions/52856496/typescript-object-keys-return-string

packages/tooltips/src/utils/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  'left-end'
+] as const;
+
+export const SIZE = ['small', 'medium', 'large', 'extra-large'] as const;
+
+export const TYPE = ['light', 'dark'] as const;
+
+export type GardenPlacement = typeof PLACEMENT[number];
+
+export type PopperPlacement = typeof POPPER_PLACEMENT[number];
```
jzempel, line 47: See https://melvingeorge.me/blog/convert-array-into-string-literal-union-type-typescript for array-to-type conversion explanation.

packages/tooltips/src/elements/Tooltip.spec.tsx
```diff
@@ -9,8 +9,8 @@ import React from 'react';
 import userEvent from '@testing-library/user-event';
 import { render, act, renderRtl } from 'garden-test-utils';
 import { DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
-
-import { Tooltip, ITooltipProps } from './Tooltip';
+import { Tooltip } from './Tooltip';
+import { ITooltipProps } from '../utils/types';
```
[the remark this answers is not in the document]
jzempel: The argument against `/types` is that there's a pretty strong pattern of only having `/elements`, `/styles`, `/utils` at the src root. (After that, our evolutionary conventions get a bit chaotic 😅). Is `/types` a strong enough concept to pull out to the src root? And do we end up creating an unnecessary decision between `/types/{file}.ts`, `types.ts`, and `/types/index.ts`? I guess this is why I opted for `/utils/types` – because I didn't really want to think through the ramifications for every package.

packages/tooltips/src/elements/Tooltip.tsx
jzempel: The component author will need to take care with this ☝️ type of reuse, as `undefined` can come through from the imported interface in a manner that makes the local styled interface feel less expressive. In this case, it could appear that `reusedProp` is mandatory. But if `ITooltipProps['prop']` is optional, then `reusedProp` may also be `undefined`.

tldr; I'm not entirely convinced this is an ideal reuse pattern.

jzempel: fwiw, there are [TypeScript utilities](https://www.typescriptlang.org/docs/handbook/utility-types.html) to help with this. It is just a matter of the author being meticulous. For example:

```ts
import { ITooltipProps } from '...';

interface IStyledTooltipProps {
  reusedProp: NonNullable<ITooltipProps['prop']>;
}
```

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -13,23 +13,17 @@ import {
...
 
 const COMPONENT_ID = 'tooltip.tooltip';
 
-export type TOOLTIP_SIZE = 'small' | 'medium' | 'large' | 'extra-large';
-export type TOOLTIP_TYPE = 'light' | 'dark';
-
-export interface IStyledTooltipProps {
-  hasArrow?: boolean;
-  size?: TOOLTIP_SIZE;
+export interface IStyledTooltipProps extends Pick<ITooltipProps, 'hasArrow' | 'size' | 'zIndex'> {
```
jzempel, line 23: After some additional noodling, I think this is a fairly decent pattern for public interface reuse within view components.

packages/tooltips/src/elements/Tooltip.spec.tsx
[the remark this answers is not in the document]
Francois-Esquire: I have been thinking about this since we first discussed this. I'm actually in favor of `types/` since it felt like `types` shouldn't categorically fit under `utils/`. Another reason why: this would make  it easier for `utils/`, `elements/` and `styles/` to `import '../types;'` instead of adding `import '../utils/types` (for `elements` and `styles`).

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -13,23 +13,17 @@ import {
...
   /** All valid [Popper.JS Placements](https://popper.js.org/popper-documentation.html#Popper.placements) */
-  placement: POPPER_PLACEMENT;
-  zIndex?: number | string;
-  type: TOOLTIP_TYPE;
+  placement: typeof POPPER_PLACEMENT[number];
+  type: NonNullable<ITooltipProps['type']>;
 }
```
Francois-Esquire, line 27: Should we move `styled/` types under the same umbrella? Or is it strictly for the public API?
If so, maybe a `types/styled.ts` would be a good way of organizing it without necessarily exporting it.

packages/tooltips/src/utils/types.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface ITooltipProps extends HTMLAttributes<HTMLDivElement> {
```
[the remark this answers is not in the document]
Francois-Esquire: I have been thinking about this as well: we can post process the results and iterate through the props per component, however for documentations sake, there are some external APIs throughout `react-components` (`react-popper` and `downshift`) that we can annotate the API with as fragments - like what we do for `dropdowns`.

packages/tooltips/src/elements/Tooltip.spec.tsx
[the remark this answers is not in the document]
jzempel: Ok, I like it. Thanks for the feedback.

packages/tooltips/src/utils/types.ts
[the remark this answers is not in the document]
jzempel: Ok, I'm happy to see how this works out in practice. In the meantime, I think the description with corresponding reference link is a solid compromise here.

packages/tooltips/src/styled/StyledTooltip.ts
jzempel: I'm opposed to re-situating these private interfaces away from their associated styled components to a central location. I feel like it communicates the wrong thing re: visibility – but that's just my personal preference and I'm happy to defer to the team's direction on this. To your point, we might discover some nice type affinity that could be gained if these are co-located 🤷.

Either way, my current tooling exercise is not going to engage with the `styled` interfaces. That's a substantial effort; my plan is to stay focused on the publicly exported element component interfaces.

hzhu (approved):

PR #1299, opened 2022-02-22
feat(dropdowns): Add `appendToElement` property

jeremy8883:
## Description

Adds a `appendToElement` property to the dropdown `Menu` component. Typically this is used to render the menu under the root element, rather than rendering it inline, eg if the dropdown menu is a descendant of an `overflow: hidden` element and is therefore getting cropped.

I used the same property name as the [Tooltip component](https://github.com/zendeskgarden/react-components/blob/c18d6cccee2158cef74249f2cb6c8f713bbba51c/packages/tooltips/src/elements/Tooltip.tsx#L27). Alternatively I could have exposed the [Popper positionFixed property](https://popper.js.org/docs/v1/#defaultspositionfixed), but I went with the consistent and slightly more flexible option. 

image

For a11y testing, I haven't used axe before. I don't think it picked up anything related to the PR, but it was a bit tricky to tell. I also tested the voiceover with Chrome Screen Reader (aka Vox), and the interaction seemed to be consistent. I played with the keyboard experience (tabbing, use of arrow keys, escape key, enter), and that also seemed to be consistent.

A story has been added to `~patterns`, to demonstrate the cropping problem and solution.

https://user-images.githubusercontent.com/4495057/155450379-549c2a98-62d5-44ba-a05b-527111de9ab6.mov

This is my first contribution to Garden, so apologies if I have missed something.

## Checklist

      designer as a reviewer)~ NA
- [X] :globe_with_meridians: demo is up-to-date (`yarn start`)
- [ ] ~:arrow_left: renders as expected with reversed (RTL) direction~ NA
- [ ] ~:metal: renders as expected with Bedrock CSS (`?bedrock`)~ NA (I think)
- [X] :wheelchair: analyzed via [axe](https://www.deque.com/axe/) and evaluated using VoiceOver
- [X] :guardsman: includes new unit tests
- [X] :memo: tested in Chrome, Firefox, Safari, Edge, and IE11 - Only introduces a new property and uses react portals, so should be fine. I gave a test in Chrome, Firefox and Safari though.


hzhu:
Hi @jeremy8883, thanks for the PR! Would you be able create a CodeSandbox demonstrating the menu being cut off so we could further review the root issue? Feel free to fork Garden's [CodeSandbox's example](https://codesandbox.io/s/lvilq?module=src/Example.tsx) of Menu. I tried applying `overflow: hidden;` to the parent element of `dropdowns.menu_wrapper` but couldn't reproduce the issue.

jeremy8883:
Ah sorry, the complete instructions involve setting the overflow, height, and position properties.

[Code sandbox is here](https://codesandbox.io/s/gallant-blackburn-7th6ji?file=/src/Example.tsx)

LMK if you would like me to add a means to test this in the story. I was thinking I could add a toggle like "Clip inside parent", which would then show some text explaining that the dev needs to then set the `appendToElement` to fix the clipping issue.

packages/dropdowns/demo/menu.stories.mdx
```diff
@@ -56,7 +56,8 @@ import { MENU_ITEMS as ITEMS } from './stories/data';
       items: ITEMS,
       selectedItems: [ITEMS[5]],
       hasMedia: false,
-      isOpen: false
+      isOpen: false,
+      appendToNode: 'undefined'
```
jzempel, line 60: This and the following addition to `argTypes` tends to muddy the default API definition and isn't strictly in line with the repo's [demo documentation](https://github.com/zendeskgarden/react-components/blob/main/docs/demo.md) guidelines. I might suggest removing these from the `Menu` story and instead adding a one-off [[patterns]](https://github.com/zendeskgarden/react-components/blob/main/docs/demo.md#patterns) story that fully illustrates what you're going for with this change. See other `[patterns]` for reference:
- https://zendeskgarden.github.io/react-components/?path=/story/packages-avatars-patterns--chrome
- https://zendeskgarden.github.io/react-components/?path=/story/packages-forms-patterns--file-upload
- https://zendeskgarden.github.io/react-components/?path=/story/packages-tags-patterns--faux-input

jzempel (changes requested):

packages/dropdowns/demo/menu.stories.mdx
jeremy8883: No worries, that `~patterns` story is created now.

I have also removed it from the main `Menu` story. With that said, I myself would find that arg handy if I was doing a11y testing, and I wanted to test all of the different permutations of the menu. It's an available prop for the `Menu` component, so it makes sense to document it in the story, at least from my experience. But up to you.

jzempel (approved):
thanks, @jeremy8883. We'll either need to wait to get this fork fully tested before merging, or you can get Zendesk member access to the repo and do a branch PR. :pray:

packages/dropdowns/src/elements/Menu/Menu.spec.tsx
```diff
@@ -81,4 +81,23 @@ describe('Menu', () => {
...
+            Item 1
+          </Item>
+        </Menu>
+      </Dropdown>
+    );
+
+    userEvent.click(getByTestId('select'));
+
+    expect(getByTestId('menu').parentElement!.parentElement!.tagName).toBe('BODY');
+  });
```
hzhu, line 102: Would you be ble to refactor this test to align with the testing-library's [guiding principles](https://testing-library.com/docs/guiding-principles/) and use [priority querying](https://testing-library.com/docs/queries/about/#priority)?.

hzhu (commented):
Thanks @jeremy8883 for the PR! API looks good. Just a few more things.

It looks like the [CI build](https://app.circleci.com/pipelines/github/zendeskgarden/react-components/5237/workflows/e4baff65-27ff-4d70-a132-d3394ea15879/jobs/9961) is breaking for this PR due to outdated snapshots. The snapshot files can be rebuilt by running `yarn build` in the package directory.

jeremy8883:
Thanks again for the review! Those updates have now been made

hzhu (approved):

mtomcal:
We can move forward with this once we have our CircleCI jobs pass. I reached out to you offline for details. Thanks!

mtomcal (approved):
Thanks @jeremy8883 the contribution! It will go out next week in our release cycle.

PR #1304, opened 2022-03-01
fix(modals): update container-modal & types

hzhu:
## Description

Updates `container-modal` version and fixes the types for `onClose` handler by using types for synthetic events from React package. 

Previously, native event types were used in `react-containers`. Since event handlers are passed to React, native event types did not work and required some instances of casting types to `any`.

## Detail

This is a follow-on fix after addressing the root type [issues](https://github.com/zendeskgarden/react-containers/pull/425) in react-containers.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/modals/demo/stories/DrawerModalStory.tsx
+++ b/packages/modals/demo/stories/DrawerModalStory.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { EventHandler, MouseEventHandler } from 'react';
+import React, { EventHandler, MouseEventHandler, KeyboardEvent, MouseEvent } from 'react';
 import { Story } from '@storybook/react';
 import { useTheme } from 'styled-components';
 import Icon from '@zendeskgarden/svg-icons/src/16/arrow-left-stroke.svg';
@@ -15,7 +15,7 @@ import { IFooterItem } from './types';
 
 interface IArgs extends IDrawerModalProps {
   onClick: MouseEventHandler<HTMLElement>;
-  onClose: EventHandler<any>;
+  onClose: EventHandler<KeyboardEvent | MouseEvent>;
   hasBody: boolean;
   body: string;
   hasClose: boolean;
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -44,5 +44,5 @@
     "utils/test/jest.d.ts"
   ],
   "include": ["packages/*/src/**/*", "packages/*/demo/**/*"],
-  "exclude": ["node_modules", "**/node_modules", "**/dist", "packages/.template/**/*"]
+  "exclude": ["node_modules", "**/node_modules", "**/dist"]
```
Francois-Esquire, line 47: Odd, this doesn't seem to be reflected in the "files changed".. for my understanding, what's the reason for removing `"packages/.template/**/*"`? Was it causing issues?

Francois-Esquire (approved):

packages/modals/demo/stories/DrawerModalStory.tsx
```diff
@@ -15,7 +15,7 @@ import { IFooterItem } from './types';
 
 interface IArgs extends IDrawerModalProps {
   onClick: MouseEventHandler<HTMLElement>;
-  onClose: EventHandler<any>;
+  onClose: EventHandler<KeyboardEvent | MouseEvent>;
```
jzempel, line 18: If the prop is properly declared on `IDrawerModalProps` then can this `onClose` be removed?

packages/modals/demo/stories/ModalStory.tsx
```diff
@@ -23,7 +23,7 @@ import { IFooterItem } from './types';
 interface IArgs extends IModalProps {
   isVisible: boolean;
   onClick: MouseEventHandler<HTMLElement>;
-  onClose: EventHandler<any>;
+  onClose: EventHandler<KeyboardEvent | MouseEvent>;
```
jzempel, line 26: Same here?

jzempel (changes requested):

packages/modals/demo/stories/DrawerModalStory.tsx
hzhu: Yep! Type is covered in `IDrawerModalProps`. I'll remove it here.

jzempel (approved):

tsconfig.json
hzhu: I added and removed this change in the current PR. Turns out it doesn't address the TypeScript warnings in CI. I'll leave addressing the warnings to a separate task since the warnings have been around before this PR.

PR #1306, opened 2022-03-11
fix(modals): use auto width for responsive layout

hzhu:
## Description

The modal is not responsive and is cut off on small mobile devices.

## Detail

This pull request fixes the issue by removing the non-static positioning and applying a width in a mobile-first, responsive design approach. The modal has a default `width: auto` and as users scale up, the explicit width for tablets + desktop sizes are applied. Width sizes on larger displays remain the same as they were before.

## Screens
**BEFORE**: notice that a modal in iPhone(s) is cut off:
Screen Shot 2022-03-10 at 12 02 06 PM

**AFTER**: notice that a modal in iPhone(s) is fully visible:
Screen Shot 2022-03-10 at 12 02 24 PM

## Note

There might be additional style enhancements on mobile in a follow up PR.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/styled/StyledModal.ts
+++ b/packages/modals/src/styled/StyledModal.ts
@@ -47,9 +47,8 @@ const sizeStyles = (props: IStyledModalProps & ThemeProps<DefaultTheme>) => {
   const largeWidth = 800;
 
   return css`
-    width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
-    @media (max-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.md}) {
-      ${props.theme.rtl ? 'right' : 'left'}: ${props.theme.space.base * 6}px;
+    @media (min-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.sm}) {
+      width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
     }
   `;
 };
--- a/packages/modals/src/styled/StyledModal.ts
+++ b/packages/modals/src/styled/StyledModal.ts
@@ -7,7 +7,12 @@
 
 import PropTypes from 'prop-types';
 import styled, { css, keyframes, ThemeProps, DefaultTheme } from 'styled-components';
-import { getColor, retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import {
+  getColor,
+  mediaQuery,
+  retrieveComponentStyles,
+  DEFAULT_THEME
+} from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'modals.modal';
 
@@ -47,7 +52,7 @@ const sizeStyles = (props: IStyledModalProps & ThemeProps<DefaultTheme>) => {
   const largeWidth = 800;
 
   return css`
-    @media (min-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.sm}) {
+    ${mediaQuery('up', props.isLarge ? 'md' : 'sm', props.theme)} {
       width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
     }
   `;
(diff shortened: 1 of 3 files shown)


packages/modals/src/styled/StyledModal.ts
```diff
@@ -47,9 +47,8 @@ const sizeStyles = (props: IStyledModalProps & ThemeProps<DefaultTheme>) => {
   const largeWidth = 800;
 
   return css`
-    width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
-    @media (max-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.md}) {
-      ${props.theme.rtl ? 'right' : 'left'}: ${props.theme.space.base * 6}px;
+    @media (min-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.sm}) {
+      width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
```
jzempel, line 51: Was an attempt made to use the [`mediaQuery`](https://garden.zendesk.com/components/utilities#mediaquery) utility here? In general, that would be preferable.

hzhu: No, but I'm looking into it now. It seems like `mediaQuery` would not support a custom breakpoint (e.g. `largeWidth - 1px` or `799px`).

I don't recall why `799px` was used for the `isLarge` breakpoint. If one of the theme breakpoints can be used instead, then media could look like: `mediaQuery('up', props.isLarge ? 'md' : 'sm', props.theme)`.

mtomcal:
This will solve some pain points for our consumers 👍 Tested this out and looks great.

packages/modals/src/styled/StyledModal.ts
mtomcal: Might require a separate PR to add some flexibility to the `mediaQuery` function.

jzempel (changes requested):

packages/modals/src/styled/StyledModal.ts
hzhu: @james-sann mind if I loop you into this for the design decision?

The current media query for the `isLarge` modal is custom and targets `@media (min-width: 799px)`. `799px` is a custom computed value. In order to use the `mediaQuery` utility suggested in this thread, the media query size must be one from Garden’s [theme object](https://garden.zendesk.com/components/theme-object#default_theme) and not custom. 

@james-sann, could you help us figure out if Garden can use `theme.breakpoints.md` instead of a custom `799px`? From the engineering perspective, I think using `theme.breakpoints.md` (aka `768px`) would be just fine. But I wanted to run it by design first since this would change the media query target of what is currently in production.

mtomcal (approved):

Francois-Esquire (approved):

james-sann (approved):
I agree we should utilize theme breakpoints. Nice work Henry!

packages/modals/src/styled/StyledModal.ts
```diff
@@ -47,9 +52,8 @@ const sizeStyles = (props: IStyledModalProps & ThemeProps<DefaultTheme>) => {
   const largeWidth = 800;
 
   return css`
-    width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
-    @media (max-width: ${props.isLarge ? `${largeWidth - 1}px` : props.theme.breakpoints.md}) {
-      ${props.theme.rtl ? 'right' : 'left'}: ${props.theme.space.base * 6}px;
+    ${mediaQuery('up', props.isLarge ? 'md' : 'sm', props.theme)} {
+      width: ${props.isLarge ? `${largeWidth}px` : `${defaultWidth}px`};
```
jzempel, line 56: In order to make this perfectly smooth responsive, this line should be

```suggestion
      width: ${props.isLarge ? props.theme.breakpoints.md : props.theme.breakpoints.sm};
```

(the `defaultWidth` and `largeWidth` variables can be removed)

jzempel (changes requested):

jzempel (approved):

PR #1307, opened 2022-03-14
fix(modals): add padding to header when close button is present

Francois-Esquire:
## Description

The `Modal` (and `DrawerModal`) `Header` overflows past the `Close` button, which causes the title text to overlap with the button.

## Detail

This PR fixes a visual bug when the `Header` text is long enough to wrap. `padding` has been added to the `Header` element for the `Modal` and `DrawerModal` components, with respect to RTL.

Before:
Screen Shot 2022-03-14 at 5 31 18 PM

After:
Screen Shot 2022-03-14 at 5 33 36 PM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/modals/src/elements/Close.spec.tsx
+++ b/packages/modals/src/elements/Close.spec.tsx
@@ -10,6 +10,7 @@ import { render } from 'garden-test-utils';
 
 import { Modal } from './Modal';
 import { Close } from './Close';
+import { ModalsContext } from '../utils/useModalContext';
 
 describe('Close', () => {
   it('passes ref to underlying DOM element', () => {
@@ -24,5 +25,35 @@ describe('Close', () => {
     );
 
     expect(getByTestId('close')).toBe(ref.current);
+  });
+
+  describe('functionality', () => {
+    it('sets isCloseButtonPresent when mounting and unmounting', () => {
+      const setCloseButtonPresent = jest.fn();
+      const fn = jest.fn();
+      const context = {
+        getTitleProps: fn,
+        getContentProps: fn,
+        getCloseProps: fn,
+        setCloseButtonPresent
+      };
+
+      const TestComponent = () => {
+        return (
+          <ModalsContext.Provider value={context}>
+            <Close />
+          </ModalsContext.Provider>
+        );
+      };
+
+      const { unmount } = render(<TestComponent />);
+
+      expect(setCloseButtonPresent).toHaveBeenCalledWith(true);
+
+      unmount();
+
+      expect(setCloseButtonPresent).toHaveBeenCalledWith(false);
+      expect(setCloseButtonPresent).toHaveBeenCalledTimes(2);
+    });
   });
 });
--- a/packages/modals/src/elements/Close.spec.tsx
+++ b/packages/modals/src/elements/Close.spec.tsx
@@ -10,7 +10,6 @@ import { render } from 'garden-test-utils';
 
 import { Modal } from './Modal';
 import { Close } from './Close';
-import { ModalsContext } from '../utils/useModalContext';
 
 describe('Close', () => {
   it('passes ref to underlying DOM element', () => {
@@ -25,35 +24,5 @@ describe('Close', () => {
     );
 
     expect(getByTestId('close')).toBe(ref.current);
-  });
-
-  describe('functionality', () => {
-    it('sets isCloseButtonPresent when mounting and unmounting', () => {
-      const setCloseButtonPresent = jest.fn();
-      const fn = jest.fn();
-      const context = {
-        getTitleProps: fn,
-        getContentProps: fn,
-        getCloseProps: fn,
-        setCloseButtonPresent
-      };
-
-      const TestComponent = () => {
-        return (
-          <ModalsContext.Provider value={context}>
-            <Close />
-          </ModalsContext.Provider>
-        );
-      };
-
-      const { unmount } = render(<TestComponent />);
-
-      expect(setCloseButtonPresent).toHaveBeenCalledWith(true);
-
-      unmount();
-
-      expect(setCloseButtonPresent).toHaveBeenCalledWith(false);
-      expect(setCloseButtonPresent).toHaveBeenCalledTimes(2);
-    });
   });
 });
--- a/packages/modals/src/styled/StyledClose.ts
+++ b/packages/modals/src/styled/StyledClose.ts
@@ -40,7 +40,7 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
   `;
 };
 
-export const baseMultipliers = {
+export const BASE_MULTIPLIERS = {
   top: 2.5,
   side: 6.5,
   size: 10
@@ -55,9 +55,9 @@ export const StyledClose = styled.button.attrs({
 })`
   display: block;
   position: absolute;
-  top: ${props => props.theme.space.base * baseMultipliers.top}px;
+  top: ${props => props.theme.space.base * BASE_MULTIPLIERS.top}px;
   ${props => (props.theme.rtl ? 'left' : 'right')}: ${props =>
-    `${props.theme.space.base * baseMultipliers.side}px`};
+    `${props.theme.space.base * BASE_MULTIPLIERS.side}px`};
   /* prettier-ignore */
   transition:
     box-shadow 0.1s ease-in-out,
@@ -68,8 +68,8 @@ export const StyledClose = styled.button.attrs({
   background-color: transparent;
   cursor: pointer;
   padding: 0;
-  width: ${props => props.theme.space.base * baseMultipliers.size}px;
-  height: ${props => props.theme.space.base * baseMultipliers.size}px;
+  width: ${props => props.theme.space.base * BASE_MULTIPLIERS.size}px;
+  height: ${props => props.theme.space.base * BASE_MULTIPLIERS.size}px;
   overflow: hidden;
   text-decoration: none;
   font-size: 0;
--- a/packages/modals/src/styled/StyledClose.ts
+++ b/packages/modals/src/styled/StyledClose.ts
@@ -40,6 +40,12 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
   `;
 };
 
+export const baseMultipliers = {
+  top: 2.5,
+  side: 6.5,
+  size: 10
+};
+
 /**
  * 1. Remove dotted outline from Firefox on focus.
  */
@@ -49,8 +55,9 @@ export const StyledClose = styled.button.attrs({
 })`
   display: block;
   position: absolute;
-  top: ${props => props.theme.space.base * 2.5}px;
-  ${props => (props.theme.rtl ? 'left' : 'right')}: ${props => `${props.theme.space.base * 6.5}px`};
+  top: ${props => props.theme.space.base * baseMultipliers.top}px;
+  ${props => (props.theme.rtl ? 'left' : 'right')}: ${props =>
+    `${props.theme.space.base * baseMultipliers.side}px`};
   /* prettier-ignore */
   transition:
     box-shadow 0.1s ease-in-out,
@@ -61,8 +68,8 @@ export const StyledClose = styled.button.attrs({
   background-color: transparent;
   cursor: pointer;
   padding: 0;
-  width: ${props => props.theme.space.base * 10}px;
-  height: ${props => props.theme.space.base * 10}px;
+  width: ${props => props.theme.space.base * baseMultipliers.size}px;
+  height: ${props => props.theme.space.base * baseMultipliers.size}px;
   overflow: hidden;
   text-decoration: none;
   font-size: 0;
--- a/packages/modals/src/styled/StyledDrawerModalClose.ts
+++ b/packages/modals/src/styled/StyledDrawerModalClose.ts
@@ -7,23 +7,22 @@
 
 import styled from 'styled-components';
 import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
-import { StyledClose, baseMultipliers as styledCloseBaseMultipliers } from './StyledClose';
+import { StyledClose, BASE_MULTIPLIERS as styledCloseBaseMultipliers } from './StyledClose';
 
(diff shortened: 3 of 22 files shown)


packages/modals/src/styled/StyledHeader.ts
```diff
@@ -28,6 +31,11 @@ export const StyledHeader = styled.div.attrs<IStyledHeaderProps>({
   margin: 0;
   border-bottom: ${props => props.theme.borders.sm} ${getColor('neutralHue', 200)};
   padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 10}px`};
+  ${props =>
+    props.isCloseButtonPresent &&
+    `padding-${props.theme.rtl ? 'left' : 'right'}: ${
+      props.theme.space.base * (baseMultipliers.size + baseMultipliers.side)
```
Francois-Esquire, line 37: The padding added only borders the `Close` button, no additional multiplier is added. We may want to add additional padding (eg `Sheet` has added `2` to the base multiplier for additional breathing room with the button)

packages/modals/src/styled/StyledDrawerModalHeader.ts
```diff
@@ -16,6 +17,11 @@ export const StyledDrawerModalHeader = styled(StyledHeader).attrs({
   'data-garden-version': PACKAGE_VERSION
 })`
   padding: ${props => props.theme.space.base * 5}px;
+  ${props =>
+    props.isCloseButtonPresent &&
+    `padding-${props.theme.rtl ? 'left' : 'right'}: ${
+      props.theme.space.base * (baseMultipliers.size + baseMultipliers.side)
```
jzempel, line 23: This should be a base-4 value, precisely the same as `Sheet`. See https://github.com/zendeskgarden/react-components/blob/028957b8a4f0f9fa63ea09fbeb3a59549381d15e/packages/chrome/src/styled/sheet/StyledSheetHeader.ts#L31.

packages/modals/src/styled/StyledHeader.ts
Francois-Esquire: Of course, that makes sense. Will update.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -40,6 +40,12 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
   `;
 };
 
+export const baseMultipliers = {
```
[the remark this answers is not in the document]
Francois-Esquire: Sure, I'll update with ALL_CAPS.

Hmm I'm not sure I follow, why do you think there isn't much to leverage for the shared export?

packages/modals/src/styled/StyledDrawerModalClose.ts
```diff
@@ -7,17 +7,23 @@
...
-import { StyledClose } from './StyledClose';
+import { StyledClose, baseMultipliers as styledCloseBaseMultipliers } from './StyledClose';
 
 const COMPONENT_ID = 'modals.drawer_modal.close';
 
+export const baseMultipliers = {
+  top: 3.5,
+  size: 8,
+  side: styledCloseBaseMultipliers.side
+};
```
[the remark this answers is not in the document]
Francois-Esquire: Sure, will update. We will need design guidance to make sure this is OK, will check offline.

packages/modals/src/styled/StyledHeader.ts
Francois-Esquire: Would we want to treat the `left`/`right` values the same here? It seems to differ from `Sheet` (`6.5` vs `2`)

jzempel: No, it should be different, due to the fact that padding is larger in `Modal` than it is in both `DrawerModal` and `Sheet`. The end result is that the close button should align with buttons in the footer for any of these components.

packages/modals/src/elements/Close.spec.tsx
```diff
@@ -25,4 +26,34 @@ describe('Close', () => {
...
+      const { unmount } = render(<TestComponent />);
+
+      expect(setCloseButtonPresent).toHaveBeenCalledWith(true);
+
+      unmount();
+
+      expect(setCloseButtonPresent).toHaveBeenCalledWith(false);
+      expect(setCloseButtonPresent).toHaveBeenCalledTimes(2);
+    });
+  });
```
hzhu, line 58: Since the source code changes involves a change to the end user UI, I think the test should resemble how users interact with the component.

I was going over the [testing-lib docs](https://testing-library.com/docs/guiding-principles) and it looks like `setCloseButtonPresent` is similar to an instance method. The guide suggests asserting against the DOM, rather than against the instance method or callbacks. The tests are less brittle when they target the end user experience rather than implementation details when testing UI. For example, if an implementation detail changed (e.g. rename `setCloseButtonPresent`), then the current test breaks.

I think the following test would better align with testing-lib's docs:

```tsx
describe('DrawerModal', () => { // DrawerModa.spec.tsx
  it('has horizontal padding when a DrawerModal.Close is present', () => {
    render(
      <DrawerModal isOpen>
        <DrawerModal.Header>Header</DrawerModal.Header>
        <DrawerModal.Close />
      </DrawerModal>
    );

    expect(screen.getByText('Header')).toHaveStyleRule('padding-right', '56px');
  });

  it('does not have horizontal padding when a DrawerModal.Close is absent', () => {
    render(
      <DrawerModal isOpen>
        <DrawerModal.Header>Header</DrawerModal.Header>
      </DrawerModal>
    );

    expect(screen.getByText('Header')).not.toHaveStyleRule('padding-right');
  });
});
```

Note: The RTL behavior is asserted in spec files under `/styled` — so no change there.

packages/modals/src/elements/Close.spec.tsx
```diff
@@ -25,4 +26,34 @@ describe('Close', () => {
 
     expect(getByTestId('close')).toBe(ref.current);
   });
+
+  describe('functionality', () => {
+    it('sets isCloseButtonPresent when mounting and unmounting', () => {
```
hzhu, line 31: Since the source code changes involve the end user UI, I think the test should resemble how users interact with the component.

I was going over the [testing-lib docs](https://testing-library.com/docs/guiding-principles) and it looks like `setCloseButtonPresent` is similar to an instance method. The guide suggests asserting against the DOM, rather than against the instance method or callbacks. The tests are less brittle when they target the end user experience rather than implementation details when testing UI. For example, if an implementation detail changed (e.g. rename `setCloseButtonPresent`), then the current test breaks.

I think the following test would better align with testing-lib's docs and less brittle:

```tsx
describe('DrawerModal', () => { // DrawerModa.spec.tsx
  it('has horizontal padding when a DrawerModal.Close is present', () => {
    render(
      <DrawerModal isOpen>
        <DrawerModal.Header>Header</DrawerModal.Header>
        <DrawerModal.Close />
      </DrawerModal>
    );

    expect(screen.getByText('Header')).toHaveStyleRule('padding-right', '56px');
  });

  it('does not have horizontal padding when a DrawerModal.Close is absent', () => {
    render(
      <DrawerModal isOpen>
        <DrawerModal.Header>Header</DrawerModal.Header>
      </DrawerModal>
    );

    expect(screen.getByText('Header')).not.toHaveStyleRule('padding-right');
  });
});
```

Note: The RTL behavior is asserted in spec files under `/styled` — so no change there.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -40,6 +40,12 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
   `;
 };
 
+export const BASE_MULTIPLIERS = {
+  top: 2.5,
+  side: 6.5,
+  size: 10
```
hzhu, line 46: It's not clear to me how these values were derived. Could you clarify where these values come from or how they were determined?

packages/modals/src/utils/useModalContext.ts
```diff
@@ -9,9 +9,11 @@ import { createContext, useContext, HTMLAttributes } from 'react';
 
 export interface IModalContext {
   isLarge?: boolean;
+  isCloseButtonPresent?: boolean;
   getTitleProps: <T>(options?: T) => T & HTMLAttributes<HTMLDivElement>;
   getContentProps: <T>(options?: T) => T & HTMLAttributes<HTMLDivElement>;
   getCloseProps: <T>(options?: T) => T & HTMLAttributes<HTMLButtonElement>;
+  setCloseButtonPresent: (isPresent: boolean) => void;
```
hzhu, line 16: Not requesting a change since I'm not sure how I'd refactor for the current approach.

Seems like the newly added values are fit nicely in this interface. Context interfaces represent a blueprint of the core/essential methods and properties that is fundamental to the component API. The newly added values are relating to implementation details.

james-sann (approved):
Looks great Mike! Thank you for addressing this.

packages/modals/src/styled/StyledDrawerModalHeader.ts
```diff
@@ -16,6 +17,13 @@ export const StyledDrawerModalHeader = styled(StyledHeader).attrs({
   'data-garden-version': PACKAGE_VERSION
 })`
   padding: ${props => props.theme.space.base * 5}px;
+  ${props =>
+    props.isCloseButtonPresent &&
+    `padding-${props.theme.rtl ? 'left' : 'right'}: ${
+      // the padding added to the Header is based on the close button size and spacing,
+      // with additional padding (+ 2) between the Header content and Close button
```
jzempel, line 24: Please follow Garden's outside block convention for CSS commenting. See https://github.com/zendeskgarden/react-components/blob/1d5b1dfe698e3692ef11a9ed5ca194cc98307274/packages/buttons/src/styled/StyledButton.ts#L47-L88 for details.

packages/modals/src/styled/StyledClose.ts
jzempel: Is the `BASE_MULTIPLIERS` juice extraction worth the squeeze? Maybe just me, but I feel like direct `props.theme.space.base` calculations combined with meaningful comments would be easier to follow.

packages/modals/src/styled/StyledHeader.ts
```diff
@@ -28,6 +31,13 @@ export const StyledHeader = styled.div.attrs<IStyledHeaderProps>({
   margin: 0;
   border-bottom: ${props => props.theme.borders.sm} ${getColor('neutralHue', 200)};
   padding: ${props => `${props.theme.space.base * 5}px ${props.theme.space.base * 10}px`};
+  ${props =>
+    props.isCloseButtonPresent &&
+    `padding-${props.theme.rtl ? 'left' : 'right'}: ${
+      // the padding added to the Header is based on the close button size and spacing,
+      // with additional padding (+ 2) between the Header content and Close button
```
jzempel, line 38: Relocate comment.

packages/modals/src/elements/Close.spec.tsx
Francois-Esquire: Good point @hzhu, I will update these tests.

packages/modals/src/styled/StyledClose.ts
Francois-Esquire: @hzhu these values were extracted from what preexisted for the CSS properties (eg `top: props.theme.space.base * 2.5`) and mapped to this const variable.

@jzempel I think there are benefits to this, without much squeezing IMO. it allows us to update the value in one place if we want to amend in the future, and the imports do tether the values and can be traced back in the code, showing the relationship with the values which is easier to read. I think a part of that would be lost without it.

If y'all don't think it's worth it and prefer literal values, fine by me.

jzempel: I like having connective tissue – we should probably have more – it's just this construct feels like it takes more comment to explain than can be derived from the code. YMMV, so fine w me if you go with your gut.

packages/modals/src/utils/useModalContext.ts
Francois-Esquire: @hzhu this one is a bit tricky for me to understand 😅 this is an internal interface and I wasn't sure but is there a standard for that? Isn't this more open-ended or are there things to consider with this?

packages/modals/src/styled/StyledDrawerModalHeader.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(container.firstChild).toHaveStyleRule('color', DEFAULT_THEME.colors.foreground);
+  });
+
+  it('renders danger styling if provided', () => {
+    const { container } = render(<StyledDrawerModalHeader isDanger />);
+
+    expect(container.firstChild).toHaveStyleRule('color', PALETTE.red[600]);
+  });
+
+  it('renders correctly when button is present', () => {
```
hzhu, line 27: This is already covered at the element spec level: `has horizontal padding when DrawerModal.Close is present`.

packages/modals/src/styled/StyledDrawerModalHeader.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+    render(<StyledDrawerModalHeader isCloseButtonPresent>Header</StyledDrawerModalHeader>);
+
+    expect(screen.getByText('Header')).toHaveStyleRule(
+      'padding-right',
+      `${DEFAULT_THEME.space.base * 14}px`
+    );
+  });
+
+  it('renders correctly in rtl mode when button is present', () => {
```
hzhu, line 36: Can probably colocate this RTL test with the one in `Header.spec.tsx`.

jzempel (approved):

packages/modals/src/styled/StyledDrawerModalHeader.spec.tsx
Francois-Esquire: Good call, just updated these and the `Modal` `Header` tests for consistency.

packages/modals/src/styled/StyledDrawerModalHeader.spec.tsx
Francois-Esquire: Removed these in [a6cf3f8](https://github.com/zendeskgarden/react-components/pull/1307/commits/a6cf3f8251f90ab9df1452314f08e94632f9fce8)!

Issue #1092, opened 2021-05-23
Cannot update a component X while rendering a different component DatepickerRange


pedropmedina:
I followed the docs and updating the state from within onChange in DatePicker is causing the error above.

Here's a snippet of the component:

```typescript
const BaseDateRange: React.FC<BaseDateRangeProps> = ({ onChangeDateRange }) => {
  const [startValue, setStartValue] = useState(new Date());
  const [endValue, setEndValue] = useState(new Date());
  const [showCalendar, setShowCalendar] = useState(false);

  const isInvalid = () => compareAsc(startValue, endValue) === 1;

  const handleShowCalendar = (show: boolean) => {
    if (!show) setShowCalendar(true);
  };

  return (
    <OutsideEvent handler={() => setShowCalendar(false)}>
      <DatepickerRange
        isCompact={isCompact}
        startValue={startValue}
        endValue={endValue}
        onChange={(changes) => {
          if (onChangeDateRange) onChangeDateRange(changes);
          if (changes.startValue) setStartValue(changes.startValue);
          if (changes.endValue) setEndValue(changes.endValue);
        }}
      >
        <div className="py-2 px-4 relative">
          <div className="flex items-center mb-4">
            <div className="px-2 text-sm">Date range:</div>
            <div>
              <Field>
                <DatepickerRange.Start>
                  <StyledInput isCompact={true} onFocus={() => handleShowCalendar(showCalendar)} />
                </DatepickerRange.Start>
              </Field>
            </div>
            <div className="px-2 text-sm">to</div>
            <div className="max-w-[130px]">
              <Field>
                <DatepickerRange.End>
                  <StyledInput
                    isCompact={true}
                    validation={isInvalid() ? 'error' : undefined}
                    onFocus={() => handleShowCalendar(showCalendar)}
                  />
                </DatepickerRange.End>
                {isInvalid() && (
                  <Message validation="error">End date must occur after the start date</Message>
                )}
              </Field>
            </div>
          </div>
          {showCalendar && (
            <div className="shadow-xl max-w-min px-6 py-4 ml-2 absolute top-[100%] left-0 bg-white">
              <DatepickerRange.Calendar />
            </div>
          )}
        </div>
      </DatepickerRange>
    </OutsideEvent>
  );
};

```

hzhu:
Hi @pedropmedina 👋🏽 . Could you create a basic CodeSandbox example demonstrating the issue you're encountering, so I can investigate further? 

[...]

Issue #1174, opened 2021-09-08
`DatepickerRange` component gives "Cannot update a component (xxx) while rendering a different component (`DatepickerRange`)" warning


DariaRadu:
## Expectations

When following the example in the [DatepickerRange](https://garden.zendesk.com/components/date-picker#date-range) documentation, the code should work without errors/warnings.

## Reality

When setting the `startValue`/ `endValue` state using the same code as the example, the following warning pops up.
![image](https://user-images.githubusercontent.com/21333467/132517458-35aaf2d7-b302-4533-ac0b-0361ddb507c2.png)
![image](https://user-images.githubusercontent.com/21333467/132517487-7b3da4be-c71e-43ae-9551-aff2be38805b.png)

## Steps to Reproduce

1. Use the code in the documented example inside of another component.
2. Change the start value/end value
3. (Optional) The current implementation has the datepicker in a dropdown. We are not sure if that might affect it.

### Fine Print

- Component: `DatepickerRange`
- Version: `8.39.0`
- Browsers: Chrome

jzempel:
Thanks @DariaRadu. This is a duplicate of #1092. Garden has this in the backlog and will schedule a fix as soon as possible.

PR #1309, opened 2022-03-25
fix(datepickers): fix React state synchronization bug

hzhu:
## Description

This pull request fixes https://github.com/zendeskgarden/react-components/issues/1092 and https://github.com/zendeskgarden/react-components/issues/1174.

## Background

The `DatepickerRange` has state synchronization issues that can cause unintended consequences. React throws a  :warning: for this. 

8ffa7d97-4aa7-49c7-aa0f-e3510b1f5893

I think this is happening:

Component A's `setState` call is called in component B's reducer function. React will not be able to deterministically update component A, while it is working on component B. For example, what if component B is unmounted — then component A's `setState` call is lost. And if component A is unmounted, React is calling `setState` on an unmounted component.

In the case of `DatepickerRage`, the consumer passes `setState` via `onChange`. This `onChange` is then called in `DatepickerRange`'s reducer function. In addition to React synchronization issues, reducer functions should be pure and side effect free, but `onChange` is inherently a side effect.

```jsx
<Component onChange={(value) => {
  setState(value); // side effect wrapped in hook
  // or
  fetchData(value).then(console.log); // free floating side effect
}} />
```

## Solution

Decouple `onChange` from state management which eliminates the React warnings 🎉 .

Now external `setState` calls are no longer called at another component's state management (`useReducer`) level. The `onChange` is now called at the event handler level and that is where side effects should happen anyways.

## Steps to reproduce

1) Pull `main` branch and make the following local changes

2) Update the `DatepickerRage` Story to be driven by local state:

```jsx
const [end, setEnd] = useState<Date | undefined>(new Date());
const [start, setStart] = useState<Date | undefined>(new Date());

const onChange = useCallback(
  ({ endValue, startValue }: { endValue?: Date; startValue?: Date }) => {
    // these set state calls are called in datepicker range's reducer, causing React synchronization warning.
    setEnd(endValue); 
    setStart(startValue);
  },
  []
);

<DatepickerRange
  endValue={end}
  startValue={start}
  onChange={onChange}
  formatDate={formatDate}
  isCompact={isCompact}
>
```

3) Select a date range

4) Notice the React error in the JavaScript console.

Note: The React warning is also visible in official Garden CodeSandbox [example](https://garden.zendesk.com/components/date-picker#date-range).

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/datepickers/src/elements/DatepickerRange/components/End.tsx
+++ b/packages/datepickers/src/elements/DatepickerRange/components/End.tsx
@@ -7,10 +7,14 @@
 
 import React, { PropsWithChildren, HTMLAttributes, useCallback } from 'react';
 import { KEY_CODES, composeEventHandlers } from '@zendeskgarden/container-utilities';
+import isValid from 'date-fns/isValid';
+import isSameDay from 'date-fns/isSameDay';
+import { parseInputValue } from '../utils/datepicker-range-reducer';
 import useDatepickerRangeContext from '../utils/useDatepickerRangeContext';
 
 export const End = (props: PropsWithChildren<HTMLAttributes<HTMLInputElement>>) => {
-  const { state, dispatch, endInputRef } = useDatepickerRangeContext();
+  const { state, dispatch, onChange, startValue, endValue, endInputRef, customParseDate } =
+    useDatepickerRangeContext();
 
   const onChangeCallback = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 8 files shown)


packages/datepickers/src/elements/DatepickerRange/components/Month.tsx
```diff
@@ -231,6 +232,37 @@ export const Month = forwardRef<HTMLDivElement, IMonthProps>(
             onClick={() => {
               if (!isDisabled) {
                 dispatch({ type: 'CLICK_DATE', value: date });
+                if (onChange) {
+                  if (state.isStartFocused) {
```
hzhu, line 236: The logical branching is pulled from take from `CLICK_DATE` switch case in the reducer function.

packages/datepickers/src/elements/DatepickerRange/components/End.tsx
```diff
@@ -30,31 +34,49 @@ export const End = (props: PropsWithChildren<HTMLAttributes<HTMLInputElement>>)
...
     },
-    [dispatch, props.children]
+    [handleBlur, props.children]
   );
 
   const childElement = React.Children.only(props.children as React.ReactElement);
 
   return React.cloneElement(childElement, {
-    value: state.endInputValue,
+    value: state.endInputValue || '',
```
hzhu, line 79: Prevents switching React input from a controlled component to an uncontrolled component. This is a band-aid fix, but necessary given the current architecture of `DatepickerRange`.

hzhu (commented):
@jzempel I added a new commit to remove prop drilling of the `onChange` prop from `Calendar` -> `<Month>`. 

Many of the other props for `Month` can be removed also the values are accessible through context. I can add a follow-on PR immediately for those since I want to keep this PR focused on the React warnings.

jzempel (approved):

jzempel:
@hzhu for CHANGELOG purposes, it might be nice if the merged title could be improved to be a bit more descriptive.

PR #1312, opened 2022-04-01
chore(demo): add StrictMode decorator

jzempel:
## Description

Adds "Strict mode" to the Storybook toolbar for enabling/disabling a `React.StrictMode` story decorator. This is being added as a convenience for component testing purposes. Strict mode is disabled by default.

## Detail

A couple of stories to try with strict mode enabled:

- Buttons ➡️ ButtonGroup
  - Click on any button
  - Result: 💥 
- Dropdowns ➡️ Menu
  -  Use keyboard to navigate menu
  - Result: all `Item` navigation requires two keystrokes instead of one

It seems that anything connected with `@zendeskgarden/containers-selection` is problematic with strict mode. But I'm sure there's more when we use this to gain in-depth insight into problematic, unsafe lifecycles.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,7 +1,7 @@
 {
   "*.{js,ts,tsx}": [
     "stylelint",
-    "eslint",
+    "eslint --max-warnings 0 --ignore-pattern !.storybook",
     "jest --config=utils/test/jest.config.js --bail --findRelatedTests",
     "prettier --write"
   ],
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -68,6 +68,7 @@ const withThemeProvider = (Story, context) => {
 };
 
 const withStrictMode = (Story, context) =>
+  /* eslint-disable-next-line new-cap */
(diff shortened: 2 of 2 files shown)


.lintstagedrc
```diff
@@ -1,7 +1,7 @@
 {
   "*.{js,ts,tsx}": [
     "stylelint",
-    "eslint",
+    "eslint --max-warnings 0 --ignore-pattern !.storybook",
```
jzempel, line 4: prevents getting burned by build errors in storybook

hzhu (approved):

.lintstagedrc
Francois-Esquire: @jzempel Would it be a good idea to include an `.eslintignore` in the future? The IDE (VS-Code, etc) gets an advantage when using that file with Eslint projects when configuring the ignores globally; devs would be able to see lint errors in the dot directory which is ignored by default, with formatting.

jzempel: ack, for sure. I hadn't seen that `.eslintignore` can be used to config an allowlist.

PR #1326, opened 2022-04-04
fix(theming, modals, colorpickers): refactor types and interfaces

jzempel:
## Description

Started as a type standardization PR for `colorpickers`, but cascaded to `modals` and then to `theming`. The result includes a number of fixes for a more readable API and improved reuse across these package dependencies.


mtomcal (approved):
Looks up to spec with the RFC to me 👍

PR #1333, opened 2022-04-07
fix(buttons): refactor types and interfaces

jzempel:
## Description

## Detail

Marked `ButtonGroup` as `@deprecated` per [website](https://garden.zendesk.com/components/button-group#:~:text=This%20is%20a%20legacy%20component%20and%20may%20be%20deprecated%20in%20the%20future.%20Garden%20does%20not%20presently%20recommend%20the%20use%20of%20Button%20groups.)

### DIFF
--- a/packages/buttons/src/types/index.ts
+++ b/packages/buttons/src/types/index.ts
@@ -1,0 +1,76 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { AnchorHTMLAttributes, ButtonHTMLAttributes, HTMLAttributes, SVGAttributes } from 'react';
+
+export const SIZE = ['small', 'medium', 'large'] as const;
+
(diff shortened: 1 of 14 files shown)


packages/buttons/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { AnchorHTMLAttributes, ButtonHTMLAttributes, HTMLAttributes, SVGAttributes } from 'react';
+
+export const SIZE = ['small', 'medium', 'large'] as const;
```
hzhu, line 10: It looks like this is `undefined` at runtime and React throws a `factoryWithTypeCheckers.jss:23 Warning: Invalid argument supplied to oneOf, expected an array` in development mode.

hzhu: I think I may have had stale `node_modules` earlier, but I'm no longer seeing this warning.

hzhu (approved):

PR #1336, opened 2022-04-12
fix(datepickers): refactor types and interfaces

jzempel:
## Description

### DIFF
--- a/packages/datepickers/src/types/index.ts
+++ b/packages/datepickers/src/types/index.ts
@@ -1,0 +1,138 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { HTMLAttributes } from 'react';
+import { Modifiers } from 'popper.js';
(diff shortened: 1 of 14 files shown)


packages/datepickers/src/types/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  ...SHARED_PLACEMENT,
+  'right',
+  'right-start',
+  'right-end',
+  'left',
+  'left-start',
+  'left-end'
+] as const;
+
+export type GardenPlacement = typeof PLACEMENT[number];
```
hzhu, line 41: @jzempel could you help me understand what the bracket notation here means? Should I read it as "string lookup on the `PLACEMENT` object"? But I see that `PLACEMENT` is an array, so maybe it's an index lookup on `PLACEMENT`? I'm still soaking up TS and a bit confused 🙏🏽 .

hzhu (approved):

packages/datepickers/src/types/index.ts
jzempel: this SO & comments has just about as good of an explanation as any I've found: https://stackoverflow.com/a/59857409

PR #1337, opened 2022-04-12
fix(notifications): refactor types and interfaces

jzempel:
## Description


hzhu (approved):

packages/notifications/src/utils/useNotificationsContext.ts
```diff
@@ -6,9 +6,10 @@
  */
 
 import { createContext, useContext } from 'react';
-import { VALIDATION_HUE } from './types';
 
-export const NotificationsContext = createContext<VALIDATION_HUE | undefined>(undefined);
+export type Hue = 'successHue' | 'warningHue' | 'dangerHue' | 'neutralHue';
```
Francois-Esquire, line 10: Should this be moved to `types.ts`?

Francois-Esquire (approved):

PR #1346, opened 2022-04-20
feat(accordions): pass icon props to default check icon

hzhu:
## Description

Allow consumers to pass in props for the default completed (check) icon.

## Detail

The default checked icon is currently not read aloud on screen readers. This PR creates a new prop so that the SVG icon element can be configured with the appropriate ARIA attributes. Alternatively, consumers can pass their own, fully configured icons, to the `icon` prop. We'll need to document both usages on the website.

Consumers can now configure the default check icon and **can make it available to screen readers**:

screen recording

Note: I will update the website + packages to reflect these changes once all Level Access tasks related to accordions package have been addressed. But updated website documentation will probably require a new task.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/accordions/demo/stepper.stories.mdx
+++ b/packages/accordions/demo/stepper.stories.mdx
@@ -35,7 +35,8 @@ import { STEPPER_STEPS as STEPS } from './stories/data';
     argTypes={{
       hasIcon: { name: 'icon', table: { category: 'Stepper.Label' } },
       isLabelHidden: { name: 'isHidden', table: { category: 'Stepper.Label' } },
-      steps: { name: 'Stepper.Step[]', table: { category: 'Story' } }
+      steps: { name: 'Stepper.Step[]', table: { category: 'Story' } },
+      iconProps: { name: 'iconProps', table: { category: 'Stepper.Label' } }
     }}
     parameters={{
       design: {
--- a/packages/accordions/demo/stepper.stories.mdx
+++ b/packages/accordions/demo/stepper.stories.mdx
@@ -22,7 +22,16 @@ import { STEPPER_STEPS as STEPS } from './stories/data';
 <Canvas>
   <Story
     name="Stepper"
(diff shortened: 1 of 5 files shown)


packages/accordions/demo/stories/StepperStory.tsx
```diff
@@ -21,7 +21,15 @@ export const StepperStory: Story<IArgs> = ({ steps, ...args }) => (
...
       <Stepper.Step key={index}>
-        <Stepper.Label icon={args.hasIcon && <Icon />} isHidden={args.isLabelHidden}>
+        <Stepper.Label
+          icon={args.hasIcon && <Icon />}
+          isHidden={args.isLabelHidden}
+          iconProps={{
+            role: 'img',
+            'aria-label': 'checked',
+            'aria-hidden': undefined
+          }}
```
jzempel, line 31: It would be good to initialize the default in the MDX `args` as a control (`args.iconProps`). That way the end user has full control for testing the rendered output.

jzempel (changes requested):

packages/accordions/demo/stepper.stories.mdx
```diff
@@ -22,7 +22,16 @@ import { STEPPER_STEPS as STEPS } from './stories/data';
...
-    args={{ hasIcon: false, isLabelHidden: false, steps: STEPS }}
+    args={{
+      hasIcon: false,
+      isLabelHidden: false,
+      steps: STEPS,
+      iconProps: {
+        role: 'img',
+        'aria-label': 'checked',
+        'aria-hidden': undefined
+      }
```
jzempel, line 33: We should see this under the "Stepper.Label" section of controls. Please add with the appropriate category to `argTypes` below.

jzempel (changes requested):

jzempel (approved):

PR #1349, opened 2022-05-10
fix(forms): conditionally render aria-describedby when field has hint

Francois-Esquire:
## Description

This PR removes `aria-describedby` attribute from an input using `Field` component when a `Hint` is not present.

## Detail

Using React context, we capture when the `Hint` component is present and add the `aria-described` prop when it is.

PR TODO:
- [ ] apply changes to the `react-dropdowns` package

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/elements/common/Field.tsx
+++ b/packages/forms/src/elements/common/Field.tsx
@@ -23,11 +23,12 @@ export interface IFieldProps extends HTMLAttributes<HTMLDivElement> {
  */
 export const Field = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
   (props, ref) => {
+    const [hasHint, setHasHint] = useState(false);
     const [isLabelActive, setIsLabelActive] = useState(false);
     const [isLabelHovered, setIsLabelHovered] = useState(false);
     const multiThumbRangeRef = useRef<HTMLDivElement>(null);
     const getMessageProps = (messageProps: any) => ({ role: 'alert', ...messageProps });
-    const propGetters = useField(props.id);
+    const { getInputProps, ...propGetters } = useField(props.id);
     const fieldProps = useMemo(
       () => ({
         ...propGetters,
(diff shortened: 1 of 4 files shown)


packages/forms/src/elements/common/Hint.tsx
```diff
@@ -18,6 +18,21 @@ export const Hint = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivEleme
...
+      }
+
+      return () => {
+        if (fieldContext) {
+          fieldContext.setHint(false);
+        }
+      };
+      // since we only want to run this once, if context exists,
+      // we use an empty dependency array
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
mtomcal, line 33: @jzempel Why does our eslint configuration flag an empty array as it is a valid case in a component's lifecycle?

packages/forms/src/elements/common/Field.tsx
```diff
@@ -36,9 +37,17 @@ export const Field = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElem
         setIsLabelActive,
         isLabelHovered,
         setIsLabelHovered,
-        multiThumbRangeRef
+        multiThumbRangeRef,
+        getInputProps: (...args: any[]) => {
+          const { 'aria-describedby': describedBy, ...inputProps } = getInputProps(...args);
+
+          if (hasHint) inputProps['aria-describedby'] = describedBy;
```
mtomcal, line 44: Would `inputProps['aria-describedby']` deliberately need to be set to `undefined` if `hasHint` is `false` on state change?

Francois-Esquire: No it shouldn't need to be explicitly set because the `getInputProps` function gets called on every render. We are overriding the useField hook outputs (`'aria-describedby'`) that to check if the component exists in this instance.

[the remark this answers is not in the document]
Francois-Esquire: Good catch! I removed the unnecessary logic and using the `isDescribed` option ([source](https://github.com/zendeskgarden/react-containers/blob/main/packages/field/src/useField.ts#L37)) to set `aria-describedby` for `react-forms` inputs.

mtomcal (approved):

PR #1356, opened 2022-05-23
feat(pane): add providerId

mtomcal:
## Description

As a consumer, I want develop a resizable-2-or-more-columns grid layout and I wish to nest a resizeable-2-or-more-rows grid layout within the parent the column grid layout.

We have introduced an optional `id` prop on `PaneProvider` and optional `providerId` prop on `Pane.Splitter` to allow nesting of Pane-driven Grids. Now, `Pane.Splitter` may control a "parent" `PaneProvider` while beneath a "child" `PaneProvider`

## Checklist

      designer as a reviewer)~


packages/grid/src/elements/pane/PaneProvider.spec.tsx
```diff
@@ -216,6 +216,7 @@ describe('PaneProvider', () => {
           "getGridTemplateColumns": [Function],
           "getGridTemplateRows": [Function],
           "getRowValue": [Function],
+          "id": undefined,
```
jzempel, line 219: Seems like there should be tests added here to suss out the new features of a nested pane provider.

packages/grid/demo/~patterns/stories/CardStory.tsx
```diff
@@ -0,0 +1,115 @@
...
+export const CardStory: Story<{ columns: IColumns[] }> = ({ columns }) => {
+  const themeContext = useContext(ThemeContext);
+  const { ref, width = 1, height = 1 } = useResizeObserver<HTMLDivElement>();
+
+  // remove 20px from the totalPanesWidth to account for gap size x 2
+  // otherwise there will be a difference between cursor and splitter position
+  return (
+    <PaneProvider
+      id="column-layout"
+      totalPanesWidth={width - 20}
```
jzempel, line 78: [nit] it might be nice to put `gap` into a `const gap = 10;` variable and then use `{width - (gap * 2)}` here (and `gap: ${gap}px` below) in order to drive the meaning of `totalPanesWidth`. In other words, express in code rather than relying (only) on the comment above.

Also, while we're at it, it is always best if you can lean into the system with Garden theming. So prefer a `12px`(?) base-4 gap size based on `theme.space.base`.

Same with `Row` above.

packages/grid/demo/~patterns/stories/CardStory.tsx
```diff
@@ -0,0 +1,115 @@
...
+              // remove the height of the storybook top bar
+              height: 'calc(100vh - 80px)',
+              gap: '10px',
+              gridTemplateRows: getGridTemplateRows(),
+              gridTemplateColumns: getGridTemplateColumns()
+            }}
+          >
+            {columns.map(column => (
+              <Pane key={column.name}>
+                {/* Make column overflow visible otherwise splitters are cut off */}
```
jzempel, line 104: Interesting. I'm seeing some styling wonkiness with the splitter `::before` pseudo highlighted by the card layout. I think we should address with refactored border styling. But definitely in a future PR.

Screen Shot 2022-05-24 at 10 43 34 AM

packages/grid/src/elements/pane/PaneProvider.spec.tsx
mtomcal: Fixed in https://github.com/zendeskgarden/react-components/pull/1356/commits/7cfdaf949e0ff030eb133e4ac672aa2403c64b0b

packages/grid/demo/~patterns/stories/CardStory.tsx
```diff
@@ -0,0 +1,115 @@
...
+    >
+      {({ getGridTemplateColumns, getGridTemplateRows }) => {
+        return (
+          <div
+            ref={ref}
+            style={{
+              direction: themeContext.rtl ? 'rtl' : 'ltr',
+              display: 'grid',
+              width: '100%',
+              // remove the height of the storybook top bar
```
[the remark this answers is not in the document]
mtomcal: Fixed in Fixed in https://github.com/zendeskgarden/react-components/pull/1356/commits/7cfdaf949e0ff030eb133e4ac672aa2403c64b0b

packages/grid/demo/~patterns/stories/CardStory.tsx
mtomcal: Fixed in https://github.com/zendeskgarden/react-components/pull/1356/commits/7cfdaf949e0ff030eb133e4ac672aa2403c64b0b

jzempel (approved):

Francois-Esquire (approved):

PR #1358, opened 2022-05-27
feat(avatars): extend status prop

Francois-Esquire:
## Description

This PR extends the `status` prop for the `Avatar` component.

## Detail

The `status` prop extension now allows for two new status types: `transfers` & `offline`, as well as adding an icon for the `away` status.

The Away status:
Screen Shot 2022-06-13 at 1 55 28 PM

The Transfers status:
Screen Shot 2022-06-13 at 1 55 44 PM

The Offline status:
Screen Shot 2022-06-13 at 3 51 44 PM

Regarding accessibility, the `active` status badge is read out on change, so it retains its existing behavior.

## Checklist

      designer as a reviewer)


mtomcal:
Screen Shot 2022-05-31 at 11 47 46 AM
Screen Shot 2022-05-31 at 11 48 01 AM

Do we want to allow a badge type of "arrow" without a "status" explicitly set?

packages/avatars/demo/avatar.stories.mdx
```diff
@@ -24,13 +24,25 @@ import { AVATAR_TYPE as TYPE } from './stories/data';
       badge: { control: 'number' },
```
mtomcal, line 24: According to the docs and types, `badge` accepts a `ReactNode`. Can we illustrate that this would accept a `number`, `string`, or a `ReactElement` of somekind? This might involve a dropdown with `number`, `string`, `element` with sample values perhaps.

packages/avatars/src/styled/StyledStatusIndicator.ts
```diff
@@ -0,0 +1,155 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStatusIndicatorProps>`
+  display: inline-flex;
+  flex: 1;
+  align-items: center;
+  justify-content: center;
+
+  transition: all ${TRANSITION_DURATION}s ease-in-out, color 0s;
+  animation: ${animation} ${TRANSITION_DURATION * 1.5}s ease-in-out;
```
mtomcal, line 144: @jzempel Im thinking with future styled components `animationStyles` becomes a separate function perhaps?

packages/avatars/src/types/index.ts
```diff
@@ -24,6 +24,14 @@ export interface IAvatarProps extends HTMLAttributes<HTMLElement> {
   size?: typeof SIZE[number];
   /** Applies status styling */
   status?: typeof STATUS[number];
-  /** Sets the badge text and applies active styling */
-  badge?: string | number;
+  /** Sets the badge and applies active styling if it is a string or number */
+  badge?: string | number | boolean | ReactNode;
```
mtomcal, line 28: `ReactNode` includes `string | number | boolean` in its type definition. Can this simplify to `badge?: ReactNode;`

Screen Shot 2022-05-31 at 12 03 24 PM
Screen Shot 2022-05-31 at 12 03 32 PM

packages/avatars/demo/stories/AvatarStory.tsx
```diff
@@ -9,25 +9,51 @@ import React from 'react';
...
+  badge,
+  badgeType,
+  badgeChildren,
+  ...args
+}) => {
+  const child = {
+    icon: args.isSystem ? <IconSystem /> : <IconUser />,
+    image: <img alt="" src={`images/avatars/${args.isSystem ? 'system' : 'user'}.png`} />,
+    text: <Avatar.Text>{children || (args.isSystem ? 'ZD' : 'G')}</Avatar.Text>
+  }[type || 'image'];
```
mtomcal, line 41: [nit] This is confusing to me. Perhaps leave the object declaration here and place the property accessing in:

```
<Avatar
    {...}
    >
      {child[type || 'image']}
    </Avatar>
```
Perhaps stick the default value of `image` in the `avatar.stories.mdx` arg declarations?

packages/avatars/demo/avatar.stories.mdx
```diff
@@ -24,13 +24,25 @@ import { AVATAR_TYPE as TYPE } from './stories/data';
...
+      },
+      badgeChildren: { name: 'badge children', control: 'text', table: { category: 'Story' } },
       type: {
         control: 'radio',
         options: TYPE,
         table: { category: 'Story' }
       }
     }}
     parameters={{
+      layout: 'centered',
```
jzempel, line 45: Why was this added? Demos are default layout, unless `placement` is used to demonstrate orientation.

jzempel (changes requested):
I didn't get too far with the review because the badge is breaking in a number of ways that were stable before the refactor:
- storybook `badge` control not working as expected (I'd like to talk this through)
- badge overflow styling is broken
- avatar of type `image` is clipping the badge

james-sann:
Hey @Francois-Esquire 👋

Just noted a few things that need amending. Reach out if you have any further questions.

PR Review notes-0100722

packages/avatars/demo/stories/AvatarStory.tsx
Francois-Esquire: I believe this was [how it was prior](https://github.com/zendeskgarden/react-components/blob/main/packages/avatars/demo/stories/AvatarStory.tsx#L26-L31), however I did move it around. Would it be less confusing if moved back or is it how the property is accessed that is?

packages/avatars/src/styled/StyledAvatar.ts
```diff
@@ -5,129 +5,59 @@
...
-    }
-
-    minWidth = height;
-  } else if (props.size === 'large') {
-    position = math(`${props.theme.shadowWidths.sm} + 1`);
-  } else if (props.size === 'small') {
-    position = math(`${props.theme.shadowWidths.sm} - 1`);
-  } else if (props.size === 'medium') {
-    position = props.theme.shadowWidths.sm;
+  const [xxs, xs, s, m, l] = SIZE;
```
jzempel, line 22: Is the smaller default badge size change intentional? The original design intent was for the `active` notification change to result in a more pronounced "bounce" effect in order to get the user's attention. The current animation does not follow original intent and is very subtle.

For comparison, visit https://zendeskgarden.github.io/react-components/?path=/story/packages-avatars-avatar--avatar&args=status:available and then set a `badge` control – observing the effect.

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -43,6 +47,24 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
         {...other}
       >
         {Children.only(children)}
+        {(badge || status) && (
```
jzempel, line 50: This code is no longer able to show a badge of `'0'` as expected. If badge is defined, then it should be displayed.

packages/avatars/src/styled/StyledStatusIndicator.ts
```diff
@@ -0,0 +1,142 @@
...
+    case available:
+    case active:
+    case away:
+    case transfers:
+      backgroundColor = getStatusColor(props.status, props.theme);
+      borderColor = backgroundColor;
+      break;
+    case offline:
+      borderColor = getStatusColor(props.status, props.theme);
+      backgroundColor = surfaceColor;
```
jzempel, line 111: Using `surfaceColor` as the background color for offline feels wrong. My hunch is that the design wants this to be `props.theme.palette.white` in every case, no matter what surface it is sitting on.

packages/avatars/src/types/index.ts
```diff
@@ -24,6 +24,6 @@ export interface IAvatarProps extends HTMLAttributes<HTMLElement> {
   size?: typeof SIZE[number];
   /** Applies status styling */
   status?: typeof STATUS[number];
-  /** Sets the badge text and applies active styling */
+  /** Sets the badge and applies active styling if it is a string or number */
```
jzempel, line 27: [nit] should this be reverted?

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -26,15 +30,15 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
     },
     ref
   ) => {
-    const computedStatus = badge === undefined ? status : 'active';
+    const computedBadge = ['string', 'number'].includes(typeof badge) ? `${badge}` : null;
+    const computedStatus = computedBadge ? 'active' : status;
```
jzempel, line 34: Should this be reverted?

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -43,6 +47,24 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
...
         {Children.only(children)}
+        {(badge || status) && (
+          <StyledStatusIndicator
+            size={size}
+            status={computedStatus}
+            backgroundColor={backgroundColor}
+            foregroundColor={foregroundColor}
+            surfaceColor={surfaceColor}
+          >
+            {typeof computedBadge === 'string' ? (
```
jzempel, line 58: Would it be simpler to just use `computedStatus === 'active'`?

jzempel (changes requested):

packages/avatars/src/elements/Avatar.tsx
james-sann: @Francois-Esquire I'm a bit confused here with the use of icons and sizes, so please explain so I can understand the implementation better. I might be missing a technical explanation but please if the following is incorrect, educate away!

It looks like you're using the `arrow-left` existing Garden 12px icon for the `transfer only` icon at both sizes? I remember walking you through the designs early on and [also stated here](https://github.com/zendeskgarden/react-components/pull/1358#issuecomment-1143114025) that that isn't optimal. We need two separate icon sizes used - a 12px variant and a 16px variant, as the grids are different when designed. The same applies for the `clock` icon. Scaling a 12px icon to 16px will cause off-pixel alignment and appear blurry on lower res screens. This why we have 12px and 16px icon variants in the first place.

We have the `clock` icon inside the current Garden icon library but not the specifically designed `transfers only` icon which has subtle differences from the Garden `arrow-left` icon.

james-sann:
@Francois-Esquire I'm a bit confused here with the use of icons and sizes, so please explain so I can understand the implementation better. I might be missing a technical explanation but please if the following is incorrect, educate away!

It looks like you're using the `arrow-left` existing Garden 12px icon for the `transfer only` icon at both sizes? I remember walking you through the designs early on and [also stated here](https://github.com/zendeskgarden/react-components/pull/1358#issuecomment-1143114025) that that isn't optimal. We need two separate icon sizes used - a 12px variant and a 16px variant, as the grids are different when designed. The same applies for the `clock` icon. Scaling a 12px icon to 16px will cause off-pixel alignment and appear blurry on lower res screens. This why we have 12px and 16px icon variants in the first place.

We have the `clock` icon inside the current Garden icon library but not the specifically designed `transfers only` icon which has subtle differences from the Garden `arrow-left` icon.

packages/avatars/src/styled/StyledAvatar.ts
Francois-Esquire: Good call on that, I reverted to the original behavior so it should be back now.

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -43,6 +47,24 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
...
+              <span>{computedBadge}</span>
+            ) : (
+              <>
+                {computedStatus === 'away' ? <ClockIcon /> : null}
+                {computedStatus === 'transfers' ? <ArrowLeftIcon /> : null}
```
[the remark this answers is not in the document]
Francois-Esquire: Updated!

Francois-Esquire:
> @Francois-Esquire I'm a bit confused here with the use of icons and sizes, so please explain so I can understand the implementation better. I might be missing a technical explanation but please if the following is incorrect, educate away!
> 
> It looks like you're using the `arrow-left` existing Garden 12px icon for the `transfer only` icon at both sizes? I remember walking you through the designs early on and [also stated here](https://github.com/zendeskgarden/react-components/pull/1358#issuecomment-1143114025) that that isn't optimal. We need two separate icon sizes used - a 12px variant and a 16px variant, as the grids are different when designed. The same applies for the `clock` icon. Scaling a 12px icon to 16px will cause off-pixel alignment and appear blurry on lower res screens. This why we have 12px and 16px icon variants in the first place.
> 
> We have the `clock` icon inside the current Garden icon library but not the specifically designed `transfers only` icon which has subtle differences from the Garden `arrow-left` icon.

@james-sann thanks for breaking this down, I made a recent update to use the different sized icons (both 12px and 16px variants). Do let me know if it looks off. [Raised a PR for the `arrow-left-sm` (AKA transfers-only)](https://github.com/zendeskgarden/svg-icons/pull/274) and once that makes it in, will replace the arrow-left icon currently in use with the new icon.

packages/avatars/src/styled/StyledAvatar.ts
jzempel: The animation is restored. But I'm 99% sure the badge resize from the original 20px down to 16px is a result of a bug in Figma where the badge was not originally sized correctly.

james-sann (approved):
Looks great Mike! Thank you for making these icon changes and adjustments

PR #1371, opened 2022-06-06
feat(grid): add pane splitter button

mtomcal:
## Description

Add the option of an expand and collapse button as a Pane.Splitter child component.

## Checklist

      designer as a reviewer)


packages/grid/src/elements/pane/components/SplitterButton.tsx
```diff
@@ -0,0 +1,113 @@
...
+        setValue(min);
+      }
+    }, [isMin, min, max, setValue]);
+
+    return (
+      <Tooltip content={label}>
+        <StyledPaneSplitterButton
+          aria-label={label}
+          {...props}
+          isBasic
```
jzempel, line 94: This `isBasic` prop probably makes more sense as a fixed `attr` on the styled component

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    &[data-garden-focus-visible] {
+      box-shadow: ${boxShadow}, ${focusRingShadow};
+    }
+
+    &:active {
+      background-color: ${activeColor};
+    }
+  `;
+};
```
jzempel, line 45: The `ChevronButton`/`IconButton` already has these styles – shouldn't need to repeat here.

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  return css`
+    top: ${top};
+    right: ${right};
+    bottom: ${bottom};
+    left: ${left};
+    width: ${buttonDimension}px;
+    min-width: ${buttonDimension}px;
+    height: ${buttonDimension}px;
```
jzempel, line 98: None of these should be specified. Prefer adding `size="small"` to styled `attrs`.

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
```diff
@@ -0,0 +1,122 @@
...
+  `;
+};
+
+export const StyledPaneSplitterButton = styled(IconButton).attrs({
```
jzempel, line 102: Was `ChevronButton` considered here? The visuals on that composite have already been vetted for system consistency and it would alleviate the `svg-icons` package dependency. Using `ChevronButton`, I would expect to see the following added to `attrs`:

- `isBasic={true}`
- `isPill={true}`
- `size='small'`

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
```diff
@@ -0,0 +1,122 @@
...
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledSplitterButtonProps>`
+  position: absolute;
+  opacity: 0;
+
+  ${StyledPaneSplitter}:hover &,
+  &[data-garden-focus-visible] {
+    opacity: 1;
```
jzempel, line 111: Opacity will definitely need transition treatment. The UI is distractingly "flicker-y" with the current styling. I'd recommend a `0.25s ease-in-out` for consistency with other system elements (the splitter itself or the table overflow button, for example). You'll probably need to pull in existing `Button` transition info for the override.

packages/grid/src/types/index.ts
```diff
@@ -219,3 +221,14 @@ export interface ISplitterProps extends HTMLAttributes<HTMLDivElement> {
   /** Determines splitter orientation within a pane */
   orientation?: Orientation;
 }
+
+export interface ISplitterButtonProps
+  extends Omit<
+    IIconButtonProps,
+    'isRotated' | 'isPrimary' | 'isDanger' | 'isBasic' | 'focusInset' | 'isNeutral' | 'size'
```
jzempel, line 228: Are we positive that `focusInset` should be knocked-out here? There may be tighter visual layouts where this focus treatment is desirable.

jzempel (changes requested):
`paneProvider.stories.mdx` is missing the addition of the `Pane.SplitterButton` subcomponent – important for seeing how the API is going to render.

packages/grid/package.json
```diff
@@ -21,11 +21,16 @@
...
     "@zendeskgarden/container-splitter": "^0.2.3",
     "@zendeskgarden/container-utilities": "^0.7.1",
+    "@zendeskgarden/react-buttons": "^8.52.0",
+    "@zendeskgarden/react-tooltips": "^8.52.0",
+    "@zendeskgarden/svg-icons": "6.32.0",
     "polished": "^4.0.0",
     "prop-types": "^15.5.7",
-    "react-merge-refs": "^1.1.0"
+    "react-merge-refs": "^1.1.0",
+    "react-popper": "^2.2.3"
```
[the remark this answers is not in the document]
mtomcal: Builds fail when adding tooltip without these libraries.

packages/grid/package.json
```diff
@@ -34,7 +39,7 @@
     "styled-components": "^4.2.0 || ^5.0.0"
   },
   "devDependencies": {
-    "@zendeskgarden/react-theming": "^8.52.0",
+    "@zendeskgarden/react-theming": "^8.1.0",
```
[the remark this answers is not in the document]
mtomcal: Accidental transposition

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
mtomcal: In our 1:1 you told me to use IconButton. I talked to Lucijan specifically about IconButton usage.

packages/grid/package.json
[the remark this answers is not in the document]
jzempel: Ah right, of course. Might highlight the opportunity to change tooltip placement based on the placement of the splitter button. But this can probably be a future conversation with design.

packages/grid/src/types/index.ts
mtomcal: @lucijanblagonic told me no focus inset

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
mtomcal: I have zero recollection of ChevronButton being considered in our 1:1 conversation only IconButton. I re-examined the ChevronButton and found it will only point up and down not left or right. I feel for the sake of time we stick with IconButton as we initially discussed as it achieves the desired behavior and does not change the developer experience.

packages/grid/src/types/index.ts
lucijanblagonic: Since button is only `24px`, has shadow and floating above the border and other UI, I thought that `focusInset` was not needed. Tables have overflow buttons with `focusInset` but are also `28px`. I don't have a strong opinion on this, but that was my initial assumption.

packages/grid/src/elements/pane/components/SplitterButton.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const onClick = useCallback(() => {
+      if (isMin) {
+        setValue(max);
+      } else {
+        setValue(min);
+      }
+    }, [isMin, min, max, setValue]);
+
+    return (
+      <Tooltip content={label}>
```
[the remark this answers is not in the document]
jzempel: @Francois-Esquire @lucijanblagonic should we punt this to a future PR fix?

packages/grid/src/elements/pane/components/Splitter.tsx
```diff
@@ -55,6 +56,14 @@ const SplitterComponent = forwardRef<HTMLDivElement, ISplitterProps>(
       ? paneProviderContext.pixelsPerFr[orientationToDimension[orientation!]]
       : 0;
 
+    const value = isRow
+      ? paneProviderContext && paneProviderContext.getRowValue(layoutKey, true)
+      : paneProviderContext && paneProviderContext.getColumnValue(layoutKey, true);
+
+    const valueInFr = isRow
+      ? paneProviderContext && paneProviderContext.getRowValue(layoutKey)
+      : paneProviderContext && paneProviderContext.getColumnValue(layoutKey);
```
Francois-Esquire, line 65: [nitpick] for code readability but non-blocking

```suggestion
    const value = isRow
      ? paneProviderContext!.getRowValue(layoutKey, true)
      : paneProviderContext!.getColumnValue(layoutKey, true);

    const valueInFr = isRow
      ? paneProviderContext!.getRowValue(layoutKey)
      : paneProviderContext!.getColumnValue(layoutKey);
```

packages/grid/src/elements/pane/components/SplitterButton.tsx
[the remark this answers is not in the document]
Francois-Esquire: @jzempel a future PR fix works!

packages/grid/package.json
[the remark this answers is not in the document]
Francois-Esquire: We should be using `"react-popper": "^1x"` to stay consistent with `react-tooltips`. It would remove the need for `@popperjs/core`. Technically `@zendeskgarden/react-tooltips` should be bringing in `react-popper` as a dependency.

packages/grid/src/elements/pane/components/Splitter.tsx
jzempel: These will need to be `?` notation. Updating...

packages/grid/package.json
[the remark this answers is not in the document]
jzempel: Fixed per https://github.com/zendeskgarden/react-components/pull/1371#discussion_r891332366

Francois-Esquire (approved):

packages/grid/src/elements/pane/components/Splitter.tsx
Francois-Esquire: Ahh right good catch, had a dev moment there.

jzempel (approved):

PR #1378, opened 2022-06-28
fix(breadcrumbs,dropdowns,modals,pagination,tabs): improved a11y with upgraded react-containers

jzempel:
## Description

PR contains the following package updates:

- `breadcrumbs` internal only updates based on `@zendeskgarden/container-breadcrumb` v1.0
-  `dropdowns` internal only updates based on `@zendeskgarden/container-selection` v2.0
- `modals` internal only updates based on `@zendeskgarden/container-focusvisible` and `@zendeskgarden/container-modal` v1.0
- `pagination` internal updates based on `@zendeskagarden/container-pagination` v1.0 along with enhanced `useText` for fallback `aria-label` previous, next, and page attributes
-  `tabs` internal only updates based on `@zendeskgarden/container-tabs` v1.0

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
@@ -83,7 +83,7 @@ describe('DrawerModal', () => {
 
     userEvent.click(getByText('Open Drawer'));
 
-    expect(getByText('title')).toHaveAttribute('id', `${DRAWER_MODAL_ID}--title`);
+    expect(getByText('title')).toHaveAttribute('id', `${DRAWER_MODAL_ID}__title`);
   });
 
   it('applies content a11y attributes to Body element', () => {
@@ -91,7 +91,7 @@ describe('DrawerModal', () => {
 
     userEvent.click(getByText('Open Drawer'));
 
-    expect(getByText('body')).toHaveAttribute('id', `${DRAWER_MODAL_ID}--content`);
+    expect(getByText('body')).toHaveAttribute('id', `${DRAWER_MODAL_ID}__content`);
   });
 
   it('closes the drawer modal when user clicks the close modal button', async () => {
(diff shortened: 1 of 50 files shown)


packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -107,8 +111,15 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
         >
-          <StyledBackdrop {...(getBackdropProps({ isAnimated: true, ...backdropProps }) as any)}>
-            <StyledDrawerModal {...modalProps} />
+          <StyledBackdrop
+            isAnimated
+            {...(getBackdropProps(backdropProps) as HTMLAttributes<HTMLDivElement>)}
+          >
+            <StyledDrawerModal
+              {...(getModalProps() as HTMLAttributes<HTMLDivElement>)}
+              {...props}
```
Francois-Esquire, line 120: Do we want user provided props to override what `getModalProps` outputs?

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -83,15 +83,15 @@ describe('DrawerModal', () => {
 
     userEvent.click(getByText('Open Drawer'));
 
-    expect(getByText('title')).toHaveAttribute('id', `${DRAWER_MODAL_ID}--title`);
+    expect(getByText('title')).toHaveAttribute('id', `${DRAWER_MODAL_ID}__title`);
```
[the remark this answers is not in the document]
jzempel: This is expected since `react-containers` was updated (to use the proper BEM notation). I think we can assume subcomponent IDs are under Garden's control and out-of-scope for end user impact.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
jzempel: Yes. This is how we handle other container hook usages. Consider both `Modal` and `TooltipModal` for example.

Francois-Esquire (approved):

PR #1381, opened 2022-07-12
Francois esquire/message validation label

Francois-Esquire:
## Description

This PR adds `validationLabel` prop to the `Message` component which is used for the `aria-label` of the SVG validation icon present in the component.

## Detail

The `validationLabel` is provided to the `useText` hook and uses the current `validation` as the default value if the `validationLabel` is not provided. For the SVG message icon, the `aria-hidden` label is removed so it can be part of the accessibility tree.

Pending until initial review:
- [ ] updating storybook

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/forms/src/types/index.ts
+++ b/packages/forms/src/types/index.ts
@@ -54,6 +54,7 @@ export interface ILabelProps extends LabelHTMLAttributes<HTMLLabelElement> {
 export interface IMessageProps extends HTMLAttributes<HTMLDivElement> {
   /** Applies validation state styling */
   validation?: Validation;
+  validationLabel?: string;
 }
 
 export interface IRadioProps extends InputHTMLAttributes<HTMLInputElement> {
--- a/packages/forms/.size-snapshot.json
+++ b/packages/forms/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 156254,
-    "minified": 107057,
-    "gzipped": 18750
+    "bundled": 156277,
+    "minified": 107076,
+    "gzipped": 18759
   },
   "index.esm.js": {
-    "bundled": 147034,
-    "minified": 98816,
-    "gzipped": 18327,
+    "bundled": 147057,
+    "minified": 98835,
+    "gzipped": 18336,
     "treeshaked": {
       "rollup": {
-        "code": 80718,
+        "code": 80737,
         "import_statements": 745
       },
       "webpack": {
-        "code": 88446
+        "code": 88465
       }
     }
   }
--- a/packages/forms/.size-snapshot.json
+++ b/packages/forms/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 155970,
-    "minified": 106878,
-    "gzipped": 18693
+    "bundled": 156254,
+    "minified": 107057,
+    "gzipped": 18750
   },
   "index.esm.js": {
-    "bundled": 146774,
-    "minified": 98659,
-    "gzipped": 18270,
+    "bundled": 147034,
+    "minified": 98816,
+    "gzipped": 18327,
     "treeshaked": {
       "rollup": {
-        "code": 80573,
-        "import_statements": 732
+        "code": 80718,
+        "import_statements": 745
       },
       "webpack": {
-        "code": 88298
+        "code": 88446
       }
     }
   }
--- a/packages/forms/src/elements/common/Message.spec.tsx
+++ b/packages/forms/src/elements/common/Message.spec.tsx
@@ -82,7 +82,24 @@ describe('Message', () => {
         );
 
         expect(getByText(text).firstChild!.nodeName).toBe('svg');
+        expect(getByText(text).firstChild).toHaveAttribute('aria-label', validation);
       });
+    });
+
+    it('renders SVG child element with a validation label', () => {
+      const validation = VALIDATION[0];
+      const validationLabel = `great ${validation}`;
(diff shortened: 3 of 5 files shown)


jzempel (changes requested):
- every related Storybook demo needs the `validationLabel` prop added under the "Message" section. Includes: Checkbox, Fieldset stories, FileUpload, Input, InputGroup, MultiThumbRange, Radio, Range, Toggle, and maybe others
- the `dropdowns` package needs corresponding support added to its `Message` component along with added demo controls

packages/forms/src/types/index.ts
```diff
@@ -54,6 +54,7 @@ export interface ILabelProps extends LabelHTMLAttributes<HTMLLabelElement> {
 export interface IMessageProps extends HTMLAttributes<HTMLDivElement> {
   /** Applies validation state styling */
   validation?: Validation;
+  validationLabel?: string;
```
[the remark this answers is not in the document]
Francois-Esquire: Updated!

packages/dropdowns/demo/autocomplete.stories.mdx
```diff
@@ -50,6 +50,15 @@ import { AUTOCOMPLETE_ITEMS as ITEMS, DROPDOWN_PLACEMENT as PLACEMENT } from './
...
       message: { name: 'children', table: { category: 'Message' } },
+      validationLabel: {
+        control: { type: 'text' },
+        table: { category: 'Message' }
+      },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Message' }
+      },
```
jzempel, line 61: This should be removed. The `Autocomplete` message validation already responds to the top-level component `validation` prop.

packages/dropdowns/demo/combobox.stories.mdx
```diff
@@ -50,6 +50,15 @@ import { COMBOBOX_ITEMS as ITEMS, DROPDOWN_PLACEMENT as PLACEMENT } from './stor
...
       message: { name: 'children', table: { category: 'Message' } },
+      validationLabel: {
+        control: { type: 'text' },
+        table: { category: 'Message' }
+      },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Message' }
+      },
```
jzempel, line 61: This should be removed. The `Combobox` message validation already responds to the top-level component `validation` prop.

packages/dropdowns/demo/multiselect.stories.mdx
```diff
@@ -53,6 +53,15 @@ import { MULTISELECT_ITEMS as ITEMS, DROPDOWN_PLACEMENT as PLACEMENT } from './s
...
       message: { name: 'children', table: { category: 'Message' } },
+      validationLabel: {
+        control: { type: 'text' },
+        table: { category: 'Message' }
+      },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Message' }
+      },
```
jzempel, line 64: This should be removed. The `Multiselect` message validation already responds to the top-level component `validation` prop.

packages/dropdowns/demo/select.stories.mdx
```diff
@@ -48,6 +48,15 @@ import { SELECT_ITEMS as ITEMS, DROPDOWN_PLACEMENT as PLACEMENT } from './storie
...
       message: { name: 'children', table: { category: 'Message' } },
+      validationLabel: {
+        control: { type: 'text' },
+        table: { category: 'Message' }
+      },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Message' }
+      },
```
jzempel, line 59: This should be removed. The `Select` message validation already responds to the top-level component `validation` prop.

packages/forms/demo/stories/common.tsx
```diff
@@ -0,0 +1,99 @@
...
+  message: { name: 'children', table: { category: 'Message' } },
+  validationLabel: {
+    control: { type: 'text' },
+    table: { category: 'Message' }
+  },
+  validation: {
+    options: ['success', 'warning', 'error'],
+    control: { type: 'radio' },
+    table: { category: 'Message' }
+  }
```
jzempel, line 33: The `validation` arg should not always be moved here. Only in the cases where the top-level form component does _not_ expose a `validation` prop. For example, this "steals" the `validation` prop from the `Input` component.

jzempel (changes requested):
Using the new `validationLabel` controls in storybook does not appear to affect the aria-label for any of the corresponding message SVGs.

packages/forms/demo/stories/common.tsx
Francois-Esquire: Gotcha, I updated this to allow the original `validation` arg to be part of the main story args.

Francois-Esquire:
> Using the new `validationLabel` controls in storybook does not appear to affect the aria-label for any of the corresponding message SVGs.

@jzempel I think this was a mix of not having the `validationLabel` value not passed to the `FieldStory` and the way `useText` works. If the `validationLabel` is set in the URL (refreshing after it was set), the prop should reflect in the `aria-label`.

jzempel:
> If the validationLabel is set in the URL (refreshing after it was set), the prop should reflect in the aria-label.

Heh, could've sworn I had refreshed the URL. I'm seeing it now. Thanks.

jzempel (approved):
Thanks for the abstractions, @Francois-Esquire. Our future selves will thank you.

PR #1386, opened 2022-07-20
fix(dropdowns): prevent enter key from firing callbacks twice

Francois-Esquire:
## Description

This PR has a fix for the `Enter` key firing `downshift.js` callbacks in the `Autocomplete` and `Combobox` components.

## Detail

This PR addresses two instances of when the `Enter` key would fire `downshift` callbacks twice:
- `Autocomplete`: would fire `onStateChange` twice
- `Combobox`: would fire `onSelect` twice

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 91128,
-    "minified": 55497,
-    "gzipped": 11894
+    "bundled": 91402,
+    "minified": 55622,
+    "gzipped": 11915
   },
   "index.esm.js": {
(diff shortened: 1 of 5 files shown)


jzempel (commented):
Are there reasonable tests that can be written to prove things are working as expected?

Francois-Esquire:
@jzempel I've added two tests to test the functionality is behaving as expected.

jzempel (approved):
Cool. Are there guarantees that the consumer-supplied event handler gets called?

Francois-Esquire:
> Cool. Are there guarantees that the consumer-supplied event handler gets called?

@jzempel as much as unit and manual tests can offer, no other guarantees than that at the moment.

PR #1389, opened 2022-07-29
fix(forms): add message id to input aria-describedby attribute

Francois-Esquire:
## Description

This PR adds the `Message` component ID in the `aria-describedby` attribute for a given `Input` (`Select`, `Radio`, etc) component within a `Field`. 

## Detail

There was an update made to `@zendeskgarden/container-field` v2.1.0 which associates the `Message` component ID to the `aria-describedby` attribute of a given `Input` when it is present. The `aria-describedby` attribute is a combination of the `Hint` and `Message` component IDs, however the ID of either of those components will not be added if they are not present as children of the `Field` component.

### Cross browser caveats

In Safari (using VoiceOver), a concatenated list of text node IDs are not all read, only the first item is. However, other browsers behave as expected.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/forms/src/elements/Input.tsx
+++ b/packages/forms/src/elements/Input.tsx
@@ -42,7 +42,7 @@ export const Input = React.forwardRef<HTMLInputElement, IInputProps>(
     }
 
     if (fieldContext) {
-      combinedProps = fieldContext.getInputProps(combinedProps, { isDescribed: true });
+      combinedProps = fieldContext.getInputProps(combinedProps);
     }
 
     return <StyledTextInput {...(combinedProps as any)} />;
--- a/packages/forms/src/elements/common/Field.tsx
+++ b/packages/forms/src/elements/common/Field.tsx
@@ -47,12 +47,12 @@ export const Field = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElem
       }),
       [
         propGetters,
+        getInputProps,
+        getMessageProps,
         isLabelActive,
         isLabelHovered,
         hasHint,
-        hasMessage,
-        getInputProps,
-        getMessageProps
+        hasMessage
       ]
     );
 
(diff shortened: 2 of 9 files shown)


packages/forms/src/elements/Input.tsx
```diff
@@ -42,7 +42,7 @@ export const Input = React.forwardRef<HTMLInputElement, IInputProps>(
     }
 
     if (fieldContext) {
-      combinedProps = fieldContext.getInputProps(combinedProps, { isDescribed: true });
```
Francois-Esquire, line 45: I found that the `isDescribed` option is `true` on multiple input components including `Input`. I found it was safe to remove after `getInputProps` was given a default value for the second function parameter. Are there any concerns with removing?

packages/forms/src/elements/common/Field.tsx
```diff
@@ -24,26 +24,36 @@ export interface IFieldProps extends HTMLAttributes<HTMLDivElement> {
...
         setIsLabelActive,
         isLabelHovered,
         setIsLabelHovered,
-        multiThumbRangeRef,
-        getInputProps: (options: any, describeOptions: any) =>
-          getInputProps(options, { ...describeOptions, isDescribed: hasHint }),
-        setHint: (hintPresent: boolean) => setHasHint(hintPresent),
-        hasHint
+        hasHint,
+        setHasHint,
```
Francois-Esquire, line 43: `setHint` was renamed to `setHasHint` for consistency, the new `setHasMessage` also follows this naming pattern.

packages/forms/src/elements/common/Message.tsx
```diff
@@ -42,8 +54,8 @@ export const Message = React.forwardRef<HTMLDivElement, IMessageProps>(
 
     let combinedProps = { validation, validationLabel, ...props };
 
-    if (fieldContext) {
-      combinedProps = fieldContext.getMessageProps(combinedProps);
+    if (typeof getMessageProps === 'function') {
```
jzempel, line 57: This seems indicative of poor typing within the context, but might not be worth addressing here.

jzempel (approved):
looks good.

packages/forms/src/elements/common/Message.tsx
Francois-Esquire: I've cleaned this up and removed the `typeof` check, I was mistakenly thinking the Field context could be misused (however it is not publicly exported).

PR #1402, opened 2022-08-15
fix(theming): allow use text to be dynamic when updated

Francois-Esquire:
## Description

This PR updates the `useText` hook to allow it to update when either the `value` of the selected `prop` (via `props[name]`) is updated or when the default `text` value changes.

## Details

For dynamic components that usually have changing labels based on user interaction or status change, the `useText` hook cannot be updated unless the component (or hook) using the `useText` hook is re-mounted (removed from the render tree and re-attached to the render tree). This prevents dynamic components that have changing text to update; this change enables the `useText` hook to remain mounted and update when the value or default value is changed.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/theming/.size-snapshot.json
+++ b/packages/theming/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "index.cjs.js": {
-    "bundled": 24395,
-    "minified": 15628,
-    "gzipped": 5689
+    "bundled": 24269,
+    "minified": 15581,
+    "gzipped": 5682
   },
   "index.esm.js": {
-    "bundled": 23496,
-    "minified": 14813,
-    "gzipped": 5558,
+    "bundled": 23385,
+    "minified": 14780,
+    "gzipped": 5552,
     "treeshaked": {
       "rollup": {
         "code": 3408,
--- a/packages/theming/src/utils/useText.spec.tsx
+++ b/packages/theming/src/utils/useText.spec.tsx
@@ -24,6 +24,40 @@ describe('useText()', () => {
     expect(text).toBe('value');
   });
 
+  it('sets default text if prop is not defined and re-renders when updated', () => {
+    let text;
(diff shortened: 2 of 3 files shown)


jgorfine-zendesk (commented):
Hey @Francois-Esquire — I believe your changes are working in the DOM as intended; however, I wanted to flag some behavior I was seeing when I tested this out on staging (via the Message part of the Input component), because I wasn't sure if it, too, was intentional.

After I cleared the `validationMessage`, I got a warning on screen. I was expecting that warning to go away as soon as I entered a new label. As you can see in this recording, though, that was not the case:

https://user-images.githubusercontent.com/93289772/185225612-f0e7ff17-b9ff-488b-b09a-a642dd92f8f6.mov

Once the warning was on screen, it was hard to tell whether or not the validationLabel was updating anymore.

Francois-Esquire:
> Hey @Francois-Esquire — I believe your changes are working in the DOM as intended; however, I wanted to flag some behavior I was seeing when I tested this out on staging (via the Message part of the Input component), because I wasn't sure if it, too, was intentional.
> 
> After I cleared the `validationMessage`, I got a warning on screen. I was expecting that warning to go away as soon as I entered a new label. As you can see in this recording, though, that was not the case:
> 
>  Screen.Recording.2022-08-17.at.3.24.00.PM.mov 
> Once the warning was on screen, it was hard to tell whether or not the validationLabel was updating anymore.

@jgorfine-zendesk Yikes, I'm sorry for the rough experience with this... this is the expected behavior of the `useText` hook (an empty string will throw an error) and how Storybook handles thrown exceptions. I do think there is serious room for improvement for the storybook experience when it comes to props tied to the `useText` hook. A temporary solution would be to refresh once the `validationLabel` is set (in the controls and URL) and it should load up as expected.

jzempel (approved):
lgtm

PR #1403, opened 2022-08-16
feat(avatars): add status label

Francois-Esquire:
## Description

This PR adds a new prop `statusLabel` to the `Avatar` component.

## Detail

The `useText` hook is used as the main source of managing the the status label. There is a default text applied that emphasizes the use case when pronunciation is taking place, which could use a language review.

- an addition to our test utils adds `getByGardenId` and `getAllByGardenId` to make it easier to select nested elements.

## Checklist

      designer as a reviewer)~


jgorfine-zendesk (changes requested):
Hey @Francois-Esquire — I took an initial pass at testing the Avatar status labels with VoiceOver, on Safari, via the staging deployment. 

**tl;dr:** All but the "Active" status labels are being announced as expected, and the AC related to reading order currently isn't being met. 

# Recordings

## Available

https://user-images.githubusercontent.com/93289772/185233099-6725477f-a98b-43f7-a775-099820532b29.mov

## Away

https://user-images.githubusercontent.com/93289772/185233202-913ceb78-f3f3-48dd-99d0-8f6381522b99.mov

## Transfers

https://user-images.githubusercontent.com/93289772/185233478-e7a6e87b-389d-4505-90c5-1def58392ea1.mov

## Offline

https://user-images.githubusercontent.com/93289772/185233546-67f83404-5a65-496c-a76a-ba378aeef3a5.mov

## Active (with badge)

https://user-images.githubusercontent.com/93289772/185233949-db74e627-1677-49f5-bc1d-dd40db075e14.mov

# Notes & observations 

## "Empty group" announcement

When I tested all of the statuses, VoiceOver announced the text label, and then added the words "empty group" at the end. Turns out this is [a known bug](https://webaim.org/discussion/mail_thread?thread=10200) that sometimes arises in Safari, when one has added an `aria-label` to a `<div>`.

The workaround for this is to move the contents of the `aria-label` into a visually-hidden `<span>` inside of the `<div>`:

Screenshot of Safari DevTools demonstrating a visually-hidden span element

After making this change, VoiceOver no longer says anything about an "empty group":

https://user-images.githubusercontent.com/93289772/185235425-fc576f81-424b-4d88-8160-3b4bfbc2340a.mov

## "Active" status hiccups

### Label bug

- Status is being read as "undefined"
- Missing the words "0 notifications"

It's worth noting that I wasn't able to test this status in isolation, like I was with the others:

https://user-images.githubusercontent.com/93289772/185232696-5808c120-50e3-4cd7-8995-1ef76864d718.mov

### CSS bug

When I first applied the badge to the Avatar, I noticed that the number (0) was not centered within the red circle:

https://user-images.githubusercontent.com/93289772/185232164-6143b06f-9b70-4623-b74e-3c1d37ef517f.mov

## Image element missing alt text & figure element missing caption

Per the AC on this ticket, I was expecting to hear the alt text for the Agent's image to be read out before the status. Turns out it's currently missing that alt text:

Screenshot of Safari DevTools demonstrating Avatar image tag with missing alt text

I noticed, also, that the `<figure>` element wrapping both the image and the status is missing a `<figcaption>`, which could provide useful context to the user. 

If we were to add both image alt text and a figure caption, the overall reading experience could sound something like this:

https://user-images.githubusercontent.com/93289772/185235668-a37e4284-92e7-403a-af6f-a5b20e0dfdbc.mov

Francois-Esquire:
@jgorfine-zendesk thank you for the comprehensive review!
- For the "group" being pronounced in Safari, I've added an update to address this.
- For the image alt text, this is an enhancement we can make in Storybook and was done in the most recent update.
- For the active status not being read out, it would be ideal to wait for #1402 to make it in (if it is accepted) to see the changes reflect. In the meantime, you can try toggling "Strict mode" when switching between statuses
Screen Shot 2022-08-19 at 5 02 02 PM

packages/avatars/src/styled/StyledSRSpan.ts
```diff
@@ -0,0 +1,31 @@
...
+  margin: -1px;
+  border: 0;
+  padding: 0;
+  width: 1px;
+  height: 1px;
+  clip: rect(1px, 1px, 1px, 1px);
+  clip-path: inset(50%);
+  overflow: hidden;
+  white-space: nowrap;
+`;
```
Francois-Esquire, line 27: I may have missed if/where we handle an `sr-only` element so I've included this in the meantime.

utils/test/garden-test-utils.tsx
```diff
@@ -37,10 +37,23 @@ RtlProvider.propTypes = {
...
+  const results = render(ui, options);
+
+  return {
+    ...results,
+    getByGardenId: (id: string, container?: HTMLElement) => {
+      return (container || results.container).querySelector(`[data-garden-id="${id}"]`);
+    },
+    getAllByGardenId: (id: string, container?: HTMLElement) => {
+      return (container || results.container).querySelectorAll(`[data-garden-id="${id}"]`);
+    }
```
Francois-Esquire, line 50: The two selectors were helpful in making the tests easier to work with when selecting Garden elements - it is by no means necessary nor is it the appropriate PR to include this, happy to remove from this PR.

jzempel: The preferred mechanism is to add a `data-test-id` which gets stripped during build. Grep for details.

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -64,8 +81,10 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
             foregroundColor={foregroundColor}
             surfaceColor={surfaceColor}
           >
+            <StyledSRSpan>{label}</StyledSRSpan>
```
jzempel, line 84: For consistency, prefer a `data-test-id` here to be used in `.spec`.

jzempel (changes requested):

utils/test/garden-test-utils.tsx
hzhu: Another alternative is to query the element in an [accessible way](https://testing-library.com/docs/queries/about#priority). In areas where `getByGardenId` is used, the `getByText` could be used instead. If a [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) was rendered, then the `getByLabelText` query could be used. For example,

```jsx
it('renders with badge and with a provided status label', () => {
  const { getByText } = render(
    <Avatar badge="2" statusLabel="two notifications">
      <img alt="lorem ipsum" />
    </Avatar>
  );

  const statusText = getByText("two notifications");

  expect(statusText).toBeInTheDocument();
});
```

There's also a bunch other accessible query functions like `getByRole`, `getByAltText`, and `getByTitle` ♿ .

jzempel: Yeah, even better ☝️. Thanks @hzhu.

packages/avatars/src/elements/Avatar.tsx
jzempel: Henry's suggestion is preferable.

packages/avatars/src/styled/StyledSRSpan.ts
jzempel: I have some CodeSandbox evidence that adding `role="img"` along with `aria-label` to the `<div>` surrounding the SVG is going to yield preferable results. And I think it falls better in line with a11y [recommendations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role). Let's pair up on this, if necessary.

packages/avatars/demo/stories/AvatarStory.tsx
```diff
@@ -22,11 +22,13 @@ export const AvatarStory: Story<IArgs> = ({ children, type, ...args }) => (
   <Avatar
     {...args}
     backgroundColor={args.backgroundColor || (type === 'image' ? undefined : PALETTE.kale[800])}
+    // for useText - avoids throwing when first inputting text arg passed as an empty string
+    statusLabel={args.statusLabel || args.status}
```
jzempel, line 26: This is not preferable, because it masks the dev console warning and expected error. Similar to other components, we want to prove these can happen, as expected, in Storybook. Therefore, these lines can be removed.

At this time, Storybook users are expected to know how to click "Refresh controls" or refresh the URL to get back to a valid state.

jzempel (changes requested):

packages/avatars/src/styled/StyledSRSpan.ts
Francois-Esquire: I think this works better for the actual experience. I have updated.

packages/avatars/demo/stories/AvatarStory.tsx
Francois-Esquire: I disagree - the `Avatar` story is not about testing the effects of the `useText` when providing an empty string. That will happen with usage of the component. For non-developers and new users to Garden, this can be a point of confusion.

Francois-Esquire: Per offline conversation, we will remove this in favor of enhancing in the future.

packages/avatars/src/elements/Avatar.tsx
```diff
@@ -42,6 +43,22 @@ const AvatarComponent = forwardRef<HTMLElement, IAvatarProps>(
...
+
+        statusMessage = `active. ${
+          count > 0 ? `${count} notification${count > 1 ? 's' : ''}` : 'no notifications'
+        }`;
+      }
+
+      return `status: ${statusMessage}`;
+    }, [computedStatus, badge]);
+
+    const statusLabel = useText(AvatarComponent, props, 'statusLabel', defaultStatusLabel);
```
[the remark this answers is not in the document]
Francois-Esquire: I updated to make `statusLabel="status"` when no status is active.

jzempel (approved):

PR #1405, opened 2022-09-02
fix(dropdowns): remove nested controls violation from select

Francois-Esquire:
## Description

Removes any focusable descendant elements from being nested within the `Select` component.

## Detail

Prior to this change, a `role="button"` was used for the wrapping container that housed a hidden `input` element, which caused the nested interactive elements violation to be flagged. To remediate, a role of `none` was used in place of `button` to remove any nesting of interactive roles.

An alternative was to add `type="hidden"` to the internal input used within the `Select` component, however this caused the menu items not to be read by the screen reader.

## Checklist

      designer as a reviewer)~

### DIFF
--- a/packages/dropdowns/.size-snapshot.json
+++ b/packages/dropdowns/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 91402,
-    "minified": 55622,
-    "gzipped": 11916
+    "bundled": 91422,
+    "minified": 55634,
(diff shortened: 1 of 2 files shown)


jzempel (approved):
Comparing against https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html, I think the downshift props are generally spreading to the wrong places. However, I think this is ok for getting us by for now until the `Combobox` v.next work is further along.

PR #1408, opened 2022-09-12
fix(dropdowns): prevent scroll when clicking dropdown menu trigger

DerJacques:
## Description

We have identified an edge case in the Dropdown Menu component.

When a Zendesk App is located in the sidebar and the app has a lot of content (content is at least 2x as high as the sidebar iframe), clicking the dropdown trigger will scroll the parent page downwards. This happens because the package positions a hidden input element `fixed` when it should be `absolute`.

The input is defined here: https://github.com/zendeskgarden/react-components/blob/main/packages/dropdowns/src/styled/select/StyledInput.ts#L15 And it gets focus here: https://github.com/zendeskgarden/react-components/blob/main/packages/dropdowns/src/elements/Trigger/Trigger.tsx#L42

In our testing we found that using position `absolute` fixes the issue without introducing any visual regression.

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/dropdowns/src/styled/select/StyledInput.ts
+++ b/packages/dropdowns/src/styled/select/StyledInput.ts
@@ -12,7 +12,7 @@ import { DEFAULT_THEME, retrieveComponentStyles } from '@zendeskgarden/react-the
 const COMPONENT_ID = 'dropdowns.input';
 
 const hiddenStyling = css`
-  position: fixed;
+  position: absolute;
   border: 0;
   clip: rect(1px, 1px, 1px, 1px);
   padding: 0;


jzempel (changes requested):
@DerJacques this conflicts with #990. I would recommend applying custom styling for this case. See https://garden.zendesk.com/components/theme-provider#targeting for details.

DerJacques:
@jzempel Thank you for the quick reply and review!
We did in fact solve it by just overriding the style, but obviously that's not fantastic.

I lack the context to come up with other solutions that would potentially work inside the table component as well, so I'll close this PR for now.

PR #1410, opened 2022-09-14
feat(avatars): add status indicator component

Francois-Esquire:
## Description

Adds the new `StatusIndicator` standalone component.

## TODO:
- [x] add Figma link for `StatusIndicator` design - @james-sann will need your assistance with this!

## Detail

This PR contains a new component `StatusIndicator` that is exported from the `avatars` package. The new component has two props `type` (`"available" | "away" | "transfers" | "offline"`) and `isCompact`.

Available `type`:
Screen Shot 2022-09-14 at 2 58 05 PM

Away `type`:
Screen Shot 2022-09-14 at 2 58 59 PM

Transfers `type`:
Screen Shot 2022-09-14 at 2 44 43 PM

Offline `type`:
Screen Shot 2022-09-14 at 2 59 33 PM

Component frame:
Screen Shot 2022-09-14 at 2 43 41 PM

Usage example:
Screen Shot 2022-09-14 at 2 45 05 PM

## Checklist

      designer as a reviewer)

### DIFF
--- a/packages/avatars/demo/statusindicator.stories.mdx
+++ b/packages/avatars/demo/statusindicator.stories.mdx
@@ -1,0 +1,39 @@
+import { Meta, ArgsTable, Canvas, Story } from '@storybook/addon-docs';
+import { StatusIndicator } from '@zendeskgarden/react-avatars';
+
+import { StatusIndicatorStory } from './stories/StatusIndicatorStory';
+
+<Meta title="Packages/Avatars/StatusIndicator" component={StatusIndicator} />
+
+# API
+
+<ArgsTable />
+
+# Demo
+
+<Canvas>
+  <Story
+    name="StatusIndicator"
+    argTypes={{
+      caption: { name: 'caption', control: 'text', table: { category: 'Story' } },
+      show: {
+        name: 'show',
+        control: 'radio',
+        options: ['default caption', 'set caption', 'no caption'],
+        table: { category: 'Story' }
+      }
+    }}
+    args={{
+      show: 'default caption'
+    }}
+    parameters={{
+      design: {
+        allowFullscreen: true,
+        type: 'figma',
+        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=102%3A110'
+      }
+    }}
+  >
+    {args => <StatusIndicatorStory {...args} />}
+  </Story>
+</Canvas>
--- a/packages/avatars/demo/statusindicator.stories.mdx
+++ b/packages/avatars/demo/statusindicator.stories.mdx
@@ -15,23 +15,10 @@ import { StatusIndicatorStory } from './stories/StatusIndicatorStory';
   <Story
     name="StatusIndicator"
     argTypes={{
-      caption: { name: 'caption', control: 'text', table: { category: 'Story' } },
-      show: {
-        name: 'show',
-        control: 'radio',
-        options: ['default caption', 'set caption', 'no caption'],
-        table: { category: 'Story' }
-      }
+      children: { control: 'text' }
     }}
     args={{
-      show: 'default caption'
-    }}
-    parameters={{
-      design: {
-        allowFullscreen: true,
-        type: 'figma',
-        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=102%3A110'
-      }
+      'aria-label': 'Label'
     }}
   >
     {args => <StatusIndicatorStory {...args} />}
--- a/packages/avatars/demo/statusindicator.stories.mdx
+++ b/packages/avatars/demo/statusindicator.stories.mdx
@@ -20,6 +20,13 @@ import { StatusIndicatorStory } from './stories/StatusIndicatorStory';
     args={{
       'aria-label': 'Label'
     }}
+    parameters={{
+      design: {
+        allowFullscreen: true,
+        type: 'figma',
+        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=10927%3A45275'
+      }
+    }}
   >
     {args => <StatusIndicatorStory {...args} />}
   </Story>
--- a/packages/avatars/demo/statusindicator.stories.mdx
+++ b/packages/avatars/demo/statusindicator.stories.mdx
@@ -18,7 +18,8 @@ import { StatusIndicatorStory } from './stories/StatusIndicatorStory';
       children: { control: 'text' }
     }}
     args={{
-      'aria-label': 'Label'
+      'aria-label': 'Label',
+      children: 'Offline'
     }}
     parameters={{
       design: {
--- a/packages/avatars/demo/statusindicator.stories.mdx
+++ b/packages/avatars/demo/statusindicator.stories.mdx
@@ -19,7 +19,7 @@ import { StatusIndicatorStory } from './stories/StatusIndicatorStory';
     }}
     args={{
       'aria-label': 'Label',
-      children: 'Offline'
+      children: 'Status'
     }}
     parameters={{
       design: {
--- a/packages/avatars/demo/stories/StatusIndicatorStory.tsx
+++ b/packages/avatars/demo/stories/StatusIndicatorStory.tsx
@@ -9,41 +9,6 @@ import React from 'react';
 import { Story } from '@storybook/react';
 import { StatusIndicator, IStatusIndicatorProps } from '@zendeskgarden/react-avatars';
 
-interface IArgs extends IStatusIndicatorProps {
-  caption?: string;
-  show?: string;
-}
-
-function renderCaption(
-  caption: IArgs['caption'],
-  show: IArgs['show'],
-  type: IStatusIndicatorProps['type']
-): string | null {
-  switch (show) {
-    case 'no caption':
-      return null;
-    case 'set caption':
-      return caption || null;
-    case 'default caption':
-    default:
-      switch (type) {
-        case 'away':
-          return 'Away';
-        case 'transfers':
-          return 'Transfers Only';
-        case 'offline':
-          return 'Offline';
-        case 'available':
-        default:
-          return 'Online';
-      }
-  }
-}
-
-export const StatusIndicatorStory: Story<IArgs> = ({ caption, show, type, ...args }) => {
-  return (
-    <StatusIndicator type={type} {...args}>
-      {renderCaption(caption, show, type)}
-    </StatusIndicator>
-  );
+export const StatusIndicatorStory: Story<IStatusIndicatorProps> = ({ ...args }) => {
+  return <StatusIndicator {...args} />;
 };
--- a/packages/avatars/demo/stories/StatusIndicatorStory.tsx
+++ b/packages/avatars/demo/stories/StatusIndicatorStory.tsx
@@ -1,0 +1,49 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { Story } from '@storybook/react';
+import { StatusIndicator, IStatusIndicatorProps } from '@zendeskgarden/react-avatars';
+
+interface IArgs extends IStatusIndicatorProps {
+  caption?: string;
+  show?: string;
+}
+
+function renderCaption(
+  caption: IArgs['caption'],
+  show: IArgs['show'],
+  type: IStatusIndicatorProps['type']
+): string | null {
+  switch (show) {
+    case 'no caption':
+      return null;
+    case 'set caption':
+      return caption || null;
+    case 'default caption':
+    default:
+      switch (type) {
+        case 'away':
+          return 'Away';
+        case 'transfers':
+          return 'Transfers Only';
+        case 'offline':
+          return 'Offline';
+        case 'available':
+        default:
+          return 'Online';
+      }
+  }
+}
+
+export const StatusIndicatorStory: Story<IArgs> = ({ caption, show, type, ...args }) => {
+  return (
+    <StatusIndicator type={type} {...args}>
+      {renderCaption(caption, show, type)}
+    </StatusIndicator>
+  );
+};
--- a/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
+++ b/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
@@ -20,7 +20,7 @@ export const StatusMenuStory: Story = ({ isCompact }) => {
         <Col textAlign="center" alignSelf="center">
           <Dropdown selectedItem={selectedType} onSelect={value => setSelectedType(value)}>
             <Trigger>
-              <Avatar status={selectedType}>
+              <Avatar status={selectedType} size={isCompact ? 'small' : 'large'}>
                 <img alt="Example User" src="images/avatars/chrome.png" />
               </Avatar>
             </Trigger>
--- a/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
+++ b/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
@@ -1,0 +1,57 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { useState } from 'react';
+import { Story } from '@storybook/react';
+import { Col, Grid, Row } from '@zendeskgarden/react-grid';
+import { Dropdown, Trigger, Menu, Item } from '@zendeskgarden/react-dropdowns';
+import { Avatar, IStatusIndicatorProps, StatusIndicator } from '@zendeskgarden/react-avatars';
+
+export const StatusMenuStory: Story = ({ isCompact }) => {
+  const [selectedType, setSelectedType] = useState<IStatusIndicatorProps['type']>('available');
+
+  return (
+    <Grid>
+      <Row style={{ height: 'calc(100vh - 80px)' }}>
+        <Col textAlign="center" alignSelf="center">
+          <Dropdown selectedItem={selectedType} onSelect={value => setSelectedType(value)}>
+            <Trigger>
+              <Avatar status={selectedType}>
+                <img alt="Example User" src="images/avatars/chrome.png" />
+              </Avatar>
+            </Trigger>
+            <Menu isCompact={isCompact}>
+              <Item value="available">
+                <StatusIndicator isCompact={isCompact} type="available">
+                  Online
+                </StatusIndicator>
+              </Item>
+
+              <Item value="transfers">
+                <StatusIndicator isCompact={isCompact} type="transfers">
+                  Transfers only
+                </StatusIndicator>
+              </Item>
+
+              <Item value="away">
+                <StatusIndicator isCompact={isCompact} type="away">
+                  Away
+                </StatusIndicator>
+              </Item>
+
+              <Item value="offline">
+                <StatusIndicator isCompact={isCompact} type="offline">
+                  Offline
+                </StatusIndicator>
+              </Item>
+            </Menu>
+          </Dropdown>
+        </Col>
+      </Row>
+    </Grid>
+  );
+};
--- a/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
+++ b/packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
@@ -20,7 +20,7 @@ export const StatusMenuStory: Story = ({ isCompact }) => {
         <Col textAlign="center" alignSelf="center">
           <Dropdown selectedItem={selectedType} onSelect={value => setSelectedType(value)}>
             <Trigger>
-              <Avatar status={selectedType} size={isCompact ? 'small' : 'large'}>
+              <Avatar status={selectedType} size={isCompact ? 'small' : 'medium'}>
                 <img alt="Example User" src="images/avatars/chrome.png" />
               </Avatar>
             </Trigger>
--- a/packages/avatars/src/elements/StatusIndicator.tsx
+++ b/packages/avatars/src/elements/StatusIndicator.tsx
@@ -46,7 +46,7 @@ const StatusIndicatorComponent = forwardRef<HTMLElement, IStatusIndicatorProps>(
         <StyledStandaloneStatusIndicator
           role="img"
           type={type}
-          isCompact={isCompact}
+          size={isCompact ? 'small' : 'medium'}
           aria-label={ariaLabel}
         >
           {type === 'away' ? <ClockIcon data-icon-status={type} aria-hidden="true" /> : null}
--- a/packages/avatars/src/elements/StatusIndicator.tsx
+++ b/packages/avatars/src/elements/StatusIndicator.tsx
@@ -1,0 +1,69 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { useText } from '@zendeskgarden/react-theming';
+import ClockIcon12 from '@zendeskgarden/svg-icons/src/12/clock-stroke.svg';
+import ClockIcon16 from '@zendeskgarden/svg-icons/src/16/clock-stroke.svg';
+import ArrowLeftIcon12 from '@zendeskgarden/svg-icons/src/12/arrow-left-sm-stroke.svg';
+import ArrowLeftIcon16 from '@zendeskgarden/svg-icons/src/16/arrow-left-sm-stroke.svg';
+
+import { IStatusIndicatorProps, STATUS } from '../types';
+import {
+  StyledStandaloneStatusIndicator,
+  StyledStandaloneStatusCaption,
+  StyledStandaloneStatus
+} from '../styled';
+
+/**
+ * @extends HTMLAttributes<HTMLElement>
+ */
+const StatusIndicatorComponent = forwardRef<HTMLElement, IStatusIndicatorProps>(
+  ({ children, type, isCompact, ...props }, ref) => {
+    let ClockIcon = ClockIcon16;
+    let ArrowLeftIcon = ArrowLeftIcon16;
+
+    if (isCompact) {
+      ClockIcon = ClockIcon12;
+      ArrowLeftIcon = ArrowLeftIcon12;
+    }
+
+    const defaultLabel = useMemo(() => ['status'].concat(type || []).join(': '), [type]);
+    const ariaLabel = useText(StatusIndicatorComponent, props, 'aria-label', defaultLabel);
+
+    return (
+      <StyledStandaloneStatus role="status" ref={ref} {...props}>
+        <StyledStandaloneStatusIndicator
+          role="img"
+          type={type}
+          isCompact={isCompact}
+          aria-label={ariaLabel}
+        >
+          {type === 'away' ? <ClockIcon data-icon-status={type} aria-hidden="true" /> : null}
+          {type === 'transfers' ? (
+            <ArrowLeftIcon data-icon-status={type} aria-hidden="true" />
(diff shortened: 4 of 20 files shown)


packages/avatars/demo/statusindicator.stories.mdx
```diff
@@ -0,0 +1,39 @@
...
+      show: {
+        name: 'show',
+        control: 'radio',
+        options: ['default caption', 'set caption', 'no caption'],
+        table: { category: 'Story' }
+      }
+    }}
+    args={{
+      show: 'default caption'
+    }}
```
jzempel, line 28: Following the convention of other stories:
- the `story` controls can be removed in favor of exposing the `children` prop. Compare with [Tag](https://zendeskgarden.github.io/react-components/?path=/story/packages-tags-tag--tag), [Button](https://zendeskgarden.github.io/react-components/?path=/story/packages-buttons-button--button), and a number of others
- add the `aria-label` override prop. Compare with [IconButton](https://zendeskgarden.github.io/react-components/?path=/story/packages-buttons-iconbutton--icon-button)

packages/avatars/demo/statusindicator.stories.mdx
```diff
@@ -0,0 +1,39 @@
...
+    parameters={{
+      design: {
+        allowFullscreen: true,
+        type: 'figma',
+        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=102%3A110'
+      }
+    }}
```
jzempel, line 35: This addition should be delayed until `StatusIndicator` is on the main Figma branch.

packages/avatars/demo/stories/StatusIndicatorStory.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import React from 'react';
+import { Story } from '@storybook/react';
+import { StatusIndicator, IStatusIndicatorProps } from '@zendeskgarden/react-avatars';
+
+interface IArgs extends IStatusIndicatorProps {
+  caption?: string;
+  show?: string;
+}
```
jzempel, line 15: To repeat, this data should simply be based on `children`. The custom `IArgs` and `renderCaption` are probably not necessary. See https://github.com/zendeskgarden/react-components/blob/main/packages/tags/demo/stories/TagStory.tsx

packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <Grid>
+      <Row style={{ height: 'calc(100vh - 80px)' }}>
+        <Col textAlign="center" alignSelf="center">
+          <Dropdown selectedItem={selectedType} onSelect={value => setSelectedType(value)}>
+            <Trigger>
+              <Avatar status={selectedType}>
+                <img alt="Example User" src="images/avatars/chrome.png" />
+              </Avatar>
```
jzempel, line 25: Similar to the `Menu` pattern under avatars, please alter the `Avatar` size so that there is visual user feedback when the `isCompact` control is toggled.

packages/avatars/src/elements/StatusIndicator.tsx
```diff
@@ -0,0 +1,69 @@
...
+          type={type}
+          isCompact={isCompact}
+          aria-label={ariaLabel}
+        >
+          {type === 'away' ? <ClockIcon data-icon-status={type} aria-hidden="true" /> : null}
+          {type === 'transfers' ? (
+            <ArrowLeftIcon data-icon-status={type} aria-hidden="true" />
+          ) : null}
+        </StyledStandaloneStatusIndicator>
+        {children && <StyledStandaloneStatusCaption>{children}</StyledStandaloneStatusCaption>}
```
jzempel, line 52: This logic properly handles rendering when children are `undefined`, `null`, or `''`. To previous comments, we can just let the story plumbing react to emptying or populating the `children` arg control.

packages/avatars/src/elements/StatusIndicator.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+StatusIndicatorComponent.displayName = 'StatusIndicator';
+
+StatusIndicatorComponent.propTypes = {
+  type: PropTypes.oneOf(STATUS).isRequired,
+  isCompact: PropTypes.bool
+};
+
+StatusIndicatorComponent.defaultProps = {
+  isCompact: false
```
jzempel, line 66: There should be a default specified for type. Please compare with https://github.com/zendeskgarden/react-components/blob/aeb2f8688ddeebca0bbfdd46ec234fdb101970e4/packages/buttons/src/elements/Button.tsx#L65 and other similar Garden prop constructs.

packages/avatars/src/styled/StyledStandaloneStatus.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const COMPONENT_ID = 'avatars.status-indicator';
+
+export const StyledStandaloneStatus = styled.figure.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  box-sizing: content-box;
+  display: inline-flex;
+  flex-flow: row wrap;
```
jzempel, line 19: The flex layout here is not to design spec. The indicator should remain in its own column when the text has to wrap.

Screen Shot 2022-09-15 at 8 55 10 AM

packages/avatars/src/styled/StyledStandaloneStatusCaption.ts
```diff
@@ -0,0 +1,36 @@
...
+    font-weight: ${props.theme.fontWeights.regular};
+  `;
+}
+
+export const StyledStandaloneStatusCaption = styled.figcaption.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  display: inline-block;
+  box-sizing: inherit;
```
jzempel, line 27: Why are either of these necessary? As I see it, the caption should be participating in a flex layout and `box-sizing` doesn't appear to be relevant.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  return css`
+    border-color: ${borderColor};
+    background-color: ${backgroundColor};
+    color: ${props.theme.palette.white};
+  `;
+};
+
+export const StyledStandaloneStatusIndicator = styled.span.attrs({
```
jzempel, line 102: Was consideration given to reusing or extending the existing `StyledStatusIndicator`? This feels like a considerable duplication of styling code.

packages/avatars/src/styled/StyledStatusIndicator.ts
```diff
@@ -17,7 +17,7 @@ export interface IStatusIndicatorProps extends Omit<IAvatarProps, 'badge' | 'isS
   borderColor?: string;
 }
 
-const COMPONENT_ID = 'avatars.status_indicator';
+const COMPONENT_ID = 'avatars.avatar.status-indicator';
```
jzempel, line 20: [nit] we try to avoid component ID renames, when possible, as it can mess with existing [targeting](https://garden.zendesk.com/components/theme-provider#targeting) customizations.

jzempel (changes requested):

packages/avatars/demo/statusindicator.stories.mdx
Francois-Esquire: Yes this is a placeholder - please see the PR description under `TODO`. I can remove it for now.

packages/avatars/src/elements/StatusIndicator.tsx
Francois-Esquire: If `type` is `required`, would it need a default value here?

jzempel: It wouldn't need to, but as a rule Garden provides a reasonable default. Let's do the same here.

jzempel: I suppose https://zendeskgarden.github.io/react-components/?path=/docs/packages-notifications-alert--alert could be a counter-argument. So let's go with whatever you think is best, @Francois-Esquire.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: The thought was there, and considered making a base status indicatorWell there are some differences (mostly in the box-shadow from visual standpoint). I will look into this further.

jzempel: Difference are fine (and expected). But it feels like there should be a significant common abstraction.

packages/avatars/src/elements/StatusIndicator.tsx
Francois-Esquire: Sure - @james-sann any thoughts on default status type? `("available" | "away" | "transfers" | "offline")`

abbelco: Agree with @jzempel on offline as default

abbelco: @Francois-Esquire moving error convo over here... I asked what happens when a browser is open and our network is done, or we can't get the data of the agent status. Do we ever see an error loading status? Or whats our fallback here?

packages/avatars/src/styled/StyledStandaloneStatus.ts
Francois-Esquire: Updated this in [ad377d7](https://github.com/zendeskgarden/react-components/pull/1410/commits/ad377d717954dc8b815886e55db728987238edad).

packages/avatars/src/elements/StatusIndicator.tsx
Francois-Esquire: @abbelco typically in this case it would be up to the consumer to handle this. For the `StatusIndicator` component it will default to the `offline` status that we agreed to set as the default.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: I gave this a try with a base component and it did raise the complexity of the different indicators, which included multiple overrides and recalculations that needed to happen regardless. I will give this more thought and see if we can abstract the styling a bit more. For the time being I was able to DRY out the SVG portion of the styling into its own utility for injection into both the styled status components.

Francois-Esquire: If we did want to reuse the existing `StyledStatusIndicator`, it would be worth trying to do and manually setting the size, as well as overriding the `box-shadow`.

packages/avatars/src/elements/StatusIndicator.tsx
abbelco: Cool thanks for explaining!

james-sann: @Francois-Esquire Agree with the others. "Offline" to be the default

james-sann (approved):
After overlooking all the feedback from JZ, nothing further on my end @Francois-Esquire! Looks great! 

FYI @Francois-Esquire — The Figma component is now in the main Figma branch to reference.

packages/avatars/demo/statusindicator.stories.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Canvas>
+  <Story
+    name="StatusIndicator"
+    argTypes={{
+      children: { control: 'text' }
+    }}
+    args={{
+      'aria-label': 'Label',
+      children: 'Offline'
```
jzempel, line 22: [nit]

```suggestion
      children: 'Status'
```

...to avoid potential confusion that the text is intrinsically tied to the `type`.

packages/avatars/demo/~patterns/stories/StatusMenuStory.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const StatusMenuStory: Story = ({ isCompact }) => {
+  const [selectedType, setSelectedType] = useState<IStatusIndicatorProps['type']>('available');
+
+  return (
+    <Grid>
+      <Row style={{ height: 'calc(100vh - 80px)' }}>
+        <Col textAlign="center" alignSelf="center">
+          <Dropdown selectedItem={selectedType} onSelect={value => setSelectedType(value)}>
+            <Trigger>
+              <Avatar status={selectedType} size={isCompact ? 'small' : 'large'}>
```
jzempel, line 23: [nit]

```suggestion
              <Avatar status={selectedType} size={isCompact ? 'small' : 'medium'}>
```

...so that the system stays intact.

packages/avatars/src/elements/StatusIndicator.tsx
```diff
@@ -0,0 +1,70 @@
...
+    if (isCompact) {
+      ClockIcon = ClockIcon12;
+      ArrowLeftIcon = ArrowLeftIcon12;
+    }
+
+    const defaultLabel = useMemo(() => ['status'].concat(type || []).join(': '), [type]);
+    const ariaLabel = useText(StatusIndicatorComponent, props, 'aria-label', defaultLabel);
+
+    return (
+      <StyledStandaloneStatus role="status" ref={ref} {...props}>
```
jzempel, line 40: Per RFC, please de-structure `aria-label` out of the props so that it is not applied to the `<figure>`.

packages/avatars/src/styled/StyledStandaloneStatusCaption.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import styled, { ThemeProps, DefaultTheme, css } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'avatars.status-indicator.caption';
+
+function sizeStyles(props: ThemeProps<DefaultTheme>) {
+  return css`
+    padding: 0 ${props.theme.space.base}px;
+    line-height: ${props.theme.space.base * 6}px;
```
jzempel, line 16: The `line-height` should be the standard 20px. Please use [getLineHeight](https://garden.zendesk.com/components/utilities#getlineheight) for Garden's unitless preference. Please see Figma for reference.

packages/avatars/src/styled/StyledStandaloneStatusCaption.ts
```diff
@@ -0,0 +1,33 @@
...
+    font-size: ${props.theme.fontSizes.md};
+    font-weight: ${props.theme.fontWeights.regular};
```
jzempel, line 18: It shouldn't be necessary to specify `font-weight` here; accepting the cascade would be desirable in most cases.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  if (props.isCompact) {
+    marginTop = props.theme.space.base * 1.5;
+    size = math(`${props.theme.space.base * 3}px - (${borderWidth} * 2)`);
+  } else {
+    size = math(`${props.theme.space.base * 4}px - (${borderWidth} * 2)`);
+  }
+
+  return css`
+    margin: ${props.theme.space.base}px;
```
jzempel, line 33: What is the impetus for margining the image? We don't see this in the Figma designs. Either the code should match the designs, or the designs should be pushed to explore inline `StatusIndicator` layouts similar to https://garden.zendesk.com/components/span#icons.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,92 @@
...
+    margin-top: ${marginTop}px;
+    border: ${borderWidth} ${props.theme.borderStyles.solid};
+    border-radius: ${size};
+    width: ${size};
+    min-width: ${size};
+    max-width: calc(2em + (${borderWidth} * 3));
+    height: ${size};
+    min-height: ${size};
+    box-sizing: inherit;
+    overflow: hidden;
```
jzempel, line 43: What purpose do these CSS properties serve?

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return css`
+    margin: ${props.theme.space.base}px;
+    margin-top: ${marginTop}px;
+    border: ${borderWidth} ${props.theme.borderStyles.solid};
+    border-radius: ${size};
+    width: ${size};
+    min-width: ${size};
+    max-width: calc(2em + (${borderWidth} * 3));
```
jzempel, line 39: I'm concerned that the use of `em` units here is brittle with respect to the system + theming.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  ${sizeStyles}
+  ${colorStyles}
+
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledStandaloneStatusIndicator.defaultProps = {
+  type: 'offline',
+  isCompact: false,
```
jzempel, line 90: Never explicitly default boolean props. Garden prefers `undefined`.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  let size = '0';
+  let marginTop = props.theme.space.base;
+
+  if (props.isCompact) {
+    marginTop = props.theme.space.base * 1.5;
+    size = math(`${props.theme.space.base * 3}px - (${borderWidth} * 2)`);
+  } else {
+    size = math(`${props.theme.space.base * 4}px - (${borderWidth} * 2)`);
+  }
```
jzempel, line 30: Are we setting up precision layout magic that might better be handled via flex?

jzempel (changes requested):

packages/avatars/src/elements/StatusIndicator.tsx
```diff
@@ -0,0 +1,70 @@
...
+          role="img"
+          type={type}
+          isCompact={isCompact}
+          aria-label={ariaLabel}
+        >
+          {type === 'away' ? <ClockIcon data-icon-status={type} aria-hidden="true" /> : null}
+          {type === 'transfers' ? (
+            <ArrowLeftIcon data-icon-status={type} aria-hidden="true" />
+          ) : null}
+        </StyledStandaloneStatusIndicator>
```
jzempel, line 51:
```suggestion
        <StyledStatusIndicator
          role="img"
          status={type}
          aria-label={ariaLabel}
          size={isCompact ? 'small' : 'medium'}
        >
          {type === 'away' ? <ClockIcon data-icon-status={type} aria-hidden="true" /> : null}
          {type === 'transfers' ? (
            <ArrowLeftIcon data-icon-status={type} aria-hidden="true" />
          ) : null}
        </StyledStatusIndicator>
```

If you try this suggestion, I think you'll see that the new `StyledStandaloneStatusIndicator` is almost entirely unnecessary – with one exception. It should only add a `margin-top` property. The pseudo-code for `margin-top` is:

`margin-top: calc((${lineHeight} - ${iconSize}) / 2);`

packages/avatars/src/styled/StyledStandaloneStatusCaption.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import styled, { ThemeProps, DefaultTheme, css } from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'avatars.status-indicator.caption';
+
+function sizeStyles(props: ThemeProps<DefaultTheme>) {
+  return css`
+    padding: 0 ${props.theme.space.base}px;
```
jzempel, line 15: Following the previous suggestion, the caption should one-way margin (not padding) the entire 8px depending on RTL. Please see the RFC codesandbox for basic layout details.

jzempel (changes requested):

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: None in particular, I will update.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: This was inherited from the initial implementation, and actually serve no purpose here. Will remove.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: None that I've spotted, will remove.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: I agree it's redundant in this case, I will remove - FYI, there are currently a few examples where default props are used in the codebase - one example `DatepickerRange` ([here](https://github.com/zendeskgarden/react-components/blob/main/packages/datepickers/src/elements/DatepickerRange/DatepickerRange.tsx#L140))

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: This would require an additional element when considering how the indicator will line up when text is overflowing. Currently the containing parent `figure` element does use `inline-flex` which lays out the indicator and caption.

packages/avatars/src/elements/StatusIndicator.tsx
Francois-Esquire: Unfortunately, `box-shadow` is another visual property that is noticeable when the background does not match the color of the  `box-shadow`. Where do you propose this line (`margin-top: calc((${lineHeight} - ${iconSize}) / 2)`) to exist?

packages/avatars/src/styled/StyledStandaloneStatusCaption.ts
Francois-Esquire: I added the RTL portion to a TODO that I will look at. The disadvantage of this approach is that flex containers will flow correctly with direction without the need to direct the margin based on RTL. Even spacing allows elements to flow naturally IMO.

packages/avatars/src/elements/StatusIndicator.tsx
jzempel: The point of the suggestion is to show how close the existing styling gets us. I think the definition of `StyledStandaloneStatusIndicator` should look something like this:

```ts
export const StyledStandaloneStatusIndicator = styled(StyledStatusIndicator).attrs({ ... })`
  margin-top: calc((${lineHeight} - ${iconSize}) / 2);
  box-shadow: none;

  ${props => retrieveStyledComponents(...)};
`;
```

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
```diff
@@ -0,0 +1,29 @@
...
+import { StyledStatusIndicatorBase } from './StyledStatusIndicatorBase';
+
+const COMPONENT_ID = 'avatars.status-indicator.indicator';
+
+export const StyledStandaloneStatusIndicator = styled(StyledStatusIndicatorBase).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledStatusIndicatorProps>`
+  position: relative;
+  margin-top: ${props => `${props.theme.space.base / (props.size === 'small' ? 1 : 2)}px`};
```
jzempel, line 21:
```suggestion
  margin-top: ${props => `calc((${props.theme.lineHeights.md} - ${props.theme.fontSizes.md}) / 2)`};
```

as this would respond more accurately to theming modifications. Another option would be to export `lineHeight` and `fontSize` values from `StyledStandaloneStatusCaption` and use them here. Perhaps another would be to move the return values out to a local `utility` function and use them in both places. 🤷

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
```diff
@@ -0,0 +1,104 @@
...
+  TRANSITION_DURATION,
+  getStatusColor,
+  getStatusBorderOffset,
+  getStatusSize,
+  IStyledStatusIndicatorProps
+} from './utility';
+
+const COMPONENT_ID = 'avatars.status-indicator.base';
+
+const [offline] = [...STATUS].reverse();
```
jzempel, line 22: This feels brittle. I think you're better off just using the `'offline'` string below. TypeScript will keep things in check.

packages/avatars/src/styled/utility.ts
```diff
@@ -6,16 +6,26 @@
...
+import { ThemeProps, DefaultTheme } from 'styled-components';
+import { math } from 'polished';
 
-import { STATUS } from '../types';
+import { STATUS, SIZE } from '../types';
 
+export const TRANSITION_DURATION = 0.25;
+
+const [xxs, xs, s, m, l] = SIZE;
 const [active, available, away, transfers, offline] = ['active', ...STATUS];
```
jzempel, line 17: These also feels brittle (and a bit too slick). Garden tends to get leverage out of TS by simply re-using the explicit union'd string values throughout.

packages/avatars/src/styled/utility.ts
```diff
@@ -6,16 +6,26 @@
...
 
+export interface IStyledStatusIndicatorProps extends ThemeProps<DefaultTheme> {
+  readonly size?: typeof SIZE[number];
+  readonly type?: typeof STATUS[number] | 'active';
```
jzempel, line 21:
```suggestion
  readonly size?: IAvatarProps['size'];
  readonly type?: IAvatarProps['status'] | 'active';
```

packages/avatars/src/styled/utility.ts
```diff
@@ -6,16 +6,26 @@
...
+}
+
 export function getStatusColor(
-  status: typeof STATUS[number] | 'active' | undefined,
-  theme = DEFAULT_THEME
+  type?: IStyledStatusIndicatorProps['type'],
+  theme?: typeof DEFAULT_THEME
```
jzempel, line 26:
```suggestion
  theme?: ThemeProps<DefaultTheme>
```

jzempel (changes requested):

packages/avatars/src/styled/utility.ts
Francois-Esquire: I think `ThemeProps<DefaultTheme>` would work here if `{ theme}` was a property on an object (AKA props). I replaced this with `IStyledStatusIndicatorProps['theme']`.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: Having a shared utility here does make more sense, thanks for spotting this. I will update this shortly.

packages/avatars/src/styled/utility.ts
Francois-Esquire: Good point about leveraging TS, updated in [37cf3bd](https://github.com/zendeskgarden/react-components/pull/1410/commits/37cf3bdeb0131ecf1f0a2da4ae655082a682133b).

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
Francois-Esquire: After a bit of trial, I found the unitless return value of `getLineHeight` didn't work well in translating over to pixels for `margin-top` calculations - . As for exporting `lineHeight` and `fontSize` values, those are based on the theme and are calculated at runtime (as opposed to static values), unless we use `props.theme.lineHeights.md` directly.

What I found did work was `calc((${props.theme.lineHeights.md} - ${getStatusSize(props, '0')}) / 2)`.

packages/avatars/src/styled/StyledStandaloneStatusIndicator.ts
jzempel: Any `false` boolean defaults should be removed (later) for consistency.

jzempel (approved):

PR #1415, opened 2022-10-03
fix(forms): add min-width to tile

Francois-Esquire:
## Description

Constrains the minimum width of the `Tile` component as a fix to prevent content from flowing out of the `Tile` when `isCentered` is `false`.

**New sizing:**
Screen Shot 2022-09-27 at 4 57 44 PM
Screen Shot 2022-09-27 at 4 57 51 PM

## Checklist

- [ ] :wheelchair: ~tested for [WCAG 2.1](https://www.w3.org/TR/WCAG21) AA compliance~


packages/forms/src/styled/tiles/StyledTile.ts
```diff
@@ -153,6 +153,7 @@ export const StyledTile = styled.label.attrs<IStyledTileProps>(props => ({
   cursor: ${props => !props.isDisabled && 'pointer'};
   padding: ${props => props.theme.space.base * 5}px;
   direction: ${props => props.theme.rtl && 'rtl'};
+  min-width: ${props => props.theme.space.base * 24}px;
```
jzempel, line 156: At these larger dimensions, the value of referencing `theme.space.base` becomes negligible. For example, we don't want to see this up to `120px` if a user overrides with base-5. In most other places, we just express the value explicitly.

Francois-Esquire: For sure, will update this with an explicit value.

jzempel (approved):

steue (approved):
thanks for these changes! lgtm! look forward to followup enhancement.

PR #1419, opened 2022-10-07
chore(deps): update testing library user events to v14

Francois-Esquire:
## Description

This PR updates `@testing-library/user-events` to v14.

## Details
Some notable differences in the migration.
- With any usage of `userEvent` in a test file, a setup function call was made `const user = userEvents.setup()`, where `user` was used in place of `userEvent`
- User events now run `async`
- For tests using fake timers, `userEvent` needed to be setup with `userEvents.setup({ delay: null })`
- Some tests were not responding as expected and the use of `fireEvent` did take their place (`datepickers` and `dropdowns`).

## Checklist

- [ ] :wheelchair: ~tested for [WCAG 2.1](https://www.w3.org/TR/WCAG21) AA compliance~

### DIFF
--- a/packages/dropdowns/src/elements/Combobox/Combobox.spec.tsx
+++ b/packages/dropdowns/src/elements/Combobox/Combobox.spec.tsx
@@ -28,28 +28,30 @@ const ExampleCombobox = () => (
 );
 
 describe('Combobox', () => {
+  const user = userEvent.setup();
+
   it('passes ref to underlying DOM element', () => {
     const { getByTestId } = render(<ExampleCombobox />);
 
     expect(getByTestId('combobox')).toBe(ref.current);
   });
 
-  it('applies correct styling on label hover', () => {
+  it('applies correct styling on label hover', async () => {
     const { getByTestId } = render(<ExampleCombobox />);
     const label = getByTestId('label');
     const combobox = getByTestId('combobox');
 
-    userEvent.hover(label);
+    await user.hover(label);
 
     expect(combobox).toHaveStyleRule('border-color', PALETTE.blue[600], { modifier: ':hover' });
   });
 
-  it('focuses input on label click', () => {
+  it('focuses input on label click', async () => {
     const { getByTestId, getByRole } = render(<ExampleCombobox />);
     const label = getByTestId('label');
     const combobox = getByRole('combobox');
 
-    userEvent.click(label);
+    await user.click(label);
 
     expect(combobox).toHaveFocus();
   });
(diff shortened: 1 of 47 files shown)


packages/dropdowns/src/elements/Combobox/Combobox.spec.tsx
```diff
@@ -215,9 +219,10 @@ describe('Combobox', () => {
         </Dropdown>
       );
       const combobox = getByTestId('combobox');
+      const input = combobox.querySelector('input')!;
 
-      await user.type(combobox, '{arrowup}');
-      await user.type(combobox, '{enter}');
+      await user.type(input, '{arrowup}');
+      await user.type(input, '{enter}');
```
Francois-Esquire, line 225: This test case has been questionable. I was only able to get it passing when targeting the input in the `Combobox` and the only test that has been "altered". When I tested manually, this behavior didn't behave as expected - nothing happened. Is there anything historical to note with this test case?

packages/notifications/src/elements/toaster/ToastSlot.spec.tsx
```diff
@@ -35,7 +35,11 @@ describe('ToastSlot', () => {
   it('pauses toast timers when placement is hovered', async () => {
     const { getByRole, getAllByText } = render(<ToastExample />);
 
-    await Promise.all(new Array(5).fill(null).map(() => user.click(getByRole('button'))));
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    for (const _ of new Array(5).fill(null)) {
+      // eslint-disable-next-line no-await-in-loop
+      await user.click(getByRole('button'));
+    }
```
Francois-Esquire, line 42: The `for of` loop was favorable here to keep the async calls originally done like so:
```
await Promise.all(new Array(5).fill(null).map(() => user.click(getByRole('button'))));
```
as it was throwing a warning on what appeared to be stacking `act` calls being done internally by the `user.click` API.

packages/notifications/src/elements/toaster/ToastProvider.spec.tsx
```diff
@@ -34,33 +34,37 @@ const ToastExample: React.FC<IToastProviderProps> = props => {
...
+  it('renders visible toasts up to provided limit', async () => {
     const { getByRole, getAllByText } = render(<ToastExample />);
 
-    for (let x = 0; x < 10; x++) {
-      userEvent.click(getByRole('button'));
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    for (const _ of new Array(10).fill(null)) {
+      // eslint-disable-next-line no-await-in-loop
+      await user.click(getByRole('button'));
     }
```
jzempel, line 68: Should this be refactored as shown by https://eslint.org/docs/latest/rules/no-await-in-loop?

packages/notifications/src/elements/toaster/ToastSlot.spec.tsx
jzempel: Can this be refactored as an internal `await` function as shown by https://eslint.org/docs/latest/rules/no-await-in-loop?

jzempel (commented):
🎗️ `userEvent` added to https://github.com/zendeskgarden/react-components/blob/30a94c50e91fe1486326605417624daca1f89ae0/packages/grid/src/elements/pane/components/Splitter.spec.tsx

packages/notifications/src/elements/toaster/ToastSlot.spec.tsx
Francois-Esquire: Having a classic for loop as it's shown above is equivalent to:

```
await Promise.all(new Array(5).fill(null).map(() => user.click(getByRole('button'))));
```

Since this mixes a sync invocation to an async operation because all the async operations are running simultaneously. This doesn't guarantee sequence nor waiting for atomic completion per invocation before the next call is started synchronously - this caused warnings to show up. An alternative to `for of` loop that would wait per call is to:

```
await user.click(getByRole('button'));
await user.click(getByRole('button'));
await user.click(getByRole('button'));
await user.click(getByRole('button'));
await user.click(getByRole('button'));
```

jzempel (changes requested):
https://github.com/zendeskgarden/react-components/blob/main/packages/grid/src/elements/pane/components/Splitter.spec.tsx needs to be updated, right?

Francois-Esquire:
> https://github.com/zendeskgarden/react-components/blob/main/packages/grid/src/elements/pane/components/Splitter.spec.tsx needs to be updated, right?

@jzempel updated in [953a723](https://github.com/zendeskgarden/react-components/pull/1419/commits/953a723fc3ea8731a6b108f440a1a7bd3d834da2)!

jzempel (approved):

PR #1443, opened 2022-10-12
feat(progress): add useText helper with default value for ARIA label

jgorfine-zendesk:
## Description

Following the convention set forth in the Status Indicator component, this PR adds the `useText` helper to the Progress component. This will both provide a default text label for the component and warn devs if a custom text label hasn't been provided.

This PR relates to [`website` PR #429](https://github.com/zendeskgarden/website/pull/429). 

## Detail

Screenshot of Progress element with "Progress" ARIA label, verified in Chrome Accessibility tools

## Checklist

### DIFF
--- a/packages/loaders/.size-snapshot.json
+++ b/packages/loaders/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 20873,
-    "minified": 16097,
-    "gzipped": 4228
+    "bundled": 21036,
+    "minified": 16201,
+    "gzipped": 4252
   },
   "index.esm.js": {
-    "bundled": 19085,
-    "minified": 14508,
-    "gzipped": 4079,
+    "bundled": 19235,
+    "minified": 14599,
+    "gzipped": 4107,
     "treeshaked": {
       "rollup": {
-        "code": 11953,
+        "code": 12031,
         "import_statements": 432
       },
       "webpack": {
-        "code": 13947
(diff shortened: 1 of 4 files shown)


jzempel (changes requested):
Please add an `aria-label` control to the demo, set to "Label", just like we see with https://zendeskgarden.github.io/react-components/?path=/story/packages-avatars-statusindicator--status-indicator. This allows us to observe component behavior in the demo.

The `args` update can be seen [here](https://github.com/zendeskgarden/react-components/blob/main/packages/avatars/demo/statusindicator.stories.mdx).

jgorfine-zendesk:
Hey @jzempel, I could use a sanity check. Just added the `'aria-label': 'Label'` line to the Progress component story; and, yet, it appears the custom label isn't making its way into the component:

Screenshot of Chrome DevTools warning us the Progress component doesn't have a custom label

At least we know `useText` is working as intended? 😅 

I feel like I'm missing something obvious here, but I'm not sure what.

jzempel:
@jgorfine-zendesk `args={{ value: 0 }}` is already declared 2 lines above. Add to that declaration like so:

```
args={{ value: 0, 'aria-label': 'Label' }}
```

jgorfine-zendesk:
Oh! Duh. 

> I feel like I'm missing something obvious here, but I'm not sure what.

Yup, that would be it. 😆 Thanks, @jzempel.

jzempel (approved):

PR #1446, opened 2022-10-19
feat(modals): add tag prop to Modal, Drawer, and Tooltip Modal components

jgorfine-zendesk:
## Description

Following the precedence set in other components (e.g., Ellipsis), this PR adds the `tag` prop to the Modal Header component. This will enable consumers to customize the underlying HTML element, with the intention being that they pass in a heading (h1-h2) tag. This would help us meet WCAG 2.1 [Success Criterion 1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships) (level A), as well as [Success Criterion 2.4.2 Page Titled](https://www.w3.org/TR/WCAG21/#page-titled) (level A).

~Separate PRs for the Drawer and Tooltip Modal to follow.~

This PR covers the Modal, Drawer, and Tooltip Modal.

### Modal

Screenshot of Chrome DevTools demonstrating the Modal component with an h2 title element

### Drawer

Screenshot of Chrome DevTools demonstrating the Drawer component with an h2 title element

### Tooltip Modal

Screenshot of Chrome DevTools demonstrating the Tooltip Modal component with an h2 title element

## Checklist

### DIFF
--- a/packages/modals/src/types/index.ts
+++ b/packages/modals/src/types/index.ts
@@ -68,6 +68,10 @@ export interface IHeaderProps extends HTMLAttributes<HTMLDivElement> {
    * Applies danger styling
    */
   isDanger?: boolean;
+  /**
+   * Updates the element's HTML tag
+   */
+  tag?: any;
 }
 
 export interface IDrawerModalProps
--- a/packages/modals/src/types/index.ts
+++ b/packages/modals/src/types/index.ts
@@ -105,3 +105,10 @@ export interface ITooltipModalProps
    */
   zIndex?: number;
 }
+
+export interface ITitleProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Updates the element's HTML tag
+   */
+  tag?: any;
+}
--- a/packages/modals/src/types/index.ts
+++ b/packages/modals/src/types/index.ts
@@ -82,6 +82,13 @@ export interface IDrawerModalProps
   isOpen?: boolean;
 }
 
+export interface IDrawerModalHeaderProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Updates the element's HTML tag
+   */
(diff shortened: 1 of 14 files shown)


jzempel (changes requested):
This is looking good @jgorfine-zendesk. re:

> Separate PRs for the Drawer and Tooltip Modal to follow.

Please also wrap `DrawerModal` and `TooltipModal` into this PR as a single unit of work. :pray:

jgorfine-zendesk:
@jzempel can do!

jgorfine-zendesk:
Hey @jzempel, just pushed the code for the Tooltip Modal and Drawer. Do we also need to include the Sheet in this work?

packages/modals/src/types/index.ts
```diff
@@ -101,3 +112,10 @@ export interface ITooltipModalProps
    */
   zIndex?: number;
 }
+
+export interface ITooltipModalTitleProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Updates the element's HTML tag
+   */
+  tag?: any;
```
jgorfine-zendesk, line 120: Do we have a mechanism for sharing types across components? If so, I'm wondering if we want to consider centralizing the `tag` interface and exporting/extending it to these components, the typography components, and any other components that may want to use this in the future.

jzempel: There is (see various forms under https://github.com/zendeskgarden/react-components/blob/main/packages/forms/src/types/index.ts), but not across packages. And when dealing with primitives, often the TypeScript juice isn't worth the squeeze.

jzempel (approved):

jzempel:
> Do we also need to include the Sheet in this work?

Let's hold off on `Sheet` and limit this PR to the modals package.

PR #1448, opened 2022-10-26
feat(anchor): add default alternative text to external anchor icon

jgorfine-zendesk:
## Description

This PR adds the `useText` helper to the `<StyledExternalIcon>` component so that we can provide default alternative text to the SVG, and also warn consumers about providing custom/translated text themselves. 

This alternative text will communicate the external link's purpose as an external link to people who rely on assistive tech.

### Before

Screenshot of Chrome DevTools showing demo Anchor hidden from accessibility tree

### After

Screenshot of Chrome DevTools showing demo Anchor in accessibility tree, with labelled icon

## Checklist

- [] :ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)


packages/buttons/demo/anchor.stories.mdx
```diff
@@ -11,7 +11,14 @@ import { Anchor } from '@zendeskgarden/react-buttons';
...
   <Story
-    args={{ children: 'Text' }}
+    args={{ children: 'Text', href: '#', externalIconTextLabel: 'Label' }}
+    argTypes={{
+      href: {
+        table: {
+          disable: true
+        }
+      }
+    }}
```
jzempel, line 21: By convention, we limit this to Story controls (only `disable` knocking-out when necessary). The `href` can be removed here and applied directly within the story below as `<Anchor href="#" {...args} />`

packages/buttons/src/types/index.ts
```diff
@@ -62,6 +62,11 @@ export interface IAnchorProps
    * safe [cross-origin destination link](https://web.dev/external-anchors-use-rel-noopener/).
    **/
   isExternal?: boolean;
+  /**
+   * Allows a customized/translated text label to be passed to the external link icon,
+   * making that icon accessible to assistive technology
+   **/
+  externalIconTextLabel?: string;
```
jzempel, line 69:
```suggestion
  externalIconLabel?: string;
```

Following other components (`<Avatar statusLabel="...">` for example), let's drop the redundant "Text" from the prop name.

packages/buttons/src/styled/StyledExternalIcon.ts
```diff
@@ -7,17 +7,25 @@
...
+  'data-garden-version': PACKAGE_VERSION,
+  role: 'img',
+  'aria-hidden': undefined,
+  'aria-label': useText(
+    StyledExternalIcon,
+    { 'aria-label': props['aria-label'] },
+    'aria-label',
+    '(opens in a new tab)'
+  )
+}))`
```
jzempel, line 28: Ideally, `useText` etc. should be applied at the element level within `Anchor.tsx`. The linked issue references a construct for handling conditionality within the hook call. I don't think we're expecting any changes to `StyledExternalIcon`.

jzempel (changes requested):

packages/buttons/src/styled/StyledExternalIcon.ts
jgorfine-zendesk: Hey @jzempel — when you say 
> I don't think we're expecting any changes to `StyledExternalIcon`
What do you mean? (Does the translated label not count as a change?)

jgorfine-zendesk: Could you please help me understand why? Reasoning about the code, I had been under the impression that the `<StyledExternalIcon>` component was the smallest/root piece (aside from the icon itself, from the icons package), which is why it made sense to me to make the changes there, rather than in the `Anchor.tsx` instance of it. 🤔

jzempel: As a rule, we want to see hooks applied at the [element API](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#element-components) level. The code here is unexpected, leading to unwanted maintenance challenges.

jgorfine-zendesk: Happy to make the requested change, to keep with codebase conventions. I guess what I'm wondering, at this point, is why the `<StyledExternalIcon>` component level doesn't qualify as the element API level?

packages/buttons/src/styled/StyledExternalIcon.ts
```diff
@@ -14,7 +14,7 @@ const COMPONENT_ID = 'buttons.external_icon';
 /**
  * Accepts all `<svg>` props
  */
-export const StyledExternalIcon = styled(NewWindowIcon).attrs({
+export const StyledExternalIcon: any = styled(NewWindowIcon).attrs({
```
Francois-Esquire, line 17: Was there an issue with the typing to use `: any`?

packages/buttons/src/styled/StyledExternalIcon.ts
```diff
@@ -28,6 +28,8 @@ export const StyledExternalIcon = styled(NewWindowIcon).attrs({
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
 
+StyledExternalIcon.displayName = 'StyledExternalIcon';
```
Francois-Esquire, line 31: Do we need to add the `displayName` here? I think it is explicitly the same because of the variable name.
Screen Shot 2022-11-03 at 12 32 25 PM

Francois-Esquire (changes requested):
PR looks good - have some comments/nitpicks.

packages/buttons/src/styled/StyledExternalIcon.ts
jgorfine-zendesk: Yeah — VSCode wasn't happy that I hadn't explicitly declared the type here. That was when I had put the `useText` hook in here, though...maybe that'll be different, now that it's in the Anchor itself? Will investigate.

packages/buttons/src/styled/StyledExternalIcon.ts
jgorfine-zendesk: The `useText` helper requires a display name as one of the arguments, so I thought that meant I had to have one explicitly declared. I can take a quick look and get back to you, though!

jgorfine-zendesk: Was able to remove it. 👍

packages/buttons/src/styled/StyledExternalIcon.ts
jgorfine-zendesk: Was able to remove it. 👍

packages/buttons/src/elements/Anchor.tsx
```diff
@@ -25,10 +26,19 @@ export const Anchor = forwardRef<HTMLAnchorElement, IAnchorProps>(
       };
     }
 
+    const iconAriaLabel = useText(
+      StyledExternalIcon,
+      { 'aria-label': externalIconLabel },
+      'aria-label',
+      '(opens in a new tab)'
+    );
```
jzempel, line 34:
```suggestion
    // Prevent useText icon label warning/error unless the anchor is external
    const checkProps = isExternal ? { externalIconLabel } : { noIconLabel: 'true' };
    const iconAriaLabel = useText(
      Anchor,
      checkProps,
      isExternal ? 'externalIconLabel' : 'noIconLabel',
      '(opens in a new tab)'
    );
```

This change ensures the correct warning is logged for the developer ("Warning: you did not provide a customized/translated `externalIconLabel` text value for &lt;Anchor&gt;. Zendesk Garden is rendering &lt;Anchor externalIconLabel="(opens in a new tab)"&gt; by default.") while ensuring the warning/error only occurs if `isExternal` is true.

jzempel (changes requested):

packages/buttons/src/elements/Anchor.tsx
jgorfine-zendesk: Can do. 👍

jgorfine-zendesk: Sometime in the future, maybe it would be worth it to move this conditional logic into the `useText` helper itself? It's a little confusing (for me, at least!) to see the name of a fake prop inside of the component and realize it's a fake prop/placeholder. 

I'm thinking this could look something like this:

Anchor.tsx
```
const iconAriaLabel = useText(
  Anchor,
  checkProps,
  'externalIconLabel',
  '(opens in a new tab)',
  isExternal
);
```

useText.ts
```
export const useText = (
  component: Pick<FC, 'displayName'>,
  props: Record<string, any>,
  name: string,
  text: string,
  shouldShowWarnings: boolean = true
): string => {
  const value = props[name];

  return useMemo(() => {
    if (shouldShowWarnings) {
      if (name === 'children') {
        // Prevent Garden from providing text as a child content default.
        throw new Error('Error: `children` is not a valid `getText` prop.');
      } else if (value === null || value === '') {
        // Prevent consumer from removing a critical text attribute.
        throw new Error(
          component.displayName
            ? `Error: you must provide a valid \`${name}\` text value for <${component.displayName}>.`
            : `Error: you must provide a valid \`${name}\` text value.`
        );
      } else if (value === undefined) {
        // Warn consumer when Garden's default text is rendered.
        if (process.env.NODE_ENV === 'development') {
          // eslint-disable-next-line no-console
          console.warn(
            component.displayName
              ? `Warning: you did not provide a customized/translated \`${name}\` text value for <${component.displayName}>. Zendesk Garden is rendering <${component.displayName} ${name}="${text}"> by default.`
              : `Warning: you did not provide a customized/translated \`${name}\` text value. Zendesk Garden is rendering ${name}="${text}" by default.`
          );
        }

        return text;
      }

      return value;
    }
  }, [component.displayName, value, name, text, shouldShowWarnings]);
};
```

Francois-Esquire (approved):

jzempel (approved):

PR #1450, opened 2022-10-26
feat(notifications): global alert component

Francois-Esquire:
## Description

This PR introduces the `GlobalAlert` component as a new member of the `notifications` family.

## Detail

The `GlobalAlert` component takes one prop `type`, which can be "info", "success", "warning" and "error".

Screen Shot 2022-10-26 at 3 58 52 PM

## Checklist


packages/notifications/src/types/index.ts
```diff
@@ -58,3 +59,25 @@ export interface IToastProviderProps {
...
+  type: Type;
+}
+
+export interface IGlobalAlertAnchorProps
+  extends AnchorHTMLAttributes<HTMLAnchorElement>,
+    Pick<IAnchorProps, 'isExternal'> {}
+
+export interface IGlobalAlertButtonProps
+  extends ButtonHTMLAttributes<HTMLButtonElement>,
+    Pick<IButtonProps, 'isBasic'> {}
```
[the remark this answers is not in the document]
jzempel: You can always check with `yarn garden cmd-docgen --pretty [path-to-GlobalAlert]`. I'd recommend doing that here and validating the JSON returns as expected. Usually Storybook docs is a reliable check, too. But `cmd-docgen` is what produces generated documentation for the website.

packages/notifications/demo/stories/GlobalAlertStory.tsx
```diff
@@ -0,0 +1,76 @@
...
+  },
+  error: {
+    title: 'Error',
+    content: 'There are bugs in the garden.'
+  },
+  info: {
+    title: 'Info',
+    content: 'Another day in the garden.'
+  }
+};
```
jzempel, line 44: Prefer strings coming straight from the `children` controls as seen with `Notification` and `Alert`. This Garden-centric content will be helpful in the upcoming website examples. However, Storybook is meant to be more utilitarian with respect to content.

packages/notifications/demo/global-alert.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+      hasClose: { name: 'Close', type: 'boolean', table: { category: 'Story' } },
+      hasTitle: { name: 'Title', type: 'boolean', table: { category: 'Story' } }
+    }}
+    parameters={{
+      design: {
+        allowFullscreen: true,
+        type: 'figma',
+        url: 'https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=103%3A25292'
+      }
+    }}
```
jzempel, line 49: Do not add the `design` parameter until the component has been checked-in to the main brain on Figma.

packages/notifications/demo/global-alert.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+    name="GlobalAlert"
+    args={{
+      type: 'info'
+    }}
+    argTypes={{
+      anchor: { name: 'children', type: 'string', table: { category: 'Anchor' } },
+      isExternal: { control: { type: 'boolean' }, table: { category: 'Anchor' } },
+      button: { name: 'children', type: 'string', table: { category: 'Button' } },
+      isBasic: { control: { type: 'boolean' }, table: { category: 'Button' } },
+      ariaLabel: { name: 'aria-label', type: 'string', table: { category: 'Close' } },
```
jzempel, line 34: Please add default `args` text for the `aria-label`. See `Modal` `Close`, `DrawerModal.Close` and `TooltipModal.Close` for details. Doing so bypasses the dev warning out of the box, and shows we are using the component as intended.

packages/notifications/demo/global-alert.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+  <Story
+    name="GlobalAlert"
+    args={{
+      type: 'info'
+    }}
+    argTypes={{
+      anchor: { name: 'children', type: 'string', table: { category: 'Anchor' } },
+      isExternal: { control: { type: 'boolean' }, table: { category: 'Anchor' } },
+      button: { name: 'children', type: 'string', table: { category: 'Button' } },
+      isBasic: { control: { type: 'boolean' }, table: { category: 'Button' } },
```
jzempel, line 33: I'd like to see the button control to be configured a lot closer to modal `FooterItem[]`. In other words, the data should describe `Button[]` array with the following by default:

```
[
  { text: 'Cancel', isBasic: true },
  { text: 'OK' }
]
```

This allows Storybook to push into the extents of button positioning, group styling, etc.

packages/notifications/demo/stories/GlobalAlertStory.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {title || globalAlertText[type].title}
+      </GlobalAlert.Title>
+    )}
+    <GlobalAlert.Content>{content || globalAlertText[type].content}</GlobalAlert.Content>
+    {hasAnchor && (
+      <GlobalAlert.Anchor href="#" isExternal={isExternal}>
+        {anchor || 'Find out more'}
+      </GlobalAlert.Anchor>
+    )}
+    {hasButton && <GlobalAlert.Button isBasic={isBasic}>{button || 'Primary'}</GlobalAlert.Button>}
```
jzempel, line 73: Please prefer control-defined `children` content rather than or'ing fallback strings.

packages/notifications/src/elements/global-alert/GlobalAlertIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const combinedProps = { 'aria-hidden': true, ...props };
+
+  const [success, warning, error, info] = TYPE;
+
+  switch (type) {
+    case success:
+      return <SuccessIcon {...combinedProps} />;
+    case warning:
+      return <WarningIcon {...combinedProps} />;
+    case error:
```
jzempel, line 27: This case is not producing the correct error icon.

packages/notifications/src/elements/global-alert/GlobalAlertTitle.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import React, { forwardRef } from 'react';
+
+import { IGlobalAlertTitleProps } from '../../types';
+import { StyledGlobalAlertTitle } from '../../styled';
+
+export const GlobalAlertTitle = forwardRef<HTMLDivElement, IGlobalAlertTitleProps>((props, ref) => {
+  return <StyledGlobalAlertTitle ref={ref} {...props} />;
+});
+
```
jzempel, line 16: Missing `propTypes` for `isRegular`.

packages/notifications/src/elements/global-alert/GlobalAlertIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+    case info:
+    default:
+      return <InfoIcon {...combinedProps} />;
+  }
+};
+
+GlobalAlertIcon.displayName = 'GlobalAlert.Icon';
```
jzempel, line 34: Since it is not part of the public API, would this just be `'GlobalAlertIcon'`?

packages/notifications/src/elements/global-alert/GlobalAlertIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [success, warning, error, info] = TYPE;
+
+  switch (type) {
+    case success:
+      return <SuccessIcon {...combinedProps} />;
+    case warning:
+      return <WarningIcon {...combinedProps} />;
+    case error:
+    case info:
+    default:
```
jzempel, line 29: Should there be a default if `type` is required?

packages/notifications/src/types/index.ts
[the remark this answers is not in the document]
jzempel: Also, the [API](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md) & [TS](https://github.com/zendeskgarden/react-components/blob/main/docs/typescript.md) docs were written to cover quirks for Storybook. However, it seems that props for subcomponents have been lost with a recent update. See `Sheet.Footer` for example: https://github.com/zendeskgarden/react-components/blob/main/docs/typescript.md (should show an `isCompact` prop).

packages/notifications/src/elements/global-alert/utility.ts
```diff
@@ -0,0 +1,16 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { createContext, useContext } from 'react';
+
+import { TYPE, IGlobalAlertProps } from '../../types';
+
+export type GlobalAlertContextProps = Pick<IGlobalAlertProps, 'type'>;
+
+export const GlobalAlertContext = createContext<GlobalAlertContextProps>({ type: TYPE[3] });
```
jzempel, line 14: This is brittle (subject to array re-ordering) and unreadable (have to go reference `TYPE[3]`). Please prefer explicit `type: 'info'` usage. TypeScript will do what is needed to keep things in check.

jzempel (changes requested):
Submitting a first-half review for type+elements+context API. I'll follow-up with a separate review for the `styled-components`.

packages/notifications/src/elements/global-alert/GlobalAlertIcon.tsx
Francois-Esquire: There shouldn't/wouldn't, I was avoiding adding an eslint rule. I will remove it.

packages/notifications/src/types/index.ts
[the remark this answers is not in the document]
Francois-Esquire: @jzempel It looks like the `yarn garden cmd-docgen --pretty packages/notifications/src/elements/global-alert` is producing the correct prop documentation.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -0,0 +1,107 @@
...
+  `;
+}
+
+function sizeStyles(props: StyledGlobalAlertProps) {
+  const height = props.theme.space.base * 13;
+  const padding = props.theme.space.base * 2;
+  const paddingStart = props.theme.space.base * 4;
+
+  return css`
+    border-bottom: 1px solid;
```
jzempel, line 58:
```suggestion
    border-bottom: ${props.theme.borders.sm};
```

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    & > svg {
+      flex-shrink: 0;
+      margin-top: ${props.theme.space.base * 2.5}px;
+    }
+
+    & > div {
+      margin-top: ${props.theme.space.base * 2}px;
+      margin-bottom: ${props.theme.space.base * 2}px;
+    }
```
jzempel, line 72: From what I'm seeing, all of the child selection here (`svg`, `div`, `button`) indicates the need to re-work the entire layout styling for the `GlobalAlert` and it's subcomponents. In summary, the 52px default height of the component needs to top/bottom pad in order to accommodate 20px line-height text. Without this, we're getting a 53px default `GlobalAlert` along with a bunch of pixel fussing, which is no bueno.

jzempel (changes requested):
Stopped styling review at `StyledGlobalAlert`. I think we need to fundamentally address base layout styling in order for the component CSS to be robust.

jzempel (changes requested):
Getting close @Francois-Esquire. A couple things I'm noticing:
- the subcomponent element components lack testing (should have "passes ref to underlying DOM element" similar to notification `Close` at a minimum)
- none of the elements have `@extends` JSDoc clauses (should see accurate "extends" key values when running `yarn garden cmd-docgen`)

jgorfine-zendesk (commented):
Did an initial review of the code and the DOM. Most of my feedback is minor, i.e., may not need to be addressed right now. There was one opportunity I saw to get ahead of potential future a11y violations, and that's in the `GlobalAlert.Title`. 👍 

I'll do a separate review in which I post recordings that demonstrate the screen reader experience on Safari x VoiceOver, Chrome x JAWS, and Firefox x NVDA (the AssistivLabs recommended combo).

packages/notifications/src/elements/global-alert/GlobalAlertTitle.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React, { forwardRef } from 'react';
+import PropTypes from 'prop-types';
+import { useGlobalAlertContext } from './utility';
+import { IGlobalAlertTitleProps } from '../../types';
+import { StyledGlobalAlertTitle } from '../../styled';
+
+export const GlobalAlertTitle = forwardRef<HTMLDivElement, IGlobalAlertTitleProps>((props, ref) => {
```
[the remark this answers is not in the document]
jzempel: This would be a future consideration for the package as a whole.

packages/notifications/src/styled/global-alert/StyledGlobalAlertContent.ts
```diff
@@ -0,0 +1,24 @@
...
+import { DEFAULT_THEME, retrieveComponentStyles } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'notifications.global-alert.content';
+
+export const StyledGlobalAlertContent = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  flex-grow: 1;
+
```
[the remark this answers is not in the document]
jzempel: It doesn't horizontally scroll with normal usage. But if we're concerned that it could, then we should change to `overflow-x: auto`.

packages/notifications/src/elements/global-alert/GlobalAlert.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import React from 'react';
+import { render, renderRtl } from 'garden-test-utils';
+
+import { Type } from '../../types';
+import { GlobalAlert } from './GlobalAlert';
+
+describe('GlobalAlert', () => {
```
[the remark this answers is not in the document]
jzempel: Garden doesn't warn in general because we can rely on TypeScript (compile-time) + PropTypes (run-time) usage.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -0,0 +1,140 @@
...
+      foregroundColor = getColor('primaryHue', 700, props.theme);
+      anchorHoverColor = getColor('primaryHue', 800, props.theme);
+      anchorActiveColor = getColor('primaryHue', 900, props.theme);
+      anchorBoxShadowColor = getColor('primaryHue', 700, props.theme, 0.35);
+      break;
+  }
+
+  // Apply a border without affecting the element's size
+  const boxShadow =
+    borderColor && `0 ${props.theme.borderWidths.sm} ${props.theme.borderWidths.sm} ${borderColor}`;
```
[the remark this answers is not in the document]
jzempel: The `GlobalAlert` is not governed by `height`, but rather `min-height`.

packages/notifications/src/styled/global-alert/StyledGlobalAlertContent.ts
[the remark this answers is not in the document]
jzempel: I think the default `flex-basis` value of `auto` implicitly gives us what we need.

packages/notifications/demo/stories/data.ts
```diff
@@ -13,3 +13,14 @@ export const TOAST_PLACEMENT_OPTIONS = [
   'top-end',
   'top-start'
 ] as const;
+
+export const GLOBAL_ALERT_BUTTONS = [
+  {
```
steue, line 18: minor question/ask. design's recommendation is to only have one action in the alert (excluding any anchors and close). could we remove this second button? i'm afraid folks might see this and consider this default, and don't want to confuse "what should" with "what's possible"

jzempel: Hey @steue, I appreciate your concern. However, we really need to lean into the purpose of Storybook here – which is to press on the extents of the layout and styling. Indeed, we use Storybook as the testing ground for "what's possible". By contrast, the website is the place to show, by example, "what should" happen with component usage.

jzempel: To put it another way...
- if we remove either button, we run the future risk of missing unintended layout updates that might impact margining
- if we remove the basic button, we run a future risk of missing unintended styling updates to that button
- if we remove the primary button, we run future risk of missing unintended styling updates to that custom button

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  return css`
+    box-shadow: ${boxShadow};
+    background-color: ${backgroundColor};
+    color: ${foregroundColor};
+
+    & a {
```
steue, line 78: found a bug, and recreated on chrome, safari, and firefox. the anchor is focusable and is in the right order, however the focus ring isn't visible.

steue (changes requested):
this looks great! thanks y'all! to summarize my comments:
- bug around focus ring not being visible
- nit to remove secondary button as default in demo

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
jzempel: @Francois-Esquire I think Steph found a critical specificity issue here.

@steue re secondary button nit: https://github.com/zendeskgarden/react-components/pull/1450#discussion_r1018101593

packages/notifications/demo/stories/data.ts
steue: yes that makes sense! definitely want to uphold the purpose of Storybook demos. our website documentation (will be) quite clear on the recommendation, so there shouldn't be any confusion.

siennagonzalez (commented):
agree with the design feedback left by @steue but everything else looks good to me!

jzempel (changes requested):
@Francois-Esquire unfortunately we can't use [transient props](https://styled-components.com/docs/faqs#transient-props-since-51) since our styled-components peer dependency is ranged `^4.2.0 || ^5.0.0` (we'll bump this peer to `^5.1.0` in v9). I'd recommend changing all of the `$type` styled props to `alertType` which will also get filtered. styled-components was letting `kind` through because it is a [valid attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track#attr-kind).

steue (approved):
this looks great! thank you @Francois-Esquire @jzempel for everything!!!!

jzempel (approved):

PR #1453, opened 2022-11-03
feat(chrome): add aria-current attribute to NavItem component, when it's current

jgorfine-zendesk:
## Description

This PR adds the `aria-current` attribute to Chrome nav item that represents the current page. This will help assistive tech users — especially screen reader users — orient themselves within the navigation. 

## Detail

https://user-images.githubusercontent.com/93289772/199769721-7f2d2273-755b-4f80-89d3-c0c543d3704c.mov

## Checklist


packages/chrome/src/elements/nav/NavItem.tsx
```diff
@@ -44,6 +46,7 @@ export const NavItem = React.forwardRef<HTMLButtonElement, INavItemProps>(
         hue={hue}
         isDark={isDark}
         isLight={isLight}
+        aria-current={ariaCurrent}
```
jzempel, line 49:
```suggestion
        aria-current={other.isCurrent}
```

We can't guarantee that these nav items are always used for page level navigation.

jzempel (changes requested):

packages/chrome/src/elements/nav/NavItem.tsx
jgorfine-zendesk: Gotcha. How do consumers add destinations (`href`s) to these elements? Is there an opportunity to conditionally set the value to `page` if we know the components are intended to behave like links?

jgorfine-zendesk: Fair enough. Is there someplace we can document this, so that consumers can see that this is the preferred approach?

jzempel (approved):

PR #1463, opened 2022-11-29
fix(accordions): relative type interface import

jzempel:
## Description

Fix incorrect path on import which causes `Cannot find module` TS error for consumers.


rprzebieglec (approved):

geotrev (approved):
Were types being bundled by webpack? (looking at size snapshot)

Francois-Esquire (approved):

PR #1481, opened 2022-12-14
fix(notifications): lets GlobalAlert.Title flow correctly in RTL documents

geotrev:
## Description

`<GlobalAlert />` is missing theme-awareness of RTL (`theme.rtl`). This PR adds it.

~~When `<GlobalAlert.Title />` is a child in `<GlobalAlert />` and the document/theme is `rtl`, The title stays left of the content (the padding does switch from left to right correctly, however).~~

~~The root issue is that inline text (in this case, a `div` that is `display: inline`) is treated as combined with all immediate sibling text nodes, and despite being flex items they naturally won't reflow.~~

## Detail

This is a pretty simple one-liner.

One extra factor noted by the reporter is that margins are incorrect in RTL, even with RTL applied via `styled-components`, But this only happens if you use an LTR language in RTL mode. So, using Arabic, for example, does show the correct margins when RTL is hooked up correctly. So no worries there.

Here's an example of Arabic + RTL property working (just the content):

Screen Shot 2022-12-15 at 11 49 59 AM

~~Setting `<GlobalAlert.Title />` to `display: inline-flex` separates the title text from content, allowing it to flow independently.~~

## ~~Risks~~

~~Due to switching `display` from `inline` to `inline-flex`, the layout has shifted to treat `Title` as block-y. Not the end of the world as content wraps just fine as shown above, but the height of the Title now stretches the full content row height of the content as opposed to when its inline, and maintains a line-height crop (as any inline element would).~~

~~Line-height is maintained unless overriden via `styled-components` (which is kind of out of our control anyway, but I digress):~~

## Checklist

- [x] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- [x] ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] ~~:wheelchair: tested for [WCAG 2.1 AA](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa) accessibility compliance~~

### DIFF
--- a/packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
+++ b/packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
@@ -43,7 +43,7 @@ export const StyledGlobalAlertTitle = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledGlobalAlertTitleProps>`
-  display: inline-flex;
+  display: inline;
   /* stylelint-disable-next-line property-no-unknown */
   margin-${props => (props.theme.rtl ? 'left' : 'right')}: ${props => props.theme.space.base * 2}px;
   font-weight: ${props =>
--- a/packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
+++ b/packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
@@ -43,7 +43,7 @@ export const StyledGlobalAlertTitle = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledGlobalAlertTitleProps>`
-  display: inline;
+  display: inline-flex;
   /* stylelint-disable-next-line property-no-unknown */
   margin-${props => (props.theme.rtl ? 'left' : 'right')}: ${props => props.theme.space.base * 2}px;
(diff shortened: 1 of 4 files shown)


packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
```diff
@@ -43,7 +43,7 @@ export const StyledGlobalAlertTitle = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledGlobalAlertTitleProps>`
-  display: inline;
+  display: inline-flex;
```
jzempel, line 46: Unfortunately, I don't think this solution scales as expected for a long title. With `inline-flex` we see the following:

Screen Shot 2022-12-15 at 9 03 16 AM

But design intent is for content to flow like this:

Screen Shot 2022-12-15 at 9 05 36 AM

jzempel (changes requested):
@geotrev I think you may need to look at CSS `direction` rather than adjusting `display`.

packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
geotrev: Updated, good catch. Hooked up `direction` to `theme.rtl` :+1:

jzempel (approved):

Francois-Esquire (approved):

PR #1489, opened 2023-01-17
fix(dropdowns): remove selected icon from next, previous and add item…

Francois-Esquire:
## Description

Removes the selected SVG icon for `AddItem`, `NextItem` and `PreviousItem` when selected.

## Detail
Prior to this bug fix, the selected icon would overlap the icon for the affected components.

Previously:
Screen Shot 2023-01-17 at 4 26 48 PM

After the fix:
Screen Shot 2023-01-17 at 4 27 34 PM

Screen Shot 2023-01-17 at 4 24 38 PM

## Checklist

### DIFF
--- a/packages/dropdowns/src/elements/Menu/Items/Item.tsx
+++ b/packages/dropdowns/src/elements/Menu/Items/Item.tsx
@@ -20,7 +20,7 @@ import { ItemContext } from '../../../utils/useItemContext';
  */
 export const Item = React.forwardRef<HTMLLIElement, IItemProps>(
   (
-    { value, disabled, isDanger, component = StyledItem, shouldHideIcon, children, ...props },
+    { value, disabled, isDanger, component = StyledItem, hasIcon, children, ...props },
     forwardRef
   ) => {
     const {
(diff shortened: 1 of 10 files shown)


packages/dropdowns/src/elements/Menu/Items/Item.tsx
```diff
@@ -55,8 +55,12 @@ export const Item = React.forwardRef<HTMLLIElement, IItemProps>(
       }
     });
 
+    // eg of a composite component: AddItem, NextItme, PreviousItem, which is passed through the `component` prop.
+    // this is used to hide the selected SVG icon when Item is in a selected state for composite components
+    const isNotCompositeComponent = !!Component.displayName;
```
jzempel, line 60: Works, but feels a bit like magic (breaks if Garden composes a new, selectable item type in the future). Is it worth being more explicit? For example if the `Component` is specifically `AddItem` || `NextItem` || `PreviousItem`?

jzempel (approved):

packages/dropdowns/src/elements/Menu/Items/Item.tsx
Francois-Esquire: I've updated this to use the exact naming of the components.

jzempel (changes requested):
oops, the additions of `displayName` can't happen because it blows up `garden cmd-docgen`.

jzempel (approved):

PR #1488, opened 2023-01-12
fix: enables accessible label in TooltipModal, ColorpickerDialog, and ColorSwatchDialog

geotrev:
## Description

This PR addresses these two problems:

1. Currently, ColorSwatchDialog and ColorpickerDialog use TooltipModal under the hood to render their dialogs. Because these two components don't use `TooltipModal.Title`, the modal container has an invalid IDREF on `aria-labelledby`.
2. There is an unnecessary `aria-describedby` on the modal container.

Solutions implemented in this PR to address the above:
1. Let TooltipModal determine which ARIA attributes to use depending on the existence of a `<Title />` child. Context is used to achieve this, similar to `isCloseButtonPresent` (also in `react-modals`). The resulting `aria-label` is passed through `useText`.
2. In both Color picker components, wrapped the (currently undocumented, but accepted) `aria-label` with `useText` to require consumers to set this label. This overrides `useText` within
3. Remove `aria-describedby` ([optional](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/#wai-aria-roles-states-and-properties-7))

## Detail

First, the fixed color picker dialogs:

**ColorpickerDialog**
Screen Shot 2023-01-12 at 2 58 06 PM

**ColorSwatchDialog**
Screen Shot 2023-01-12 at 2 57 48 PM

Here are the three possibilities for the `aria-label`/`aria-labelledby` configuration.

**With Title (`aria-labelledby`)**  (as it is today)
Screen Shot 2023-01-12 at 3 11 46 PM

**Without Title, default `aria-label`**
Screen Shot 2023-01-12 at 3 46 07 PM

**Without Title, user-defined `aria-label`**
Screen Shot 2023-01-12 at 3 21 26 PM

## Checklist

- [ ] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- [x] :globe_with_meridians: demo is up-to-date (`yarn start`)
- [ ] ~~:arrow_left: renders as expected with reversed (RTL) direction~~
- [ ] ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] :guardsman: includes new unit tests. Maintain [existing coverage](https://coveralls.io/github/zendeskgarden/react-components?branch=main) (always >= 96%)
- [x] :wheelchair: tested for [WCAG 2.1 AA](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa) accessibility compliance
[...]

PR #1491, opened 2023-01-19
fix(modals): enables accessible label on Modal when no Header is given

geotrev:
## Description

When no `<Header />` child is present on `<Modal />`, the dialog element's `aria-labelledby` should be replaced with a descriptive `aria-label`. Consumers will be nudged to provide a label via console log, if none is given.

This is an identical fix to what recently merged for `TooltipModal` ([link](https://github.com/zendeskgarden/react-components/pull/1488)).

## Detail

A small change was also made to `TooltipModal` to check modal container props instead of consumer props for `aria-labelledby`. 

See [Detail](https://github.com/zendeskgarden/react-components/pull/1488#:~:text=(optional)-,Detail,-First%2C%20the%20fixed) on 1488 (link above) for exact change and its DOM effects.

## Checklist

### DIFF
--- a/packages/modals/demo/stories/ModalStory.tsx
+++ b/packages/modals/demo/stories/ModalStory.tsx
@@ -53,8 +53,8 @@ export const ModalStory: Story<IArgs> = ({
   dialogAriaLabel,
   ...args
 }) => {
-  // Using `aria-label={undefined}` when `hasTitle` is `true` appears to
-  // void the fallback value in Storybook, resulting in no rendered attribute
+  // Using `aria-label={undefined}` when `hasHeader` is `true` appears to
+  // void the fallback value in Storybook, resulting in no rendered attribute.
   const ariaProp: Record<string, any> = hasHeader
     ? {}
     : {
--- a/packages/modals/demo/stories/ModalStory.tsx
+++ b/packages/modals/demo/stories/ModalStory.tsx
@@ -32,6 +32,8 @@ interface IArgs extends IModalProps {
   isDanger: boolean;
   tag: string;
   header: string;
+  closeAriaLabel: string;
+  dialogAriaLabel: string;
 }
 
 export const ModalStory: Story<IArgs> = ({
@@ -47,7 +49,8 @@ export const ModalStory: Story<IArgs> = ({
   header,
   isDanger,
   tag,
-  'aria-label': ariaLabel,
+  closeAriaLabel,
+  dialogAriaLabel,
   ...args
 }) => (
   <>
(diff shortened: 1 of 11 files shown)


packages/modals/demo/stories/ModalStory.tsx
```diff
@@ -47,42 +49,53 @@ export const ModalStory: Story<IArgs> = ({
...
-            ))}
-          </Footer>
-        )}
-        {hasClose && <Close aria-label={ariaLabel} />}
-      </Modal>
-    )}
-  </>
-);
+}) => {
+  // Using `aria-label={undefined}` when `hasTitle` is `true` appears to
```
jzempel, line 56: [nit] *`hasHeader`

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
```diff
@@ -86,7 +86,7 @@ const TooltipModalComponent = React.forwardRef<HTMLDivElement, ITooltipModalProp
 
     // Derive aria attributes from props
     const attribute = hasTitle ? 'aria-labelledby' : 'aria-label';
-    const defaultValue = hasTitle ? props['aria-labelledby'] : 'Modal dialog';
+    const defaultValue = hasTitle ? modalProps['aria-labelledby'] : 'Modal dialog';
```
jzempel, line 89: Should we differentiate by defaulting to "Tooltip modal dialog" here?

jzempel (commented):
Looks good, pending test fixes

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
geotrev: I debated that as well and I'm not sure. The components will functionally be treated as a modal regardless, but not all SR users are fully blind and may appreciate the extra identification. I'll check in with a couple folks and get some opinions. 🙂

geotrev: Talking with Jen, we agreed that the concepts of "Drawer modal" and "Tooltip modal" likely wouldn't have a ton of value beyond simply using "modal dialog." What do you think? I could go either way, but leaning towards the generic label.

jzempel (approved):

packages/modals/src/elements/TooltipModal/TooltipModal.tsx
jzempel: I can appreciate that perspective. Let's remain generic.

PR #1497, opened 2023-01-20
fix(grid): hide pane content when not visible

Francois-Esquire:
## Description

This PR is an accessibility fix that hides `Pane.Content` when a `Pane` has a dimension less than `1` for either the width or height.

## Detail

`useResizeObserver` from the npm package `use-resize-observer` has been added to the `Pane` component and is used to observe the component element to check its dimensions. When the `width` or `height` dimension of a `Pane` is less than `1`, it sets the `isVisible` context property to `false`. The `isVisible` context property is consumed by the `Pane.Content` component that is nested in the `Pane` component and sets the `hidden` attribute, which causes the `Pane.Content` to `display: none` in `StyledPaneContent`.

The `useResizeObserver` was added to the `Pane` instead of the `Pane.Content` due to the `Pane.Content` not being displayed when hidden (which causes `ResizeObserver` to behave in unexpected ways).

While there is a polyfilled version of `use-size-observer`, it is only used in testing and mapped to `moduleNameMapper` in `jest.config.js` due to ResizeObserver being a Browser API which doesn't exist in node environment. Without this mapping, a handful of tests that import the `Pane` component throw `ReferenceError`s.

It may be worth noting, `ResizeObserver` is not available in the `node.js` runtime, it would help to document this fact. This would not affect SSR, since a DOM ref is required to invoke the `ResizeObserver` API.

## Checklist

### DIFF
--- a/packages/grid/package.json
+++ b/packages/grid/package.json
@@ -28,7 +28,7 @@
     "polished": "^4.0.0",
     "prop-types": "^15.5.7",
     "react-merge-refs": "^1.1.0",
-    "use-resize-observer": "9.1.0"
+    "use-resize-observer": "^9.1.0"
   },
   "peerDependencies": {
     "@zendeskgarden/react-theming": "^8.1.0",
--- a/packages/grid/package.json
+++ b/packages/grid/package.json
@@ -27,7 +27,8 @@
     "@zendeskgarden/react-tooltips": "^8.62.1",
     "polished": "^4.0.0",
     "prop-types": "^15.5.7",
-    "react-merge-refs": "^1.1.0"
+    "react-merge-refs": "^1.1.0",
+    "use-resize-observer": "9.1.0"
   },
   "peerDependencies": {
     "@zendeskgarden/react-theming": "^8.1.0",
@@ -36,8 +37,7 @@
     "styled-components": "^4.2.0 || ^5.0.0"
   },
   "devDependencies": {
-    "@zendeskgarden/react-theming": "^8.62.1",
-    "use-resize-observer": "9.1.0"
+    "@zendeskgarden/react-theming": "^8.62.0"
   },
   "keywords": [
     "components",
--- a/packages/grid/src/elements/pane/Pane.spec.tsx
+++ b/packages/grid/src/elements/pane/Pane.spec.tsx
@@ -6,10 +6,23 @@
  */
 
 import React from 'react';
-import { render } from 'garden-test-utils';
+import { act, render } from 'garden-test-utils';
+import { Content } from './components/Content';
 import { Splitter } from './components/Splitter';
 import { Pane } from './Pane';
 import { PaneProvider } from './PaneProvider';
+
+// '@juggle/resize-observer' is the polyfill used by 'use-resize-observer'
+// we tap into the polyfilled ResizeObserver to trigger the useResizeObserver hook
+// this way we can test when the Pane is collapsed and ensure the Pane.Content is hidden
+jest.mock('@juggle/resize-observer', () => ({
+  ResizeObserver: function ResizeObserver(cb: () => void) {
+    const o = jest.requireActual('@juggle/resize-observer');
+
+    // @ts-expect-error resize observer callback is an artificial method added for testing
+    return new o.ResizeObserver((window.resizeObserverCallback = jest.fn(cb)));
+  }
+}));
 
 const UncontrolledTestSplitter = () => {
   return (
@@ -21,6 +34,7 @@ const UncontrolledTestSplitter = () => {
     >
       {() => (
         <Pane>
+          <Content>Content</Content>
           <Splitter layoutKey="a" min={0} max={1} />
           <Splitter layoutKey="b" min={0} max={1} />
         </Pane>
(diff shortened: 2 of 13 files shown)


packages/grid/src/elements/pane/Pane.tsx
```diff
@@ -16,17 +18,33 @@ const PaneComponent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>
...
 
+    const observerRef = useRef<HTMLDivElement>(null);
+    const { width = 1, height = 1 } = useResizeObserver<HTMLDivElement>({ ref: observerRef });
+
+    const isVisible = useMemo(() => {
+      if (observerRef.current) {
+        if (width <= 1) {
+          return false;
+        } else if (height <= 1) {
+          return false;
```
jzempel, line 29: Might be splitting hairs (pixels), but why aren't we comparing with `0`?

packages/grid/src/elements/pane/Pane.spec.tsx
```diff
@@ -6,11 +6,24 @@
...
 import { PaneProvider } from './PaneProvider';
 
+// '@juggle/resize-observer' is the polyfill used by 'use-resize-observer'
+// we tap into the polyfilled ResizeObserver to trigger the useResizeObserver hook
+// this way we can test when the Pane is collapsed and ensure the Pane.Content is hidden
+jest.mock('@juggle/resize-observer', () => ({
+  ResizeObserver: function ResizeObserver(cb: () => void) {
+    const o = jest.requireActual('@juggle/resize-observer');
+
+    // @ts-expect-error resize observer callback is an artificial method added for testing
```
Francois-Esquire, line 22:
```suggestion
    // @ts-expect-error resize observer callback is a non-standard method added for testing
```
Improvement in language?

packages/grid/src/elements/pane/Pane.spec.tsx
```diff
@@ -6,11 +6,24 @@
...
+    return new o.ResizeObserver((window.resizeObserverCallback = jest.fn(cb)));
```
Francois-Esquire, line 23: Setting the `resizeObserverCallback` to a property on the `window` allows us to capture the internal handler used by `useResizeObserver` and trigger updates to the hook with custom entries ([ResizeObserverEntry - MDN](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry))

packages/grid/src/elements/pane/Pane.tsx
```diff
@@ -16,17 +18,33 @@ const PaneComponent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>
...
     const [paneId, setPaneId] = useState<string>();
 
+    const observerRef = useRef<HTMLDivElement>(null);
+    const { width = 1, height = 1 } = useResizeObserver<HTMLDivElement>({ ref: observerRef });
+
+    const isVisible = useMemo(() => {
+      if (observerRef.current) {
+        if (width <= 1) {
+          return false;
+        } else if (height <= 1) {
```
Francois-Esquire, line 28: Would the threshold of `1` need additional considerations?

jzempel (changes requested):
Now that we're committed to `useResizeObserver`, I wonder if there is some burden we can remove from the consumer by handling `totalPanesWidth` and `totalPanesHeight` for `PaneProvider` under the covers? [maybe not](https://github.com/zendeskgarden/react-components/blob/main/packages/grid/demo/~patterns/stories/CardStory.tsx#L18-L24)? 🤷 Anyway, topic for another time :wink:

packages/grid/package.json
```diff
@@ -27,7 +27,8 @@
     "@zendeskgarden/react-tooltips": "^8.62.0",
     "polished": "^4.0.0",
     "prop-types": "^15.5.7",
-    "react-merge-refs": "^1.1.0"
+    "react-merge-refs": "^1.1.0",
+    "use-resize-observer": "9.1.0"
```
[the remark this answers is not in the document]
Francois-Esquire: Good catch, I've updated!

packages/grid/src/elements/pane/Pane.tsx
Francois-Esquire: Fair point, this was an err on my end from the default value of `1` set on the `width` and `height` properties provided by the `useResizeObserverHook`. I've updated to use `0` instead.

packages/grid/src/elements/pane/Pane.tsx
```diff
@@ -16,17 +18,33 @@ const PaneComponent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>
...
+    const isVisible = useMemo(() => {
+      if (observerRef.current) {
+        if (width < 1) {
+          return false;
+        } else if (height < 1) {
+          return false;
+        }
+      }
+
+      return true;
```
jzempel, line 33:
```suggestion
      return observerRef.current ? width >= 0 && height >= 0 : true;
```

might be nice to compact this a bit 🤷

jzempel (approved):

PR #1512, opened 2023-02-07
fix(modals/drawer): correctly set focus on trigger element when focus is restored

Francois-Esquire:
## Description

This PR fixes an edge case found in the `Drawer` component with the focus management being handled in all cases.

## Detail

The underlying bug is around how focus is returned to the trigger element that opened the modal and `restoreFocus` - if an element has focus other than the `document.body` (or no focus) as the `Drawer` is rendered, it will lock out the actual trigger element from reclaiming focus.

## Checklist

### DIFF
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
@@ -183,7 +183,7 @@ describe('DrawerModal', () => {
 
       await user.click(getByText('Open Drawer'));
 
-      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+      expect(getByRole('dialog')).toBe(document.activeElement);
     });
 
     it('restores focus to the trigger button after close', async () => {
@@ -191,13 +191,13 @@ describe('DrawerModal', () => {
 
       await user.click(getByText('Open Drawer'));
 
-      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+      expect(getByRole('dialog')).toBe(document.activeElement);
 
       await user.type(getByRole('dialog'), '{escape}');
 
       await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
 
-      expect(getByText('Open Drawer')).toStrictEqual(document.activeElement);
+      expect(getByText('Open Drawer')).toBe(document.activeElement);
     });
   });
 });
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
@@ -177,19 +177,43 @@ describe('DrawerModal', () => {
     await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
   });
 
-  it('correctly focuses on the Drawer when open and restores focus to the trigger button after close', async () => {
-    const { getByText, getByRole, queryByRole } = render(
-      <Example backdropProps={{ 'data-test-id': 'backdrop' } as any} />
-    );
+  describe('focus management', () => {
+    it('correctly focuses on the Drawer when open', async () => {
+      const { getByText, getByRole } = render(<Example />);
 
-    await user.click(getByText('Open Drawer'));
+      await user.click(getByText('Open Drawer'));
 
-    expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+    });
 
-    await user.type(getByRole('dialog'), '{escape}');
+    it('restores focus to the trigger button after close', async () => {
+      const { getByText, getByRole, queryByRole } = render(<Example />);
 
-    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+      await user.click(getByText('Open Drawer'));
 
-    expect(getByText('Open Drawer')).toStrictEqual(document.activeElement);
+      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+
+      await user.type(getByRole('dialog'), '{escape}');
+
+      await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+
+      expect(getByText('Open Drawer')).toStrictEqual(document.activeElement);
+    });
+
+    it('handles restoreFocus prop change and releases the trigger from being refocused after close', async () => {
+      const { getByText, getByRole, queryByRole, rerender } = render(<Example />);
+
+      await user.click(getByText('Open Drawer'));
+
+      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+
+      rerender(<Example restoreFocus={false} />);
+
+      await user.type(getByRole('dialog'), '{escape}');
+
+      await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+
+      expect(getByText('Open Drawer')).not.toStrictEqual(document.activeElement);
+    });
   });
 });
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
@@ -199,21 +199,5 @@ describe('DrawerModal', () => {
 
       expect(getByText('Open Drawer')).toStrictEqual(document.activeElement);
     });
-
-    it('handles restoreFocus prop change and releases the trigger from being refocused after close', async () => {
-      const { getByText, getByRole, queryByRole, rerender } = render(<Example />);
-
-      await user.click(getByText('Open Drawer'));
-
-      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
-
-      rerender(<Example restoreFocus={false} />);
-
-      await user.type(getByRole('dialog'), '{escape}');
-
-      await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
-
-      expect(getByText('Open Drawer')).not.toStrictEqual(document.activeElement);
-    });
   });
 });
--- a/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
+++ b/packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
@@ -176,4 +176,20 @@ describe('DrawerModal', () => {
 
     await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
   });
+
+  it('correctly focuses on the Drawer when open and restores focus to the trigger button after close', async () => {
+    const { getByText, getByRole, queryByRole } = render(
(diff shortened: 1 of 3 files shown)


packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -49,12 +58,32 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
+            modalRef.current.focus();
+          }
+        }
+
+        if (triggerRef.current && !isOpen) {
+          triggerRef.current.focus();
+          triggerRef.current = null;
+        }
+      }
+    }, [environment, modalRef, triggerRef, restoreFocus, focusOnMount, isOpen]);
```
Francois-Esquire, line 85: The implementation is intentionally verbose for the sake of readability. If we find this unnecessary, we can compact this further.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -49,12 +58,32 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
 
+    /* [1:c] */
+    useEffect(() => {
+      if (environment) {
+        if (modalRef.current && isOpen) {
+          if (restoreFocus === undefined ? true : restoreFocus) {
+            triggerRef.current = activeElement(environment) as HTMLElement;
+          }
+
+          if (focusOnMount === undefined ? true : focusOnMount) {
```
Francois-Esquire, line 75: Both `restoreFocus` and `focusOnMount` revert to `true` when undefined to preserve the existing behavior that the `useFocusJail` + `useModal` containers had in place.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -22,6 +22,7 @@ import { ThemeContext } from 'styled-components';
 import { useModal } from '@zendeskgarden/container-modal';
 import { useDocument, useText } from '@zendeskgarden/react-theming';
 import { useFocusVisible } from '@zendeskgarden/container-focusvisible';
+import activeElement from 'dom-helpers/activeElement';
```
Francois-Esquire, line 25: We may need to look into filtering out nested paths in our `external` config for `rollup` since this dependency is being bundled.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
Francois-Esquire: Alternatively, we can use defaultProps in this case if we think thats best.

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -176,4 +176,20 @@ describe('DrawerModal', () => {
...
+    await user.click(getByText('Open Drawer'));
+
+    expect(getByRole('dialog')).toStrictEqual(document.activeElement);
+
+    await user.type(getByRole('dialog'), '{escape}');
+
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+
+    expect(getByText('Open Drawer')).toStrictEqual(document.activeElement);
+  });
```
geotrev, line 194: This can probably be two tests with a `beforeEach` that opens the dialog.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -49,12 +58,32 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
+        restoreFocus: false /* [1:b] */,
         environment,
         onClose
       });
 
+    /* [1:c] */
+    useEffect(() => {
+      if (environment) {
+        if (modalRef.current && isOpen) {
+          if (restoreFocus === undefined ? true : restoreFocus) {
```
geotrev, line 71: nit: `restoreFocus === undefined || restoreFocus` also works

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
geotrev: Default props makes sense to me as long as this prop reference is consistent everywhere else too.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -49,12 +58,32 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
+            triggerRef.current = activeElement(environment) as HTMLElement;
+          }
+
+          if (focusOnMount === undefined ? true : focusOnMount) {
+            modalRef.current.focus();
+          }
+        }
+
+        if (triggerRef.current && !isOpen) {
```
geotrev, line 80: What happens if `modalRef`, `triggerRef`, `restoreFocus`, or `focusOnMount` change but neither `triggerRef` nor `isOpen` does and this condition still passes?

geotrev: I'm also wondering if this should include a `restoreFocus` check.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
Francois-Esquire: Agreed, will update.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
Francois-Esquire: True, good call. We probably wouldn't need this if we use defaultProps though and it would just be `if (restoreFocus)`.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
Francois-Esquire: > What happens if `modalRef`, `triggerRef`, `restoreFocus`, or `focusOnMount` change but neither `triggerRef` nor `isOpen` does and this condition still passes?

That is a good question, let me confirm with additional test cases.

> I'm also wondering if this should include a `restoreFocus` check.

I don't think we necessarily need it. In this case, `triggerRef.current` would only be assigned if `restoreFocus` is `true`, which is why it's not necessary to check since the triggerRef indicates that focus was moved away from it when the modal was opened.

geotrev: I see - so if another ref changes for some reason, it's pretty safe to assume it's due to modal closing and not something else?

Francois-Esquire: In this case, the `isOpen` prop is what mainly drives this effect and as that is toggled, it would used to point the focus where it should be. With React refs are attached before effects are ran so we can safely assume it is present when `CSSTransition` mounts the `StyledDrawer` to the DOM. I realized after we don't need to add the refs to the dependency array so I removed them.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -162,6 +196,11 @@ DrawerModalComponent.propTypes = {
   isOpen: PropTypes.bool
 };
 
+DrawerModalComponent.defaultProps = {
+  focusOnMount: true,
+  restoreFocus: true
```
Francois-Esquire, line 201: Compare to https://github.com/zendeskgarden/react-components/blob/main/packages/modals/src/elements/TooltipModal/TooltipModal.tsx#L158-L159

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
geotrev: I would be surprised if rollup doesn't have a mechanism for this already - export paths are a native feature nowadays.

geotrev (approved):
Looks sane to me - I didn't spin up the demo yet though. I'll do that in a minute :)

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -162,6 +197,11 @@ DrawerModalComponent.propTypes = {
   isOpen: PropTypes.bool
 };
 
+DrawerModalComponent.defaultProps = {
+  focusOnMount: true /* [1:d] */,
+  restoreFocus: true /* [1:d] */
```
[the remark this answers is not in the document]
Francois-Esquire: Yes, the default behavior that was inherited from `useModal` (and `useFocusJail`) was to set both `focusOnMount` and `restoreFocus` to `true` when it was `undefined`. So this does align to that and should not cause any breaking change in behavior.

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
```diff
@@ -176,4 +176,28 @@ describe('DrawerModal', () => {
...
     await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
   });
+
+  describe('focus management', () => {
+    it('correctly focuses on the Drawer when open', async () => {
+      const { getByText, getByRole } = render(<Example />);
+
+      await user.click(getByText('Open Drawer'));
+
+      expect(getByRole('dialog')).toStrictEqual(document.activeElement);
```
jzempel, line 186:
```suggestion
      expect(getByRole('dialog')).toBe(document.activeElement);
```

suggested [nit] applies to changes below as well.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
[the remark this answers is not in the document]
jzempel: In that case, clarifying the `defaultProps` should also be done on the other `Modal` components. I'd recommend handling that under a separate PR.

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
```diff
@@ -49,12 +59,37 @@ const DrawerModalComponent = forwardRef<HTMLDivElement, IDrawerModalProps>(
...
+          triggerRef.current.focus();
+        }
+      }
+
+      return () => {
+        if (!(restoreFocus && isOpen)) {
+          triggerRef.current = null;
+        }
+      };
+    }, [environment, restoreFocus, focusOnMount, isOpen]);
```
jzempel, line 91: I'm curious, why was `useFocusJail` not enhanced accordingly? It seems like the logic would be more beneficial down there.

jzempel (approved):

packages/modals/src/elements/DrawerModal/DrawerModal.tsx
Francois-Esquire: That was the initial intention and as we discussed offline, that is definitely a future goal.

packages/modals/src/elements/DrawerModal/DrawerModal.spec.tsx
Francois-Esquire: Updated!

PR #1516, opened 2023-02-14
fix(tables): hide content of Table cell without breaking table layout

geotrev:
## Description

Inadvertently introduced a style regression in `Cell` (react-tables) by using visually hidden styles when the `hidden` prop is used. Namely, CSS properties like `position`, `width`, and removal of `padding` when the cell is a header causes column width issues.

This fix conditionally renders an inner `div` with visually hidden styles instead of manipulating the outer `td`/`th`.

## Detail

*Note: the left column in the below images has a static width, which was being overridden by visually hidden styles*

Before:
Screen Shot 2023-02-13 at 5 59 18 PM

After:
Screen Shot 2023-02-13 at 5 57 20 PM

There are really two ways to go about this fix:
1. Use an inner element with hidden styles
2. Apply a `text-indent` with `overflow: hidden` to push content out of the content box.

There are trade offs to both approaches, namely that 1 adds a new element, and 2 has page flow implications.

I went with 1 because it is somewhat more bullet proof in that a consumer won't be able to change the visibility by applying non-theme spacing (padding), which would break the effect if we went with 2.

For folks who don't want to nest content inside a cell but still have accessible text, they can still add `aria-label`.

## Checklist

- [x] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~

### DIFF
--- a/packages/tables/src/elements/Cell.tsx
+++ b/packages/tables/src/elements/Cell.tsx
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import { hideVisually } from 'polished';
 import { StyledCell } from '../styled';
 import { useTableContext } from '../utils/useTableContext';
 import { ICellProps } from '../types';
@@ -14,11 +15,17 @@ import { ICellProps } from '../types';
 /**
  * @extends TdHTMLAttributes<HTMLTableCellElement>
  */
-export const Cell = React.forwardRef<HTMLTableCellElement, ICellProps>((props, ref) => {
-  const { size } = useTableContext();
+export const Cell = React.forwardRef<HTMLTableCellElement, ICellProps>(
+  ({ hidden, ...props }, ref) => {
+    const { size } = useTableContext();
 
-  return <StyledCell ref={ref} size={size} {...props} />;
-});
+    return (
+      <StyledCell ref={ref} size={size} {...props}>
+        {hidden ? <div style={hideVisually()}>{props.children}</div> : props.children}
+      </StyledCell>
+    );
+  }
+);
 
 Cell.displayName = 'Cell';
 
--- a/packages/tables/src/elements/Cell.tsx
+++ b/packages/tables/src/elements/Cell.tsx
@@ -7,8 +7,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { hideVisually } from 'polished';
-import { StyledCell } from '../styled';
+import { StyledCell, StyledHiddenCell } from '../styled';
 import { useTableContext } from '../utils/useTableContext';
 import { ICellProps } from '../types';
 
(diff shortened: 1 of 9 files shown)


packages/tables/src/styled/StyledCell.ts
```diff
@@ -68,10 +68,6 @@ export const StyledCell = styled.td.attrs<IStyledCellProps>({
   ${props => sizeStyling(props)};
   ${props => props.isTruncated && truncatedStyling};
 
-  &[hidden] {
-    ${hideVisually()}
-  }
```
jzempel, line 73: Let's keep this here rather than spreading the logic around to element cell `style`. Use something like `&[hidden] > div` or similar.

jzempel (changes requested):

packages/tables/src/styled/StyledHiddenCell.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { hideVisually } from 'polished';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+
+const COMPONENT_ID = 'tables.cell';
```
jzempel, line 12:
```suggestion
const COMPONENT_ID = 'tables.hidden_cell';
```

This currently conflicts with `StyledCell`. True story: we used to keep a custom linter to ensure unique `COMPONENT_ID`s, but it got too unwieldy to maintain.

jzempel (changes requested):

packages/tables/src/styled/StyledHiddenCell.ts
geotrev: Fixed 🚀

jzempel (approved):

packages/tables/src/elements/Cell.tsx
```diff
@@ -7,18 +7,28 @@
...
-  const { size } = useTableContext();
+export const Cell = React.forwardRef<HTMLTableCellElement, ICellProps>(
+  ({ hidden, ...props }, ref) => {
+    const { size } = useTableContext();
 
-  return <StyledCell ref={ref} size={size} {...props} />;
-});
+    return (
+      <StyledCell ref={ref} size={size} {...props}>
+        {hidden && props.children ? (
```
Francois-Esquire, line 23: #nitpick, is the `&& props.children` necessary here?

Francois-Esquire (approved):
Looks solid!

packages/tables/src/elements/Cell.tsx
geotrev: If a consumer decides to use an `aria-label` on the cell instead of children, this prevents an empty `div`.

PR #1517, opened 2023-02-15
fix(forms): reset margin start for text input in input group

Francois-Esquire:
## Description

Resets the start margin of the `TextInput` styled component in an `InputGroup`, fixing a bug where `overflow: hidden` in a parent element will display the `Input` incorrectly.

## Detail

Before:
Screenshot 2023-02-15 at 3 38 03 PM

After:
Screenshot 2023-02-15 at 3 38 09 PM

## Checklist

### DIFF
--- a/packages/forms/src/styled/input-group/StyledInputGroup.ts
+++ b/packages/forms/src/styled/input-group/StyledInputGroup.ts
@@ -23,8 +23,6 @@ interface IStyledInputGroupProps {
  */
 const positionStyles = (props: ThemeProps<DefaultTheme> & IStyledInputGroupProps) => {
   const topMargin = `${props.theme.space.base * (props.isCompact ? 1 : 2)}px`;
-  const startDirection = props.theme.rtl ? 'right' : 'left';
-  const endDirection = props.theme.rtl ? 'left' : 'right';
 
   return css`
     /* stylelint-disable */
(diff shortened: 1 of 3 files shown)


packages/forms/src/styled/input-group/StyledInputGroup.ts
```diff
@@ -79,6 +80,10 @@ const itemStyles = (props: ThemeProps<DefaultTheme>) => {
       z-index: -1;
     }
 
+    & > ${StyledTextInput}:first-child {
```
jzempel, line 83:
```suggestion
    & > *:first-child {
```

This override is important no matter what kind of component (`Input`, `Button`) is first.

packages/forms/src/styled/input-group/StyledInputGroup.ts
```diff
@@ -79,6 +80,10 @@ const itemStyles = (props: ThemeProps<DefaultTheme>) => {
...
+      margin-${horizontal}: 0 !important; /* [2] */
+    }
+
```
jzempel, line 86: Is this entire block necessary, or does `!important` just need to be added to the `& > *:first-child:not(:last-child)` block below?

packages/forms/src/styled/input-group/StyledInputGroup.ts
```diff
@@ -79,6 +80,10 @@ const itemStyles = (props: ThemeProps<DefaultTheme>) => {
       z-index: -1;
     }
 
+    & > ${StyledTextInput}:first-child {
+      margin-${horizontal}: 0 !important; /* [2] */
```
geotrev, line 84: Would moving this to the end of the style cascade make `!important` unnecessary? 🤔

jzempel (changes requested):

packages/forms/src/styled/input-group/StyledInputGroup.ts
Francois-Esquire: Just updated changes to remove `!important` altogether.

jzempel (approved):
Checked with clipped demo in both LTR/RTL – looks solid.

PR #1518, opened 2023-02-16
fix(buttons): correct icon button icon animation transitioning

Francois-Esquire:
## Description

Synchronizes the color change for icons within the `IconButton` to change at the same time the transition starts.

## Detail

Due to how `currentColor` and the CSS transition spec outlines, computed values will be delayed which initially caused the SVG color change to delay until the color value of the parent element was done.

## Checklist

### DIFF
--- a/packages/buttons/src/styled/StyledIcon.ts
+++ b/packages/buttons/src/styled/StyledIcon.ts
@@ -41,7 +41,7 @@ export const StyledIcon = styled(({ children, isRotated, theme, ...props }) =>
   'data-garden-version': PACKAGE_VERSION
 })<IStyledIconProps>`
   transform: ${props => props.isRotated && `rotate(${props.theme.rtl ? '-' : '+'}180deg)`};
-  transition: transform 0.25s ease-in-out, color 0.25s ease-in-out;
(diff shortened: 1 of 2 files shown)


packages/buttons/src/styled/StyledIcon.ts
```diff
@@ -41,7 +41,7 @@ export const StyledIcon = styled(({ children, isRotated, theme, ...props }) =>
   'data-garden-version': PACKAGE_VERSION
 })<IStyledIconProps>`
   transform: ${props => props.isRotated && `rotate(${props.theme.rtl ? '-' : '+'}180deg)`};
-  transition: transform 0.25s ease-in-out, color 0.25s ease-in-out;
+  transition: transform 0.25s ease-in-out;
```
jzempel, line 44: This knocks out the intended color transition for the icons. The problem can be seen more clearly with `disable` toggling. See attached...

transition

jzempel (changes requested):

Francois-Esquire:
Closing until we come up with the right solution to correctly transition the color in all button states.

PR #1542, opened 2023-05-01
feat(theming): add focus styling utilities

jzempel:
## Description

This PR adds `xs` units to the `DEFAULT_THEME` to account for updated "spacer" focus styling, along with three new utilities:

- `focusStyles` (preferred) provides a CSS block with focus `outline` reset, focus-visible `box-shadow` styling, and [transparent outline hack](https://css-tricks.com/quick-tips-for-high-contrast-mode/) for high contrast mode
- `getFocusBoxShadow` (fallback) in the case where existing code would be a tedious/surgical refactor, this utility provides the CSS value for the `box-shadow` property
- `SELECTOR_FOCUS_VISIBLE` constant that can be used together with the `selector` parameter on `focusStyles`:

```js
focusStyles({ theme: props.theme, selector: `&:focus-within, ${SELECTOR_FOCUS_VISIBLE}` });
```

## Detail

For performance concerns, the `getColor` utility was memoized, resulting in an extensive diff due to the `memoize` function wrap.

This PR contains two commits (rolled back), demonstrating real utility usage within a couple of key styled components:
- [StyledTextInput](https://github.com/zendeskgarden/react-components/pull/1542/commits/d1ea6a352faf989f4e98a11aba4d778fc5e4fa5f) using `focusStyles` (note that the existing `:focus` block should've also been removed)
- [StyledButton](https://github.com/zendeskgarden/react-components/pull/1542/commits/290b4f6e1c08cc377515f58eb302ea78bb3e1454) using `getFocusBoxShadow` (`focusStyles` could be applied with extra effort)

## Checklist

### DIFF
--- a/packages/theming/src/utils/focusStyles.ts
+++ b/packages/theming/src/utils/focusStyles.ts
@@ -1,0 +1,38 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { css } from 'styled-components';
+import { IGardenTheme } from '../types';
+import DEFAULT_THEME from '../elements/theme';
+import { getColor, DEFAULT_SHADE, Hue } from './getColor';
+
+type FocusStyles = {
+  focusInset?: boolean;
+  hue?: Hue;
+  shade?: number;
+  shadowWidth?: 'sm' | 'md';
+  spacerWidth?: 'xs' | 'sm';
+  theme?: IGardenTheme;
+};
+
+export const focusStyles = ({
+  focusInset = false,
+  hue = 'primaryHue',
+  shade = DEFAULT_SHADE,
+  shadowWidth = 'md',
+  spacerWidth = 'xs',
+  theme = DEFAULT_THEME
+}: FocusStyles) => {
+  const color = getColor(hue, shade, theme);
+  const boxShadow = `
+    ${focusInset ? 'inset' : ''} ${theme.shadows[spacerWidth](theme.colors.background)},
+    ${focusInset ? 'inset' : ''} ${theme.shadows[shadowWidth](color!)}`;
+
+  return css`
+    box-shadow: ${boxShadow};
+  `;
+};
--- a/packages/theming/src/utils/focusStyles.ts
+++ b/packages/theming/src/utils/focusStyles.ts
@@ -9,8 +9,11 @@ import { css, CSSObject } from 'styled-components';
 import { math } from 'polished';
 import { FocusBoxShadowParameters, getFocusBoxShadow } from './getFocusBoxShadow';
 
+export const SELECTOR_FOCUS_VISIBLE = '&:focus-visible, &[data-garden-focus-visible="true"]';
+
 type FocusStylesParameters = FocusBoxShadowParameters & {
   condition?: boolean;
+  selector?: string;
   styles?: CSSObject;
 };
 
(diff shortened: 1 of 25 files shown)


packages/theming/src/utils/getColor.ts
```diff
@@ -37,57 +38,53 @@ const adjust = (color: string, expected: number, actual: number) => {
...
-) {
-  let retVal;
-
-  if (isNaN(shade)) {
-    return undefined;
-  }
-
-  const palette = theme && theme.palette ? theme.palette : DEFAULT_THEME.palette;
-  const colors = theme && theme.colors ? theme.colors : DEFAULT_THEME.colors;
+export const getColor = memoize(
```
Francois-Esquire, line 42: Were there any performance concerns that prompted the `lodash.memoize` utility usage?

packages/theming/src/utils/focusStyles.ts
```diff
@@ -0,0 +1,77 @@
...
+ *   box-shadow: 0 0 0 {1px} #fff,
+ *               0 0 0 {3px} {blue};
+ *   outline: {2px} solid transparent;
+ *   outline-offset: {1px};
+ *   // additional {styles} here...
+ * }
+ * ```
+ */
+export const focusStyles = ({
+  condition = true,
```
Francois-Esquire, line 46: I'm curious, what advantage does the `condition` argument bring? The `condition` argument feels forced and can be done in the context of the code where needed (which reads clearer and less surface area to debug).

packages/theming/src/utils/focusStyles.ts
```diff
@@ -0,0 +1,77 @@
...
+  const outline = `${math(
+    `${theme.shadowWidths[shadowWidth]} - ${theme.shadowWidths[spacerWidth]}`
+  )} solid transparent`;
+  const boxShadow = condition
+    ? getFocusBoxShadow({ shadowWidth, spacerWidth, theme, ...options })
+    : undefined;
+
+  /*
+   * 1. Browser reset
+   * 2. High contrast mode hack
```
Francois-Esquire, line 62: - Can you please share your resource/research on the high contrast mode hack? Very curious to learn more about this approach.
- Will there need to be other parts to this mechanism? Maybe an extension to the theme colors in the future to set the outline color w/ query?

packages/theming/src/utils/focusStyles.ts
jzempel: See [commit](https://github.com/zendeskgarden/react-components/pull/1542/commits/d1ea6a352faf989f4e98a11aba4d778fc5e4fa5f) linked in description

packages/theming/src/utils/focusStyles.ts
jzempel: See [link](https://css-tricks.com/quick-tips-for-high-contrast-mode/) in description

packages/theming/src/utils/getColor.ts
jzempel: Yes, calling out to these utilities results in repeat `getColor` calls in a lot of existing cases. But I think there's a massive savings potential overall.

geotrev (commented):
Nothing major to comment on. I dig it.

geotrev (approved):

packages/theming/src/utils/getColor.ts
```diff
@@ -37,57 +38,53 @@ const adjust = (color: string, expected: number, actual: number) => {
...
-  }
+    if (transparency) {
+      retVal = rgba(retVal, transparency);
+    }
 
-  return retVal;
-}
+    return retVal;
+  },
+  (...args) => JSON.stringify(args)
```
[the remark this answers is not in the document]
jzempel: Yeah, I'm also concerned about this resolver as it pertains to the `theme` object. I pushed in on this pretty hard, and profiling seems to indicate the performance is plenty fast. I'm of two minds here 1) rollback the update, 2) push the update and be prepared to rollback if there are unintended performance consequences. I'm leaning toward 2.

packages/theming/src/utils/focusStyles.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { css, CSSObject } from 'styled-components';
+import { math } from 'polished';
+import { FocusBoxShadowParameters, getFocusBoxShadow } from './getFocusBoxShadow';
+
+export const SELECTOR_FOCUS_VISIBLE = '&:focus-visible, &[data-garden-focus-visible="true"]';
```
geotrev, line 12: `:focus-visible` 🚀 🚀

packages/theming/src/utils/getColor.ts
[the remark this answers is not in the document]
geotrev: Seems fair to me. Knowing the theme object is pretty shallow also mitigates my deepest concern anyway (no pun intended).

Francois-Esquire (approved):
Looks sharp!

PR #1548, opened 2023-05-09
feat(components): use new focus styles for interactive elements

Francois-Esquire:
## Description

This PR introduces new focus ring styling using the `focusStyles` utility.

## Detail

Interactive components throughout the component library have been updated with the new focus ring. 

#### sweeping changes and additions
- transition for focus ring has been added to all affected components

#### focus with `outline` vs `box-shadow`
While most components use the conventional `box-shadow` focus ring, there are some exceptions:
- `Anchor`: `outline` was used to better visualize wrapping anchor text when encountering a line break.
- `NavItem`: `outline` worked with the transparent styling within the component

#### focusing within
We want to move components away from the `isFocused` prop in favor of modern CSS focus selectors (eg `:focus-within`) where possible. The `Tiles.Tile` component was updated with that in mind. This was made possible with the `:focus-within` CSS selector to allow these wrapping components to display focus when their children receive it.

Other components that were not affected:
- `Accordion.Header`: `:focus-within` is being used, however `isFocused` is needed due to nested interactive elements.
- `FauxInput`: due to `isFocused` being a public API, the prop was not removed, however `:focus-within` is used.

#### components not using the `focusStyles` utility
- `colorpickers` color range
-  `forms` `Range` component

#### Accordion
Screenshot 2023-05-09 at 3 42 09 PM

#### Forms

Screenshot 2023-05-09 at 3 42 34 PM
Screenshot 2023-05-09 at 3 42 43 PM
Screenshot 2023-05-09 at 3 43 47 PM

##### text Input
Screenshot 2023-05-09 at 3 44 05 PM
Screenshot 2023-05-09 at 3 49 21 PM
Screenshot 2023-05-09 at 3 48 40 PM
Screenshot 2023-05-09 at 3 48 53 PM
Screenshot 2023-05-09 at 3 48 27 PM

## Checklist


Francois-Esquire:
Thanks for the feedback @lucijanblagonic !
- `Breadcrumb` demo now allows the `Anchor`s to receive focus
- `Button` in danger state now has correct border color
- `Tag` border size has been adjusted
- `Select` now has the correct focus inset styles
- `Multiselect` has been fixed to remove the inner focus
- `SkipNavLink` underline has been removed when focused

A few of the items that include the `Chrome` header navigation items and `SplitButton` did raise some concerns that I would appreciate your input on. Will reach out offline!

packages/forms/src/styled/text/StyledTextFauxInput.ts
```diff
@@ -18,6 +24,26 @@ export interface IStyledTextFauxInputProps extends IStyledTextInputProps {
   isReadOnly?: boolean;
 }
 
+export function getValidationHue(
```
Francois-Esquire, line 27: This local utility can be used across forms components. As of now it is a style helper for now.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -23,11 +28,10 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
...
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${props.theme.shadows.md(
-        getColor(backgroundColor, 600, props.theme, 0.35) as string
-      )};
-    }
+    ${focusStyles({
+      theme: props.theme,
+      hue: backgroundColor
+    })}
```
geotrev, line 34:
```suggestion
    ${focusStyles({ theme: props.theme, hue: backgroundColor })}
```

packages/grid/src/styled/pane/StyledPaneSplitterButton.ts
```diff
@@ -49,14 +54,14 @@ const colorStyles = ({ theme }: IStyledSplitterButtonProps & ThemeProps<DefaultT
...
   return css`
     box-shadow: ${boxShadow};
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${focusBoxShadow}, ${boxShadow};
-    }
+    ${focusStyles({
+      theme,
+      boxShadow
+    })}
```
geotrev, line 64:
```suggestion
    ${focusStyles({ theme, boxShadow })}
```

packages/forms/src/styled/text/StyledTextFauxInput.ts
```diff
@@ -32,8 +58,24 @@ export const StyledTextFauxInput = styled(
   cursor: ${props => (props.mediaLayout && !props.isDisabled ? 'text' : 'default')};
   overflow: hidden;
 
+  ${props =>
+    focusStyles({
+      theme: props.theme,
+      inset: props.focusInset,
+      // removes focus ring when `isBare`
```
geotrev, line 65: Unnecessary comment. :)

packages/forms/src/styled/range/StyledSliderThumb.ts
```diff
@@ -38,16 +42,16 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
...
-
     &:hover,
     &[data-garden-hover='true'] {
       border-color: ${hoverBorderColor};
       background-color: ${hoverBackgroundColor};
     }
 
+    ${focusStyles({
+      theme: props.theme
+    })}
```
geotrev, line 53: Destructuring `theme` would condense further.

```suggestion
    ${focusStyles({ theme: props.theme })}
```

packages/forms/src/styled/file-upload/StyledFileUpload.ts
```diff
@@ -43,9 +43,10 @@ const colorStyles = (props: ThemeProps<DefaultTheme> & IStyledFileUploadProps) =
       color: ${hoverColor};
     }
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${boxShadow};
-    }
+    ${focusStyles({
+      theme: props.theme,
+      hue: baseColor
+    })}
```
geotrev, line 49:
```suggestion
    ${focusStyles({ theme: props.theme, hue: baseColor })}
```

packages/chrome/src/styled/header/StyledHeaderItem.ts
```diff
@@ -44,14 +49,13 @@ export const StyledHeaderItem = styled(StyledBaseHeaderItem as 'button').attrs({
...
-  }
-
-  &[data-garden-focus-visible] {
-    box-shadow: ${props => props.theme.shadows.md(getColor('chromeHue', 400, props.theme, 0.35)!)};
-  }
+  ${props =>
+    focusStyles({
+      theme: props.theme,
+      inset: props.maxY
+    })}
```
geotrev, line 56:
```suggestion
  ${p => focusStyles({ theme: p.theme, inset: props.maxY })}
```

packages/chrome/src/styled/sheet/StyledSheetClose.ts
```diff
@@ -29,11 +34,9 @@ const colorStyles = (props: ThemeProps<DefaultTheme>) => {
...
     }
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${props.theme.shadows.md(
-        getColor(backgroundColor, 600, props.theme, 0.35) as string
-      )};
-    }
+    ${focusStyles({
+      theme: props.theme
+    })}
```
geotrev, line 39:
```suggestion
    ${focusStyles({ theme: props.theme })}
```

packages/colorpickers/src/styled/ColorSwatch/StyledSwatchButton.ts
```diff
@@ -28,10 +28,10 @@ export const StyledSwatchButton = styled(StyledButtonPreview).attrs<IStyleButton
...
   padding: 0;
 
-  &[data-garden-focus-visible] {
-    box-shadow: ${props =>
-      props.theme.shadows.md(getColor('primaryHue', 600, props.theme, 0.35) as string)};
-  }
+  ${props =>
+    focusStyles({
+      theme: props.theme
+    })}
```
geotrev, line 34:
```suggestion
  ${p => focusStyles({ theme: p.theme })}
```

packages/chrome/src/styled/StyledSkipNav.ts
```diff
@@ -57,6 +59,12 @@ const colorStyles = (theme: DefaultTheme) => {
     &:focus {
       color: ${color};
     }
+
+    ${focusStyles({
+      theme,
+      inset: true,
+      boxShadow
+    })}
```
geotrev, line 67:
```suggestion
    ${focusStyles({ theme, inset: true, boxShadow })}
```

packages/accordions/src/styled/accordion/StyledHeader.ts
```diff
@@ -22,15 +21,19 @@ export const StyledHeader = styled.div.attrs<IStyledHeader>({
...
   &:hover {
     cursor: ${props => (props.isCollapsible || !props.isExpanded) && 'pointer'};
   }
 
+  ${props =>
+    focusStyles({
+      theme: props.theme,
+      inset: true,
+      selector: '&:focus-within'
+    })}
```
geotrev, line 35:
```suggestion
  ${p => focusStyles({ theme: p.theme, inset: true, selector: '&:focus-within' })}
```

geotrev (changes requested):
Very minor style changes/suggestions and a few open questions in areas. Going to spin up the branch and test locally next.

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -192,47 +209,46 @@ const groupStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
...
-      border-top-width: 0;
-      border-bottom-width: 0;
+    }
+
+    /* stylelint-disable property-no-unknown, property-case */
+    &:not(:first-of-type) {
+      /* prettier-ignore */
+      margin-${rtl ? 'right' : 'left'}: ${
+    isPrimary ? props.theme.borderWidths.sm : `-${props.theme.borderWidths.sm}`
+  };
```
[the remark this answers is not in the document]
Francois-Esquire: I was hoping prettier would ignore that line. Will clean this up.

packages/chrome/src/styled/StyledSkipNav.ts
```diff
@@ -102,8 +110,8 @@ export const StyledSkipNav = styled.a.attrs({
 
   ${props => sizeStyles(props)};
 
-  &:focus {
-    outline: none;
+  ${SELECTOR_FOCUS_VISIBLE} {
+    text-decoration: none;
```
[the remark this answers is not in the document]
Francois-Esquire: The significance here is that we did not want the `SkipNav` link to be underlined when focused. and the `focusStyles` utility adds `outline: none` for the `:focus` selector.

packages/drag-drop/src/styled/draggable/StyledDraggable.ts
```diff
@@ -83,9 +73,13 @@ const colorStyles = (props: IStyledDraggableProps) => {
       background-color: ${hoverBackgroundColor};
     }
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${focusBoxShadow};
-    }
+    ${focusStyles({
+      theme: props.theme,
+      inset: focusInset,
+      // removes focus ring when `isBare`
```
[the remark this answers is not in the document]
Francois-Esquire: Agreed, on second read, the code speaks for itself. I'll remove where I placed these.

packages/chrome/src/styled/header/StyledHeaderItem.ts
```diff
@@ -64,25 +68,6 @@ export const StyledHeaderItem = styled(StyledBaseHeaderItem as 'button').attrs({
...
-        ${props.theme.shadowWidths.md} 0 0 0 ${getColor('chromeHue', 400, props.theme, 0.35)},
-        inset ${props.theme.shadows.lg(
-          `-${props.theme.shadowWidths.md}`,
-          '0',
-          getColor('chromeHue', 400, props.theme, 0.35)!
-        )},
-        -${props.theme.shadowWidths.md} 0 0 0 ${getColor('chromeHue', 400, props.theme, 0.35)};
-      }
-  `}
-
```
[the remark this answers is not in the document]
Francois-Esquire: Indeed, with the `maxY` prop set, we want to display an inset focus ring which the `focusStyles` utility handles.

packages/chrome/src/styled/StyledSkipNav.ts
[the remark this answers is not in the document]
geotrev: Oh I see. This makes total sense.

packages/accordions/src/styled/accordion/StyledHeader.spec.tsx
```diff
@@ -7,24 +7,38 @@
...
 
     expect(container.firstChild).not.toHaveStyleRule('box-shadow');
   });
 
-  it('renders isFocused styling correctly', () => {
-    const { container } = render(<StyledHeader isFocused />);
+  it('renders focus styling correctly', async () => {
+    const { container } = render(
+      <StyledHeader>
+        <button />
```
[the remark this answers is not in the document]
geotrev: Got it OK. And in Garden's implementation that inner button/focusable element is an implementation detail (if memory serves) similar to how button is needed in this test, right?

geotrev:
Accordion focus ring appears duplicated when tabbing to inner icons:

branch-acc

main-acc

packages/forms/src/styled/range/StyledSliderThumb.ts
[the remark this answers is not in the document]
Francois-Esquire: Yes! So we ideally want to follow "LVHFA" ordering (`hover`, `focus`, `active`) ([found here](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules)) and I took the opportunity to reorder the files I've touched being that it involves `focus`.

geotrev:
I noticed a few cases where the focus box shadow animates in/out for some components but not others. `Anchor` for instance doesn't seem to animate focus. If I find others I'll note them here.

packages/grid/src/styled/pane/StyledPaneSplitter.ts
```diff
@@ -175,6 +182,10 @@ export const StyledPaneSplitter = styled.div.attrs({
     content: '';
   }
 
+  ${SELECTOR_FOCUS_VISIBLE} {
+    z-index: 2;
```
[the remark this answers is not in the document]
Francois-Esquire: This was due to overlap of other `Pane.Splitter`s covering the focus ring when a given `Splitter` received focus. This was an existing issue prior as seen here:
Screenshot 2023-05-16 at 3 47 24 PM

packages/modals/src/styled/StyledClose.ts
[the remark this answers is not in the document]
Francois-Esquire: Correct, the `focusStyles` utility actually adds:
```css
&:focus {
  outline: none;
}
```
along with the focus ring and HCM styling. So it is safe to replace it in place and preferable for the sake of LVHFA.
Another reason we willingly replace `data-garden-focus-visible` is because `focusStyles` uses it and results in the same code (along with adding the native CSS selector `:focus-visible`).

packages/grid/src/styled/pane/StyledPaneSplitter.ts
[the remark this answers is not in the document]
Francois-Esquire: I will add an inline comment for explanation.

packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts
```diff
@@ -63,9 +68,11 @@ function colorStyles(props: ThemeProps<DefaultTheme> & IStyledGlobalAlertButtonP
       background-color: ${hoverBackgroundColor};
     }
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${boxShadowColor && props.theme.shadows.md(boxShadowColor)};
-    }
+    ${focusStyles({
+      theme: props.theme,
+      hue: focusColor,
+      shade: props.alertType === 'info' ? 600 : 800
```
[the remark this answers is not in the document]
Francois-Esquire: Yes it does vary across the board, maybe @lucijanblagonic or @steue would have more insight into this?

packages/forms/src/utils/getValidationHue.ts
```diff
@@ -0,0 +1,21 @@
...
+    case 'success':
+      return 'successHue';
+    case 'warning':
+      return 'warningHue';
+    case 'error':
+      return 'dangerHue';
+    default:
+      return defaultHue;
+  }
+}
```
jzempel, line 21:
```suggestion
const VALIDATION_HUES = {
  success: 'successHue',
  warning: 'warningHue',
  error: 'dangerHue'
}

function getValidationHue(validation?: Validation, defaultHue = 'primaryHue') {
  if (validation) {
    return VALIDATION_HUES[validation];
  }
  
  return defaultHue;
  }
}
```

Since this is only used by StyledFauxInput, I would recommend moving it locally to that file.

packages/forms/src/styled/select/StyledSelect.ts
```diff
@@ -18,9 +18,11 @@ const colorStyles = (props: IStyledTextInputProps & ThemeProps<DefaultTheme>) =>
 
   /* prettier-ignore */
   return css`
+    /* stylelint-disable-next-line no-duplicate-selectors */
     &:hover + ${StyledTextMediaFigure},
     &:focus + ${StyledTextMediaFigure},
-    &[data-garden-focus-visible='true'] + ${StyledTextMediaFigure} {
+    &:focus-visible + ${StyledTextMediaFigure},
+    &[data-garden-focus-visible='true'] ${StyledTextMediaFigure} {
```
jzempel, line 25:
```suggestion
    &[data-garden-focus-visible='true'] + ${StyledTextMediaFigure} {
```

looks like this changed ... intentionally?

packages/chrome/src/styled/subnav/StyledSubNavItem.tsx
```diff
@@ -7,42 +7,44 @@
...
       background-color: ${hoverColor};
     }
 
-    &[data-garden-focus-visible] {
-      opacity: 1;
-      box-shadow: ${props.theme.shadows.md(focusColor)};
-    }
+    ${focusStyles({
+      theme,
+      hue: isLight ? DARK : LIGHT,
```
jzempel, line 44: Focus styling for collapsible subnav items with children appears to be broken.

packages/chrome/src/styled/StyledSkipNav.ts
jzempel, line 114: This has a similar problem as my comment re: anchors, in general. A hover should display the underline.

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -337,17 +353,17 @@ export const StyledButton = styled.button.attrs<IButtonProps>(props => ({
     padding: 0;
   }
 
+  &:hover {
+    text-decoration: ${props => (props.isLink ? 'underline' : 'none')}; /* [2] */
+  }
+
```
jzempel, line 359: While this shuffle honors LVHFA, it represents a change in CSS logic and the text is no longer showing underline on hover when focused.

jzempel (commented):
Anchor styling across the board is broken with respect to Garden system rules for state styling. That is:
- hover should show hover state
- focus should show focus state
- hover+focus should show both

This should work with/without Bedrock CSS (and https://zendeskgarden.github.io/css-components/?path=/story/components-bedrock--bedrock probably needs to be updated as a follow-on).

jzempel:
> Here's a list of non-animating focus rings but there might be a few more. It doesn't seem common which is why I think it could be a CSS error/bug.

Agree with @geotrev, in general on getting the transition animations right. However, please don't transition `ColorSwatch`. Due to the proximity of the controls, we determined that a focus transition made the whole component feel too sluggish.

packages/accordions/src/styled/accordion/StyledHeader.spec.tsx
```diff
@@ -7,24 +7,38 @@
...
+
+    await user.tab();
 
     expect(container.firstChild).toHaveStyleRule(
       'box-shadow',
-      `${DEFAULT_THEME.shadows.md(
-        getColor('primaryHue', 600, DEFAULT_THEME, 0.35) as string
-      )} inset`
+      getFocusBoxShadow({ theme: DEFAULT_THEME, inset: true })
+        // normalize string output to match
```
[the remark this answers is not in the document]
Francois-Esquire: Makes sense, I'll update!

Francois-Esquire:
> There's a handful of components that don't animate their focus ring, but the vast majority of Garden components _do_ animate.
> 
> Here's a list of non-animating focus rings but there might be a few more. It doesn't seem common which is why I think it could be a CSS error/bug.
> 
> * Breadcrumbs
> * ColorSwatch
> * Alert (close button)
> * Notification (close button)



> > Here's a list of non-animating focus rings but there might be a few more. It doesn't seem common which is why I think it could be a CSS error/bug.
> 
> Agree with @geotrev, in general on getting the transition animations right. However, please don't transition `ColorSwatch`. Due to the proximity of the controls, we determined that a focus transition made the whole component feel too sluggish.

That makes sense, I will add the missing transitions here to unify the experience.

packages/forms/src/styled/select/StyledSelect.ts
Francois-Esquire: Good catch! Will update this.

packages/chrome/src/styled/subnav/StyledSubNavItem.tsx
Francois-Esquire: Could you explain which part appears to be broken?

packages/accordions/src/styled/accordion/StyledHeader.spec.tsx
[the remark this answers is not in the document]
Francois-Esquire: Correct! We expect to have an `Accordion.Label` (a `<button>`) at all times with each `Accordion` section. When the `Label` is nested in the `Header`, we see `:focus-within` pseudo-selector and the `isFocused` prop working as expected.
See:
https://github.com/zendeskgarden/react-components/blob/main/packages/accordions/src/elements/accordion/components/Label.tsx

packages/notifications/src/styled/global-alert/StyledGlobalAlertClose.ts
```diff
@@ -66,9 +71,11 @@ export const colorStyles = (props: ThemeProps<DefaultTheme> & IStyledGlobalAlert
...
     }
 
-    &[data-garden-focus-visible] {
-      box-shadow: ${props.theme.shadows.md(boxShadowColor!)};
-    }
+    ${focusStyles({
+      theme: props.theme,
+      hue: focusColor,
+      shade: props.alertType === 'info' ? 600 : 800
+    })}
```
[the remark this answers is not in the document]
Francois-Esquire: We should be okay with these values:
https://github.com/zendeskgarden/react-components/pull/1548#discussion_r1197727645

packages/forms/src/styled/select/StyledSelect.ts
Francois-Esquire: Updated!

Francois-Esquire:
> Accordion focus ring appears duplicated when tabbing to inner icons:
> 
> branch-acc main-acc

Good catch! I've updated this to only focus when the `Accordion.Header` and the nested `Accordion.Label` is focused.

lucijanblagonic:
Noticed these two things:

## **Focused Anchor + hover**
image
Focused anchor is missing hover color while being focused.

## **Global alert Anchor changes color**
image
image
image
This might be related with the first item on the list. Focused Anchor shouldn't change the color of the text.

And two nitpicks... :) 

## **Draggable focus grabbed + shadow glitch**
draggable-focus-animation-glitch
You can see the shadow flickering once the item is focused + grabbed. This is probably as the box-shadow property is animated.

## **`FauxInput` and `isFocused` not consistent**
image
Currently, it's showing just the border, but it should show the whole outline ([this is how it currently works](https://zendeskgarden.github.io/react-components/?path=/story/packages-forms-fauxinput--faux-input&args=isFocused:true)). Not sure if this is a bug, but visually it's not consistent with what we have today.

Francois-Esquire:
Good catches @lucijanblagonic ! I've updated all the points you've brought to my attention except the drag-drop shadow. There seems to be a constraint with `dnd-kit` which we need to work through - that would be best addressed in a future effort.

lucijanblagonic:
Thanks for the fixes. Nitpicking, but I see some weird _sluggish_ animation style happening on Anchors/Breadcrumbs, but I think this is because of ease-in-out which makes it look "slow". I tested in the GIF below with that default animation and with removing the animation after. Do you see it also? Does this _feel_ weird, or is it just me? 🤷

video

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -124,18 +132,24 @@ export const StyledGlobalAlert = styled.div.attrs({
...
 
-    &:focus {
+    /* [1] */
+    ${SELECTOR_FOCUS_VISIBLE} {
+      text-decoration: none;
+    }
+
+    /* [1] */
+    &:hover {
       text-decoration: underline;
```
jzempel, line 149: If design intent is _not_ to have the underline on focus, then it should not be reapplied on focus+hover.

packages/chrome/src/styled/subnav/StyledSubNavItem.tsx
jzempel: Select a light `hue` and tab through the `SubNav`.

Kapture 2023-06-01 at 15 14 32

jzempel (changes requested):

Francois-Esquire:
@jzempel (RE: [sub nav header item color](https://github.com/zendeskgarden/react-components/pull/1548#discussion_r1213573524)) I believe the `SubNavHeaderItem` was a previously existing issue. We would need to add `useChromeContext` within the `CollapsibleSubNavItem` ([here](https://github.com/zendeskgarden/react-components/blob/main/packages/chrome/src/elements/subnav/CollapsibleSubNavItem.tsx#L23)) and pass in `isLight` to the internal styled component. Any opposition in doing a parallel PR to address this?

packages/chrome/src/styled/subnav/StyledSubNavItem.tsx
Francois-Esquire: Added #1558 to address this!

Francois-Esquire:
> Thanks for the fixes. Nitpicking, but I see some weird _sluggish_ animation style happening on Anchors/Breadcrumbs, but I think this is because of ease-in-out which makes it look "slow". I tested in the GIF below with that default animation and with removing the animation after. Do you see it also? Does this _feel_ weird, or is it just me? 🤷
> 
> video

No you are spot on @lucijanblagonic, this was something I observed as well. It should be fixed now!

packages/theming/src/utils/focusStyles.ts
```diff
@@ -23,7 +23,7 @@ type FocusStylesParameters = FocusBoxShadowParameters & {
  * @param {boolean} [options.condition=true] Supplies an optional condition that can be used to prevent the focus `box-shadow`
  * @param {boolean} [options.inset=false] Determines whether the `box-shadow` is inset
  * @param {string|Object} [options.hue='primaryHue'] Provides a theme object `palette` hue or `color` key, or any valid CSS color notation
- * @param {string} [options.selector=SELECTOR_FOCUS_VISIBLE] Provides a subsitute `:focus-visible` pseudo-class CSS selector.
+ * @param {string} [options.selector=SELECTOR_FOCUS_VISIBLE] Provides a substitute `:focus-visible` pseudo-class CSS selector.
```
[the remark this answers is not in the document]
Francois-Esquire: Good call, I've updated!

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
Francois-Esquire: Oof, good catch.

jzempel: This treatment appears to be fixed, but the problem still exists on `<Anchor>` in the buttons package.

jzempel (changes requested):

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
jzempel: ...or is `<Anchor>` focus+hover working as intended by design, @lucijanblagonic?

lucijanblagonic: I would keep the underline on focus, to be consistent with how `Anchor` work.

Francois-Esquire:
Thanks for the feedback @steue ! I've corrected the `ButtonGroup` and `SplitButton` styling and it should be as intended. This does include the transition, which was indeed missing (good catch).

For the `TextArea`, good eye for catching however it isn't a trivial task because it is a built-in browser (and vendor) specific piece of UI. We will need to do some research to implement this correctly moving forward. Off the bat, the use of `::-webkit-resizer` was a challenge to control and we will need to think this through.

lucijanblagonic (approved):
LGTM! 😍

geotrev (approved):
Other than remaining feedback, everything looks good.

packages/accordions/src/elements/accordion/components/Header.tsx
```diff
@@ -38,6 +39,18 @@ const HeaderComponent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement
...
+
+    setTimeout(() => {
+      const isAccordionButton = e.target.getAttribute('data-garden-id') === buttonGardenId;
+      const isFocusVisible = e.target.getAttribute('data-garden-focus-visible');
+
+      if (isAccordionButton && isFocusVisible) {
+        setIsFocused(true);
+      }
+    }, 0);
+  };
```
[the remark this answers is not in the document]
Francois-Esquire: I think this may be a warp in git history 👽 because this was the original code within the `Accordion.Header` component being restored after it was removed early in this PR. Currently we don't modify this file at all, however this is something we can talk more about. Come to think of it, there is history I am unaware of of why we implemented in this fashion.

steue (approved):
awesome @Francois-Esquire ! this looks great! ✨

PR #1549, opened 2023-05-10
feat(theming): enhance focus utilities with additional parameter overrides

jzempel:
## Description

- feat: add `boxShadow` retention to `getFocusBoxShadow`
- fix: prevent `focusStyles` from overriding existing `styles: { boxShadow: ... }`
- feat: add `spacerHue` and `spacerShade` color overrides to `getFocusBoxShadow` and `focusStyles`

### DIFF
--- a/packages/theming/src/utils/getFocusBoxShadow.ts
+++ b/packages/theming/src/utils/getFocusBoxShadow.ts
@@ -10,6 +10,7 @@ import { IGardenTheme } from '../types';
 import { DEFAULT_SHADE, Hue, getColor } from './getColor';
 
 export type FocusBoxShadowParameters = {
+  boxShadow?: string;
   inset?: boolean;
   hue?: Hue;
   shade?: number;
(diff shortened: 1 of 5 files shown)


packages/theming/src/utils/getFocusBoxShadow.ts
```diff
@@ -22,21 +25,27 @@ export type FocusBoxShadowParameters = {
  * Get a CSS `box-shadow` property value for focus state styling. The `hue` and
  * `shade` are used to determine the color of the focus ring.
  *
+ * @param {Object} [options.boxShadow] Provides an existing `box-shadow` (a drop shadow, for example) to be retained along with the focus ring
```
Francois-Esquire, line 28:
```suggestion
 * @param {string} [options.boxShadow] Provides an existing `box-shadow` (a drop shadow, for example) to be retained along with the focus ring
```

PR #1559, opened 2023-06-13
feat: add refactored `Combobox` component

jzempel:
## Description

Garden's current suite of combobox-like components – `Autocomplete`, `Combobox`, `Multiselect` and `Select` – is plagued with deep-seated accessibility issues, costly support maintenance, and APIs that challenging to navigate. This PR adds a new `@zendeskgarden/dropdowns.next` package with a fully refactored `Combobox` that is built to replace the existing component set with 100% WCAG 2.1 and WAI-ARIA 1.2 accessibility. The non-accessible component set will be removed in a future major release.

## Detail

Key capabilities include:

- **Controllable**: The `Combobox` functions in both [uncontrolled and
  controlled](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components) modes. Controlled mode enables aspects, such as input value, selection value(s),
  listbox expansion, and current option active index, to share and adapt to the
  surrounding UI.
- **Autocomplete-able**: Denotes the `Combobox` with [list
  autocomplete](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/). Filtering implementation is left to the API consumer.
- **Selectable**: The `Combobox` API ensures the selection of one or more
  listbox option values, while also supporting the W3C [no autocomplete
  example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-none/) for use cases like search.
- **Multi-selectable**: This feature enables the `Combobox` to provide WAI-ARIA
  [multi-select listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/#ex2_label) functionality with option-as-tag value rendering.
- **Non-editable**: The `Combobox` supports [select-only mode](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/), where the user cannot modify the `<input>`.
- **Filterable**: The `Combobox` offers various filtering methods for listbox
  options. Details of the filtering implementation are left to the API consumer.
- **Markup-able**: The `Combobox` can convert input value text to rich HTML
  markup on blur in single-selection mode.
- **Decorate-able**: The `Combobox` allows adding start and end media (SVG icons).
  Certain features will replace end media with Garden's standard dropdown chevron
  treatment.
- **Group-able**: The `Combobox` API utilizes fully accessible `<OptGroup>`
  components for grouping, similar to the corresponding HTML element.
- **Compactible**: Like other form elements, the `Combobox` supports compact
  sizing.
- **Field-able**: The `Combobox` builds on Garden’s Field API context to
  establish accessible relationships with corresponding Label, Hint, and Message
  components.
- **Validate-able**: The `Combobox` provides validation styling and
  accessibility comparable to other Garden form components.
- **RTL theme-able**: Functionality displays and operates correctly for
  left-to-right and right-to-left layouts.

## Checklist


packages/dropdowns.next/README.md
```diff
@@ -0,0 +1,82 @@
...
+npm install react react-dom styled-components @zendeskgarden/react-theming
+```
+
+## Usage
+
+### Combobox
+
+```jsx
+import { ThemeProvider } from '@zendeskgarden/react-theming';
+import { Field, Label, Combobox } from '@zendeskgarden/react-dropdowns.next';
```
Francois-Esquire, line 24:
```suggestion
import { Field, Label, Combobox, Option } from '@zendeskgarden/react-dropdowns.next';
```
`<Option />` is missing from the imports.

Francois-Esquire (commented):
`views` often represent compositions of a user interface rather than a component (or its primitives). Views tend to be associated with application level thinking (like the MVC pattern) as well. What was the reason for `src/views` (instead of the  conventional `src/styled`)?

packages/dropdowns.next/src/elements/combobox/Option.tsx
```diff
@@ -0,0 +1,106 @@
...
+      option,
+      ref: mergeRefs([optionRef, ref])
+    }) as LiHTMLAttributes<HTMLLIElement>;
+
+    return (
+      <OptionContext.Provider value={contextValue}>
+        <StyledOption
+          isActive={isActive}
+          isCompact={isCompact}
+          $type={type}
```
Francois-Esquire, line 72: Since we are still supporting v4 of `styled-components` (`"styled-components": "^4.2.0 || ^5.0.0"`), we should avoid transient props.

packages/dropdowns.next/src/elements/combobox/OptGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+      !label /* condition */
+    );
+    const optGroupProps = getOptGroupProps({
+      'aria-label': (groupAriaLabel || label)!
+    }) as HTMLAttributes<HTMLUListElement>;
+
+    return (
+      <StyledOption
+        isCompact={isCompact}
+        $type="group"
```
Francois-Esquire, line 46: Same as with `Option`. Since we are still supporting v4 of `styled-components` (`"styled-components": "^4.2.0 || ^5.0.0"`), we should avoid transient props.

packages/dropdowns.next/src/elements/combobox/Option.tsx
jzempel: I need to change the package dependency to `^5.0.0` – which is not a breaking change for this new package.

jzempel:
> What was the reason for `src/views` (instead of the conventional `src/styled`)?

@Francois-Esquire in light of future development plans, I want to establish a new/improved dev structure that more closely aligns with the [documented API](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#view-components). In Garden, we historically refer to a "container-view-element" architecture. This structure reinforces that and takes weight off of the current choice of `styled-components` as a means for expressing view components.

geotrev (commented):
No strong thoughts so far. The code looks solid. I'll take a look at the tests & storybook next.

packages/dropdowns.next/src/elements/combobox/Combobox.tsx
```diff
@@ -0,0 +1,361 @@
...
+  onChange: PropTypes.func,
+  placeholder: PropTypes.string,
+  renderExpandTags: PropTypes.func,
+  renderValue: PropTypes.func,
+  selectionValue: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
+  startIcon: PropTypes.any,
+  validation: PropTypes.oneOf(VALIDATION)
+};
+
+Combobox.defaultProps = {
```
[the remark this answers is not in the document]
geotrev: They can be moved into default args of the component props.

lucijanblagonic:
Here are a couple of things I noticed:

text-jump-on-focus
The text _jumps_ on focus. It should stay in the same line.

---

image
Focus styling on `Tags` not up-to-date with updated outlines.

---

image
Validation `Warning` outline color doesn't match to updated [focus outlines](https://zendeskgarden.github.io/react-components/?path=/story/packages-dropdowns-autocomplete--autocomplete&args=isOpen:true;validation:warning).

---

input-clear-after-escape
In [this case](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#support-notice-header) I see that the Input clears text on the second `Escape` press. Do you think we should follow that behavior here?

---

Is there an enhancement opportunity where the keyboard can move the focus between tags with the arrow keys? In situations where you have multi-select + editable, it could be appropriate. 🤔

packages/dropdowns.next/src/elements/combobox/Combobox.tsx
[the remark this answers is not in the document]
jzempel: TIL. We'll have a bit of work to do when this lands. The `defaultProps` serves the critical function of exposing these values via docgen.

packages/dropdowns.next/src/elements/combobox/OptGroup.tsx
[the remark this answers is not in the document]
jzempel: `styled-components` (currently) knows how to filter out non-HTML attributes from leaking into the render. We use [transient props](https://styled-components.com/docs/api#transient-props) only when the collision isn't automatically handled. If/when `styled-components` changes this, Garden will have a lot of work to do 😱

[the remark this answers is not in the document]
geotrev: v5 of styled-components requires us to use transient props. RIP

packages/dropdowns.next/src/elements/combobox/Combobox.tsx
```diff
@@ -0,0 +1,361 @@
...
+    const { hasHint, hasMessage, labelProps, setLabelProps } = useFieldContext();
+    const [isLabelHovered, setIsLabelHovered] = useState(false);
+    const [optionTagProps, setOptionTagProps] = useState<Record<string, IOptionProps['tagProps']>>(
+      {}
+    );
+    const options = useMemo(() => {
+      const tagProps: Record<string, IOptionProps['tagProps']> = {};
+      const retVal = toOptions(children, tagProps);
+
+      setOptionTagProps(value => ({ ...value, ...tagProps }));
```
Francois-Esquire, line 92: What was the reason for spreading the previous state?

A small concern I have about spreading the previous state is when it comes dynamic options, which is an edge case. It would result in persisting stale values and a possible memory leak.

packages/dropdowns.next/src/elements/combobox/Tag.tsx
```diff
@@ -0,0 +1,79 @@
...
+        ref={ref}
+      >
+        {children || <span>{text}</span>}
+        {!option.disabled &&
+          (removeLabel ? (
+            /* istanbul ignore next */
+            <Tooltip appendToNode={doc?.body} content={removeLabel} zIndex={tooltipZIndex}>
+              <StyledTag.Close aria-label={removeLabel} onClick={handleClick} />
+            </Tooltip>
+          ) : (
```
Francois-Esquire, line 55: If the `removeLabel` is not supplied, would tags not be removable?

packages/dropdowns.next/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+export { Combobox } from './elements/combobox/Combobox';
+export { Field } from './elements/combobox/Field';
+export { Hint } from './elements/combobox/Hint';
+export { Label } from './elements/combobox/Label';
+export { Message } from './elements/combobox/Message';
+export { Option } from './elements/combobox/Option';
+export { OptGroup } from './elements/combobox/OptGroup';
+export { Tag } from './elements/combobox/Tag';
```
Francois-Esquire, line 15: How to use the `Tag` component is not very clear, how is this used within the context of `Combobox`?

packages/dropdowns.next/src/views/combobox/StyledHint.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { Hint } from '@zendeskgarden/react-forms';
+
+const COMPONENT_ID = 'dropdowns.hint';
```
Francois-Esquire, line 12: #thought how do we feel about naming notation and directory structure? Should elements namespaced with `*.combobox.*` (eg `dropdowns.combobox.*`) live exclusively under the directory `src/views/combobox`?

packages/dropdowns.next/src/views/combobox/StyledSeparator.ts
```diff
@@ -0,0 +1,47 @@
...
+    margin: ${margin} 0;
+    height: ${height};
+  `;
+};
+
+export const StyledSeparator = styled.li.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<ThemeProps<DefaultTheme>>`
+  border: none;
```
Francois-Esquire, line 35: Is `border` needed?

Francois-Esquire (commented):
Solid work @jzempel 🙌 , left some feedback.

I noticed some visual tug-o-war between the scroll bar and the `"Trigger"` wrapping the input and tags. When multi-select is enabled with 2 or more tags and limited space, I'm seeing some overlap which may be undesirable:
 
Without `focusInset`:
with-tags

With `focusInset`:
with-focus-inset

jzempel:
@lucijanblagonic thanks for the review. Here are replies to your comments above.

> The text jumps on focus. It should stay in the same line.

I'm digging into this and will re-request your review when I have a solution committed.

> Focus styling on Tags not up-to-date with updated outlines.

I hadn't merged in the latest `main` with updated tag styling - done now.

> Validation Warning outline color doesn't match to updated focus outlines

Fixed in latest deploy.

> I see that the Input clears text on the second Escape press. Do you think we should follow that behavior here?

I actually have this [commented as a TODO](https://github.com/zendeskgarden/react-containers/blob/3676836c2463bee4163139299f156bffa7d26a39/packages/combobox/src/useCombobox.ts#L223) in the underlying container 😉. We can explore this option as a follow-on. But, since [the keyboard interaction is optional](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboard_interaction), I think the default should match what we currently have in Garden.

> Is there an enhancement opportunity where the keyboard can move the focus between tags with the arrow keys?

I don't think so. This results in an accessibility violation with the current `<Multiselect>`. I do think there is an opportunity for design research that opens up improved possibilities for multiselectable tag keyboarding while staying true to a11y specs.

packages/dropdowns.next/src/elements/combobox/Combobox.tsx
jzempel: Without this, `tagProps` get lost when options are filtered out – resulting in visual changes to currently selected tags. After careful consideration, I decided a "leak" wasn't a concern as the object will remain stable at N total options.

packages/dropdowns.next/src/elements/combobox/Tag.tsx
jzempel: No, the `<StyledTag.Close>` renders on line 56.

packages/dropdowns.next/src/index.ts
jzempel: It isn't ever used directly. We export here in order to generate the downstream prop table so that consumers have a reference for `tagProps` on the `Option` API. We'll make this clear on the website.

packages/dropdowns.next/src/views/combobox/StyledSeparator.ts
jzempel: Probably not. This is leftover from when I originally used an `hr` for this component

packages/dropdowns.next/src/views/combobox/StyledHint.ts
jzempel: Agreed. We could do with a more formalized/documented taxonomy throughout. But I'll update here.

jzempel:
@Francois-Esquire re:

> When multi-select is enabled with 2 or more tags and limited space, I'm seeing some overlap which may be undesirable

I'm not super concerned about this as the native scrollbar is better left alone – and isn't present by default (without enabling in system settings).

jzempel:
@lucijanblagonic the latest commit addresses...

> The text jumps on focus. It should stay in the same line.

PR is ready for your re-review

geotrev (approved):

PR #1561, opened 2023-06-27
fix(tables): persist overflow button visibility

Francois-Esquire:
## Description

Persists `OverflowButton` visibility on `Table` rows.

## Detail

Removes all conditional render logic when it comes to opacity toggling.
There have been additional updates to the `OverflowButton.spec.tsx` to account for this change.
Prior to this change, `OverflowButton`s would be visible only when its given row was being interacted with.

**Before:**
Table with ephemeral overflow button

**Current:**
table with persistent overflow buttons

## Checklist

### DIFF
--- a/packages/tables/.size-snapshot.json
+++ b/packages/tables/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "index.esm.js": {
-    "bundled": 25353,
-    "minified": 18456,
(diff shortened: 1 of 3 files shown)


geotrev (approved):
LGTM. Thanks for including test updates!

jzempel (approved):

PR #1562, opened 2023-06-28
fix(combobox): correct peer dependencies and label pixel shift

jzempel:
## Description

A few minor fixes, detailed below...

## Detail

- revert the underlying `Label`'s `vertical-align: center` which isn't playing nice with the `Combobox` flex layout
- bump the `theming` peer to 8.67.0, since we're taking advantage of `focusStyles` override parameters from that release
- correct the `styled-components` peer to 5.1 as that is the [transient props](https://styled-components.com/docs/api#transient-props) feature release

### DIFF
--- a/packages/dropdowns.next/.size-snapshot.json
+++ b/packages/dropdowns.next/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 51294,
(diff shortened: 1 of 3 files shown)


Francois-Esquire (approved):

geotrev (approved):
LGTM

PR #1570, opened 2023-07-10
fix(accordions): allow dynamic height for stepper content

Francois-Esquire:
## Description

Use a pure CSS implementation to animate/transition height for the active `Step` in `Stepper` component. This fixes a bug found with dynamically changing heights in a `Stepper.Step`, since there are instances where changing the height after the `Stepper.Content` has been rendered does not change the height.

## Detail

There is long-standing issue with animating `height` change that affects the web, and the implementation here makes use of `grid` (specifically `transition`ing the `grid-template-rows` CSS property) to handle that animation.

This treatment can be applied across the `accordions` package where this is relevant.

## Checklist

### DIFF
--- a/packages/accordions/src/styled/stepper/StyledContent.ts
+++ b/packages/accordions/src/styled/stepper/StyledContent.ts
@@ -26,6 +26,8 @@ const sizeStyles = (props: IStyledContent & ThemeProps<DefaultTheme>) => {
   return css`
     margin: ${marginVertical}px ${marginRight}px ${marginVertical}px ${marginLeft}px;
     padding: 0 ${paddingRight}px ${paddingBottom}px ${paddingLeft}px;
+    min-width: ${space.base * 30}px;
+    height: auto;
   `;
 };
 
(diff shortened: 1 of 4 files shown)


packages/accordions/src/styled/stepper/StyledContent.ts
```diff
@@ -33,10 +33,15 @@ export const StyledContent = styled.div.attrs<IStyledContent>({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledContent>`
-  ${sizeStyles}
+  display: grid;
+  grid-template-rows: ${props => (props.isActive ? 1 : 0)}fr;
+  transition: grid-template-rows 0.25s ease-in-out;
   min-width: ${props => props.theme.space.base * 30}px;
+  height: auto;
```
jzempel, line 40: [nit] both `min-width` and `height` should be moved to `sizeStyles`.

jzempel (approved):

geotrev (approved):
Glad this worked and love the simplicity of the approach. :shipit:

PR #1573, opened 2023-07-12
fix(combobox): prevent page scroll on input focus

jzempel:
## Description

Bump `react-containers` with the Combobox fix for preventing page scroll on focus.

## Detail

zendeskgarden/react-containers#550


geotrev (approved):
:shipit:

PR #1582, opened 2023-07-24
refactor(accordions): uplift for react v18

### LINKED ISSUE #1462: Stepper not expanding steps in `vite`
## Expectations
Expecting `Steps` to expand based on `activeIndex`

## Reality
Steps are not expanding no matter what I do, tried hardcoding the index, removing it etc. and cannot get the step to expand. I am using `vite`.

## Steps to Reproduce
1. Copied the example code from docs

### Fine Print

- Component: `<Stepper />`
- Version: 8.60.0
- Browsers: all

Francois-Esquire:
## Description

Refactoring the `Accordion` and `Stepper` for React v18 compatibility.

## Detail

This PR updates the two component sets in the `accordions` family:
- `Stepper`: replaces the `ref` for ordinal counting of `Stepper.Step`s with the `StepContext` that wraps each `Stepper.Step` (or child, if wrapped) with a pre-formed context value computed from the current step value - the context contains properties used throughout each `Step` (and the nested components) and it's current value.
- `Accordion`: replaces the `ref` for ordinal counting of `Accordion.Section`s with the `SectionContext` which will wrap the `Accordion.Section` child (or any wrapping element). The `SectionContext` contains the section value as the context value as it was prior. There was an uplift of `@zendeskgarden/container-accordion` to v3 and adopted the new API for `useAccordion` as well.

Fixes #1462

## Checklist


packages/accordions/src/utils/useSectionContext.ts
```diff
@@ -17,7 +17,7 @@ export const useSectionContext = () => {
   const context = useContext(SectionContext);
 
   if (context === undefined) {
-    throw new Error('This component must be rendered within a Accordion.Section component');
+    throw new Error('This component must be rendered within a Accordion component');
```
jzempel, line 20:
```suggestion
    throw new Error('This component must be rendered within an Accordion component');
```

[nit]

packages/accordions/src/elements/stepper/Stepper.tsx
```diff
@@ -5,33 +5,46 @@
...
     return (
       <StepperContext.Provider value={stepperContext}>
-        <StyledStepper ref={ref} isHorizontal={isHorizontal} {...props} />
+        <StyledStepper ref={ref} isHorizontal={isHorizontal} {...props}>
+          {useMemo(
+            () =>
+              Children.toArray(children).map((child, index) => (
+                <StepContext.Provider
+                  key={index}
+                  // eslint-disable-next-line react/jsx-no-constructed-context-values
```
jzempel, line 34: Could you extract `const value: IStepContext = { ... };` to the top of this function to get rid of this `eslint-disable`?

packages/accordions/src/elements/accordion/components/Panel.tsx
```diff
@@ -55,16 +53,16 @@ const PanelComponent = forwardRef<HTMLElement, HTMLAttributes<HTMLElement>>((pro
...
         ref: mergeRefs([panelRef, ref]),
-        index,
+        value: sectionValue,
         isBare,
         isCompact,
         isExpanded,
         isAnimated,
         ...props
-      })}
+      } as any) as any)}
```
jzempel, line 65: I think `{...getPanelProps() as HTMLAttributes<HTMLElement>}` would be preferable over the use of `any` here.

packages/accordions/src/elements/accordion/components/Header.tsx
```diff
@@ -76,7 +75,7 @@ const HeaderComponent = forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement
           onMouseOver: composeEventHandlers(onMouseOver, () => setIsHovered(true)),
           onMouseOut: composeEventHandlers(onMouseOut, () => setIsHovered(false)),
           ...other
-        })}
+        } as any) as any)}
```
jzempel, line 78: prefer `{...getHeaderProps() as HTMLAttributes<HTMLDivElement>}` vs overuse of `any`.

packages/accordions/src/elements/accordion/Accordion.tsx
```diff
@@ -5,46 +5,66 @@
...
     },
     ref
   ) => {
+    const { sections, sectionChildren } = useMemo(
+      () =>
+        Children.toArray(children)
+          .map((child, index) => (
+            <SectionContext.Provider key={index} value={index}>
+              {child}
+            </SectionContext.Provider>
```
jzempel, line 41: I'd like to hear a little bit more about this children wrapping vs. the previous context setup for this component. Should there be concerns about a more brittle DOM structure that is unable to handle intervening nodes?

Also, if this sticks, can `isValidElement` be used to eliminate the `any` casting? Ref: https://github.com/zendeskgarden/react-components/blob/3b332883cfe7fd23d285f7358a4b671c83fa3948/packages/dropdowns.next/src/elements/combobox/utils.ts#L38-L67

jzempel (changes requested):

packages/accordions/src/elements/stepper/Stepper.tsx
Francois-Esquire: Moving the variable doesn't get rid of the `eslint-disable` rule unfortunately. The rule states that context value should be wrapped in a `useMemo` to avoid re-renders- being that it is already in a `useMemo` it's not permitted nor does it give any benefit.

packages/accordions/src/elements/accordion/Accordion.tsx
Francois-Esquire: The idea here is that we expect:
```jsx
<accordion>
  <section />
  <section />
  <div>
    <section />
  </div>
</accordion>
```
would give each group the context it needs. This of course relies on the children to take this into account. It actually provides more flexibility when it comes to unique DOM structure requirements, so long as the top level children follow the given pattern.

packages/accordions/src/elements/accordion/components/Panel.tsx
```diff
@@ -55,16 +53,16 @@ const PanelComponent = forwardRef<HTMLElement, HTMLAttributes<HTMLElement>>((pro
...
         ref: mergeRefs([panelRef, ref]),
-        index,
+        value: sectionValue,
         isBare,
         isCompact,
         isExpanded,
         isAnimated,
         ...props
-      })}
+      } as Omit<HTMLAttributes<HTMLElement>, 'role'> & { value: number }) as any)}
```
jzempel, line 65: Similar here. Solution is:

```jsx
    <StyledPanel
      isBare={isBare}
      isCompact={isCompact}
      isExpanded={isExpanded}
      isAnimated={isAnimated}
      {...getPanelProps({
        role: role === undefined ? null : 'region',
        ref: mergeRefs([panelRef, ref]),
        value: sectionValue,
        ...props
      }) as HTMLAttributes<HTMLElement>}
    >
```

Change `const panelRef = useRef<HTMLElement>(null);` above and extract `role` from the component props.

packages/accordions/src/elements/stepper/Stepper.tsx
```diff
@@ -5,33 +5,48 @@
...
-import { StepperContext } from '../../utils';
+import { StepContext, StepperContext } from '../../utils';
 import { Step } from './components/Step';
 import { Label } from './components/Label';
 import { Content } from './components/Content';
 
 const StepperComponent = forwardRef<HTMLOListElement, IStepperProps>(
-  ({ isHorizontal, activeIndex, ...props }, ref) => {
-    const currentIndexRef = useRef(0);
+  ({ isHorizontal, activeIndex, children, ...props }, ref) => {
```
jzempel, line 17:
```suggestion
  ({ isHorizontal, activeIndex = DEFAULT_ACTIVE_INDEX, children, ...props }, ref) => {
```

...where `const DEFAULT_ACTIVE_INDEX = 0;` is defined above and also applied below as...

```js
StepperComponent.defaultProps = {
  activeIndex: DEFAULT_ACTIVE_INDEX
};
```

packages/accordions/src/elements/stepper/Stepper.tsx
```diff
@@ -5,33 +5,48 @@
...
     const stepperContext = useMemo(
       () => ({
         isHorizontal: isHorizontal || false,
-        activeIndex: activeIndex!,
-        currentIndexRef
+        activeIndex: activeIndex!
```
jzempel, line 21:
```suggestion
        activeIndex
```

...with `DEFAULT_ACTIVE_INDEX` set above.

jzempel (changes requested):

packages/accordions/src/elements/stepper/Stepper.tsx
Francois-Esquire: Updated! About `defaultProps`, setting a default value for `activeIndex` serves the same purpose, does it not?

packages/accordions/src/elements/stepper/Stepper.tsx
```diff
@@ -5,44 +5,57 @@
...
       </StepperContext.Provider>
     );
   }
 );
 
 StepperComponent.displayName = 'Stepper';
 
-StepperComponent.defaultProps = {
-  activeIndex: 0
-};
```
jzempel, line 44: This needs to be retained for runtime checking.

jzempel (changes requested):

jzempel (approved):

PR #1583, opened 2023-07-24
chore(dropdowns): tag and document deprecated components

jzempel:
## Description

Add programmatic and demo pointers to the new `Combobox` component.

### DIFF
--- a/packages/dropdowns/demo/select.stories.mdx
+++ b/packages/dropdowns/demo/select.stories.mdx
@@ -22,6 +22,10 @@ import README from '../README.md';
 
 # API
 
+**DEPRECATED:** use `@zendeskgarden/react-dropdowns.next`
+[Combobox](https://zendeskgarden.github.io/react-components/?path=/docs/packages-dropdowns-next-combobox)
+instead.
+
 <ArgsTable />
 
 # Demo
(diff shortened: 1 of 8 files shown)


geotrev (approved):
Is there any way to test this out before merging, or will it be testable from `website.git` when the submodule is updated in a PR?

packages/dropdowns/demo/select.stories.mdx
```diff
@@ -22,6 +22,10 @@ import README from '../README.md';
 
 # API
 
+**DEPRECATED:** use `@zendeskgarden/react-dropdowns.next`
+[Combobox](https://zendeskgarden.github.io/react-components/?path=/docs/packages-dropdowns-next-combobox)
+instead.
```
Francois-Esquire, line 27: Should we be linking to the `Select` component (in `forms`) instead?

jzempel: Nope. `Combobox` is the [replacement](https://garden.zendesk.com/components/combobox#select-only) for this custom `Select`. The `Select` in forms is a native `<select>` with no listbox styling.

jzempel:
> Is there any way to test this out before merging

@geotrev I checked imports in the related `Story` files to ensure folks will see the intended deprecation message.

Screenshot 2023-07-24 at 3 11 24 PM

Francois-Esquire (approved):

PR #1600, opened 2023-08-07
fix(accordions): enable dynamic height for panel

Francois-Esquire:
## Description

Use a pure CSS implementation to animate/transition height for the expanded `Accordion.Panel` in `Accordion` component.

## Detail

There is long-standing issue with animating height change that affects the web, and the implementation here makes use of grid (specifically transitioning the grid-template-rows CSS property) to handle that animation.

Relates to #1570 

## Checklist

- [ ] :guardsman: includes new unit tests. Maintain [existing coverage](https://coveralls.io/github/zendeskgarden/react-components?branch=main) (always >= 96%)


packages/accordions/src/styled/accordion/StyledInnerPanel.ts
```diff
@@ -6,37 +6,21 @@
...
   'data-garden-version': PACKAGE_VERSION
-})<IStyledInnerPanel>`
-  transition: ${props => props.isAnimated && 'max-height 0.25s ease-in-out'};
-  /* stylelint-disable-next-line declaration-no-important */
-  max-height: ${props => !props.isExpanded && '0 !important'}; /* [1] */
-  overflow: hidden;
-  line-height: ${props => getLineHeight(props.theme.space.base * 5, props.theme.fontSizes.md)};
-  font-size: ${props => props.theme.fontSizes.md};
+})`
+  overflow: inherit;
```
jzempel, line 17: `overflow` is not an inherited property. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#formal_definition

packages/accordions/src/styled/accordion/StyledPanel.ts
```diff
@@ -17,38 +22,56 @@ interface IStyledPanel {
...
-  if (props.isExpanded === false) {
+  if (isExpanded === false) {
     paddingTop = 0;
     paddingBottom = 0;
   }
 
   return css`
-    transition: ${props.isAnimated && 'padding 0.25s ease-in-out'};
+    grid-template-rows: ${isExpanded ? 1 : 0}fr;
+    border-bottom: ${theme.borders.sm};
```
jzempel, line 53: Recommend using `border-bottom-width` here.

packages/accordions/src/styled/accordion/StyledPanel.ts
```diff
@@ -17,38 +22,56 @@ interface IStyledPanel {
...
     padding: ${paddingTop}px ${paddingHorizontal}px ${paddingBottom}px;
+    line-height: ${getLineHeight(base * 5, theme.fontSizes.md)};
+    font-size: ${theme.fontSizes.md};
   `;
 };
 
+/**
+ * 1. reverse order of `sizeStyles` and `colorStyles` to give `border-bottom-color` styling precedence
+ */
```
jzempel, line 62: With the previous recommendation, I think we can remove this comment – especially since the updated order now honors the documented [rules](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules).

jzempel (changes requested):

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
Francois-Esquire: I believe it is ([please see spec](https://www.w3.org/TR/css-cascade-4/#inheriting)). Also in the MDN docs that you shared, I do see `inherit` listed under `global values`.

packages/accordions/src/styled/accordion/StyledPanel.ts
Francois-Esquire: Good call, that was a mixup on my end.

packages/accordions/src/styled/accordion/StyledPanel.ts
Francois-Esquire: Hmm setting `border-bottom-width` didn't give us the same styling as `isBare` styling which needs additional properties. Setting `border-bottom` seems to be the most economical here.
Screenshot 2023-08-07 at 3 09 31 PM

packages/accordions/src/styled/accordion/StyledInnerPanel.ts
Francois-Esquire: One of the reasons I prefer `inherit` here is that the `InnerPanel` component can be removed in the next major, and treating it as a pass through will make that migration easier in the future.

Francois-Esquire: Updated to explicitly set to `hidden` for shadow DOM use cases!

packages/accordions/src/styled/accordion/StyledPanel.ts
jzempel: We tend to get desired results by separating out all of the `border` shorthand – `border-color`, `border-width`, and `border-style`. See https://github.com/zendeskgarden/react-components/blob/main/packages/drag-drop/src/styled/dropzone/StyledDropzone.ts for reference.

jzempel (approved):

PR #1610, opened 2023-08-17
fix(loaders): optimize performance of skeleton loader

geotrev:
## Description

Improves animation performance in `Skeleton` loader.

## Detail

CSS properties like `left` will cause a reflow/repaint when they change, resulting in heavy DOM thrashing when animated. There are some [recommended](https://web.dev/animations-guide/) properties we can leverage instead.

Here is a before and after snapshot of animating using `left`, then switching to `translateX`:

### `left`

Screenshot 2023-08-17 at 2 00 29 PM

Layout/styles recalc is 120 times per second:

Screenshot 2023-08-17 at 2 07 07 PM

### `translateX`

Screenshot 2023-08-17 at 2 02 44 PM

Layout recalc is now zero, and styles recalc is at a much more reasonable 4-8 times per second, max:

Screenshot 2023-08-17 at 2 06 31 PM

The improvement in `layouts / sec` and `style calcs / sec` in the second image in particular shows a much better persistence of the loader.

As an added bonus, there's substantial CPU reduction as well.

## Checklist

- [ ] ~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~
- [ ] ~:guardsman: includes new unit tests. Maintain [existing coverage](https://coveralls.io/github/zendeskgarden/react-components?branch=main) (always >= 96%)~
- [ ] ~:wheelchair: tested for [WCAG 2.1 AA](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa) accessibility compliance~

### DIFF
--- a/packages/loaders/.size-snapshot.json
+++ b/packages/loaders/.size-snapshot.json
@@ -1,20 +1,20 @@
 {
   "index.cjs.js": {
-    "bundled": 20732,
-    "minified": 16106,
-    "gzipped": 4232
+    "bundled": 20801,
+    "minified": 16175,
+    "gzipped": 4233
   },
   "index.esm.js": {
-    "bundled": 19197,
-    "minified": 14618,
-    "gzipped": 4134,
+    "bundled": 19266,
+    "minified": 14687,
+    "gzipped": 4132,
     "treeshaked": {
       "rollup": {
-        "code": 11907,
+        "code": 11976,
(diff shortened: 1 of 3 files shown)


jzempel (changes requested):
nice 🙌

jzempel (approved):
🤦

PR #1612, opened 2023-08-24
feat(typography): create AccessibleText component

jgorfine-zendesk:
## Description

This PR introduces a new component to the Typography package: `AccessibleText`. This component provides a straightforward means of adding text to a UI that's both visually hidden and available to assistive technology.

It can be used in a static way:
```jsx
<AccessibleText tag="h2">Heading element for navigating the page with assistive tech</AccessibleText>
```

Or even as a live region:
```jsx
<AccessibleText role="status">Notification that only assistive tech users would perceive</AccessibleText>
```

## Detail

There are lots of ways to hide elements from the UI; however, there's only 1 tried and true way that makes elements invisible, completely void, _and_ available to assistive tech. 

Features | `display: none;` | `visibility: hidden;` | [hideVisually()](https://polished.js.org/docs/#hidevisually)
--- | --- | --- | ---
Invisible | yes ✅ | yes ✅ | yes ✅
Doesn't take up space | yes ✅ | no 🚫 | yes ✅
Accessible | no 🚫 | yes ✅ | yes ✅

## Checklist


jzempel (changes requested):
For consistency with the rest of the Garden system (ref [Label](https://garden.zendesk.com/components/input#label) for example), we'd rather see this implemented as a `hidden` prop over the existing [Span](https://garden.zendesk.com/components/span#span) component rather than adding a new component.

jgorfine-zendesk:
Hey @jzempel, thanks for the feedback.
 
> we'd rather see this implemented as a `hidden` prop over the existing [Span](https://garden.zendesk.com/components/span#span) component rather than adding a new component.

If we took that approach, would a consumer be able to pass in the kind of element they want to have visually-hidden, using a `tag` or `as` prop? I've seen some places in Agent Workspace, for example, where it would be helpful to have visually-hidden heading elements.

jzempel:
> I can have a go at implementing this. Would it be helpful if I closed this PR and opened another one for that work? Or can I continue doing it on this branch?

I think either can work.

jgorfine-zendesk:
You got it!

jzempel:
Closing in favor of #1677

PR #1618, opened 2023-09-05
fix(combobox): stabilize production SSR ID generation

jzempel:
## Description

Fix ID mismatch with `Combobox` when rendered in production SSR environments


geotrev (approved):
As far as I can tell the IDs look good. 👍🏻

smritimadan:
Looks good in the deployment 👍 ! 
Do we need additional tests to prevent in the future?

jzempel:
> Do we need additional tests to prevent in the future?

@smritimadan at this time, I'm not able to identify a straightforward test that reliably reproduces and checks the issue

PR #1623, opened 2023-09-21
feat(dropdowns.next): add refactored menu component

geotrev:
## Description

This PR adds a refactored menu component that is 100% WCAG 2.1 & WAI-ARIA 1.2 compliant. It offers a parity feature-set with the existing `Dropdown`, `Trigger`, and `Menu` components. The release of this component marks the deprecation of `Dropdown` package legacy Menu components. The `@zendeskgarden/react-dropdowns` package will be fully removed in a future major version.

## Detail

- **Controllable**: The `Menu` functions in both [uncontrolled and
  controlled](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components) modes. Controlled mode enables aspects, such as expansion, selection items, and focused value.
- **Nestable**: The `Menu` will have first-class N+ nesting support when configured with `onChange`.
- **Decorate-able**: The `Menu` continues to allow adding media (SVG icons) to `Item`s & `ItemGroup` headings.
- **Group-able**: The `Menu` API utilizes a fully accessible `<ItemGroup>` component for grouping.
- **Checkable**: The `ItemGroup` API optionally enables both radio and checkbox selection.
- **Compactible**: The `Menu` supports compact sizing.
- **RTL theme-able**: Functionality displays and operates correctly for left-to-right and right-to-left layouts.

## Checklist


packages/dropdowns.next/src/elements/menu/Menu.tsx
```diff
@@ -5,42 +5,103 @@
...
+      selectedItems,
+      items,
+      menuRef,
+      triggerRef,
+      onChange
+    });
+
+    const triggerProps: IMenuButtonProps = {
+      ...getTriggerProps({ type: 'button' }),
+      ...(isCompact && { size: 'small' }),
```
geotrev, line 70: My reasoning for applying sizing here is inspired by Combobox - where `isCompact` changes size of both input and listbox. 

However, I'm not opposed to excluding it from here and requiring a consumer to apply `size` to a custom `button` instance, in favor of a strict trigger/menu prop separation (like all the other props).

packages/dropdowns.next/src/elements/menu/utils.ts
```diff
@@ -33,45 +50,76 @@ type FloatingPlacement = (typeof FLOATING_PLACEMENT)[number];
...
+    return fallbackPlacements.map(fallback => toFloatingPlacement(isRtl, fallback) as Placement);
+  }
+
+  const side = placement.split('-')[0];
+  const sameSideFallbacks = [...Placements[side]];
+  const oppositeSideFallbacks = Placements[oppositeSideMap[side]];
+
+  // Remove the primary placement from the list of same-side fallbacks
+  // to prevent extra work for Floating-UI
+  sameSideFallbacks.splice(sameSideFallbacks.indexOf(placement as Placement), 1);
```
geotrev, line 118: There were some oddities with including crossAxis fallback placements, so by default, `Menu` won't offer that (this is in parity with the legacy Menu). If cross-axis becomes a necessary feature for consumers, the use of Garden-flavor placement values in `fallbackPlacements` is a way to achieve that behavior.

packages/dropdowns.next/src/types/index.ts
```diff
@@ -252,3 +226,100 @@ export interface ITagGroupProps {
   /** Provides the current selection */
   selection: IOption[];
 }
+
+export interface IMenuListProps extends HTMLAttributes<HTMLUListElement> {
```
geotrev, line 230: I'm not sure the best way to handle this so doc gen ignores it.  Storybook does, but not sure about the website.

jzempel: Always run `garden cmd-docgen` against new components or components with updated doc to verify that the output matches expectations. In this case `yarn garden cmd-docgen --pretty packages/dropdowns.next/src/elements/menu/Menu.tsx` etc.

jzempel (changes requested):

packages/dropdowns.next/src/views/combobox/StyledFloatingListbox.ts
```diff
@@ -37,7 +36,6 @@ export const StyledFloatingListbox = styled.div.attrs({
     menuStyles(props.position, {
       theme: props.theme,
       hidden: props.isHidden,
-      childSelector: `> ${StyledListbox}`,
```
jzempel, line 40: I'm curious, was this removal necessary to get animation working?

packages/dropdowns.next/src/elements/menu/Menu.tsx
geotrev: Makes sense. So in that scenario, a consumer could Choose Your Own Adventure™?

- Set `button` (string) + `buttonProps`; or
- `button` (callback)

packages/dropdowns.next/src/types/index.ts
geotrev: Very handy, thanks. I found a couple errors running this. :+1:

packages/dropdowns.next/demo/~patterns/stories/MenuNestedStory.tsx
```diff
@@ -0,0 +1,53 @@
...
+                }
+
+                if ('value' in item) {
+                  return (
+                    <Item key={item.value} {...item}>
+                      {item.value}
+                      {item.type === 'next' && <StyledHiddenText>Go to submenu</StyledHiddenText>}
+                      {item.type === 'previous' && (
+                        <StyledHiddenText>Go to previous menu</StyledHiddenText>
+                      )}
```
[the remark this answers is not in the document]
geotrev: Good call. The [proposed `AccessibleText` component](https://github.com/zendeskgarden/react-components/pull/1612) would be great here too.

packages/dropdowns.next/src/types/index.ts
```diff
@@ -252,3 +226,98 @@ export interface ITagGroupProps {
...
+  placement?: GardenPlacement;
+  /** Provides ref access to the associated trigger element */
+  triggerRef: RefObject<HTMLElement>;
+  /** Sets the `z-index` of the menu */
+  zIndex?: IListboxProps['zIndex'];
+}
+
+export interface IMenuButtonProps extends IButtonProps {
+  ref: RefObject<HTMLButtonElement>;
+}
```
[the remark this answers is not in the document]
geotrev: Kind of realizing I'm not sure the best way to write `jsdoc` for string or callback. Do you have any suggestions? I'm thinking it would probably be better to inline that type right on the `button` prop instead of exporting a totally separate interface.

packages/dropdowns.next/src/views/combobox/StyledFloatingListbox.ts
geotrev: In this case, yes. Since `FloatingMenu` extends `FloatingListbox`, and both these containers always have one child, removing this line allows the [direct descendent fallback](https://github.com/zendeskgarden/react-components/blob/main/packages/theming/src/utils/menuStyles.ts#L47) to work for both components.

packages/dropdowns.next/src/elements/menu/Menu.tsx
```diff
@@ -65,8 +68,11 @@ export const Menu = forwardRef<HTMLUListElement, IMenuProps>(
       onChange
     });
 
-    const triggerProps: IMenuButtonProps = {
-      ...getTriggerProps({ type: 'button' }),
+    const { onClick, onKeyDown, ...buttonProps } = _buttonProps;
+
+    const triggerProps: IButtonProps & { ref: RefObject<HTMLButtonElement> } = {
+      ...buttonProps,
```
jzempel, line 74: `isCompact` should spread before the new `buttonProps` so the consumer can manage the override.

packages/dropdowns.next/src/types/index.ts
```diff
@@ -250,15 +250,13 @@ export interface IMenuListProps extends HTMLAttributes<HTMLUListElement> {
...
-  ref: RefObject<HTMLButtonElement>;
-}
-
 export interface IMenuProps extends HTMLAttributes<HTMLUListElement> {
   /** Appends the menu to the element provided */
   appendToNode?: IMenuListProps['appendToNode'];
   /** Sets the menu button label or renders a provided trigger element */
-  button: string | ((props: IMenuButtonProps) => ReactNode);
+  button: string | ((props: IButtonProps & { ref: RefObject<HTMLButtonElement> }) => ReactNode);
+  /** Provides additional props to the menu button */
```
jzempel, line 258:
```suggestion
  /** Provides additional props to the menu [Button](/components/button#button) */
```

...for website cross-reference. Compare with https://garden.zendesk.com/components/multiselect#multiselect `renderItem`.

jzempel (changes requested):

jzempel (approved):
Fantastic work @geotrev :tada:

steue (approved):
looks great @geotrev!

PR #1625, opened 2023-09-29
chore(dropdowns): Add menu to readme, mark legacy dropdowns deprecated

geotrev:
## Description

- Adds Menu example to README
- Updates deprecation links & marks Menu/Dropdown/Trigger as deprecated.

## Checklist

### DIFF
--- a/packages/dropdowns/README.md
+++ b/packages/dropdowns/README.md
@@ -3,6 +3,10 @@
 This package includes components relating to dropdowns in the
 [Garden Design System](https://zendeskgarden.github.io/).
 
+**DEPRECATED:** use [`@zendeskgarden/react-dropdowns.next`][dropdowns.next] instead.
+
+[dropdowns.next]: https://github.com/zendeskgarden/react-components/tree/main/packages/dropdowns.next
+
 ## Installation
 
 ```sh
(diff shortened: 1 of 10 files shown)


packages/dropdowns/README.md
```diff
@@ -3,6 +3,10 @@
 This package includes components relating to dropdowns in the
 [Garden Design System](https://zendeskgarden.github.io/).
 
+**DEPRECATED:** use [`@zendeskgarden/react-dropdowns.next`][dropdowns.next] instead.
+
+[dropdowns.next]: https://github.com/zendeskgarden/react-components/tree/main/packages/dropdowns.next
```
jzempel, line 8: I'm tentative about adding this one, since our current intent (pending plan details) is to move `dropdowns.next` to `dropdowns` in the next major. As a result, we don't plan to run `npm deprecate` the `@zendeskgarden/react-dropdowns` package.

geotrev: That's a good point, and I couldn't totally remember if we were doing it that way. I'll remove it from the package README :)

geotrev: Updated: 47c6b0dfc

jzempel (approved):

PR #1673, opened 2023-12-12
chore: dark mode POC

jzempel:
## Description

This PR is a brute force POC (not intended for merge) that demonstrates how the existing `theme.colors.base` value might be used to impact styling for the [how-to-react sandbox](https://codesandbox.io/s/github/zendesk-garden/how-to-react). Ultimately, we'll look at a combination of CSS variables together with a `'light'` vs `'dark'` toggle to handle a richer color theming experience.

## Detail

The story can be found under "Chrome -> [patterns] -> Dark". A `hue` control is included to observe the affect of Garden's existing branding control.

### DIFF
--- a/packages/buttons/src/styled/StyledButton.ts
+++ b/packages/buttons/src/styled/StyledButton.ts
@@ -29,18 +29,26 @@ const getBorderRadius = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
   return props.theme.borderRadii.md;
 };
 
-const getDisabledBackgroundColor = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
-  return getColor('neutralHue', 200, props.theme);
+const getDisabledBackgroundColor = (
+  props: IButtonProps & ThemeProps<DefaultTheme>,
+  shades: Record<string, any>
+) => {
+  return getColor('neutralHue', shades.disabledBgColor, props.theme);
 };
 
 export const getHeight = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
+  const { sizes } = props.theme.variables.buttons;
+  let height;
+
   if (props.size === 'small') {
-    return `${props.theme.space.base * 8}px`;
+    height = sizes.sm.height;
   } else if (props.size === 'large') {
-    return `${props.theme.space.base * 12}px`;
+    height = sizes.lg.height;
+  } else {
+    height = sizes.md.height;
   }
 
-  return `${props.theme.space.base * 10}px`;
+  return `${height}px`;
 };
 
 /**
@@ -50,6 +58,7 @@ export const getHeight = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
  * 4. set default outline-color for smooth transition without artifacts
  */
 const colorStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
+  const { shades } = props.theme.variables.buttons;
   let retVal;
   let hue;
 
@@ -64,13 +73,13 @@ const colorStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
     hue = 'primaryHue';
   }
 
-  const shade = 600;
+  const shade = shades.base;
   const baseColor = getColor(hue, shade, props.theme);
-  const hoverColor = getColor(hue, shade + 100, props.theme);
-  const activeColor = getColor(hue, shade + 200, props.theme);
+  const hoverColor = getColor(hue, shades.hover, props.theme);
+  const activeColor = getColor(hue, shades.active, props.theme);
   const focusColor = getColor('primaryHue', shade, props.theme);
-  const disabledBackgroundColor = getDisabledBackgroundColor(props);
-  const disabledForegroundColor = getColor(hue, shade - 200, props.theme);
+  const disabledBackgroundColor = getDisabledBackgroundColor(props, shades);
+  const disabledForegroundColor = getColor(hue, shades.disabledColor, props.theme);
 
   if (props.isLink) {
     retVal = css`
@@ -209,6 +218,7 @@ const colorStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
  * 2. reset icon button with border
  */
 const groupStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
+  const { shades } = props.theme.variables.buttons;
   const { theme, isPrimary, isBasic, isSelected, isPill, focusInset } = props;
   const { rtl, borderWidths, borders } = theme;
   const startPosition = rtl ? 'right' : 'left';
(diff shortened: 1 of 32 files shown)


packages/chrome/src/styled/header/StyledHeader.ts
```diff
@@ -5,21 +5,55 @@
...
+  const fontSize = props.theme.fontSizes.md;
+  const height = getNavItemHeight(props);
+  const padding = `${props.theme.space.base}px`;
+
+  return css`
+    border-bottom: ${border};
+    padding: 0 ${padding};
+    height: ${height};
+    font-size: ${fontSize};
+  `;
```
jzempel, line 52: This is an example of following the [documented](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#view-components) `colorStyles` and `sizeStyles` for view components. We'll expect to see this structure more fastidiously followed during dark mode development

packages/modals/src/styled/StyledHeader.ts
```diff
@@ -17,35 +17,73 @@ import { BASE_MULTIPLIERS } from './StyledClose';
...
-      props.theme.space.base * (BASE_MULTIPLIERS.size + BASE_MULTIPLIERS.side + 2)
-    }px;`} /* [1] */
-  line-height: ${props => getLineHeight(props.theme.lineHeights.md, props.theme.fontSizes.md)};
-  color: ${props =>
-    props.isDanger ? getColor('dangerHue', 600, props.theme) : props.theme.colors.foreground};
-  font-size: ${props => props.theme.fontSizes.md};
   font-weight: ${props => props.theme.fontWeights.semibold};
 
+  ${sizeStyles};
+  ${colorStyles};
```
jzempel, line 85: ditto re: `sizeStyles` & `colorStyles` refactor

packages/theming/src/types/index.ts
```diff
@@ -129,13 +129,13 @@ export interface IGardenTheme {
...
 
-export interface IThemeProviderProps extends Partial<ThemeProps<IGardenTheme>> {
+export interface IThemeProviderProps extends Partial<ThemeProviderProps<IGardenTheme>> {
   /**
    * Provides values for component styling. See styled-components
    * [`ThemeProvider`](https://styled-components.com/docs/api#themeprovider)
    * for details.
    */
-  theme?: IGardenTheme;
+  theme?: IGardenTheme | ((theme: IGardenTheme) => IGardenTheme);
```
jzempel, line 138: Updated to allow the same `theme` typing as styled-components

packages/theming/src/elements/ThemeProvider.tsx
```diff
@@ -20,7 +20,7 @@ export const ThemeProvider = ({
   ...other
 }: PropsWithChildren<IThemeProviderProps>) => {
```
jzempel, line 21: Ultimately, I think we'll expect to see [createGlobalStyle](https://styled-components.com/docs/api#createglobalstyle) here with a listing of CSS variables pulled from the default theme.

packages/chrome/demo/~patterns/stories/DarkStory.tsx
```diff
@@ -0,0 +1,47 @@
...
+    ({
+      ...parentTheme,
+      colors: {
+        ...parentTheme.colors,
+        base: 'dark',
+        background: getColor('neutralHue', 900, parentTheme),
+        foreground: getColor('neutralHue', 200, parentTheme),
+        primaryHue: '#50a1e1'
+      }
+    }) as IGardenTheme;
```
jzempel, line 32: For the POC, this is the main theming setup for dark mode – affecting a major portion of the UI as expected.

packages/theming/src/elements/theme/global-styles.ts
```diff
@@ -0,0 +1,35 @@
...
+        --${prefix}--buttons_md--padding: ${buttons.sizes.md.padding}px;
+        --${prefix}--buttons_md--height: ${buttons.sizes.md.height}px;
+        --${prefix}--buttons_lg--fontSize: ${buttons.sizes.lg.fontSize};
+        --${prefix}--buttons_lg--padding: ${buttons.sizes.lg.padding}px;
+        --${prefix}--buttons_lg--height: ${buttons.sizes.lg.height}px;
+        --${prefix}--buttons--shade: ${buttons.shades.base};
+        --${prefix}--buttons--shade--hover: ${buttons.shades.hover};
+        --${prefix}--buttons--shade--active: ${buttons.shades.active};
+        --${prefix}--buttons--shade--color--disabled: ${buttons.shades.disabledColor};
+        --${prefix}--buttons--shades--bg--disabled: ${buttons.shades.disabledBgColor};
```
geotrev, line 31: These are set after semantic variables are created in ThemeProvider.

packages/theming/src/elements/ThemeProvider.tsx
```diff
@@ -5,31 +5,72 @@
...
+  _variables?: (theme: IGardenTheme) => ISemanticVariables
+): IGardenTheme => {
+  let variables;
+
+  if (_variables) {
+    variables = _variables(theme);
+  } else {
+    variables = {
+      buttons: getButtonVariables(theme)
+    };
```
geotrev, line 39: This is an example of the expected return value should a consumer provide their own variables getter function to ThemeProvider.

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -319,27 +333,35 @@ const sizeStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
...
+          --${prefix}--buttons_${size}--padding,
+          ${em(
+            math(`${sizes[size].padding}px - ${props.theme.borderWidths.sm}`),
+            sizes[fontSize].fontSize
+          )}
+        );
+      height: var(--${prefix}--buttons_${size}--height, ${height});
       line-height: ${lineHeight};
-      font-size: ${fontSize};
+      font-size: var(--${prefix}--buttons_${fontSize}--fontSize, ${sizes[fontSize].fontSize});
```
geotrev, line 364: It might serve us well to isolate custom property definitions like this in a separate file, so we can reference exactly which properties a component has tied to theming. Especially if we have custom properties not tied directly to specific properties (i.e. shades).

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -49,7 +58,8 @@ export const getHeight = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
  * 3. shifting :focus-visible from LVHFA order to preserve `color` on hover
  * 4. set default outline-color for smooth transition without artifacts
  */
-const colorStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
+const colorStyles = (props: IButtonProps & ThemeProps<IGardenTheme>) => {
+  const { shades } = props.theme.variables.buttons;
```
geotrev, line 62: Not married to shades being themeable this way - but I found it clunky and anti-thetical to how `getColor` works to define colors, with shades, in semantic variable generation. If we move that generation step into the component, this might be less of an issue.

packages/theming/src/elements/theme/global-styles.ts
geotrev: I'm also realizing now, we if were to lazily generate semantic variables in component instances, it would throw a wrench in generating these variables on page load. Not sure how/if we can wiggle around that issue.

packages/buttons/src/styled/StyledButton.ts
geotrev: Regardless of how that works, it felt natural to have the custom properties take priority should a consumer want to override them from that direction.

PR #1676, opened 2024-01-19
fix(forms): prevent `Textarea` autosize infinite loop

### LINKED ISSUE #1563: Maximum update depth exceeded with textarea component
## Expectations

Textarea expands with content

## Reality

Maximum update depth exceeded – plugin "crashes" – see video:

https://github.com/zendeskgarden/react-components/assets/380914/812c15e0-a8aa-4b84-9d3a-d1b625125f5a

We were able to pinpoint the setState loop to this code:

image

## Steps to Reproduce

For reproduction steps, we have a Zendesk textarea component like this inside a plugin:

```
<Textarea minRows={4} maxRows={10} isCompact {...field} />
```

Typing enough enters so that the textarea expands causes a setState loop inside of the Textarea component. This is causing our shared customers to lose their data on a regular basis.

### Fine Print

- Component: Textarea
- Version: 8.62.1
- Browsers: Chrome+Brave, maybe all

jzempel:
## Description

This PR removes a `useEffect` that puts a React 18, _controlled_, autosized (with `minRows` & `maxRows` defined) `Textarea` into an infinite loop after the window has been resized. Testing shows that the existing `useLayoutEffect` is sufficient for handling both controlled and uncontrolled component resizing.

## Detail

closes #1563

### DIFF
--- a/packages/forms/src/elements/Textarea.tsx
+++ b/packages/forms/src/elements/Textarea.tsx
@@ -5,21 +5,10 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, {
-  useRef,
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useContext,
-  useState
-} from 'react';
+import React, { useRef, useCallback, useLayoutEffect, useState } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


smritimadan (approved):
👍

PR #1677, opened 2024-01-19
feat(typography): add support for accessible, visually hidden `Span` styling

jzempel:
## Description

Similar to other Garden components (such as [Label](https://garden.zendesk.com/components/input#label)), this PR adds the ability to visually hide text – via `<Span hidden>` – without hiding the text from screen readers.

## Detail

Simplified pivot from #1612

## Checklist

### DIFF
--- a/packages/typography/demo/span.stories.mdx
+++ b/packages/typography/demo/span.stories.mdx
@@ -20,6 +20,7 @@ import README from '../README.md';
     name="Span"
     args={{ children: 'Text', hasIcon: false, hasStartIcon: false }}
     argTypes={{
+      hidden: { control: 'boolean' },
       hue: { control: 'color' },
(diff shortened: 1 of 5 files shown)


smritimadan (approved):
Looks good 👍

jgorfine-zendesk:
This is going to be so useful! Thanks for making it happen. 🙌

jgorfine-zendesk:
@jzempel my one question is: if the expectation is that consumers use the `hidden` attribute to make this component visually-hidden, what should they do if they want/need to make it fully hidden?

jzempel:
> what should they do if they want/need to make it fully hidden?

@jgorfine-zendesk in that case, I wouldn't recommend using this component as it provides no value to the `display: none` use case

PR #1697, opened 2024-02-05
chore: fix config to support `next` prerelease

jzempel:
## Description

The config updates and fixes in the PR allow the positional tag `version` specifier to work as intended. For example: `npm run tag preminor` would ensure that the version is set with `-next.x` notation and that it is published under the `next` tag.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
     steps:
       - *attach_workspace
       - run: npm set //registry.npmjs.org/:_authToken=$NPM_TOKEN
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     steps:
       - *attach_workspace
       - run: npm set //registry.npmjs.org/:_authToken=$NPM_TOKEN
-      - run: npm exec -- lerna publish from-git --ignore-scripts --no-verify-access --yes
```
jzempel, line 55: removes a deprecated Lerna flag – all access is now verified by default

geotrev (approved):

PR #1699, opened 2024-02-06
fix(combobox): prevent disabled combobox from swapping static/input values on click

jzempel:
## Description

Currently, a disabled `Combobox` is swapping the static value with the input value on click (focus). Note that keyboard focus is not available for disabled comboboxes.

## Detail

Here is an example of the current bug where "cherry" is a custom `renderValue` that gets displaced by the empty input value on click focus.

https://github.com/zendeskgarden/react-components/assets/143773/2b40db14-9375-4bc8-824c-23cd77828f0c

## Checklist


geotrev (approved):
🚀

PR #1712, opened 2024-02-20
feat: renames Pagination to OffsetPagination with revised API

geotrev:
## Description

- Introduces `OffsetPagination` (renamed `Pagination`) with API and accessibility improvements.
- Migration doc updated

## Detail

- `transformPageProps` prop has been removed
- `labels` prop added to more simply customize pagination element labels
- `container-pagination` removed (no arrow key navigation)
- HTML structure & accessibility improvements:
  - Wrapping `nav` tag with translatable `aria-label`.
  - Translatable `aria-label` recommended for ellipsis/page gap element
  - List items contain `button`s
  - Numbered page elements now indicated with `aria-current="page"`
  - Numbered page elements consistently labeled as `Page X` (`current` word removed).

## Checklist


packages/pagination/demo/OffsetPagination.stories.mdx
jzempel, line 1: File should be named `offsetPagination.stories.mdx` (lowercase "o") for consistency with others.

packages/pagination/src/elements/OffsetPagination/components/Gap.tsx
```diff
@@ -6,13 +6,23 @@
...
-  <StyledGap {...props} ref={ref}>
-    …
-  </StyledGap>
-));
+const GapComponent = forwardRef<HTMLLIElement, LiHTMLAttributes<HTMLLIElement>>((props, ref) => {
+  const ariaLabel = useText(
+    GapComponent,
+    props,
+    'aria-label',
+    'Ellipsis indicating non-visible pages'
```
jzempel, line 17: Is it worth passing a `count` (naming tbd) prop to the gap component so that this default label can be customized with "Ellipsis indicating ${count} non-visible pages"? Certainly non-essential; perhaps nice to have.

packages/pagination/demo/OffsetPagination.stories.mdx
```diff
@@ -14,8 +13,18 @@ import README from '../README.md';
...
+    name="OffsetPagination"
+    args={{
+      currentPage: 1,
+      totalPages: 11,
+      labels: {
+        gap: 'Ellipsis indicating non-visible pages',
+        page: p => `Page ${p}`,
+        next: 'Next page',
+        previous: 'Previous page'
+      },
```
jzempel, line 25: Not sure, due to the type mix. But can this be setup so it can be controlled via the story?

packages/pagination/src/styled/OffsetPagination/StyledList.ts
```diff
@@ -8,13 +8,13 @@
...
 
-const COMPONENT_ID = 'pagination.pagination_view';
+const COMPONENT_ID = 'pagination.list';
 
 /**
  * 1. List reset.
  * 2. Text truncation.
  */
-export const StyledPagination = styled.ul.attrs({
+export const StyledList = styled.ul.attrs({
```
jzempel, line 17: Not sure, due to the arrows, but is it more semantically appropriate for this to be a `styled.ol`?

packages/pagination/src/styled/OffsetPagination/StyledPage.ts
```diff
@@ -17,7 +17,7 @@ const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
 
   return css`
     min-width: ${height};
-    max-width: ${math(`${height} * 2`)}; /* [1] */
+    max-width: ${math(`${height} * 2`)};
```
jzempel, line 20: Should we get rid of the expensive `math` function by making `height` a number calculation and appending `'px'` where necessary?

jzempel (changes requested):
There should be a `migration.md` entry for the rename, type `extend` change, and prop update.

packages/pagination/src/elements/OffsetPagination/components/Gap.tsx
geotrev: I like that idea. Looks like the calculation is all there already, too.

packages/pagination/src/styled/OffsetPagination/StyledList.ts
geotrev: The resources I cited in the RFC seemed consistent with `ul` for the list element, so I ran with that. My worry with `ol` would also be that the announcement of the `li` number would create confusion next to the page number. `list item 3, page 1`, that sort of thing.

packages/pagination/src/elements/OffsetPagination/components/Gap.tsx
geotrev: Actually, the calculation is a bit more involved than I thought. Definitely something to consider in the future though.

jzempel: Ok, I'm totally fine to defer. But we'll probably need to punt to a future major, because we'd want the `labels` API to follow suit with a function.

packages/pagination/src/styled/OffsetPagination/StyledList.ts
jzempel: Ah, right. Thanks for doing the thorough research on this.

jzempel (changes requested):

docs/migration.md
```diff
@@ -59,6 +59,13 @@ consider additional positioning prop support on a case-by-case basis.
 
 - `TooltipModal`: removed `popperModifiers` prop (see [note](#breaking-changes))
 
+#### @zendeskgarden/react-pagination
+
+- `Pagination`: renamed to `OffsetPagination`
+  - changed type export to `HTMLAttributes<HTMLElement>`
```
jzempel, line 65:
```suggestion
  - changed type export from `HTMLAttributes<HTMLUListElement>` to `HTMLAttributes<HTMLElement>`
```

this minor change let's users know if they really need to care or not about the change

jzempel (changes requested):
A slight content change suggestion in addition to the file rename request :pray:

docs/migration.md
geotrev: Aha, thanks for the clarification. Added. :)

packages/pagination/demo/OffsetPagination.stories.mdx
geotrev: Dang. I really thought I pushed this one up. 🤦🏻

jzempel (approved):

PR #1728, opened 2024-02-29
fix: allows menu to position correctly in RTL

geotrev:
## Description

Configures RTl override in Floating UI (FUI).

## Detail

FUI determines RTL from `direction: rtl` in the style cascade. This PR sets an override to the [`platform` utility function](https://floating-ui.com/docs/platform#isrtl) that now points to `theme.rtl`.

## Checklist

- [ ] ~~:guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)~~


jzempel (changes requested):
We don't want the `direction` approach as the current `Button` (and legacy `Menu`) don't use that. Instead, the Garden-to-Popper placement logic needs refinement for RTL.

jzempel (approved):
Nice. Please update this PR's description accordingly.

PR #1729, opened 2024-03-01
feat(dropdowns)!: replaces label prop with legend in OptGroup

geotrev:
## Description

Adds `legend` prop to `OptGroup` + deprecates `label`

## Checklist

- [x] :metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] :guardsman: includes ~~new~~ updated unit tests. Maintain existing coverage (always >= 96%)

### DIFF
--- a/docs/migrations/v8.md
+++ b/docs/migrations/v8.md
@@ -303,6 +303,5 @@ isolation.
 
 #### @zendeskgarden/react-utilities
 
-- This package has been deprecated This package has been removed.and will be removed in a future
-  major release.
+- This package has been deprecated and will be removed in a future major release.
 - Migrate to `@zendeskgarden/container-utilities` to continue receiving updates.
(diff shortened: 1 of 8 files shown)


packages/dropdowns/src/types/index.ts
```diff
@@ -185,6 +185,8 @@ export interface IOptGroupProps extends Omit<LiHTMLAttributes<HTMLLIElement>, 'c
   /** Accepts an icon to display */
   icon?: ReactElement;
   /** Sets the text label of the option group */
+  legend?: string;
+  /** @deprecated use `IOptGroupProps['legend']` instead */
```
geotrev, line 189: I was looking around in the gatsby prop sheet code and couldn't quickly see the best way to ensure `label` is marked deprecated or ignored at build time. I also saw `@ignore`, but that has a different meaning I guess.

jzempel: Per sequence doc, let's pull the bandaid off and remove `label` rather than deprecating.

jzempel (changes requested):

docs/migrations/v8.md
```diff
@@ -303,6 +303,5 @@ isolation.
 
 #### @zendeskgarden/react-utilities
 
-- This package has been deprecated This package has been removed.and will be removed in a future
-  major release.
+- This package has been deprecated and will be removed in a future major release.
```
geotrev, line 306: Not sure how this typo fell in. This is a copy-paste revert from the migration doc on `main`.

jzempel (approved):

PR #1746, opened 2024-03-11
chore(theming): rename `getColor` => `getColorV8`

jzempel:
## Description

This update clears the way for a new, enhanced `getColor` utility that will take updated theming (`variables`) values into account. All instances of `getColorV8` will eventually be removed throughout the v9 development cycle.

## Detail

Material changes are limited to the `theming` package, while all others represent a global find-n-replace.


geotrev (approved):
Incremental development go brrrrrr 🎉

PR #1752, opened 2024-03-21
feat: updates icon component and prop types for cross-package consistency

geotrev:
## Description

Updates multiple packages so their `icon` prop types are consistent with other packages. Also updates several component type signatures as well.

See `migration.md` for full changelog.

## Detail

There are several icon components still using `div` styled component that nest an `svg` icon inside them. I didn't touch these as they're still spreading props and setting their ref to the styled `div`, not the `svg` directly.

## Checklist

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -34,7 +34,10 @@ consider additional positioning prop support on a case-by-case basis.
 
 #### @zendeskgarden/react-accordions
 
-- Removed `IItem` type export. Use `ITimelineItemProps` instead.
+- The following React component prop types have changed:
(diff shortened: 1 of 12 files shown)


jzempel (approved):
@geotrev this is a good start and good for what this PR is. But I think we're looking for more out of the sequence task before it's considered complete for v9. Let's sync up on the internal doc comments.

jzempel (changes requested):
Now I think we're missing `Span` icons. Maybe others. Be sure to scour for `styled` component usage with `cloneElement`. Fwiw, I think we should expose a utility `StyledIcon` (or something similar) base class. But that's TBD.

jzempel (approved):

PR #1754, opened 2024-03-22
feat: adds an internal `getColorV8` utility

jzempel:
## Description

This PR lays the foundation to support staged package migration to v9 where this theming utility will function over the v8 color palette (and be marked as `@deprecated`).


geotrev (approved):
⛏️

PR #1755, opened 2024-03-25
feat(theming)!: add `variables` to theme `colors`

jzempel:
## Description

This PR adds a new `variables` section to Garden's `colors` theme. These variables will provide the precedent lookup for most Garden component colors. The BREAKING CHANGE is the removal of `colors.background` and `colors.foreground` – now obviated by the v9 variables.

## Detail

The forthcoming `getColor` utility replacement PR will utilize the new `variables` to obtain light/dark mode-based colors.

## Checklist

### DIFF
--- a/packages/colorpickers/src/styled/ColorSwatch/StyledColorSwatchLabel.ts
+++ b/packages/colorpickers/src/styled/ColorSwatch/StyledColorSwatchLabel.ts
@@ -7,7 +7,12 @@
 
 import styled, { DefaultTheme, ThemeProps } from 'styled-components';
 import { parseToRgb, readableColor } from 'polished';
-import { DEFAULT_THEME, focusStyles, retrieveComponentStyles } from '@zendeskgarden/react-theming';
+import {
+  DEFAULT_THEME,
+  focusStyles,
+  getColorV8,
+  retrieveComponentStyles
+} from '@zendeskgarden/react-theming';
 import { StyledButtonPreview } from '../ColorPickerDialog/StyledButtonPreview';
 import { IRGBColor } from '../../types';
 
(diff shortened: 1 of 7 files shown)


packages/colorpickers/src/styled/ColorSwatch/StyledColorSwatchLabel.ts
jzempel, line 1: FYI this file is new on the `next` branch and did not receive the `getColorV8('background/foreground')` treatment with #1756 on `main`.

packages/theming/src/types/index.ts
```diff
@@ -75,14 +75,24 @@ export interface IGardenTheme {
...
+        background: Record<string, string>;
+        border: Record<string, string>;
+        foreground: Record<string, string>;
+      };
+      light: {
+        background: Record<string, string>;
+        border: Record<string, string>;
+        foreground: Record<string, string>;
+      };
+    };
```
geotrev, line 95: How are you feeling about the `variables` property name? I was thinking about the [CSS naming](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) the other day (e.g., `scheme` from `prefers-color-scheme`). I don't feel particularly strongly, just curious if that's come up for you at all.

geotrev: Also thinking about how this writes in our components:

- `p.theme.colors.variables.light.foreground`
- `p.theme.colors.scheme.light.foreground`
- etc

😅

jzempel: > How are you feeling about the `variables` property name? I was thinking about [CSS naming](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) the other day (e.g., `scheme` from `prefers-color-scheme`). I don't feel particularly strongly, just curious if that's come up for you at all.

I don't like `scheme` as this object is not a scheme – it's a collection of variables. We're leaning into the design language here and making room for additional themes such as "lightHighContrast", "darkDimmed", etc.

jzempel: > Also thinking about how this writes in our components:
> 
> * `p.theme.colors.variables.light.foreground`
> * `p.theme.colors.scheme.light.foreground`
> * etc
> 
> 😅

We will never reference values in this manner. It should make more sense when the `getColor` PR lands, but the function call will be something like `getColor({ theme, variable: 'foreground.default' });`

geotrev: > I don't like scheme as this object is not a scheme

Makes sense, especially with high contrast and other themes in the variable list. 👌🏻

geotrev (approved):

PR #1761, opened 2024-03-27
feat(chrome)!: detect Footer automatically with Context

geotrev:
## Description

Removes `hasFooter` prop from `react-chrome`'s `Body` component. Uses React context similar to how other components detect when a child is present in the JSX tree.

## Checklist

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -49,6 +49,7 @@ consider additional positioning prop support on a case-by-case basis.
 - Removed `PRODUCT` type export. Use `IHeaderItemProps['product']` instead.
 - Removed `hasFooter` prop for `Body` (no replacement needed)
 - The following React component types have changed:
+  - Removed `IBodyProps` type export.
   - Renamed `ICollapsibleSubNavItemProps` type export to `ISubNavCollapsibleItemProps`.
   - `Header.ItemIcon`: `HTMLAttributes<HTMLElement>` -> `SVGAttributes<SVGElement>`
   - `Nav.ItemIcon`: `HTMLAttributes<HTMLElement>` -> `SVGAttributes<SVGElement>`
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -47,6 +47,7 @@ consider additional positioning prop support on a case-by-case basis.
 #### @zendeskgarden/react-chrome
 
 - Removed `PRODUCT` type export. Use `IHeaderItemProps['product']` instead.
+- Removed `hasFooter` prop for `Body` (no replacement needed)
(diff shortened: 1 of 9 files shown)


docs/migration.md
```diff
@@ -47,6 +47,7 @@ consider additional positioning prop support on a case-by-case basis.
 #### @zendeskgarden/react-chrome
 
 - Removed `PRODUCT` type export. Use `IHeaderItemProps['product']` instead.
+- Removed `hasFooter` prop for `Body` (no replacement needed)
```
jzempel, line 50: Should we call out that `IBodyProps` is no longer exported?

packages/chrome/src/elements/body/Body.tsx
```diff
@@ -5,27 +5,27 @@
...
 import { BodyContext } from '../../utils/useBodyContext';
 
 /**
  * @extends HTMLAttributes<HTMLDivElement>
  */
-export const Body = React.forwardRef<HTMLDivElement, IBodyProps>(({ hasFooter, ...props }, ref) => {
-  const bodyContextValue = useMemo(() => ({ hasFooter: !!hasFooter }), [hasFooter]);
+export const Body = React.forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
+  (props, ref) => {
+    const [hasFooter, setHasFooter] = useState(true);
```
jzempel, line 17: Shouldn't this initialize to `useState(false)`?

geotrev: I'm retaining the previous behavior which is [true by default](https://github.com/zendeskgarden/react-components/blob/next/packages/chrome/src/utils/useBodyContext.ts#L15). I'm OK shifting to `false` if that makes more sense though. EIther way the initial state has a 50/50 chance to not match the instructed UI. 🤔

docs/migration.md
geotrev: Yep, good call.

jzempel (approved):

packages/chrome/src/elements/body/Body.tsx
jzempel: Weird. I think the previous was wrong. The `<Body>` definitely doesn't have a footer .. until it does 😅. There are plenty of areas that are `<Footer>`-less. Let's make sure the layout doesn't clip/overflow for any of these cases.

geotrev: I agree. Makes more sense to retain a "no footer" base case.

PR #1763, opened 2024-03-28
feat: add refactored `getColor` utility

jzempel:
## Description

Provides a fully refactored `getColor` utility that will support v9 color redesigns (incl. dark mode) along with supporting tests and Storybook demo.

## Checklist


packages/theming/src/utils/getColor.ts
```diff
@@ -0,0 +1,225 @@
...
+  if (typeof retVal === 'string') {
+    return retVal;
+  } else if (retVal === undefined) {
+    throw new ReferenceError(`Error: color variable "${path}" is not defined`);
+  } else {
+    throw new TypeError(`Error: unexpected '${typeof retVal}' type for color variable "${path}"`);
+  }
+};
+
+type ColorParameters = {
```
geotrev, line 131: 💡 Wondering if there is consumer value in externalizing this type.

jzempel: > 💡 Wondering if there is consumer value in externalizing this type.

Yeah, I was wondering the same thing. Let's hold off for now, as we don't export types for the other utilities. If a consumer wants it, we can release the exported type.

packages/theming/src/utils/getColor.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  if (typeof _hue === 'object') {
+    retVal = toHex(_hue, shade, offset, scheme);
+  } else if (valid(_hue)) {
+    if (shade === undefined) {
+      retVal = _hue;
+    } else {
+      const _colors = scale([PALETTE.white, _hue, PALETTE.black])
+        .correctLightness()
+        .colors(PALETTE_SIZE + 2);
```
ze-flo, line 97: It might be good to explain that this generates a 14-shade color palette that includes `white` and `black`. Further down `white` and `black` is excluded to create a standardized 12-shade palette.

packages/theming/src/utils/getColor.ts
```diff
@@ -0,0 +1,225 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import { scale, valid } from 'chroma-js';
```
ze-flo, line 8: While [chroma-js](https://gka.github.io/chroma.js/) is a wonderful library, it does add [15.7k](https://bundlephobia.com/package/chroma-js@2.4.2) to the bundle size. And because it's not tree-shakeable, even pulling in a couple methods will import the whole package.

In the future, we could attempt to reproduce something close to `scale` to reduce the JS footprint.

As for `valid`, we could potentially leverage [CSS.support](https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports_static) (browser only). Ex: 

```
CSS.support('color', 'red') // true
CSS.support('color', '#fff') //true

CSS.support('color', 'bad') // false

```

packages/theming/src/utils/getColor.ts
```diff
@@ -0,0 +1,225 @@
...
+    _hue = palette[_hue]; /* ex. `hue` = 'grey' */
+  }
+
+  if (typeof _hue === 'object') {
+    retVal = toHex(_hue, shade, offset, scheme);
+  } else if (valid(_hue)) {
+    if (shade === undefined) {
+      retVal = _hue;
+    } else {
+      const _colors = scale([PALETTE.white, _hue, PALETTE.black])
```
ze-flo, line 95: Chaining `.mode('lab')` could result in more vibrant colors in some cases. 

Here's a [sandbox](https://stackblitz.com/edit/stackblitz-starters-sgcjtk?file=src%2FApp.tsx) highlighting the differences.

packages/theming/src/utils/getColor.ts
jzempel: Great callout @ze-flo. Can I request that we allow this "brand backfill" feature of `getColor` through for now? I'd like to lean into the RFC process for evaluating efficacy of other approaches (let's take a closer look at [color2k](https://color2k.com/), etc). Specifically, let's add a `scale` research task to our v9 dev sequence. That way, we can get this into the hands of early adopters while continuing forward with the v9 effort.

packages/theming/src/utils/getColor.ts
jzempel: Awesome. Similar to the previous comment, I'd like to continue forward and get some of this comparative analysis into our designer's hands for evaluation. To my eyes, the LAB side has a tendency to yellow/brown-out the colors. We know the designers are getting good leverage out of [Leonardo](https://github.com/adobe/leonardo/blob/main/packages/contrast-colors/lib/color.js), which defaults to RGB and then adds some [OKLCH sugar](https://github.com/adobe/leonardo/blob/ac2237b82cae7e10e5248d104c697ac70c168d14/packages/contrast-colors/lib/color.js#L120-L129) on top of `chroma.js`. But we also know that standard CSS advancements are happening fast, and Leonardo only [just graduated](https://www.npmjs.com/package/@adobe/leonardo-contrast-colors?activeTab=versions) from alpha after having been [dormant for nearly 2 years](https://github.com/adobe/leonardo/blob/main/packages/contrast-colors/CHANGELOG.md). Anyway, more to research in this space.

packages/theming/src/utils/getColor.ts
ze-flo: Absolutely! 💯

packages/theming/src/utils/getColor.ts
ze-flo: > To my eyes, the LAB side has a tendency to yellow/brown-out the colors.

Good observation! 👁️ 

> defaults to RGB and then adds some [OKLCH sugar](https://github.com/adobe/leonardo/blob/ac2237b82cae7e10e5248d104c697ac70c168d14/packages/contrast-colors/lib/color.js#L120-L129) on top of chroma.js

Very interesting. Thanks for sharing this.


I agree. We can iteratively fine-tune this as we see fit. Great work!

geotrev (approved):

PR #1776, opened 2024-04-04
fix(grid): ensure `Pane.SplitterButton` is rendered as a sibling, rather than child, of `Pane.Splitter`

jzempel:
## Description

Currently the structure detailed below currently fails WCAG accessibility by violating the following:

> Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies

This PR ensures that the `<Pane.SplitterButton>` will be rendered as a sibling after `<Pane.Splitter>` rather than a child. The updated positioning logic prevents the `::before` pseudo backing "dish" hack behind the transparent `<ChevronButton>`. As a result positioning, box-shadow, stacking, and opacity CSS was moved from `StyledPaneSplitterButton` to a new backing `StyledPaneSplitterButtonContainer` div. The result is an overall simplification of the styling structure.

## Detail

The JSX structure remains the same (non-breaking)

```jsx
<Pane>
  <Pane.Content>...</PaneContent>
  <Pane.Splitter>
    <Pane.SplitterButton />
  <Pane.Splitter>
</Pane>
```

## Checklist

- [x] :guardsman: ~includes new unit tests. Maintain existing coverage (always >= 96%)~


geotrev (approved):
Smooth as butta'

PR #1779, opened 2024-04-04
fix: update theming peer dependencies to ^9.0.0

jzempel:
## Description

Title sez it all ☝️


geotrev (approved):
I had to double check that this would match our `-next` tags. Good to see it does!

PR #1784, opened 2024-04-09
feat(chrome)!: improves markup structure for `Nav` items

geotrev:
## Description

Introduces `Nav.List` subcomponent for a semantically accurate navigation list.

## Detail

- introduced `Nav.List` as a wrapper for `Nav.Item`
- React Context is used to detect if `Nav.Item` is within a `Nav.List`, and if so, includes a styled `li`
- Requires consumers to keep logo/brandmark items out of `Nav.List`

New composition API:

```jsx
<Chrome>
  <Nav> // <nav>
    <Nav.Item hasLogo /> // div
    <Nav.List> // <ul>
      <Nav.Item /> // <li>
      <Nav.Item />
      <Nav.Item />
    </Nav.List>
    <Nav.Item hasBrandmark /> // div
  </Nav>
</Chrome>
```

## Checklist


packages/chrome/src/styled/nav/StyledNavList.ts
```diff
@@ -0,0 +1,30 @@
...
+const COMPONENT_ID = 'chrome.nav_list';
+
+export const StyledNavList = styled.ul.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  order: -1;
```
ze-flo, line 20: CSS `order` rules can lead to accessibility problems and I've had mixed experiences with it. Do we still need these rules or should we leave it to the user to specify the correct order (potential breaking change 💥)?

https://developer.mozilla.org/en-US/docs/Web/CSS/order#accessibility_concerns
https://www.w3.org/TR/css-flexbox-1/#order-accessibility

geotrev: Actually, you just jogged my senses a bit. Since this element is just a `ul` and it's a container, it naturally shouldn't land in any tab/render order. So, we can probably just remove it I think. 🤔 

but your concern is valid and I don't think we need to include `order` in these elements since the original tab oder already prescribes the correct navigation behavior. I'll give it a quick test to confirm, and we can discuss inconsistencies if there are any!

geotrev: I think I see why `order` is used here.

It's to ensure that a child composition of:

```jsx
<Nav>
  <Nav.List>
    <Nav.Item hasBrandmark />
    <Nav.Item>
      <Nav.ItemText />
      <Nav.ItemIcon />
    </Nav.Item>
    <Nav.Item hasLogo />
  </Nav.List>
</Nav>
```

Renders in the expected way in the DOM (using React components for simplicity):

```jsx
<nav>
  <ul>
    <li class="StyledLogoItem />
    <li class="StyledItem">
      <span class="StyledItemIcon" />
      <span class="StyledItemText" />
    </li>
    <li class="StyledBrandmarkItem" />
  </ul>
</nav>
```

E.g., a break-prevention measure. I can't speak to if this is still desirable though. I assume so but maybe JZ or design have further insights. I would agree it is a bit heavy-handed but I can understand the logic after toying with it.

ze-flo: Thanks for your excellent explanation. 💯  It looks like removing `order` isn't so trivial after all and definitely falls out of scope for this PR. Perhaps something we can revisit in the future.

geotrev: Yeah, I agree on revisiting if it makes sense!

jzempel: The `order` CSS throughout Chrome is intentional, and despite the a11y concerns we keep it because enforcing consistent UX tab ordering despite consumer usage is the more important concern. Perhaps if/when a grid layout vs. flex is explored in the future, we can reevaluate.

jzempel (changes requested):
In accordance with the a11y issue we're attempting to address, we're looking for logo and brandmark items to live _outside_ the list. Expecting to see:

```jsx
<Nav>
  <Nav.Item hasLogo />
  <Nav.List>
    <Nav.Item ... />
    <Nav.Item ... />
    <Nav.Item ... />
  </Nav.List>
  <Nav.Item hasBrandmark />
</Nav>
```

This means `<Nav.Item>` will need context to understand whether to render a `<li>` or not.

geotrev:
Thanks @jzempel - that detail wasn't immediately clear to me from our previous conversations. I recall you mentioned context but I didn't connect that to the logo/brandmark items being excluded from the list. Should be updated now.

jzempel (approved):
Approved, with minor URL comment for consideration.

docs/migration.md
```diff
@@ -76,6 +76,9 @@ consider additional positioning prop support on a case-by-case basis.
   - `CollapsibleSubNavItem` -> `SubNav.CollapsibleItem`
   - `SubNavItem` -> `SubNav.Item`
   - `SubNavItemText` -> `SubNav.ItemText`
+- Added `Nav.List` as a semantic wrapper for `Nav.Item`. See
+  [README](https://github.com/zendeskgarden/react-components/blob/main/packages/chrome/README.md#usages)
```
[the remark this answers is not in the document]
geotrev: Oh, I wasn't sure if that was possible here. Good call. I'll update and test it before I merge!

geotrev:
Dang @ze-flo amazing snipe on that pattern file! haha

PR #1786, opened 2024-04-11
fix: update theming peer dependency to ^8.75.0

jzempel:
## Description

This should've happened in conjunction with #1754 where all component packages now rely on the `getColorV8` utility. Next best is to have this go out with the next patch release. In accordance with Garden [ADR-001](https://github.com/zendeskgarden/react-components/blob/main/docs/adrs/001-theming-peer-dependency.md), this is not considered a breaking change.


ze-flo (approved):
Good catch! 💯

PR #1789, opened 2024-04-12
fix(deps): use -next versions for react-theming peer dep

geotrev:
## Description

`^9.0.0` version range doesn't capture prereleases and seems to still require the `--force` flag when running an install. Using `>=9.0.0-next` which captures only v9 next versions. We can revert back to `^9.0.0` right before v9 is fully published.

Separately, updates a few missed peer dep versions from #1779.


ze-flo (approved):
:shipit:

jzempel (approved):

PR #1790, opened 2024-04-17
fix(Notifications): improve screen-reader support

ze-flo:
## Description

- Sets `Nofitication`'s role to `alert` to improve compatibility with JAWS, Narrator, and NVDA.
- Set `aria-hidden` to Notification's `Close` button to improve UX. 

## Detail
I've tested the following combinations using the latest version of the browser and screen-reader:

- Chrome + JAWS
- Chrome + NVDA
- Edge + Narrator
- Chrome + Narrator
- Chrome + VoiceOver
- Safari + VoiceOver

Screen Shot 2024-04-17 at 12 19 49 PM

### Two methods were compared:

1. "Incremental notifications": This is current pattern where a role of `status` or `alert` is applied to `Notification`. When the Notification incrementally mounts, the SR will read out loud its content.
2. "Persistent `aria-live` region": This region has an `aria-live` attribute set to either `polite` or `assertive` and is mounted by the `ToastProvider` on load. Changes to its content will be announced incrementally by setting the attribute `aria-atomic` to `false`.

## Conclusion

There is no perfect solution but `role="alert"` on `Notification` seems best.

### TL;DR
The "Persistent `aria-live` region" offers wider SR compatibility with with `aria-live=polite` but completely fails with Chrome + Narrator. Additionally, the assertive behavior of JAWS and NVDA remains `polite` regardless of `aria-live` setting.

Setting `role="alert"` on `Notification` provides more consistent results. Most SR's except `Narrator` would announce the notification with an alert prefix or a chime (VoiceOver) before reading its content. That helps users better distinguish such events but at the cost of the occasional announcement interruption.

## Checklist

### DIFF
--- a/packages/notifications/src/elements/Notification.tsx
+++ b/packages/notifications/src/elements/Notification.tsx
@@ -20,7 +20,7 @@ export const Notification = forwardRef<HTMLDivElement, INotificationProps>((prop
   const hue = props.type && validationHues[props.type];
 
   return (
-    <StyledNotification ref={ref} type={props.type} isFloating {...props} role="alert">
+    <StyledNotification ref={ref} type={props.type} isFloating role="alert" {...props}>
       {props.type && (
         <StyledIcon hue={hue}>
           <Icon />
--- a/packages/notifications/src/elements/Notification.tsx
+++ b/packages/notifications/src/elements/Notification.tsx
@@ -15,30 +15,22 @@ import { validationIcons, validationHues } from '../utils/icons';
 /**
  * @extends HTMLAttributes<HTMLDivElement>
  */
-export const Notification = forwardRef<HTMLDivElement, INotificationProps>(
-  ({ role, ...props }, ref) => {
-    const Icon = props.type ? validationIcons[props.type] : InfoStrokeIcon;
-    const hue = props.type && validationHues[props.type];
+export const Notification = forwardRef<HTMLDivElement, INotificationProps>((props, ref) => {
+  const Icon = props.type ? validationIcons[props.type] : InfoStrokeIcon;
+  const hue = props.type && validationHues[props.type];
 
-    return (
-      <StyledNotification
-        ref={ref}
-        type={props.type}
-        isFloating
-        {...props}
-        role={role === undefined ? 'status' : role}
-      >
-        {props.type && (
-          <StyledIcon hue={hue}>
-            <Icon />
-          </StyledIcon>
-        )}
+  return (
+    <StyledNotification ref={ref} type={props.type} isFloating {...props} role="alert">
+      {props.type && (
+        <StyledIcon hue={hue}>
+          <Icon />
+        </StyledIcon>
+      )}
 
-        {props.children}
-      </StyledNotification>
-    );
-  }
-);
+      {props.children}
+    </StyledNotification>
+  );
+});
 
 Notification.displayName = 'Notification';
 
--- a/packages/notifications/src/elements/Notification.tsx
+++ b/packages/notifications/src/elements/Notification.tsx
@@ -15,22 +15,24 @@ import { validationIcons, validationHues } from '../utils/icons';
 /**
  * @extends HTMLAttributes<HTMLDivElement>
  */
-export const Notification = forwardRef<HTMLDivElement, INotificationProps>((props, ref) => {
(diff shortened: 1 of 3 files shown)


packages/notifications/src/elements/content/Close.tsx
```diff
@@ -20,7 +20,7 @@ export const Close = React.forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HT
     const hue = useNotificationsContext();
 
     return (
-      <StyledClose ref={ref} hue={hue} aria-label={ariaLabel} {...props}>
+      <StyledClose ref={ref} hue={hue} aria-label={ariaLabel} {...props} aria-hidden>
```
ze-flo, line 23: This is a controversial quality-of-life enhancement. With the assumption that most notifications are auto-dismissed and because we are against focus stealing, it's unlikely screen-reader users will be able to use the `Close` button. Therefore, I think it's better to remove it from the a11y tree to prevent its announcement.

Additionally, I could first check that the `autoDismiss` is `true` before applying `aria-hidden`. 

Thoughts? 🤔

packages/notifications/src/elements/Notification.tsx
```diff
@@ -15,30 +15,22 @@ import { validationIcons, validationHues } from '../utils/icons';
...
-        {...props}
-        role={role === undefined ? 'status' : role}
-      >
-        {props.type && (
-          <StyledIcon hue={hue}>
-            <Icon />
-          </StyledIcon>
-        )}
+  return (
+    <StyledNotification ref={ref} type={props.type} isFloating {...props} role="alert">
```
jzempel, line 23: Please retain the existing `role` override so that consumers can continue to cater to their specific needs.

packages/notifications/src/elements/content/Close.tsx
jzempel: This doesn't feel right to me. Garden's approach to accessibility is a pragmatic one. If a close exists, there shouldn't be harm in announcing it. I'm also not aware of this being flagged during a11y audit – so best not to touch it.

geotrev: My only concern is your wording: "most notificaitons" - implying there are some notifications that may still receive focus, right?

packages/notifications/src/elements/Notification.tsx
ze-flo: From my testing, any role other than `alert` will not trigger an announcement on JAWS or NVDA. Leaving it up to the user could create an opportunity for an A11Y bug. Can `Notification` be used outside of a "toast" pattern?

packages/notifications/src/elements/content/Close.tsx
ze-flo: Sounds good. I'll revert.

ze-flo:
> Narrator is not part of Garden's testing matrix. We follow [Zendesk's recommended screen readers and browsers](https://www.zendesk.com/company/agreements-and-terms/accessibility/) – which includes Safari + VO. What is the affect on that combo?

Thanks for the helpful link. I'll test that combo. 👍

packages/notifications/src/elements/Notification.tsx
jzempel: > Can `Notification` be used outside of a "toast" pattern?

By definition, yes. Conversely, `ToastProvider` can receive other types of content. Let's stick with Garden's approach of a) doing the right thing, b) giving the user control over the corner cases.

packages/notifications/src/elements/Notification.tsx
```diff
@@ -15,30 +15,22 @@ import { validationIcons, validationHues } from '../utils/icons';
...
-        {...props}
-        role={role === undefined ? 'status' : role}
-      >
-        {props.type && (
-          <StyledIcon hue={hue}>
-            <Icon />
-          </StyledIcon>
-        )}
+  return (
+    <StyledNotification ref={ref} type={props.type} isFloating role="alert" {...props}>
```
jzempel, line 23:
```suggestion
    <StyledNotification ref={ref} type={type} isFloating role="alert" {...props}>
```

Not exactly related to this PR, but should `type` be structured out in the parameters above so that it doesn't (re)spread?

jzempel (approved):
Approved, but left a minor cleanup that was present before this work.

PR #1791, opened 2024-04-17
feat: adds styled icon base component

geotrev:
## Description

Adds the base styled icon component to `react-theming`, along with updates to two components to demonstrate usage.

A follow up PR will update the remaining component instances (13 in total).

## Detail

The primary goals of this base component:

- Make the `cloneElement` call centralized
- Omit `theme` from the resulting prop spread

Separately, using this PR as an opportunity for [transient props](https://styled-components.com/docs/faqs#transient-props-since-51) to prevent further prop-to-DOM leaking in the affected components.

## Checklist

### DIFF
--- a/packages/theming/src/index.ts
+++ b/packages/theming/src/index.ts
@@ -24,6 +24,7 @@ export { useWindow } from './utils/useWindow';
 export { useText } from './utils/useText';
 export { default as menuStyles } from './utils/menuStyles';
 export { focusStyles, SELECTOR_FOCUS_VISIBLE } from './utils/focusStyles';
+export { StyledBaseIcon } from './styled/StyledBaseIcon';
 
 export {
   ARROW_POSITION,
--- a/packages/theming/src/index.ts
+++ b/packages/theming/src/index.ts
@@ -24,7 +24,7 @@ export { useWindow } from './utils/useWindow';
 export { useText } from './utils/useText';
 export { default as menuStyles } from './utils/menuStyles';
 export { focusStyles, SELECTOR_FOCUS_VISIBLE } from './utils/focusStyles';
-export { StyledBaseIcon } from './styled/StyledBaseIcon';
+export { StyledBaseIcon } from './utils/StyledBaseIcon';
 
 export {
   ARROW_POSITION,
--- a/packages/theming/src/styled/StyledBaseIcon.ts
+++ b/packages/theming/src/styled/StyledBaseIcon.ts
@@ -1,0 +1,27 @@
+/**
+ * Copyright Zendesk, Inc.
+ *
+ * Use of this source code is governed under the Apache License, Version 2.0
+ * found at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+import styled from 'styled-components';
+import React, { Children } from 'react';
+import retrieveComponentStyles from '../utils/retrieveComponentStyles';
+import DEFAULT_THEME from '../elements/theme';
+
+const COMPONENT_ID = 'icon.base';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const StyledBaseIcon = styled(({ children, theme, ...props }) =>
+  React.cloneElement(Children.only(children), props)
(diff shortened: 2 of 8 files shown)


packages/theming/src/styled/StyledBaseIcon.ts
```diff
@@ -0,0 +1,27 @@
...
+import styled from 'styled-components';
+import React, { Children } from 'react';
+import retrieveComponentStyles from '../utils/retrieveComponentStyles';
+import DEFAULT_THEME from '../elements/theme';
+
+const COMPONENT_ID = 'icon.base';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const StyledBaseIcon = styled(({ children, theme, ...props }) =>
+  React.cloneElement(Children.only(children), props)
```
ze-flo, line 17: `props.theme` appears to be [excluded](https://github.com/styled-components/styled-components/blob/v6.1.8/packages/styled-components/src/models/StyledComponent.ts#L146) already. Was it still landing in the DOM?

geotrev: Strange, it still shows up for me 🤷🏻

geotrev: Oh, I see what's happening. `theme` is omitted in a normal `styled.div` (or any tag) call. But because of the `cloneElement`, we need to manually omit it.

jzempel, line 17: [nit] prefer a `{ cloneElement }` import over `React.cloneElement`.

packages/theming/src/index.ts
```diff
@@ -24,6 +24,7 @@ export { useWindow } from './utils/useWindow';
 export { useText } from './utils/useText';
 export { default as menuStyles } from './utils/menuStyles';
 export { focusStyles, SELECTOR_FOCUS_VISIBLE } from './utils/focusStyles';
+export { StyledBaseIcon } from './styled/StyledBaseIcon';
```
jzempel, line 27: Should this just live in `utils`? We have other `...Styles` utilities in there based heavily on `styled-components`. I'm thinking we can avoid this one-off.

packages/theming/src/styled/StyledBaseIcon.ts
```diff
@@ -0,0 +1,27 @@
...
+).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+`;
+
+StyledBaseIcon.defaultProps = {
+  theme: DEFAULT_THEME
+};
```
jzempel, line 27: Recommend removing from the base.

packages/theming/src/index.ts
geotrev: Yeah that feels better.

packages/theming/src/utils/StyledBaseIcon.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars,garden-local/require-default-theme
+export const StyledBaseIcon = styled(({ children, theme, ...props }) =>
+  React.cloneElement(Children.only(children), props)
+).attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})`
+  /* stylelint-disable no-empty-block */
+`;
```
jzempel, line 21:
```suggestion
export const StyledBaseIcon = styled(({ children, theme, ...props }) =>
  React.cloneElement(Children.only(children), props)
)`
  /* stylelint-disable no-empty-block */
`;
```

The `attrs` block should not be needed. Every Garden extension will apply their own. And we shouldn't assume these attributes for any, potential external consumers. Also, the component ID is only used for `retrieveComponentStyles` CSS overrides.

jzempel (changes requested):

PR #1795, opened 2024-04-22
fix(theming): update palette color values

jzempel:
## Description

Updated v9 palette color values and added an associated story demo.

## Checklist


packages/theming/demo/utilities.stories.mdx
```diff
@@ -18,6 +19,21 @@ import README from '../README.md';
 
 # Demo
 
+## PALETTE
+
+<Canvas>
+  <Story
+    name="PALETTE"
+    args={{ palette: PALETTE }}
```
ze-flo, line 27: Should we let the user interact with the arguments for this story? If so, the current implementation will only track changes to the outer color keys. That's because `Hue` gets the colors from the [theme.palette](https://github.com/zendeskgarden/react-components/pull/1795/files#diff-fe006ad1caedbe3ffabf042e74a6e9a5b8f3d63ad21cd74525f7521381442877R74) and not the palette object injected as an argument. 

Screen Shot 2024-04-22 at 12 58 56 PM

packages/theming/demo/stories/PaletteStory.tsx
```diff
@@ -0,0 +1,133 @@
...
+        {rows.map((row, index) => {
+          const hue1 = row[0];
+          const hue2 = row[1];
+
+          return (
+            <StyledRow key={index}>
+              <Grid.Col sm>
+                <Hue hue={hue1} />
+              </Grid.Col>
+              {hue1 !== 'product' && <StyledCol sm>{hue2 && <Hue hue={hue2} />}</StyledCol>}
```
ze-flo, line 126: That's always `true` since `product` is [excluded](https://github.com/zendeskgarden/react-components/pull/1795/files#diff-fe006ad1caedbe3ffabf042e74a6e9a5b8f3d63ad21cd74525f7521381442877R101).

packages/theming/demo/utilities.stories.mdx
jzempel: I'm not super concerned about whether or not the story is interactive – more interested in demonstrating where the values are coming from. However, there is some copy/paste jank in here from the `website` implementation. Now cleaned up. 🧹

ze-flo (approved):
Colors for the win! 🌈 🤩

PR #1797, opened 2024-04-24
feat(theming): add system `opacity` values to the theme

jzempel:
## Description

This PR introduces a consistent set of themed opacity values and updates `getColor` to receive either one of the theme's `100`-`1200` keys or a value between `0` and `1`. Garden will strive to use `theme.opacity` keys for all v9 opacity/transparency applications.

## Detail

I weighed adding `opacity` under the `colors` theme structure, but stuck with root placement since CSS opacity applies to more than just color.

## Checklist


geotrev (approved):
I was thinking about fallback values when `theme.opacity` may contain invalid entries, but it may be too heavy handed to force consumers down an unexpected path vs letting `getColor` provide users with an error.

jzempel:
> I was thinking about fallback values when `theme.opacity` may contain invalid entries, but it may be too heavy handed to force consumers down an unexpected path vs letting `getColor` provide users with an error.

Ah, that reminds me, I wanted to handle invalid `opacity` references before the `rgba` function throws a `PolishedError`. Added now.

PR #1799, opened 2024-04-29
fix(forms): updates Message to receive margin when intermediaries are present

geotrev:
## Description

Implements a general sibling combinator for `Message`s in a `Field`. This way an `Input` and`Message` can have intermediary elements but the latter can still space correctly (like in the case of `DatePicker`).

## Detail

Currently, `DatePicker` doesn't hvae any spacing when a `Message` is present. This happens because the menu wrapper sits between them, and the existing adjacent sibling combinator invalidates this scenario:

Screenshot 2024-04-29 at 1 32 05 PM

Compare to `Input`, which has the correct `margin-top` when a `Message` is used:

Screenshot 2024-04-29 at 1 29 22 PM

Now using the general sibling combinator:

Screenshot 2024-04-29 at 1 32 15 PM

## Checklist

### DIFF
--- a/packages/datepickers/demo/datepicker.stories.mdx
+++ b/packages/datepickers/demo/datepicker.stories.mdx
@@ -40,7 +40,7 @@ import README from '../README.md';
       },
       validationLabel: {
         control: { type: 'text' },
-        table: { category: 'Input' }
+        table: { category: 'Message' }
       }
     }}
     parameters={{
--- a/packages/datepickers/demo/datepicker.stories.mdx
+++ b/packages/datepickers/demo/datepicker.stories.mdx
@@ -16,7 +16,12 @@ import README from '../README.md';
 <Canvas>
   <Story
     name="Datepicker"
-    args={{ dateStyle: DATE_STYLE_OPTIONS[1], eventsEnabled: true, isAnimated: true }}
+    args={{
+      dateStyle: DATE_STYLE_OPTIONS[1],
+      eventsEnabled: true,
+      isAnimated: true,
+      message: 'Message'
+    }}
     argTypes={{
       value: { control: 'date' },
       minValue: { control: 'date' },
@@ -26,15 +31,16 @@ import README from '../README.md';
         options: DATE_STYLE_OPTIONS,
         table: { category: 'Story' }
       },
+      hasMessage: { name: 'Message', control: { type: 'boolean' }, table: 'Story' },
       message: { name: 'children', control: { type: 'text' }, table: { category: 'Message' } },
       validation: {
         options: ['success', 'warning', 'error'],
         control: { type: 'radio' },
-        table: { category: 'Message' }
+        table: { category: 'Input' }
       },
       validationLabel: {
         control: { type: 'text' },
-        table: { category: 'Message' }
+        table: { category: 'Input' }
       }
     }}
     parameters={{
--- a/packages/datepickers/demo/datepicker.stories.mdx
+++ b/packages/datepickers/demo/datepicker.stories.mdx
@@ -25,6 +25,16 @@ import README from '../README.md';
         control: 'radio',
         options: DATE_STYLE_OPTIONS,
         table: { category: 'Story' }
+      },
+      message: { name: 'children', control: { type: 'text' }, table: { category: 'Message' } },
+      validation: {
+        options: ['success', 'warning', 'error'],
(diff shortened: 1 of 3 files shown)


packages/datepickers/demo/stories/DatepickerStory.tsx
```diff
@@ -8,28 +8,45 @@
...
       <Row style={{ height: 'calc(100vh - 80px)' }}>
-        <Col textAlign="center" alignSelf="center">
-          <Field>
-            <Label hidden>{Datepicker.displayName}</Label>
-            <Datepicker {...args} formatDate={formatDate} isCompact={isCompact}>
-              <Input isCompact={isCompact} style={{ width: isCompact ? 256 : 320 }} />
-            </Datepicker>
-          </Field>
+        <Col alignSelf="center">
+          <div style={{ display: 'flex', justifyContent: 'center' }}>
```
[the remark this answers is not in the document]
geotrev: Nice 👍🏻 I looked at the `Col` props and missed that prop for `Row`.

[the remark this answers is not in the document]
geotrev: Lookin good I think. Removed the hardcoded width so the input could adjust with the column.

packages/datepickers/demo/datepicker.stories.mdx
```diff
@@ -25,6 +25,16 @@ import README from '../README.md';
         control: 'radio',
         options: DATE_STYLE_OPTIONS,
         table: { category: 'Story' }
+      },
+      message: { name: 'children', control: { type: 'text' }, table: { category: 'Message' } },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Message' }
+      },
```
jzempel, line 34: Let's restructure this to behave more like [Input](https://zendeskgarden.github.io/react-components/?path=/story/packages-forms-input--uncontrolled) or [Combobox](https://zendeskgarden.github.io/react-components/?path=/story/packages-dropdowns-combobox--combobox)
- default the message `children` to "Message"
- add a **Story** category section control for toggling the `Message` on
- move `validation` to an **Input** category section and allow it to control validation styling for both the input and the message together (as there is no design recommendation for seeing them happen separately)

packages/datepickers/demo/stories/DatepickerStory.tsx
```diff
@@ -8,27 +8,42 @@
...
+}) => {
   const formatDate = (date: Date) =>
     new Intl.DateTimeFormat(args.locale, { dateStyle }).format(date);
 
   return (
     <Grid>
-      <Row style={{ height: 'calc(100vh - 80px)' }}>
-        <Col textAlign="center" alignSelf="center">
+      <Row justifyContent="center" style={{ height: 'calc(100vh - 80px)' }}>
+        <Col alignSelf="center" xs={12} md={4}>
```
jzempel, line 36:
```suggestion
        <Col textAlign="center" alignSelf="center">
```

...let's avoid the change in order to remain consistent with all other centered non-responsive dropdown type stories

jzempel (changes requested):

packages/datepickers/demo/datepicker.stories.mdx
```diff
@@ -25,6 +30,17 @@ import README from '../README.md';
...
+      hasMessage: { name: 'Message', control: { type: 'boolean' }, table: 'Story' },
+      message: { name: 'children', control: { type: 'text' }, table: { category: 'Message' } },
+      validation: {
+        options: ['success', 'warning', 'error'],
+        control: { type: 'radio' },
+        table: { category: 'Input' }
+      },
+      validationLabel: {
+        control: { type: 'text' },
+        table: { category: 'Input' }
```
jzempel, line 43:
```suggestion
        table: { category: 'Message' }
```

[nit] the `validationLabel` actually belongs to the Message's icon. See https://garden.zendesk.com/components/input#message

jzempel (changes requested):

packages/datepickers/demo/datepicker.stories.mdx
geotrev: Ah, shoot, I see it now in the Input story too. Fixing.

jzempel (approved):

PR #1804, opened 2024-05-09
fix(Buttons): correctly support overriding default `data-garden-id` attribute

ze-flo:
## Description

~~### ❗❗ BREAKING CHANGES ❗❗~~

~~Users targeting `Button` and `IconButton` using the `data-garden-id` attribute (commonly for testing purposes) should acknowledge these id changes:~~

<details><summary>See previous breaking changes</summary>

 **Module**                           | **Component**         | **Previous ID**       | **New ID**                          
--------------------------------------|-----------------------|-----------------------|-------------------------------------
 `@zendeskgarden/react-notifications` | `GlobalAlert.Close`   | `buttons.button`      | `notifications.global-alert.button` 
 `@zendeskgarden/react-colorpickers`          | `ColorSwatchDialog`   | `buttons.button`      | `colorpickers.colordialog_button`   
 `@zendeskgarden/react-dropdowns` (formerly `react-dropdowns.next`)     | `Menu`                | `buttons.button`      | `dropdowns.menu.button`             
 `@zendeskgarden/react-grid`          | `Pane.SplitterButton` | `buttons.icon_button` | `pane.splitter_button`              
 `@zendeskgarden/react-notifications` | `GlobalAlert`         | `buttons.icon_button` | `notifications.global-alert.close`  
</details>

~~We can avoid the breaking change by renaming the new ids to match the previous ones.~~

## Detail
- Fix support for overriding default `data-garden-id` for each button component exported from `@zendeskgarden/react-buttons`

## Example

https://github.com/zendeskgarden/react-components/blob/717a8d56d9a722c9e0cb76b6f4e7b8c77dc3791b/packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts#L20

https://github.com/zendeskgarden/react-components/blob/717a8d56d9a722c9e0cb76b6f4e7b8c77dc3791b/packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts#L92-L100

Screen Shot 2024-05-08 at 5 20 19 PM

## Checklist


packages/buttons/src/styled/StyledAnchor.ts
```diff
@@ -24,7 +25,7 @@ export const StyledAnchor = styled(StyledButton).attrs(props => ({
 }))`
   direction: ${props => props.theme.rtl && 'rtl'};
 
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${props => retrieveComponentStyles(props['data-garden-id'], props)};
```
ze-flo, line 28: `props` always include the ones set in `.attrs`. If we pursue with this type of migration for every component and drop the `componentId` param from `retrieveComponentStyles, we could rewrite this to:

```suggestion
  ${retrieveComponentStyles};
```

jzempel (changes requested):
I don't think this is necessary. Please compare `menu` with `combobox` in the `dropdowns.next` package. While it doesn't always get sorted in dev, the attrs work out as expected in production.

packages/buttons/src/styled/StyledIconButton.ts
```diff
@@ -69,17 +69,17 @@ const iconStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
   `;
 };
 
-export const StyledIconButton = styled(StyledButton as 'button').attrs({
-  'data-garden-id': COMPONENT_ID,
+export const StyledIconButton = styled(StyledButton).attrs<IButtonProps>(props => ({
+  'data-garden-id': props['data-garden-id'] || COMPONENT_ID,
```
ze-flo, line 73: As mentioned in this [comment](https://github.com/zendeskgarden/react-components/pull/1807/files/3464347fc3856858345009856e5bde2901e032f5#r1602045804), this pattern causes a regression with `IconButton`. `StyledIconButton` now inherits the `COMPONENT_ID` `buttons.button` from `StyledButton`. It should be: 

https://github.com/zendeskgarden/react-components/blob/52822988c09ec0bbcf0a82f63d3d8622b74d0b98/packages/buttons/src/styled/StyledIconButton.ts#L14

Screen Shot 2024-05-15 at 7 37 03 AM


For `styled-components` extending another `styled-component`, we must first check the `data-garden-id` doesn't match the one set on the `styled-component` it extends.

```suggestion
export const StyledIconButton = styled(StyledButton).attrs<IButtonProps>(props => ({
  'data-garden-id':
    props['data-garden-id'] && props['data-garden-id'] !== 'buttons.button'
      ? props['data-garden-id']
      : COMPONENT_ID,
```

This [sandbox](https://codesandbox.io/p/sandbox/overriding-attrs-2-l3y6mj?file=%2Fsrc%2FExample.tsx%3A131%2C17) highlights all the possible combinations, what works and what doesn't. There's a lot there to unfold. You've been warned. 😅

ze-flo:
# Summary



Here are some rules-of-thumb when creating Styled components with `data-garden-id`:

1. `styled-components` that apply styles to an HTML element should uses the `attrs` [Prop Factory](https://styled-components.com/docs/api#.attrs) fn to support overriding the default `data-garden-id` attribute if it receives one.
2. `styled-components` that extend another `styled-component` should:
  a. Use the `attrs` 'Prop Factory' fn to support overriding the default `data-garden-id` attribute if it receives one. 
  b. AND first checks that the received `data-garden-id` doesn't match the one set on the styled-component it extends.


```js

const BASIC_BTN_ID = 'basic-btn';
/**
 * styled-component that apply styles to an HTML element should
 * uses the `attrs` 'Prop Factory' fn to support overriding
 * the default `data-garden-id` attribute if it receives one.
 * @see https://styled-components.com/docs/api#.attrs
 * 
 */
const StyledBasicBtn = styled.button.attrs((props) => ({
  'data-garden-id': props['data-garden-id'] || BASIC_BTN_ID
}))`
  ...
`;

CONST PINK_BTN_ID = 'pink-btn';

/**
 * styled-components that extend another styled-component should:
 * 1. use the `attrs` 'Prop Factory' fn to support overriding
 *    the default `data-garden-id` attribute if it receives one.
 * 2. First checks that the received `data-garden-id`
 *    doesn't match the one set on the styled-component it extends.
 */
const StyledPinkBtn = styled(StyledBasicBtn).attrs((props) => ({
  'data-garden-id':
    props['data-garden-id'] && props['data-garden-id'] !== BASIC_BTN_ID // Very important!
      ? props['data-garden-id']
      : PINK_BTN_ID
}))`
  background: hotpink;
`;
```

This [sandbox](https://codesandbox.io/p/sandbox/overriding-attrs-2-l3y6mj?file=%2Fsrc%2FExample.tsx%3A131%2C17) highlights all the possible combinations, what works , and what doesn't.

overriding_attrs

packages/buttons/src/styled/StyledAnchor.ts
jzempel: I would discourage the rewrite at this point since a v10 goal is to consolidate packages. At that point, we should visit styled component re-wrapping in earnest and make sure we have the right solution for CSS [targeting](https://garden.zendesk.com/components/theme-provider#targeting).

packages/dropdowns/src/elements/menu/Menu.tsx
```diff
@@ -87,12 +88,12 @@ export const Menu = forwardRef<HTMLUListElement, IMenuProps>(
...
-        <StyledButton {...triggerProps}>
+        <Button {...triggerProps} data-garden-id={BTN_COMPONENT_ID}>
           {button}
-          <StyledButton.EndIcon isRotated={isExpanded}>
+          <Button.EndIcon isRotated={isExpanded}>
             <ChevronIcon />
-          </StyledButton.EndIcon>
-        </StyledButton>
+          </Button.EndIcon>
+        </Button>
```
jzempel, line 96: Let's keep `StyledButton` restored rather than making architecture assertions in this PR.

jzempel (changes requested):
- needs `migration.md` update similar to https://github.com/zendeskgarden/react-components/pull/1807/commits/3464347fc3856858345009856e5bde2901e032f5
- consider using simple `data-garden-id` comparison strings and avoiding the added export/import clunkiness (i.e. `toHaveAttribute('data-garden-id', 'pane.splitter_button')`)

In general, let's really try to limit the spread here. I don't think we want this pattern to be normative. The backing and overarching goal is simply to retain the `modal.close` component ID while being able to reuse `Button` 😅

packages/dropdowns/src/views/menu/StyledButton.ts
```diff
@@ -1,23 +0,0 @@
...
-import { Button } from '@zendeskgarden/react-buttons';
-
-const COMPONENT_ID = 'dropdowns.menu.button';
-
-export const StyledButton = styled(Button).attrs({
-  'data-garden-id': COMPONENT_ID,
-  'data-garden-version': PACKAGE_VERSION
-})`
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
-`;
```
ze-flo, line 19: RE: https://github.com/zendeskgarden/react-components/pull/1804#discussion_r1603476175

After removing all code causing a breaking-change, we would be left with:

```
export const StyledButton = styled(Button)``
```

So I think it's better to delete.

ze-flo:
Verified that breaking-changes have been removed and that `data-garden-id` match with V8 for the following components:

[ ] GlobalAlert.Button
[ ] GlobalAlert.Close
[ ] ColorSwatchDialog
[ ] Menu
[ ] Pane.SplitterButton

This also aligns with updated specs.

Screenshot 2024-05-16 at 10 37 19 AM

packages/buttons/src/styled/StyledAnchor.ts
jzempel, line 1: Maybe I'm missing something, but I'm not seeing the need for changes to `StyledAnchor` in this PR.

packages/buttons/src/styled/StyledButton.ts
jzempel, line 1: Can this rollback now that there are no more competing `styled(Button)` extensions? The remaining `styled(StyledButton)` extension still works as expected without this, right?

ze-flo: If I'm understanding your questions correctly, rolling back would create a breaking-change with the `Modal.Close` button in https://github.com/zendeskgarden/react-components/pull/1808. As you can see below, the `data-garden-id` will be `buttons.icon_button` rather than `modals.close`.

https://github.com/zendeskgarden/react-components/assets/6879688/485f6235-eed8-4b08-986e-36a1d1d175e7


You can see this in action by updating `line 51` in the [Sandbox](https://codesandbox.io/p/sandbox/overriding-attrs-2-l3y6mj?file=%2Fsrc%2FExample.tsx%3A131%2C17).

Screen Recording 2024-05-17 at 5 41 52 AM

packages/buttons/src/styled/StyledAnchor.ts
jzempel: Yes, please. Again, let's keep this somewhat hack as minimal as possible so that we have less to unwind when package consolidation is a thing.

packages/buttons/src/styled/StyledButton.ts
jzempel: Ah, got it. Thanks for the clarification/reminder. Sorry, my brain is slow on accepting this (anti)pattern 😜

ze-flo: The crazy magic built into `styled-components` took me a while to grasp and is forcing us down a less-than-ideal path. I definitely wanna revisit this in the near future to find a better solution.

jzempel (approved):

PR #1806, opened 2024-05-09
feat(theming): enhance `getColor` to accept rgba and literal color variables

jzempel:
## Description

Mainly adds support for `'rgba(param1, param2')` notation to theming color variable values, where `param1` is either a property reference (i.e. `neutralHue.200`) or a literal color and `param2` is either a theming opacity value (i.e. `600`) or a floating point number between 0 and 1.

## Detail

The added unit tests are probably the best explanation of the new capabilities. A new `background.disabled` has been added to the theming color variables. The PR is temporarily on hold for finalized `disabled` values from Garden design.


packages/theming/src/utils/getColor.ts
```diff
@@ -135,6 +135,56 @@ const toProperty = (object: object, path: string) => {
   }
 };
 
+/* derive the property + transparency from the given rgba variable value */
+const fromRgba = (value: string) => {
+  let retVal;
+  const regex =
+    /rgba\s*\(\s*(?<property>[#\w.]+)\s*,\s*(?<alpha>[\w.]+)\s*\)/gu; /* ex. 'rgba(primaryHue.700, 600)' */
```
geotrev, line 142: Would it be worth making the space before the alpha value optional? Feels like an easy thing to  miss and spend a long time debugging. 🤔  E.g., `\s?`.

geotrev: Unless we have appetite to create a lint rule for this, which feels like an acceptable path as well.

jzempel: Isn't it already optional? Perhaps you can give me the valid string that fails. Here's a link for testing: https://regex101.com/r/td33LV/1

geotrev: Oops - I missed the asterisk. All good on my end!

geotrev (approved):

PR #1807, opened 2024-05-13
feat(buttons): new light/dark mode colors

jzempel:
## Description

...rooted in semantic variables.

## Detail

Verified stateful (default, hover, active, focused) color treatment for the following in both light and dark modes – with icons enabled wherever applicable:

- `<Anchor>`
  - `<Anchor isDanger>`
- `<Button isLink>`
  - `<Button isLink isNeutral>` (no change)
  - `<Button isLink isDanger>`
  - `<Button isLink isDanger focusInset>`
  - `<Button isLink disabled>`
- `<Button isBasic>`
  - `<Button isBasic isNeutral>`
  - `<Button isBasic isNeutral focusInset>`
  - `<Button isBasic isDanger>`
  - `<Button isBasic isDanger focusInset>`
  - `<Button isBasic disabled>`
- `<Button isPrimary>`
  - `<Button isPrimary isNeutral>`
  - `<Button isPrimary isNeutral focusInset>`
  - `<Button isPrimary isDanger>`
  - `<Button isPrimary isDanger focusInset>`
  - `<Button isPrimary disabled>`
- `<Button>`
  - `<Button isNeutral>`
  - `<Button isNeutral focusInset>`
  - `<Button isDanger>`
  - `<Button isDanger focusInset>`
  - `<Button disabled>`
- `<SplitButton>`
  - `isBasic`
  - `isBasic isNeutral`
  - `isBasic isDanger`
  - `isPrimary`
  - `isPrimary isNeutral`
  - `isPrimary isDanger`
  - `isNeutral`
  - `isDanger`
  - `disabled`
- `<ToggleButton>`
  - `<TogglButton isPressed>`

## Checklist


packages/buttons/src/styled/StyledButton.ts
```diff
@@ -49,88 +45,104 @@ export const getHeight = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
...
-  if (props.disabled || (props.isNeutral && props.isPrimary && !props.isDanger)) {
-    hue = 'neutralHue';
-  } else if (props.isDanger) {
-    hue = 'dangerHue';
-  } else {
-    hue = 'primaryHue';
-  }
+  const disabledBackgroundColor = getColor({ theme, variable: 'background.disabled' });
+  const disabledForegroundColor = getColor({ theme, variable: 'foreground.disabled' });
+  const offset100 = { dark: { offset: -100 }, light: { offset: +100 } };
```
ze-flo, line 60: Do we need the [unary +](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus) operator?

```suggestion
  const offset100 = { dark: { offset: -100 }, light: { offset: 100 } };
```

packages/buttons/src/styled/StyledButton.ts
```diff
@@ -49,88 +45,104 @@ export const getHeight = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
...
+  const offset200 = { dark: { offset: -200 }, light: { offset: +200 } };
+
+  if (isLink) {
+    /*
+     * Anchor / link button styling
+     */
+    const options = { theme, variable: `foreground.${isDanger ? 'danger' : 'primary'}` };
```
ze-flo, line 67: While elegant, this pattern might complicate searching for variable names. For ex, searching for all instances of `foreground.primary`.

```suggestion
    const options = { theme, variable: isDanger ? 'foreground.isDanger' : 'foreground.primary' };
```

packages/buttons/src/styled/StyledButton.ts
geotrev: Agreed - shouldn't be necessary if the base operation in theming is just `+ offset` - the math works itself out 😁

packages/buttons/src/styled/StyledButton.ts
geotrev: I think I agree with this as well... we'll also probably thank ourselves by helping the vars be searchable.

packages/buttons/src/styled/StyledButton.ts
jzempel: I was preferring the unary for the sake of offset clarity and scannability. But if this feels odd to the team, I'll remove.

packages/buttons/src/styled/StyledIconButton.ts
```diff
@@ -69,10 +69,10 @@ const iconStyles = (props: IButtonProps & ThemeProps<DefaultTheme>) => {
   `;
 };
 
-export const StyledIconButton = styled(StyledButton as 'button').attrs({
-  'data-garden-id': COMPONENT_ID,
+export const StyledIconButton = styled(StyledButton as 'button').attrs(props => ({
+  'data-garden-id': (props as any)['data-garden-id'] || COMPONENT_ID,
```
ze-flo, line 73: I dug further into https://github.com/zendeskgarden/react-components/pull/1804 and found some regressions with this method. I'll share my findings and recommendation in that PR. Meanwhile, here's a summary.

Notice how `IconButton` now inherit the `COMPONENT_ID` `buttons.button` from `StyledButton`. 

Screen Shot 2024-05-15 at 7 37 03 AM

When extending a styled-component with `attrs` already set, the props will include such attributes. In this case `props['data-garden-id'] === 'buttons.button'` and takes precedence over the `StyledIconButton`'s `COMPONENT_ID`.

One way around that is to first check that the `data-garden-id` exist and doesn't match the id from the styled-component being extended:

```suggestion
export const StyledIconButton = styled(StyledButton as 'button').attrs(props => ({
  'data-garden-id': (props as any)['data-garden-id'] && (props as any)['data-garden-id'] !== 'buttons.button' ?
  (props as any)['data-garden-id'] : COMPONENT_ID,
```

I can't recommend this pattern. It's brittle and error-prone. My hunch is that we should set `data-garden-id` on the components we expose to the public and only rely on `attrs` for `data-garden-version` or other immutable attributes.

I'm happy to discuss further.

jzempel: > I can't recommend this pattern.

Thanks for the carefule review. I would say that, within the context of a package, this 2nd pattern (although somewhat awkward) is totally acceptable.

ze-flo: The approach I highlighted above is not perfect but it's a good stepping stone.

I [updated all instances](https://github.com/zendeskgarden/react-components/pull/1804/commits/93dfd81e310563b44770202191c1915d70ac2978) in https://github.com/zendeskgarden/react-components/pull/1804. 

How should we move forward with merging these changes into this PR?

lucijanblagonic:
Everything else look good, colors is correctly applied based on all the props tested. 🙌 

Unrelated to the new palette and semantic variables, I see some weird pixel glitching for `SplitButton` (visible in current version as well).
image

lucijanblagonic (approved):

packages/buttons/src/styled/StyledIconButton.ts
jzempel: I've reverted the changes here (and updated PR title, description, and tags) in favor of reviewing #1804

geotrev (approved):

ze-flo (approved):

PR #1808, opened 2024-05-14
feat(Modal): new light/dark mode colors

ze-flo:
## Description
Adds dark / light mode support to `Modal`

## Detail

- Adds dark / light mode support to `Modal`
- Reuse `IconButton` to remove duplicated css for `StyledClose` button

### Light
Screenshot 2024-05-16 at 2 10 54 PM

### Dark
Screenshot 2024-05-16 at 2 10 10 PM

## Checklist


packages/modals/src/styled/StyledClose.spec.tsx
```diff
@@ -8,17 +8,26 @@
...
 import { render, renderRtl } from 'garden-test-utils';
 import { StyledClose } from './StyledClose';
+import XStrokeIcon from '@zendeskgarden/svg-icons/src/16/x-stroke.svg';
 
 describe('StyledClose', () => {
   it('renders default styling', () => {
-    const { container } = render(<StyledClose />);
+    const { container } = render(
+      <StyledClose>
+        <XStrokeIcon />
```
ze-flo, line 17: `StyledClose` extends `IconButton` and now requires a child.

packages/modals/src/styled/StyledClose.ts
```diff
@@ -5,89 +5,26 @@
...
 };
 
-/**
- * 1. Remove dotted outline from Firefox on focus.
- */
-export const StyledClose = styled.button.attrs({
+export const StyledClose = styled(IconButton).attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })`
```
ze-flo, line 23: Keeping as is for now even though attributes set on `StyledIconButton` take precedence.

https://github.com/zendeskgarden/react-components/blob/52822988c09ec0bbcf0a82f63d3d8622b74d0b98/packages/buttons/src/styled/StyledIconButton.ts#L72-L74

packages/modals/src/styled/StyledClose.ts
```diff
@@ -5,89 +5,26 @@
...
-import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
-import {
-  getColorV8,
-  retrieveComponentStyles,
-  DEFAULT_THEME,
-  focusStyles
-} from '@zendeskgarden/react-theming';
+import styled from 'styled-components';
+import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { IconButton } from '@zendeskgarden/react-buttons';
```
ze-flo, line 10: `import { StyledIconButton } from '../../../buttons/src/styled/StyledIconButton';`

Importing `StyledIconButton` using relative paths would allows us to override `data-garden-id` without causing breaking changes.

Is that an acceptable solution?

packages/modals/src/styled/StyledClose.ts
jzempel: Hey @ze-flo, I have a [commit](https://github.com/zendeskgarden/react-components/pull/1807/commits/0171530849f0fe2408491af36caaf8de39667150) on the buttons PR that resolves the issue 🙌

packages/modals/src/styled/StyledModal.ts
```diff
@@ -77,9 +83,11 @@ export const StyledModal = styled.div.attrs<IStyledModalProps>({
   flex-direction: column;
   animation-delay: 0.01s;
   margin: ${props => (props.isCentered ? '0' : `${props.theme.space.base * 12}px`)};
+  border: ${({ theme }) =>
+    `${theme.borders.sm} ${getColor({ theme, variable: 'border.default' })}`};
   border-radius: ${props => props.theme.borderRadii.md};
   box-shadow: ${boxShadow};
-  background-color: ${props => getColorV8('background', 600 /* default shade */, props.theme)};
+  background-color: ${({ theme }) => getColor({ theme, variable: 'background.raised' })};
```
jzempel, line 90: ditto re: extracting `colorStyles`

jzempel (commented):
Let's remove all breaking change (title, description, tags) mention on this PR in favor of having that discussion in #1804 which will resolve the matter. Also, let's get #1807 merged before obtaining designer review here.

geotrev:
Do we want to toggle the system `color-scheme: dark` property so scrollbars look correct? https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does

ze-flo:
> Do we want to toggle the system `color-scheme: dark` property so scrollbars look correct? https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does

That's a great idea! 🔥 

Should individual components (in this case `Modal.Body`) or `ThemeProvider` dynamically set the `color-scheme: dark | light`?

Setting it at the component level is more error-prone since we may forget to add the rule.

Leaving that responsibility to `ThemeProvider`, by creating a global CSS rule via `createGlobalStyle`, offers a great benefit: every component will automatically inherit the `color-scheme` related styles.

jzempel:
> > Do we want to toggle the system `color-scheme: dark` property so scrollbars look correct? https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does
> 
> That's a great idea! 🔥
> 
> Should individual components (in this case `Modal.Body`) or `ThemeProvider` dynamically set the `color-scheme: dark | light`?
> 
> Setting it at the component level is more error-prone since we may forget to add the rule.
> 
> Leaving that responsibility to `ThemeProvider`, by creating a global CSS rule via `createGlobalStyle`, offers a great benefit: every component will automatically inherit the `color-scheme` related styles.

I was going to say no, since consumers can use components without a theming parent. But then they could never toggle dark mode 😅. So I think ThemeProvider is a solid choice that also benefits custom components rendered within the theme.

ze-flo:
> > > Do we want to toggle the system `color-scheme: dark` property so scrollbars look correct? https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does
> > 
> > 
> > That's a great idea! 🔥
> > Should individual components (in this case `Modal.Body`) or `ThemeProvider` dynamically set the `color-scheme: dark | light`?
> > Setting it at the component level is more error-prone since we may forget to add the rule.
> > Leaving that responsibility to `ThemeProvider`, by creating a global CSS rule via `createGlobalStyle`, offers a great benefit: every component will automatically inherit the `color-scheme` related styles.
> 
> I was going to say no, since consumers can use components without a theming parent. But then they could never toggle dark mode 😅. So I think ThemeProvider is a solid choice that also benefits custom components rendered within the theme.

@steue `Modal` is ready for your VQA. Don't worry about the intensely-bright scrollbar in dark mode. This issue isn't solely connected to this component and will be addressed a follow-up PR.

Screenshot 2024-05-16 at 2 36 05 PM

packages/modals/src/styled/StyledClose.ts
```diff
@@ -5,89 +5,26 @@
...
-
-  &::-moz-focus-inner {
-    border: 0; /* [1] */
-  }
-
-  ${props => colorStyles(props)}
-
-  & > svg {
-    vertical-align: middle;
-  }
```
geotrev, line 90: 🧹

geotrev (approved):
Very nice!

jzempel (changes requested):
[nit] consider setting up local variables in your `colorStyles` functions rather than inlining in the generated CSS. It benefits CSS readability and future debug maintenance.

packages/modals/src/styled/StyledModal.ts
```diff
@@ -47,14 +47,26 @@ const animationStyles = (props: IStyledModalProps) => {
...
 
-const boxShadow = (props: ThemeProps<DefaultTheme>) => {
-  const { theme } = props;
-  const { space, shadows } = theme;
-  const offsetY = `${space.base * 5}px`;
-  const blurRadius = `${space.base * 7}px`;
-  const color = getColorV8('neutralHue', 800, theme, 0.35);
+const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
+  const offsetY = `${theme.space.base * 5}px`;
+  const blurRadius = `${theme.space.base * 7}px`;
```
jzempel, line 52: The spread is different for dark mode. Please compare with in-progress branch: https://github.com/zendeskgarden/react-components/blob/9f82a0296642f4f10ab6b5327f904d71f2aa519d/packages/theming/src/utils/menuStyles.ts#L67-L71

jzempel (changes requested):

packages/modals/src/styled/StyledModal.ts
ze-flo: Good catch! 💯 There is difference in Dark mode. I'll update. 

Unrelated... but should we raise these 1px shadow spread differences with design? They seem questionable. 🤔

jzempel: Already did 😉; @lucijanblagonic confirmed the difference is intentional.

steue:
took a peak and it looks gr8!! just to clarify, i am just reviewing `Modal` and not tooltip and drawer? if so, then LGTM! thanks @ze-flo

PR #1811, opened 2024-05-22
feat(tooltips): new light/dark mode colors

ze-flo:
## Description

Adds light / dark mode support for Tooltips

## Detail

Screenshot 2024-05-23 at 2 00 13 PM
Screenshot 2024-05-23 at 2 00 36 PM
Screenshot 2024-05-23 at 2 01 35 PM
Screenshot 2024-05-23 at 2 01 54 PM

## Checklist

### DIFF
--- a/packages/tooltips/src/styled/StyledTooltip.ts
+++ b/packages/tooltips/src/styled/StyledTooltip.ts
@@ -9,10 +9,10 @@ import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
 import {
   arrowStyles,
   retrieveComponentStyles,
-  getColorV8,
   DEFAULT_THEME,
   getLineHeight,
-  getArrowPosition
+  getArrowPosition,
+  getColor
 } from '@zendeskgarden/react-theming';
 import { Placement } from '@floating-ui/react-dom';
 import { ITooltipProps } from '../types';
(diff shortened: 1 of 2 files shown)


ze-flo:
The drop-shadow casts a shadow on the tooltip's arrow tip. This is particularly visible in Dark mode.

Screenshot 2024-05-23 at 2 18 39 PM

The problem should automatically be resolved once either https://github.com/zendeskgarden/react-components/pull/1812 or https://github.com/zendeskgarden/react-components/pull/1812 is merged.

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -121,30 +121,49 @@ const sizeStyles = ({
...
+
+  let backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
+    light: { shade: 900 },
+    dark: { shade: 700 }
+  });
+
+  let borderColor = backgroundColor;
+  let color = theme.palette.white;
```
jzempel, line 144:
```suggestion
  let color = getColor({ hue: 'white' });
```

Ensure a best possible future-proof lookup.

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -121,30 +121,49 @@ const sizeStyles = ({
...
-  let color = getColorV8('background', 600 /* default shade */, theme);
+
+  let backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
+    light: { shade: 900 },
+    dark: { shade: 700 }
+  });
+
+  let borderColor = backgroundColor;
```
jzempel, line 143:
```suggestion
  let borderColor = 'transparent';
```

[nit] this is Garden's usual approach for retaining border size without styling.

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -121,30 +121,49 @@ const sizeStyles = ({
...
   );
-  let backgroundColor = getColorV8('chromeHue', 700, theme);
-  let color = getColorV8('background', 600 /* default shade */, theme);
+
+  let backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
+    light: { shade: 900 },
+    dark: { shade: 700 }
+  });
```
jzempel, line 141: Feels like all this assignment is premature and could be micro-optimized by moving to an `else` block below `if (type === 'light')`. Also ends up being slightly more readable.

jzempel (approved):
Pending design approval

ze-flo:
> The drop-shadow casts a shadow on the tooltip's arrow tip. This is particularly visible in Dark mode.
> 
> Screenshot 2024-05-23 at 2 18 39 PM
> 
> The problem should automatically be resolved once either #1812 or #1814 is merged.
---

> [!NOTE]
> https://github.com/zendeskgarden/react-components/pull/1814 fixed the drop-shadow arrow issue. 🥳 💯 

Screenshot 2024-05-30 at 12 50 35 PM

sieeeeeenna (approved):
Everything looks good to me! The only thing I noticed is the small light tooltip has a larger shadow and the fg/subtle text instead of fg/default (see screenshot for the Figma version). I don't believe the small light tooltip is ever used and Garden doesn't recommend it anyway, so if that it doesn't matter then disregard. Other than that, looks good!!! :)
Screenshot 2024-05-30 at 4 24 10 PM

ze-flo:
> Everything looks good to me! The only thing I noticed is the small light tooltip has a larger shadow and the fg/subtle text instead of fg/default (see screenshot for the Figma version). I don't believe the small light tooltip is ever used and Garden doesn't recommend it anyway, so if that it doesn't matter then disregard. Other than that, looks good!!! :) Screenshot 2024-05-30 at 4 24 10 PM

@sieeeeeenna You are spot-on for calling this out. 💯 The component's API isn't as restrictive and allows for combinations design doesn't support. But I think it's safe to assume the use case you described will never happen IRL.

PR #1812, opened 2024-05-22
fix: refine `arrowStyles` CSS

jzempel:
## Description

Dark mode is highlighting a shadow overcast problem with Garden arrows. This PR update swaps the `::after` border "peek-through" with the following:

- `border` is inherited on the front side `::before` pseudo (`box-shadow` continues to be inherited on the back side `::after`)
- border inheritance is set to `none` for the non-arrow sides of the element
- the border radius hack for IE11 is removed – `clip-path` does the job for all supported browsers

The result is that the arrow must be sized and positioned in order to lay precisely over the base component's border line. All other changes are a result of the need for precision sizing/positioning.

## Checklist

### DIFF
--- a/packages/theming/src/utils/arrowStyles.ts
+++ b/packages/theming/src/utils/arrowStyles.ts
@@ -37,7 +37,7 @@ const animationStyles = (position: ArrowPosition, modifier: string) => {
 const positionStyles = (position: ArrowPosition, size: string, inset: string) => {
   const margin = math(`${size} / -2`);
   const placement = math(`${margin} + ${inset} - 1`);
-  const _position = math(`${size} + 4`);
+  const offset = math(`${size} + 4`);
   let clipPath;
   let borderCss;
   let positionCss;
@@ -50,8 +50,8 @@ const positionStyles = (position: ArrowPosition, size: string, inset: string) =>
     `;
     positionCss = css`
       top: ${placement};
-      right: ${position === 'top-right' && _position};
-      left: ${position === 'top' ? '50%' : position === 'top-left' && _position};
+      right: ${position === 'top-right' && offset};
+      left: ${position === 'top' ? '50%' : position === 'top-left' && offset};
       margin-left: ${position === 'top' && margin};
     `;
   } else if (position.startsWith('right')) {
@@ -61,9 +61,9 @@ const positionStyles = (position: ArrowPosition, size: string, inset: string) =>
       border-left: none;
     `;
     positionCss = css`
-      top: ${position === 'right' ? '50%' : position === 'right-top' && _position};
+      top: ${position === 'right' ? '50%' : position === 'right-top' && offset};
       right: ${placement};
-      bottom: ${position === 'right-bottom' && _position};
+      bottom: ${position === 'right-bottom' && offset};
       margin-top: ${position === 'right' && margin};
     `;
   } else if (position.startsWith('bottom')) {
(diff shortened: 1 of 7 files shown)


packages/theming/src/utils/arrowStyles.ts
```diff
@@ -54,65 +36,77 @@ const animationStyles = (position: ArrowPosition, modifier: string) => {
 
 const positionStyles = (position: ArrowPosition, size: string, inset: string) => {
   const margin = math(`${size} / -2`);
-  const placement = math(`${margin} + ${inset}`);
+  const placement = math(`${margin} + ${inset} - 1`);
+  const _position = math(`${size} + 4`);
```
ze-flo, line 40: Where does the 4 come from?

Perhaps renaming `_position` to `offset` could help better differentiate it from arrow `position`? 🤷‍♂️

packages/theming/src/utils/arrowStyles.ts
```diff
@@ -54,65 +36,77 @@ const animationStyles = (position: ArrowPosition, modifier: string) => {
...
-   *    (improved layout for IE which doesn't support 'clip-path').
-   * 2. Clip portion of the foreground square opposite the arrow tip so that it
+   * 1. Clip portion of the foreground square opposite the arrow tip so that it
    *    doesn't interfere with container content.
+   * 2. Knock out border opposite the arrow tip.
    * 3. Arrow positioning on the base element.
    */
   return css`
     &::before {
-      ${propertyRadius!}: 100%; /* [1] */
```
ze-flo, line 109: ❤️

packages/theming/src/utils/arrowStyles.ts
```diff
@@ -54,65 +36,77 @@ const animationStyles = (position: ArrowPosition, modifier: string) => {
...
-      clip-path: ${clipPath}; /* [2] */
+      clip-path: ${clipPath}; /* [1] */
+      ${borderCss}; /* [2] */
     }
 
     &::before,
     &::after {
```
ze-flo, line 108: We might not even need an `::after` after all! 😉 🙌 

Screenshot 2024-05-23 at 7 01 53 AM

packages/theming/src/utils/arrowStyles.ts
```diff
@@ -150,32 +144,28 @@ const positionStyles = (position: ArrowPosition, size: string, inset: string) =>
...
     &::before {
       /* [2] */
-      border-width: inherit;
-      border-style: inherit;
-      border-color: transparent;
+      border: inherit;
       background-clip: content-box;
     }
 
     &::after {
```
ze-flo, line 166: Same as above. `::after` might be superfluous.

ze-flo:
Bummer. 😢 Probably due to pixel rounding, there's a tiny gap on low-res screen at zoom 100% with `inset === 0`.

Screenshot 2024-05-23 at 7 57 16 AM

jzempel:
> Bummer. 😢 Probably due to pixel rounding, there's a tiny gap on low-res screen at zoom 100% with `inset === 0`.
> 
> Screenshot 2024-05-23 at 7 57 16 AM

Ugh. Nice find.

packages/theming/src/utils/arrowStyles.ts
jzempel: > Where does the 4 come from?

It was the unit needed to keep consistent with previous as it provided the means to bury a bigger arrow into the base element. TBH I'm not super happy about it.

packages/theming/src/utils/arrowStyles.ts
jzempel: We need `::after` to retain the `box-shadow` inheritance. Otherwise, the arrow pokes out un-shadowed. Not terrible, but the original designs wanted the shadow to follow the arrow.

ze-flo: Totally missed that but after some CSS tweaking I see it now. Thanks!

Screenshot 2024-05-23 at 10 27 29 AM

jzempel:
Closing in favor of #1814

PR #1818, opened 2024-05-29
feat(loaders): new light/dark mode colors

ze-flo:
## Description

Adds dark / light mode support for `loaders`.

## Detail
- Adds dark / light mode support for `loaders`
- Removes duplicate specs

> [!IMPORTANT]
> ~~`Spinner`, `Dots`, and `Inline` now default to the agreed mode-specific color rather than inheriting it from the parent container.~~

## Checklist


packages/loaders/src/elements/Progress.spec.tsx
```diff
@@ -6,9 +6,14 @@
...
-import { render } from 'garden-test-utils';
+import { render, renderDark } from 'garden-test-utils';
 import { Progress } from './Progress';
-import { PALETTE_V8 } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, PALETTE } from '@zendeskgarden/react-theming';
+import { rgba } from 'polished';
+
+function getRenderFn(mode: 'dark' | 'light') {
+  return mode === 'dark' ? renderDark : render;
+}
```
ze-flo, line 16: Would it be useful to move this into `garden-test-utils`? I reduces verbosity when using `it.each`:

```js
it.each<['light' | 'dark', string]>([
      ['light', PALETTE.grey[700]],
      ['dark', PALETTE.grey[500]]
    ])('applies the default colors in "%s mode', (mode, color) => {
      const { container } = getRenderFn(mode)(<Component ... />);
      ...
})
```

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -48,14 +47,16 @@ const retrieveSkeletonBackgroundColor = ({
...
-  if (isLight) {
-    return css`
-      background-color: ${rgba(getColorV8('background', 600 /* default shade */, theme)!, 0.2)};
-    `;
-  }
+  const backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
+    transparency: theme.opacity[200],
+    light: { shade: isLight ? 500 : 700 },
```
ze-flo, line 54: Should we deprecate `props.isLight` is favor of a more relevant `props.isInversed`?

From my research, it's only used once internally. The plan would be to support both until until v10.

packages/loaders/src/types/index.ts
```diff
@@ -14,7 +14,7 @@ export type Size = (typeof SIZE)[number];
 export interface IDotsProps extends SVGAttributes<SVGSVGElement> {
   /** Sets the height and width in pixels. Inherits the parent's font size by default. */
   size?: string | number;
-  /** Sets the fill color. Inherits the parent's `color` by default. */
+  /** Overrides the default fill color. */
```
ze-flo, line 17: As mentioned ☝️ , breaking change?

packages/loaders/demo/stories/SkeletonStory.tsx
```diff
@@ -50,7 +50,7 @@ export const SkeletonStory: Story<IArgs> = ({ backgroundColor, count = 1, typesc
   return (
     <div
       style={{
-        backgroundColor: backgroundColor || (args.isLight ? PALETTE.kale[600] : undefined),
+        backgroundColor: backgroundColor || (args.isLight ? PALETTE.kale[800] : undefined),
```
ze-flo, line 53: Bumping the shade in accordance to the handy-dandy v8-to-v9 color mapping tool. 👍

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -89,9 +90,13 @@ const retrieveSkeletonGradient = ({
...
-      ${isLight
-        ? getColorV8('chromeHue', 700, theme, 0.4)
-        : rgba(getColorV8('background', 600 /* default shade */, theme)!, 0.6)},
+      ${getColor({
+        theme,
+        hue: 'neutralHue',
+        transparency: theme.opacity[200],
+        light: { shade: isLight ? 500 : 700 },
+        dark: { shade: isLight ? 700 : 500 }
+      })},
```
ze-flo, line 99: @steue @lucijanblagonic We used to support a [Light Skeleton](https://garden.zendesk.com/components/skeleton#light) for dark surfaces. Now with Dark mode, it will be the inverse shade. 

See in action:
https://6657851b9c7aef2ddba8ea98--zendeskgarden-react-components.netlify.app/?path=/story/packages-loaders-skeleton--skeleton&args=isLight:!true

Hope that works with you.

lucijanblagonic:
> Spinner, Dots, and Inline now default to the agreed mode-specific color rather than inheriting it from the parent container.

@ze-flo Could you expand a bit on why that is the case?

> Skeleton `isLight` [loader](https://6657851b9c7aef2ddba8ea98--zendeskgarden-react-components.netlify.app/?path=/story/packages-loaders-skeleton--skeleton&args=isLight:!true;typescale:2x-large&globals=backgrounds.value:!hex(151a1e))
image

The color on this one seems a bit off, it should be `white, .16%`. I see we didn't have that specced up in our inventory.

image
_(default switched to dark on the top, `isLight` in dark mode on the bottom)_

Overall, the `isLight` in dark mode is a bit ligther in dark mode when it's using white. Is that doable?

And another question, is there an easy way to see which variable is being used and where in Storybook? Inspector is showing the color value only, and Variables section shows main variables only.

packages/loaders/src/elements/Progress.spec.tsx
jzempel: I'm ok with adding to the test utils.

packages/loaders/src/elements/Dots.tsx
```diff
@@ -90,7 +90,6 @@ Dots.propTypes = {
 
 Dots.defaultProps = {
   size: 'inherit',
-  color: 'inherit',
```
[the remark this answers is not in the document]
jzempel: We need to revisit this removal together with design. I'm not convinced this is the right direction. The updated designs should've been informed about the foreground color inheritance nature of loaders.

packages/loaders/src/styled/StyledSkeleton.ts
jzempel: I think that could make sense. Let's see how the PR plays out with design feedback. If we do the rename, let's just rip the bandage and make it happen in v9.

packages/loaders/src/types/index.ts
jzempel: Again, I don't think this is something we should do.

jzempel (changes requested):

packages/loaders/src/elements/Progress.spec.tsx
ze-flo: To avoid blowing scope, I'll do that refactor in a small separate PR since I'm also using this pattern in [Tooltip](https://github.com/zendeskgarden/react-components/pull/1811/files#diff-9ca09c7b035971f09ea9d6cf8f106c45fdfa61041e2986f36ae778e3dc78731aR15-R18).

ze-flo:
> > Spinner, Dots, and Inline now default to the agreed mode-specific color rather than inheriting it from the parent container.
> 
> @ze-flo Could you expand a bit on why that is the case?
> 
> > Skeleton `isLight` [loader](https://6657851b9c7aef2ddba8ea98--zendeskgarden-react-components.netlify.app/?path=/story/packages-loaders-skeleton--skeleton&args=isLight:!true;typescale:2x-large&globals=backgrounds.value:!hex(151a1e))
> > image
> 
> The color on this one seems a bit off, it should be `white, .16%`. I see we didn't have that specced up in our inventory.
> 
> image _(default switched to dark on the top, `isLight` in dark mode on the bottom)_
> 
> Overall, the `isLight` in dark mode is a bit ligther in dark mode when it's using white. Is that doable?
> 
> And another question, is there an easy way to see which variable is being used and where in Storybook? Inspector is showing the color value only, and Variables section shows main variables only.

---

> @ze-flo Could you expand a bit on why that is the case?

Based on @jzempel's comment, we need to align on whether we should keep the color inheritance or set the default color ourselves.

> Overall, the `isLight` in dark mode is a bit ligther in dark mode when it's using white. Is that doable?

Absolutely doable! 💯 

Let me recap for clarity:

### Regular Skeleton
#### Light mode
`neutral[700] / 16%`

#### Dark mode 
`neutral[500] / 16%`


### Light Skeleton
#### Light mode
`neutral[500] / 16%` // the regular Skeleton's color in Dark mode

#### Dark mode

`white / 16%`

@lucijanblagonic Is this correct?


> And another question, is there an easy way to see which variable is being used and where in Storybook? Inspector is showing the color value only, and Variables section shows main variables only.

Sadly, it's impossible at this time. 😭 Garden components do not reference CSS variables. Instead they get their calculated color values at runtime from our `getColor` function.

lucijanblagonic:
> ### Light Skeleton
> #### Light mode
> `neutral[500] / 16%` // the regular Skeleton's color in Dark mode
> 
> #### Dark mode
> `white / 16%`

Light mode should also be `white / 16%`. The reason is that `isLight` might be used to place a Skeleton on a filled (color) surface and we don't want to overlay grey loader on top of it. Light skeleton is a deliberate styling decision so it should be white at all times.

> Sadly, it's impossible at this time. 😞 Garden components do not reference CSS variables. Instead they get their calculated color values at runtime from our getColor function.

OK, so can only review colors, and we are counting on eng to implement correct variables scope (e.g. progress loaders use `bg` group, not `border`)?

packages/loaders/src/styled/StyledProgress.ts
```diff
@@ -30,15 +30,34 @@ interface IStyledProgressBackgroundProps {
...
+  color
+}: IStyledProgressBackgroundProps & ThemeProps<DefaultTheme>) => {
+  const backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
+    transparency: theme.opacity[200],
+    light: { shade: 700 },
+    dark: { shade: 500 }
+  });
+  const foregroundColor = color || getColor({ theme, variable: 'border.successEmphasis' });
```
ze-flo, line 44: > OK, so we can only review color values, and we are counting on eng to implement correct variables (e.g. progress loaders might use bg group, not border and it should be 100% clear from the inventory)?

@lucijanblagonic I always define colors using the variables found in the inventory first.

For `Progress bar`, the `foregroundColor` uses `border.successEmphasis`. It exists in our collection of border variables so I used it.

The `backgroundColor` uses `WIP_opacity/neutral (700 : 500)/200`, which is the same as `foreground.subtle`. However, because we don't have a background variable that matches the hue / shade combo per mode, I build it using advanced `getColor` options.

I hope this answers your question. If this isn't the right approach, let me know. 

Thanks!

packages/loaders/src/styled/StyledInline.ts
```diff
@@ -5,19 +5,23 @@
...
+import { retrieveComponentStyles, DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'loaders.inline';
 
-const PULSE_ANIMATION = keyframes`
-  0%, 100% {
-    opacity: .2;
-  }
+const retrieveAnimation = ({ theme }: ThemeProps<DefaultTheme>) => keyframes`
+0% {
```
geotrev, line 14: I think we may have lost a tab for this style block. :D

packages/loaders/src/styled/StyledProgress.ts
```diff
@@ -30,15 +30,34 @@ interface IStyledProgressBackgroundProps {
 
 const PROGRESS_BACKGROUND_COMPONENT_ID = 'loaders.progress_background';
 
+const colorStyles = ({
+  theme,
+  color
+}: IStyledProgressBackgroundProps & ThemeProps<DefaultTheme>) => {
+  const backgroundColor = getColor({
+    theme,
+    hue: 'neutralHue',
```
geotrev, line 39: Should this inherit from a background variable? 🤔

ze-flo: As discussed with Design, we're not tying this one to a variable, but rather build it from scratch.

packages/loaders/src/styled/StyledInline.ts
```diff
@@ -5,18 +5,22 @@
...
   }
 
   50% {
-    opacity: .8;
+    opacity: ${theme.opacity[600]};
+  }
+
+  100% {
+    opacity: ${theme.opacity[200]};
   }
```
jzempel, line 24: Why did the 0% portion of the animation change?

packages/loaders/src/styled/StyledSkeleton.ts
```diff
@@ -79,22 +85,14 @@ const retrieveSkeletonAnimation = ({ theme }: ThemeProps<DefaultTheme>) => {
   `;
 };
 
-const retrieveSkeletonGradient = ({
-  theme,
-  isLight
-}: IStyledSkeletonProps & ThemeProps<DefaultTheme>) => {
-  // Disabling stylelint due to conflicts with prettier and linear-gradient formatting
+const gradienStyles = (props: IStyledSkeletonProps & ThemeProps<DefaultTheme>) => {
```
jzempel, line 88:
```suggestion
const gradientStyles = (props: IStyledSkeletonProps & ThemeProps<DefaultTheme>) => {
```

[nit]

packages/loaders/src/styled/StyledProgress.ts
jzempel: fwiw, I think @geotrev's point of tying this to `foreground.subtle` actually does make sense as the background of this loader actually functions as a subtle foreground color when `Progress` is applied to a page. Let's not make any adjustments now, but I want to reserve space to triple-check semantics with design before v9 wraps up.

jzempel (changes requested):

packages/loaders/src/styled/StyledInline.ts
ze-flo: Aligning with Figma, where the first dot has 100% opacity.

Screenshot 2024-06-06 at 7 26 45 AM

jzempel: Ok, I checked with @lucijanblagonic and he wasn't expecting the animation to materially change based on Figma. And now we're seeing some odd animation with the `Inline` loader. I think we should revert here (while retaining the `theme.opacity` usage).

jzempel (approved):

lucijanblagonic (approved):

PR #1819, opened 2024-05-31
test: add `getRenderFn` utility

ze-flo:
## Description

Adds a small utility fn to `garden-test-utils` to simplify testing across color modes. `getRenderFn` returns the correct `render` fn for the `mode`.

## Detail
`getRenderFn` aims to reduce boilerplate code when testing colors in multiple modes using `it.each()`. 

See this [comment](https://github.com/zendeskgarden/react-components/pull/1818#discussion_r1619349923).

PR #1820, opened 2024-06-01
feat(typography): new light/dark mode colors (excluding CodeBlock)

ze-flo:
## Description

Adds `light` / `dark` mode support for our Typography components, excluding `CodeBlock`

## Detail

> [!NOTE]
> ~~`StyledFont` defaults to `foreground.default` color if no `hue` is provided. The following components extend it and now inherit the default color:~~
> ~~- `SM`~~
> ~~- `MD`~~
> ~~- `LG`~~
> ~~- `XL`~~
> ~~- `XXL`~~
> ~~- `XXXL`~~
> ~~- `Blockquote`~~
> ~~- `OrderedList`~~
> ~~- `UnorderedList`~~
> ~~- `Span`~~
>
> ~~`Ellipsis` and `Paragraph`'s color also defaults to `foreground.default`.~~

## Checklist


packages/tooltips/src/elements/Tooltip.spec.tsx
```diff
@@ -12,6 +12,7 @@ import { PALETTE } from '@zendeskgarden/react-theming';
 import { Tooltip } from './Tooltip';
 import { ITooltipProps } from '../types';
 
+// TODO: remove once `getRenderFn` is in `garden-test-utils`
```
ze-flo, line 15: To do once https://github.com/zendeskgarden/react-components/pull/1819 has merged.

packages/typography/src/styled/StyledCode.spec.tsx
```diff
@@ -60,18 +65,46 @@ describe('StyledCode', () => {
...
+      // light mode
+      ['grey', 'light', { color: PALETTE.grey[900], bgColor: PALETTE.grey[200] }],
+      ['green', 'light', { color: PALETTE.green[900], bgColor: PALETTE.green[100] }],
+      ['red', 'light', { color: PALETTE.red[900], bgColor: PALETTE.red[100] }],
+      ['yellow', 'light', { color: PALETTE.yellow[900], bgColor: PALETTE.yellow[100] }],
+      // dark mode
+      ['grey', 'dark', { color: PALETTE.grey[300], bgColor: PALETTE.grey[900] }],
+      ['green', 'dark', { color: PALETTE.green[300], bgColor: PALETTE.green[1000] }],
+      ['red', 'dark', { color: PALETTE.red[300], bgColor: PALETTE.red[1000] }],
+      ['yellow', 'dark', { color: PALETTE.yellow[300], bgColor: PALETTE.yellow[1000] }]
```
ze-flo, line 87: If this is too dense, I can expand it.

packages/typography/src/styled/StyledCode.ts
```diff
@@ -6,17 +6,40 @@
...
+      bgColorArgs.variable = 'background.warning';
+      fgColorArgs.variable = 'foreground.warningEmphasis';
+      break;
+    // includes grey
+    default:
+      fgColorArgs.variable = 'foreground.default';
+      bgColorArgs.hue = 'neutralHue';
+      bgColorArgs.dark = { shade: 900 };
+      bgColorArgs.light = { shade: 200 };
+      break;
```
ze-flo, line 38: Unsupported hues default to grey. If a hue isn't recognized, `getColor` throws.

packages/typography/src/styled/StyledFont.tsx
```diff
@@ -29,43 +29,46 @@ export const THEME_SIZES: Record<TypographySize, ThemeSize> = {
...
-  } else if (props.isBold === false || props.size !== 'inherit') {
-    fontWeight = props.theme.fontWeights.regular;
+  if (isBold === true) {
+    fontWeight = theme.fontWeights.semibold;
+  } else if (isBold === false || size !== 'inherit') {
+    fontWeight = theme.fontWeights.regular;
   }
 
-  if (props.hue) {
-    const shade = props.hue === 'yellow' ? 700 : 600;
```
ze-flo, line 66: No longer needed since the default shade for all hues is now `700` in `light` mode.

packages/typography/src/styled/StyledFont.tsx
```diff
@@ -29,43 +29,46 @@ export const THEME_SIZES: Record<TypographySize, ThemeSize> = {
...
+  theme
+}: IStyledFontProps & ThemeProps<DefaultTheme>) => {
+  const monospace = isMonospace && ['inherit', 'small', 'medium', 'large'].indexOf(size!) !== -1;
+  const fontFamily = monospace && theme.fonts.mono;
+  const direction = theme.rtl ? 'rtl' : 'ltr';
   let fontSize;
   let fontWeight;
   let lineHeight;
-  let color;
+  let color = getColor({ theme, variable: 'foreground.default' });
```
ze-flo, line 45: Sets default color correctly based on mode.

packages/typography/src/styled/StyledEllipsis.ts
```diff
@@ -18,6 +18,7 @@ export const StyledEllipsis = styled.div.attrs({
   text-overflow: ellipsis;
   white-space: nowrap;
   direction: ${props => (props.theme.rtl ? 'rtl' : 'ltr')};
+  color: ${props => getColor({ theme: props.theme, variable: 'foreground.default' })};
```
ze-flo, line 21: Sets default color correctly based on mode.

packages/typography/src/styled/StyledParagraph.ts
```diff
@@ -19,6 +19,7 @@ export const StyledParagraph = styled.p.attrs({
   margin: 0;
   padding: 0;
   direction: ${props => (props.theme.rtl ? 'rtl' : 'ltr')};
+  color: ${props => getColor({ theme: props.theme, variable: 'foreground.default' })};
```
ze-flo, line 22: Sets default color correctly based on mode.

packages/typography/src/elements/span/Span.spec.tsx
```diff
@@ -47,35 +52,47 @@ describe('Span', () => {
...
+
+      return arr;
+    }, []);
+
+    it.each(cases)('renders with a "%s" hue in "%s" mode', (hue, mode) => {
+      const { container } = getRenderFn(mode)(<Span hue={hue} />);
+
+      expect(container.firstChild).toHaveStyleRule(
+        'color',
+        (PALETTE as any)[hue][mode === 'light' ? 700 : 500]
```
ze-flo, line 85: Follows `getColor`'s logic:

https://github.com/zendeskgarden/react-components/blob/fcf78927e812df6a5dd7c07505cb2a13144c77e3/packages/theming/src/utils/getColor.ts#L34

packages/typography/src/elements/span/Span.spec.tsx
```diff
@@ -47,35 +52,47 @@ describe('Span', () => {
...
+      );
     });
 
-    it('handles yellow hue with specialized shading', () => {
```
ze-flo, line 75: Not needed since every hue's shade is `700` in light mode.

packages/typography/src/styled/StyledFont.tsx
```diff
@@ -29,43 +29,46 @@ export const THEME_SIZES: Record<TypographySize, ThemeSize> = {
...
+  } else if (isBold === false || size !== 'inherit') {
+    fontWeight = theme.fontWeights.regular;
   }
 
-  if (props.hue) {
-    const shade = props.hue === 'yellow' ? 700 : 600;
-
-    color = getColorV8(props.hue, shade, props.theme);
+  if (hue) {
+    color = getColor({ hue, theme });
```
ze-flo, line 71: Using default shades:

https://github.com/zendeskgarden/react-components/blob/fcf78927e812df6a5dd7c07505cb2a13144c77e3/packages/theming/src/utils/getColor.ts#L34

packages/typography/src/elements/Code.spec.tsx
```diff
@@ -45,30 +44,4 @@ describe('Code', () => {
...
-      expect(container.firstChild).toHaveStyleRule('background-color', PALETTE_V8.red[200]);
-    });
-
-    it('renders yellow hue if provided', () => {
-      const { container } = render(<Code hue="yellow" />);
-
-      expect(container.firstChild).toHaveStyleRule('background-color', PALETTE_V8.yellow[200]);
-    });
-  });
 });
```
ze-flo, line 47: Moved specs to `StyledCode.spec.tsx`.

lucijanblagonic:
Everything looks good. Why was the `Code` and `CodeBlock` excluded?

lucijanblagonic (approved):

packages/typography/src/styled/StyledCode.ts
jzempel: This move unintentionally breaks the `Code` hue API. The expectation is that `hue` can receive any value that `getColor`'s hue argument can receive. So that includes all Garden hue values (lime, fuschia, etc), hex, etc. So a few things need to happen here:
- continue to give special treatment to Garden's primary colors
- get feedback from design on what shades to fallback to (perhaps dark:900 & light: 200)
- allow `getColor` to do it's thing – including throwing an error if someone provides a bogus value

Please see [StyledTag](https://github.com/zendeskgarden/react-components/blob/next/packages/tags/src/styled/StyledTag.ts) for comparison.

packages/typography/src/styled/StyledEllipsis.ts
jzempel: I don't think we want this; the `color` should be inherited.

packages/typography/src/styled/StyledParagraph.ts
jzempel: ditto ..  paragraph `color` should be inherited

jzempel (changes requested):

packages/typography/src/styled/StyledCode.ts
ze-flo: `StyledCode` is "loosely" typed with `hue?: string`

https://github.com/zendeskgarden/react-components/blob/cffdb06680e0a63bd8faad7cc62085ec83b2f835/packages/typography/src/styled/StyledCode.ts#L31-L33

In our case, only `Code` uses this component and restricts possible `hue` values to only `grey`, `red`, `green`, and `yellow`.

https://github.com/zendeskgarden/react-components/blob/327ea52612f08bf0861d23853827cc2d8a590a7d/packages/typography/src/types/index.ts#L10

https://github.com/zendeskgarden/react-components/blob/327ea52612f08bf0861d23853827cc2d8a590a7d/packages/typography/src/types/index.ts#L84-L89

Therefore, I don't think it's a breaking change, but rather a call to improve clarity by redefining the `hue` type in `StyledCode` to match its real-life usage. I must emphasize this suggestion only make sense because `StyledCode` is exclusively used by `Code`.

```
interface IStyledCodeProps extends Omit<IStyledFontProps, 'size'> {
  hue?: (typeof HUE)[number];
  size?: ICodeProps['size'];
}
```
 
Toughts?

packages/typography/src/elements/Code.spec.tsx
ze-flo: Thank you for clarifying. I wasn't sure about that. I'll move the specs back to this file instead.

packages/typography/src/styled/StyledCode.ts
```diff
@@ -6,17 +6,40 @@
...
+    case 'yellow':
+      bgColorArgs.variable = 'background.warning';
+      fgColorArgs.variable = 'foreground.warningEmphasis';
+      break;
+    // includes grey
+    default:
+      fgColorArgs.variable = 'foreground.default';
+      bgColorArgs.hue = 'neutralHue';
+      bgColorArgs.dark = { shade: 900 };
+      bgColorArgs.light = { shade: 200 };
```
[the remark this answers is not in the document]
ze-flo: Design opted for a custom component variable for this one:

```
WIP_code-bg-default

#E8EAEC
```

I chose to explicitly build the color to reflect their choice and highlight the lack for variable for this hue / shade combo.

I think modifying an existing variable can lead to side-effects and increased maintenance overhead. 

For ex, if Design were to change the shade of `background.subtle`, ENG would have to review all uses of this variable to ensure the resulting shade from offsets is still the desired one.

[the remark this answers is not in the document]
jzempel: We already have guidance from design to use `getColor`'s offset feature when possible. They wouldn't be using `WIP` variables if Figma provided similar functionality. Let's double check the semantics with them. As far as any `variables` change goes, we would have to review all uses – so I don't think that's a major issue here.

packages/typography/src/styled/StyledCode.ts
jzempel: Ah, interesting – I was looking at the wrong (`Styled`) API and testing on v8 with usage like `<Code hue="fuschia">`. In this case, restricting the `hue` color values makes sense. I still think allowing `getColor` to throw on invalid usage – as it now does in other v9 components – is appropriate and better than protecting the consumer from themselves.

packages/typography/src/styled/StyledCode.ts
```diff
@@ -29,14 +52,15 @@ const colorStyles = (props: IStyledCodeProps & ThemeProps<DefaultTheme>) => {
 };
 
 interface IStyledCodeProps extends Omit<IStyledFontProps, 'size'> {
-  hue?: string;
+  hue?: ICodeProps['hue'];
```
ze-flo, line 55: `hue?: string;` was causing confusion. So I refined the type to emphasize that `Code`, `StyledCode`'s only consumer, strictly supports `grey`, `red`, `green`, `yellow`, and `undefined`. 

https://github.com/zendeskgarden/react-components/blob/327ea52612f08bf0861d23853827cc2d8a590a7d/packages/typography/src/types/index.ts#L10

Passing an unsupported hue would trigger a TS error or an invalid `PropType` error at runtime.

packages/typography/src/styled/StyledCode.ts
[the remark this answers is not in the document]
ze-flo: Following our convo, @jzempel and I decided it was best to leave as-is to better handle future themes.

packages/typography/src/styled/StyledCode.spec.tsx
```diff
@@ -1,77 +0,0 @@
-/**
```
ze-flo, line 1: Consolidated into `Code.spec.tsx`.

packages/typography/src/styled/StyledCode.ts
```diff
@@ -29,14 +52,15 @@ const colorStyles = (props: IStyledCodeProps & ThemeProps<DefaultTheme>) => {
...
   size?: ICodeProps['size'];
 }
 
 export const StyledCode = styled(StyledFont as 'code').attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  as: 'code'
+  as: 'code',
+  isMonospace: true
```
ze-flo, line 63: `isMonospace: true` is required by `StyledFont` to apply the correct styles. 

Defining it as a `defaultProp` didn't make that obvious.

This pattern aligns with other components. For ex:

https://github.com/zendeskgarden/react-components/blob/cffdb06680e0a63bd8faad7cc62085ec83b2f835/packages/buttons/src/styled/StyledAnchor.ts#L18-L25

steue (changes requested):
this is looking good! have some clarifications on bg treatment for `code` components. design did some optical adjustments to the BG and shifted the treatments up 1 step. this would only affect `success`, `danger`, and `warning`. `default` looks good.

image

packages/typography/src/styled/StyledCode.ts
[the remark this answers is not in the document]
jzempel: Let's be sure to double-check intent with design.

packages/typography/src/styled/StyledFont.tsx
ze-flo: @jzempel To maintain the current behavior, do we need to keep color as `undefined`? If so, the color would be inherited from the parent container. It will also be the user's responsibility to set the correct color based on the mode they're in.

jzempel: Yes, the current API structure needs to be kept (aka `undefined`)

jzempel (approved):
nice – hits API expectations

sieeeeeenna:
Looks good to me! These colors are a nice improvement!

steue (approved):
looks awesome thank you!

PR #1830, opened 2024-06-03
fix(tabs): prevent `isVertical` styling cascade

### LINKED ISSUE #1817: Nested Tabs isVertical inherit
## Expectations
I nest a horizontal tab inside a vertical tab and the layout remains as shown on the demo page.

## Reality
If a parent Tabs has isVertical, all other nested instances of Tabs will inherit it even if it is not there or is false.

## Steps to Reproduce
Here's the [sandbox](https://codesandbox.io/p/devbox/zendesk-tabs-8dcvcq?file=%2Fapp%2Fpage.tsx%3A4%2C60) to recreate it.

1. click on first vertical tab

### Fine Print

- Component: Tabs
- Version: react-tabs 8.76.2
- Browsers: all

jzempel:
## Description

The previous `verticalStyling` catch-all function in `StyledTabs` was not well-structured, allowing component CSS to leak into child `Tabs` so that if any parent sets the `isVertical` prop, all children would be vertically styled regardless of their `isVertical` prop setting.

This PR eradicates the `verticalStyling` function by portioning all of its properties out to relevant components, conditionally and precisely applying CSS depending on the context. The updated styles follow Garden's [API rules](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules) by utilizing `sizeStyles` and `colorStyles` accordingly.

## Detail

closes #1817 

## Checklist

- [ ] :guardsman: ~includes new unit tests. Maintain existing coverage (always >= 96%)~

### DIFF
--- a/packages/tabs/src/styled/StyledTabs.ts
+++ b/packages/tabs/src/styled/StyledTabs.ts
@@ -5,64 +5,14 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled, { css, ThemeProps, DefaultTheme } from 'styled-components';
+import styled from 'styled-components';
 import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
-import { StyledTab } from './StyledTab';
-import { StyledTabPanel } from './StyledTabPanel';
-import { StyledTabList } from './StyledTabList';
 
 const COMPONENT_ID = 'tabs.tabs';
 
 interface IStyledTabsProps {
-  /**
-   * Displays vertical TabList styling
-   */
   isVertical?: boolean;
 }
-
-const verticalStyling = ({ theme }: ThemeProps<DefaultTheme>) => {
-  return css`
-    display: table;
-
-    ${StyledTabList} {
-      display: table-cell;
-      margin-bottom: 0;
-      border-bottom: none;
-      vertical-align: top;
-    }
-
-    ${StyledTab} {
-      display: block;
-      margin-bottom: ${theme.space.base * 5}px;
-      margin-left: ${theme.rtl && '0'};
-      border-left: ${theme.rtl && '0'};
-      border-bottom-style: none;
-      /* stylelint-disable property-case, property-no-unknown */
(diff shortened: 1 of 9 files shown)


packages/tabs/src/styled/StyledTabs.ts
```diff
@@ -5,78 +5,26 @@
...
-
 /**
  * Accepts all `<div>` props
  */
 export const StyledTabs = styled.div.attrs<IStyledTabsProps>({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledTabsProps>`
-  display: block;
+  display: ${props => (props.isVertical ? 'table' : 'block')};
```
geotrev, line 24: I'm wondeirng if using flex might be preferable for handling flow. 🤔 `display: table` can impact the element's role for some screen readers ([source](https://www.456bereastreet.com/archive/201110/using_displaytable_has_semantic_effects_in_some_screen_readers/)). Granted, it's a small subset of SRs, so maybe not a huge deal?

geotrev (commented):
Looks good. Just the one comment.

One unrelated thing I noticed (present on `main`): trying to use arrow keys backward through tabs requires one extra arrow key press to loop from the first to last tab. Guessing that's in the tabs container logic though. We can track that separately but not sure if anyone else noticed.

packages/tabs/src/styled/StyledTabs.ts
geotrev: I also confirmed using a fake table demo in the link above that Chrome + JAWS treats CSS tables as actual tables (so long as they have more than one cell). 😬

jzempel: @geotrev all valid concerns for sure. For this PR I want to retain exact parity with the previous styling, since that's already been through a11y audit.

> trying to use arrow keys backward through tabs requires one extra arrow key press to loop from the first to last tab

Looks like that might be related to the inclusion of a `disabled` tab in the Storybook example. Again, let's plan to address under a separate effort.

geotrev (approved):

PR #1833, opened 2024-06-05
feat(tables): new light/dark mode colors

geotrev:
## Description

Adds `light` / `dark` mode support for Table components.

## Detail

In addition to updating to light/dark compatibility, this PR also:

- Uses base styling from `IconButton` for `OverflowButton` (with height/width overrides)
- Removes `isHovered`, `isFocused`, and `isActive` props from `OverflowButton`

## Checklist

### DIFF
--- a/packages/tables/src/elements/OverflowButton.tsx
+++ b/packages/tables/src/elements/OverflowButton.tsx
@@ -22,7 +22,7 @@ export const OverflowButton = forwardRef<
   const { size } = useTableContext();
 
   return (
-    <StyledOverflowButton type="button" size={size} ref={ref} {...props}>
+    <StyledOverflowButton type="button" $size={size} ref={ref} {...props} focusInset>
       <OverflowStrokeIcon />
     </StyledOverflowButton>
   );
--- a/packages/tables/src/elements/OverflowButton.tsx
+++ b/packages/tables/src/elements/OverflowButton.tsx
@@ -5,57 +5,27 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { ButtonHTMLAttributes, useState, forwardRef } from 'react';
-import PropTypes from 'prop-types';
-import { composeEventHandlers } from '@zendeskgarden/container-utilities';
-import OverflowStrokeIcon from '@zendeskgarden/svg-icons/src/16/overflow-stroke.svg';
-import { StyledOverflowButton, StyledOverflowButtonIconWrapper } from '../styled';
+import React, { ButtonHTMLAttributes, forwardRef } from 'react';
+import OverflowStrokeIcon from '@zendeskgarden/svg-icons/src/16/overflow-vertical-stroke.svg';
+import { StyledOverflowButton } from '../styled';
 import { useTableContext } from '../utils/useTableContext';
-
-export interface IOverflowButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
-  /** @ignore Applies hover styling */
-  isHovered?: boolean;
-  /** @ignore Applies active styling */
-  isActive?: boolean;
-  /** @ignore Applies focus styling */
-  isFocused?: boolean;
-}
 
 /**
  * @deprecated use `Table.OverflowButton` instead
  *
  * @extends ButtonHTMLAttributes<HTMLButtonElement>
  */
-export const OverflowButton = forwardRef<HTMLButtonElement, IOverflowButtonProps>(
-  ({ onFocus, onBlur, isFocused: focused, ...other }, ref) => {
-    const [isFocused, setIsFocused] = useState(false);
-    const { size } = useTableContext();
+export const OverflowButton = forwardRef<
+  HTMLButtonElement,
+  ButtonHTMLAttributes<HTMLButtonElement>
+>((props, ref) => {
+  const { size } = useTableContext();
 
-    return (
-      <StyledOverflowButton
-        onFocus={composeEventHandlers(onFocus, () => {
-          setIsFocused(true);
-        })}
-        onBlur={composeEventHandlers(onBlur, () => {
-          setIsFocused(false);
-        })}
-        size={size}
-        isFocused={typeof focused === 'undefined' ? isFocused : focused}
-        ref={ref}
-        {...other}
-      >
-        <StyledOverflowButtonIconWrapper>
-          <OverflowStrokeIcon />
-        </StyledOverflowButtonIconWrapper>
-      </StyledOverflowButton>
-    );
-  }
-);
+  return (
+    <StyledOverflowButton type="button" size={size} ref={ref} {...props}>
+      <OverflowStrokeIcon />
+    </StyledOverflowButton>
+  );
+});
 
 OverflowButton.displayName = 'OverflowButton';
-
-OverflowButton.propTypes = {
-  isHovered: PropTypes.bool,
-  isActive: PropTypes.bool,
-  isFocused: PropTypes.bool
-};
--- a/packages/tables/src/elements/Row.spec.tsx
+++ b/packages/tables/src/elements/Row.spec.tsx
@@ -9,7 +9,7 @@ import React from 'react';
 import { css } from 'styled-components';
 import userEvent from '@testing-library/user-event';
 import { render } from 'garden-test-utils';
-import { getColorV8, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, PALETTE, getColor } from '@zendeskgarden/react-theming';
 
 import { Table } from './Table';
 import { Body } from './Body';
@@ -56,16 +56,12 @@ describe('Row', () => {
 
     await user.click(row);
 
-    expect(row).toHaveStyleRule(
-      'box-shadow',
-      `inset 3px 0 0 0 ${getColorV8('primaryHue', 600, DEFAULT_THEME)}`,
-      {
-        /* prettier-ignore */
-        /* stylelint-disable */
-        modifier: css`${StyledCell}:first-of-type` as any
-        /* stylelint-enable */
-      }
-    );
+    expect(row).toHaveStyleRule('box-shadow', `inset 3px 0 0 0 ${PALETTE.blue[700]}`, {
+      /* prettier-ignore */
+      /* stylelint-disable */
+      modifier: css`${StyledCell}:first-of-type` as any
+      /* stylelint-enable */
+    });
   });
 
   it('does not apply focus styling when table is readonly', async () => {
@@ -80,16 +76,12 @@ describe('Row', () => {
 
     await user.click(row);
 
-    expect(row).not.toHaveStyleRule(
-      'box-shadow',
-      `inset 3px 0 0 0 ${getColorV8('primaryHue', 600, DEFAULT_THEME)}`,
-      {
-        /* prettier-ignore */
-        /* stylelint-disable */
-        modifier: css`${StyledCell}:first-of-type` as any
-        /* stylelint-enable */
-      }
-    );
+    expect(row).not.toHaveStyleRule('box-shadow', `inset 3px 0 0 0 ${PALETTE.blue[700]}`, {
+      /* prettier-ignore */
+      /* stylelint-disable */
+      modifier: css`${StyledCell}:first-of-type` as any
+      /* stylelint-enable */
+    });
   });
 
   it('removes focus styling when blurred', async () => {
@@ -105,14 +97,10 @@ describe('Row', () => {
     await user.click(row);
     await user.tab();
 
-    expect(row).not.toHaveStyleRule(
-      'box-shadow',
-      `inset 3px 0 0 0 ${getColorV8('primaryHue', 600, DEFAULT_THEME)}`,
-      {
-        /* prettier-ignore */
-        modifier: css`${StyledCell}:first-of-type` as any
-      }
-    );
+    expect(row).not.toHaveStyleRule('box-shadow', `inset 3px 0 0 0 ${PALETTE.blue[700]}`, {
+      /* prettier-ignore */
+      modifier: css`${StyledCell}:first-of-type` as any
+    });
   });
 
   it('renders striped styling', () => {
@@ -126,7 +114,12 @@ describe('Row', () => {
 
     expect(getByTestId('row')).toHaveStyleRule(
       'background-color',
-      getColorV8('neutralHue', 100, DEFAULT_THEME)
+      getColor({
+        variable: 'background.emphasis',
+        transparency: DEFAULT_THEME.opacity[100],
+        light: { offset: -300 },
+        theme: DEFAULT_THEME
+      })
     );
   });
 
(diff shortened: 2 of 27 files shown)


packages/tables/src/styled/StyledRow.ts
```diff
@@ -23,36 +23,79 @@ export interface IStyledRowProps {
...
+const baseRowColorStyles = ({ theme, isStriped }: IStyledRowProps & ThemeProps<DefaultTheme>) => {
+  return css`
+    border-bottom: ${theme.borders.sm} ${getColor({ variable: 'border.subtle', theme })};
+    background-color: ${isStriped &&
+    getColor({
+      variable: 'background.emphasis',
+      transparency: theme.opacity[100],
+      light: { offset: -300 },
+      theme
+    })};
```
ze-flo, line 35: Given the context and how `background.subtle` is used in other components, wouldn't `background.subtle` fit better semantically?

```suggestion
    getColor({
      variable: 'background.subtle',
      transparency: theme.opacity[100],
      light: { offset: 300 },
      dark: { offset: -600 },
      theme
    })};
```

ze-flo: I would be good to understand Design's intent for using a custom color / opacity combo.

With Chrome's color-picker tool, I got the "flattened" HEX value for the stripped row background color. That value corresponds to the color a visual user perceives when the row color (neutral (400:600)/0.08%) is applied over our default light / dark background.

Using this [tool](https://colordesigner.io/compare-colors) I compared their DeltaE:


### Light

custom: #f9f9fa
`background.subtle`: #f8f9f9

Screenshot 2024-06-06 at 10 18 22 AM


### Dark

custom: #1e2328
`background.subtle`: #1c2227

Screenshot 2024-06-06 at 10 19 19 AM

As you can see, the colors are nearly identical.


> Typically, a Delta E of less than 1 is not perceptible to the human eye.


If `Table` is always rendered on top of a backdrop with a `background.default`, would it make more sense to reuse `background.subtle`?

packages/tables/src/elements/Row.spec.tsx
```diff
@@ -126,7 +114,12 @@ describe('Row', () => {
 
     expect(getByTestId('row')).toHaveStyleRule(
       'background-color',
-      getColorV8('neutralHue', 100, DEFAULT_THEME)
+      getColor({
+        variable: 'background.emphasis',
+        transparency: DEFAULT_THEME.opacity[100],
+        light: { offset: -300 },
+        theme: DEFAULT_THEME
+      })
```
ze-flo, line 122: Isn't it preferred to reference expected values from PALETTE?

packages/tables/src/styled/StyledHead.ts
```diff
@@ -16,30 +16,40 @@ interface IStyledHeadProps {
...
+  const borderColor = getColor({ variable: 'border.default', theme });
+  const backgroundColor = getColor({ variable: 'background.default', theme });
 
   return css`
-    position: sticky;
-    top: 0;
-    z-index: 1; /* [1] */
-    box-shadow: inset 0 -${props.theme.borderWidths.sm} 0 ${borderColor}; /* [2] */
-    background-color: ${getColorV8('background', 600 /* default shade */, props.theme)};
+    box-shadow: inset 0 -${theme.borderWidths.sm} 0 ${borderColor}; /* [1] */
```
jzempel, line 26: Just checking, `theme.shadows.xs(borderColor)` doesn't work here, right?

packages/tables/src/styled/StyledHeaderRow.ts
```diff
@@ -23,19 +23,38 @@ const getHeaderRowHeight = (props: IStyledRowProps & ThemeProps<DefaultTheme>) =
...
   'data-garden-version': PACKAGE_VERSION
 })`
-  border-bottom-color: ${props => getColorV8('neutralHue', 300, props.theme)};
-  height: ${getHeaderRowHeight};
-  vertical-align: bottom;
   font-weight: ${props => props.theme.fontWeights.semibold};
 
+  ${colorStyles}
+
+  ${sizeStyles}
```
jzempel, line 54: ✨

packages/tables/src/styled/StyledRow.ts
```diff
@@ -23,36 +23,79 @@ export interface IStyledRowProps {
   size?: ITableProps['size'];
 }
 
+const baseRowColorStyles = ({ theme, isStriped }: IStyledRowProps & ThemeProps<DefaultTheme>) => {
+  return css`
+    border-bottom: ${theme.borders.sm} ${getColor({ variable: 'border.subtle', theme })};
```
jzempel, line 28: [nit] the preference is usually to separate `border-bottom` from `border-bottom-color`. And moving these values out to local variables makes the CSS more readable.

packages/tables/src/styled/StyledRow.ts
```diff
@@ -23,36 +23,79 @@ export interface IStyledRowProps {
...
+    getColor({
+      variable: 'background.emphasis',
+      transparency: theme.opacity[100],
+      light: { offset: -300 },
+      theme
+    })};
+  `;
+};
+
 export const StyledBaseRow = styled.tr<IStyledRowProps>`
```
jzempel, line 39: Fwiw, this combined `Styled...` components in one file is a bad code smell. Might be good to clean it up here.

jzempel (changes requested):

lucijanblagonic:
There is an inconsistency with `border-bottom-color` on `Hover`. Our designs specify border bottom color to be `opacity/primary/200`
image

**Current implementation**
image

**Updated example**
image

Not sure if there are any other implications to the border-color, @steue please take a look as well.

packages/tables/src/elements/Row.spec.tsx
geotrev: Is there an easy way to do that when opacity is in play? I didn't see one. 🤔

geotrev: Oh, nvm, JZ answered it below. :D

packages/tables/src/elements/OverflowButton.tsx
```diff
@@ -5,57 +5,27 @@
...
-      >
-        <StyledOverflowButtonIconWrapper>
-          <OverflowStrokeIcon />
-        </StyledOverflowButtonIconWrapper>
-      </StyledOverflowButton>
-    );
-  }
-);
+  return (
+    <StyledOverflowButton type="button" size={size} ref={ref} {...props}>
```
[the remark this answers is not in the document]
geotrev: Yep, good call!

packages/tables/src/styled/StyledHead.ts
geotrev: I don't think so. `xs` returns `0 0 0 ${shadowWidths.xs} ${color}`, so even if we add  `inset`, the axis/fill is incorrect.

packages/tables/src/styled/StyledRow.ts
geotrev: And thank you for the detailed breakdown / comparison with other components in this use case. TIL!

packages/tables/src/elements/OverflowButton.tsx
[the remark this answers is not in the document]
geotrev: Ah, small correction: `size` is definitely needed for calculating the margin / centering the button in the row. This can be set as a transient prop though to prevent DOM leaking.

packages/tables/src/styled/StyledSortableButton.ts
```diff
@@ -57,6 +57,78 @@ interface IStyledSortableButtonProps {
...
+    fill = fgInactive;
+  } else if (sort === 'desc') {
+    color = fgInactive;
+    fill = fgActive;
+  }
+
+  return css`
+    ${StyledSortableFillIconWrapper} {
+      color: ${color};
+      fill: ${fill};
```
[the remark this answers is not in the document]
geotrev: Can you clarify what an empty sort arrow is? I set `foreground.subtle` for sortable arrows over in `StyledSortableButton.ts` for both active and inactive states, but I'm not sure if that's the same as what you're referring to. 😅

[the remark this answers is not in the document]
jzempel: Current deployment

Screenshot 2024-06-10 at 11 28 33 AM

vs. Figma design

Screenshot 2024-06-10 at 11 27 58 AM

packages/tables/src/styled/StyledSortableButton.ts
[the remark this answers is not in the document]
geotrev: I'll set up a ticket so we can prioritize it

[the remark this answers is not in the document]
jzempel: Yes, when you turn `Table.SortableCell` off, they get placed in their correct positions.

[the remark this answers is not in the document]
geotrev: Got it. I'll plan to see if that's addressable in a follow-on PR on `main` so we can pull it over into `next` after.

packages/tables/src/styled/StyledSortableButton.ts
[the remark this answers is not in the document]
geotrev: Gotcha. Looks like the color isn't inheriting correctly. Will fix.

packages/tables/src/styled/StyledRow.ts
geotrev: Something I'm thinking about is what it means for the color and its semantics when we need to do a 6 level offset (plus/minus `600`). Does the color respect its design intent by the time its shade is offset by this much? @lucijanblagonic perhaps a question for you.

geotrev: The backgrounds of both GroupRow and Row (when striped) are set to the below value, per Flo's suggestion, which is the reason for my inquiry/curiosity around color intent with major offsets:

```
 getColor({
      variable: 'background.subtle',
      transparency: theme.opacity[100],
      light: { offset: 300 },
      dark: { offset: -600 },
      theme
})
```

steue: > That value corresponds to the color a visual user perceives when the row color (neutral (400:600)/0.08%) is applied over our default light / dark background.

that is exactly the intent! we wanted a value that visually matches the v8 group/striped bg treatment. i've attached an image to visually describe the points below:

image

1. at first glance, `background.subtle` was the first logical replacement, but that is intentionally a solid fill, and a solid fill would not respect the bg subtle, raised, and recessed treatments in both light and dark modes.
2. with that considered we opted for the existing neutral opacity, which targets 700 and 500 @ .8, but in light mode it was too dark of a bg (albeit dark mode bg was just fine)
3. then opted for the grey-400 / grey-600 @ .08 as it was the closest match.

if there's a more efficient way to obtain that visual parity, then i'm all for it!

packages/tables/src/styled/StyledCell.ts
```diff
@@ -16,7 +16,7 @@ const COMPONENT_ID = 'tables.cell';
 
 export interface IStyledCellProps
   extends Pick<ICellProps, 'isMinimum' | 'isTruncated' | 'hasOverflow' | 'width'> {
-  size?: ITableProps['size'];
+  $size?: ITableProps['size'];
```
jzempel, line 19: Please note that the picked props above will continue to leak in `styled-components` v6.

packages/tables/src/elements/Row.tsx
```diff
@@ -54,10 +57,14 @@ export const Row = forwardRef<HTMLTableRowElement, IRowProps>(
...
-        isReadOnly={isReadOnly}
-        isFocused={computedFocused}
+        $size={size}
+        $isReadOnly={isReadOnly}
+        $isFocused={computedFocused}
+        $isHovered={isHovered}
+        $isStriped={isStriped}
+        $isSelected={isSelected}
         ref={ref}
+        tabIndex={isReadOnly ? undefined : -1}
```
jzempel, line 67: I like the element component move, but in order to match previous the `tabIndex` should follow `{...otherProps}` so that it wins.

packages/tables/src/styled/StyledHeaderRow.ts
```diff
@@ -5,37 +5,61 @@
...
-})`
-  border-bottom-color: ${props => getColorV8('neutralHue', 300, props.theme)};
-  height: ${getHeaderRowHeight};
-  vertical-align: bottom;
+})<IStyledHeaderRowProps>`
   font-weight: ${props => props.theme.fontWeights.semibold};
 
+  ${colorStyles}
+
+  ${sizeStyles}
```
jzempel, line 59:
```suggestion
  ${sizeStyles}

  ${colorStyles}
```

[nit] per API docs, these should be switched .. although it shouldn't happen, color should hold the more significant position if there happens to be a cascade.

packages/tables/src/styled/StyledRow.ts
```diff
@@ -96,14 +113,19 @@ const colorStyles = (props: IStyledRowProps & ThemeProps<DefaultTheme>) => {
...
 };
 
-export const StyledRow = styled(StyledBaseRow).attrs<IStyledRowProps>(props => ({
+const sizeStyles = (props: IStyledRowProps & ThemeProps<DefaultTheme>) => {
+  return css`
+    height: ${getRowHeight(props)};
+  `;
+};
+
+export const StyledRow = styled(StyledBaseRow).attrs<IStyledRowProps>({
```
jzempel, line 122:
```suggestion
export const StyledRow = styled(StyledBaseRow).attrs({
```

[nit] type should no longer be needed for the `attrs`.

jzempel (changes requested):
Looks like `Table.Row` `isSelected` styling got partially lost. Perhaps due to a `$` prop spread? :shrug:

Screenshot 2024-06-11 at 10 12 00 AM

geotrev:
@jzempel think you found a limitation with how backgrounds/borders intermingle in cases of truncation re: `isSelected`. Will wait for design to weigh in (left a comment in the design inventory).

packages/tables/src/styled/StyledCell.ts
geotrev: I went ahead and converted the remaining props to transient props and I _think_ captured all the remaining spreadable props as explicitly converted. Doing another self review now, as well.

jzempel (approved):
:tada:

steue (commented):
the accommodations look good in light mode! but the borders look too dark when `isSelected` is set to true. it looks like it is set to blue-900? we have it set to blue-500 @  16%. was this the border issue you were speaking of? 
 
image

i was able to target it directly in inspect to get the desired effect.

image

geotrev:
@steue Per the [bug JZ reported above](https://github.com/zendeskgarden/react-components/pull/1833#pullrequestreview-2110587601), using a border color with opacity will result in truncated rows hiding their borders. 😅 

That means all states need solid colors for the row border to render consistently.

steue:
😓 in light mode `isSelected` treatment looks good. but it still looks too dark than intended in dark mode. since we can not use opacity, could we use a bespoke value to achieve the same effect?
image

lucijanblagonic:
It's probably best to use colors that we have available within those 12 steps, and accept the technical limitations.

* consistent color for border-bottom-color in both states (selected, selected + hover)
* use `blue-300` for light mode
* use `blue-900` for dark mode

fixed-blue-300-900

A big change in saturation and lightness in dark mode (like using blue-800/700) doesn't look good, and it's OK to have a subtle border as the background is much more prominent.

steue (approved):
i think this is an acceptable workaround for the limitation posed.

PR #1836, opened 2024-06-11
ci: migrate from CircleCI to Github Actions

ze-flo:
## Description
Migrate from CircleCI to Github Actions! 🤖 🚀 

## Detail
- Converts CCI's to GHA config
- Leverage parallel jobs to speed up workflow
- Follows approach to "build first, discard if tests fail"
- Persists only the necessary files across jobs using [actions/upload-artifact@v4](https://github.com/actions/upload-artifact) + [actions/download-artifact@v4](https://github.com/actions/download-artifact)
- Reduces average CI workflow duration on "non-main" branches from ~16min to ~11min (~30% faster! 🚀 )

Screenshot 2024-06-28 at 11 29 39 AM

## Checklist

- ~~[ ] :ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- ~~[ ] :globe_with_meridians: demo is up-to-date (`npm start`)~~
- ~~[ ] :arrow_left: renders as expected with reversed (RTL) direction~~
- ~~[ ] :metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- ~~[ ] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)~~
- ~~[ ] :wheelchair: tested for WCAG 2.1 AA accessibility compliance~~
- ~~[ ] :memo: tested in Chrome, Firefox, Safari, and Edge~~


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages
+
+      - name: Test
+        run: npm run test:ci
+
+      - name: Coveralls
+        if: env.COVERALLS_REPO_TOKEN != ''
```
ze-flo, line 94: Matches the former CCI config.

Is that conditional statement still needed?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Download dist artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages
```
ze-flo, line 88: Retrieves `./dist` folders from `build` job.

If we find a way to run the `test` job independently, we'll speed up the CI workflow by ~3min. 🚅 

Screenshot 2024-06-11 at 2 01 09 PM

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - name: Deploy
+        run: utils/scripts/deploy.mjs
+        env:
+          NETLIFY_TOKEN: ${{ secrets.NETLIFY_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+          # Maintain compatibility with @zendeskgarden/scripts
+          CIRCLE_PROJECT_USERNAME: zendeskgarden
+          CIRCLE_PROJECT_REPONAME: react-components
+          CIRCLECI: true
```
ze-flo, line 129: Temporary solution to avoid errors. Ultimately, we could update `githubRepository` to support [HTTP URLs](https://github.com/zendeskgarden/react-components/actions/runs/9462583243/job/26066350814#step:6:9).

https://github.com/zendeskgarden/scripts/blob/82b001563247320e98315a30c6c4fe45db9f8950/src/github/repository/index.ts#L34-L52

package.json
```diff
@@ -12,6 +12,7 @@
     "format:package_json": "lerna exec -- prettier-package-json",
     "postinstall": "husky",
     "lint": "npm run lint:css && npm run lint:js && npm run lint:md",
+    "lint:ci": "npm run lint && npm run format:all && npm exec tsc",
```
ze-flo, line 15: Spliting linting from all tests to speed up workflow.

package.json
```diff
@@ -20,7 +21,7 @@
     "start": "storybook dev --no-version-updates -p 6006",
     "tag": "utils/scripts/tag.mjs",
     "test": "jest --config=utils/test/jest.config.js",
-    "test:ci": "npm run lint && npm run format:all && npm exec tsc && npm test -- --coverage",
+    "test:ci": "npm test -- --coverage --runInBand --silent",
```
ze-flo, line 24: `--silent`: muting console warnings to avoid [massive and unhelpful logs](https://github.com/zendeskgarden/react-components/actions/runs/9462583243/job/26066206181#step:6:127).

I'm still researching the best options to speed up Jest on GH runners.

utils/build/styled.d.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import 'styled-components';
-import { IGardenTheme } from '../../packages/theming';
+import type { IGardenTheme } from '../../packages/theming/src/index';
```
ze-flo, line 9: Pointing to the source rather than the `./dist` output allows the `lint` job (`npm exec tsc`) to run in parallel with `build`.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - name: Download dist artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages
+
+      # - name: Publish to npm
+      #   run: npm exec -- lerna publish from-git --ignore-scripts --yes
+      #   env:
+      #     NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
ze-flo, line 159: The million 💰 question. Will this work? 🤷‍♂️ 

Based on my [research](https://github.com/actions/checkout/issues/217#issuecomment-634802933) `fetch-depth: 0` is all we need for `lerna publish` to work.

> [!NOTE]
> Fetching the entire Git history + tags takes [~2min](https://github.com/zendeskgarden/react-components/actions/runs/9462583243/job/26066351150#step:2:52).


Do you have any recommendations on how to test this?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
```
ze-flo, line 41: If you're wondering why this isn't DRY by using YAML anchors and aliases, Github [doesn't support](https://github.com/actions/runner/issues/1182) such syntax.

.circleci/config.yml
```diff
@@ -2,73 +2,15 @@ version: 2.1
 orbs:
   node: circleci/node@5.2.0
 
-references:
-  executor: &executor
+jobs:
+  check:
```
ze-flo, line 6: Since it appears CCI is a required check, it might be good to temporarily add a dummy config for a smoother transition.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
```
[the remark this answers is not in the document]
geotrev: Does GHA automatically install nvm as part of the orb setup?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+        run: npm ci --ignore-scripts
+
+      - name: Build
+        run: npm exec -- lerna run build --concurrency=2
+
+      - name: Upload dist artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages/**/dist
```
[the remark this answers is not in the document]
geotrev: Is this something we had set previously for CCI? I didn't see an equivalent, so I could go either way. The config route does feel more obvious for us to easily grok in the future, but the settings live independent of package commits.

.github/workflows/ci.yaml
geotrev: womp womp

.github/workflows/ci.yaml
geotrev: I could be wrong on intent, but `test` feels like a great parallel job with `lint` and `build`.

.github/workflows/ci.yaml
geotrev: I'm curious if there's a real scenario where the token isn't/couldn't be loaded, given it's defined in our secrets. 🤔

.circleci/config.yml
jzempel, line 1: My understanding was that we would leave the CircleCI config untouched for now, allowing it to run in parallel with GH Actions so that we can make side-by-side comparisons. Also, I think we still need `publish` capabilities here until we smoke-test the GH action (more on this below).

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
+name: CI
+
+on: push
```
jzempel, line 3: All of these CI workflows should only happen on `pull_request`. That's the [typical setting](https://app.circleci.com/settings/project/github/zendeskgarden/react-containers/advanced?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fzendeskgarden%2Freact-containers) for all CircleCI repos. We just turned it off temporarily on `react-components` because there was no way to get `next` to build in addition to `main`.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
jzempel: Let's definitely rely on the setting rather than specify here. I want to be able to control this outside the scope of a PR.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
```
jzempel, line 64: Is there a way to run install only once, as we do with CircleCI, and subsequently use the installed `node_modules` cache for subsequent test, deploy, and publish steps?

.github/workflows/ci.yaml
jzempel: If GH actions provides standard owner & repo string in the env, then we should definitely get `scripts` updated for use with this PR.

.github/workflows/ci.yaml
jzempel: For the smoke test, let's run `npm whoami`. If the result comes back with `zendesk-garden`, then we have confidence that the configuration is solid.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+        env:
+          NETLIFY_TOKEN: ${{ secrets.NETLIFY_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+          # Maintain compatibility with @zendeskgarden/scripts
+          CIRCLE_PROJECT_USERNAME: zendeskgarden
+          CIRCLE_PROJECT_REPONAME: react-components
+          CIRCLECI: true
+
+  publish:
```
jzempel, line 131: How can this step be configured to only run for users with [maintainers](https://github.com/orgs/zendeskgarden/teams/maintainers) permissions?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - name: Test
+        run: npm run test:ci
+
+      - name: Coveralls
+        if: env.COVERALLS_REPO_TOKEN != ''
+        run: npm exec -- coveralls < .cache/coverage/lcov.info
+        env:
+          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+
+  deploy:
```
jzempel, line 99: How can deploy be configured to only run for users with [writers](https://github.com/orgs/zendeskgarden/teams/writers) permission?

package.json
jzempel: After the `.circleci/config.yml` is restored, please be sure to include `lint:ci` in the test portion of that workflow.

jzempel (changes requested):

.github/workflows/ci.yaml
ze-flo: I've never used the `environment` option but I think this is what we're looking for:

https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#creating-an-environment

.github/workflows/ci.yaml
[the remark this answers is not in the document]
ze-flo: `ubuntu-latest` latest correspond to this runner [config](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md). `nvm` isn't installed. Instead if `actions/setup-node@v4` recognizes a [supported syntax](https://github.com/actions/setup-node?tab=readme-ov-file#supported-version-syntax) in the `node-version-file`, it installs the appropriate Node version.

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

.github/workflows/ci.yaml
ze-flo: > Note It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci

https://github.com/actions/cache/blob/main/examples.md#node---npm

.github/workflows/ci.yaml
ze-flo: It does within the Github Context.

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/workflows/ci.yaml
ze-flo: See above ☝️

.circleci/config.yml
ze-flo: Should I remove the `deploy` step to avoid deploying twice, or is that not a concern?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,163 @@
...
+          name: demo-artifact
+          path: ./demo
+
+      - name: Deploy
+        run: utils/scripts/deploy.mjs
+        env:
+          NETLIFY_TOKEN: ${{ secrets.NETLIFY_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+          REPO_SLUG: ${{ github.repository }}
```
ze-flo, line 125: See https://github.com/zendeskgarden/scripts/pull/215

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,190 @@
...
+          cache: npm
+          registry-url: https://registry.npmjs.org # Sets the registry in the project level .npmrc
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Who am i?
+        run: npm whoami
```
ze-flo, line 151: > For the smoke test, let's run `npm whoami`. If the result comes back with `zendesk-garden`, then we have confidence that the configuration is solid.

Looks good. ✅ 
https://github.com/zendeskgarden/react-components/actions/runs/9523125471/job/26254182167#step:5:8

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,190 @@
...
+
+      - name: Download dist artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages
+
+      - name: Check dist theming artifact
+        run: ls -Rla ./packages/theming
```
ze-flo, line 160: That's what we're expecting.

https://github.com/zendeskgarden/react-components/actions/runs/9523125471/job/26254182167#step:7:16

https://github.com/zendeskgarden/react-components/actions/runs/9523125471/job/26254182167#step:7:40

.github/workflows/ci.yaml
jzempel: But, just like we currently have with CircleCI, the `npm ci` shouldn't need to run more than once. So I don't believe this statement would apply. I'd like to see this setup start by mirroring current CI as closely as possible and then work out from there.

.github/workflows/ci.yaml
jzempel: The script should be using GH Actions [default variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)
```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,190 @@
...
+        run: npm ci --ignore-scripts
+
+      - name: Download demo artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: demo-artifact
+          path: ./demo
+
+      - name: Deploy
+        run: utils/scripts/deploy.mjs
```
jzempel, line 120: Are we good with GitHub Pages deployments on `main` after June 30? https://docs.github.com/en/pages/quickstart

jzempel (changes requested):

.circleci/config.yml
jzempel: Yeah, I think knocking `deploy` out of the CircleCI workflow would be good.

.github/workflows/ci.yaml
ze-flo: I'm following best-practices as defined in the [docs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#comparing-artifacts-and-dependency-caching). That approach also aligns with Github Primer React's [config](https://github.com/primer/react/blob/c093411ec6caa4d4ce1720a68a22e492d5f39458/.github/workflows/ci.yml).

> Artifacts and caching are similar because they provide the ability to store files on GitHub, but each feature offers different use cases and cannot be used interchangeably.

> - Use caching when you want to reuse files that don't change often between jobs or workflow runs, such as build dependencies from a package management system.
> - Use artifacts when you want to save files produced by a job to view after a workflow run has ended, such as built binaries or build logs.

To mirror the CCI config, we would have to upload the ~1GB workspace as an artifact and then download it for every job. That will probably take longer than executing `npm ci --no-scripts` _with_ caching.

> Storing artifacts uses storage space on GitHub 

[Source](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#comparing-artifacts-and-dependency-caching:~:text=Storing%20artifacts%20uses%20storage%20space%20on%20GitHub)

One CI run with an stored 1GB artifact would set up above our [free 500MB monthly allowance](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes). That could become costly.

.github/workflows/ci.yaml
ze-flo: I'll remove once I get confirmation. Thanks!

.github/workflows/ci.yaml
geotrev: Does this config/GHA now allow us to only build on PR for select branches? 🤞🏻

ze-flo: Yes, it supports it.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request

```
on:
  pull_request:
    branches:
      - 'releases/**'
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,186 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install dependencies
```
jzempel, line 17: Can we forego `name` throughout? We're accustomed to seeing commands in CI, and the name tends to hide that info under a section that has to be expanded to view. And similar to CCI, GH Actions that we depend on will provide their own.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,186 @@
...
+      - name: Build
+        run: npm exec -- lerna run build --concurrency=2
+
+      - name: Upload dist artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist-artifact
+          path: packages/**/dist
+
+  build-demo:
```
jzempel, line 29: I'd like to see this folded directly into `deploy`. We're not getting a ton of value here, and the new workflow could be simplified with:

- build
  - lint
  - test
  - deploy
    - publish

where (lint, test, deploy) run in parallel. Furthermore, it makes good sense to wait for a successful `build` before doing the `deploy`.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,186 @@
...
+      - name: Test
+        run: npm run test:ci
+
+      - name: Coveralls
+        run: npm exec -- coveralls < .cache/coverage/lcov.info
+        env:
+          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+
+  deploy:
+    needs: [build-demo, lint, test]
```
jzempel, line 100:
```suggestion
    needs: [build]
```

...per previous comment, and just scoop the `build:demo` into this job.

.github/workflows/ci.yaml
jzempel: Ok, I'm on board with the current approach.

.github/workflows/ci.yaml
ze-flo: Looks great and it's fast! 🔥 🚀 

Screenshot 2024-06-18 at 11 48 04 AM

On `main`, do we still want to deploy even if `test`& `lint` have not passed?

.github/workflows/ci.yaml
ze-flo: Much better! ✅ 

Screenshot 2024-06-18 at 12 16 31 PM

.github/workflows/ci.yaml
jzempel: > On main, do we still want to deploy even if test& lint have not passed?

Hmm, good Q. No I don't think we do.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - main
+      - next
+
+permissions: write-all
```
ze-flo, line 11: Even with `permissions: write-all`, `garden github-teams` cannot access team resources. 

Screenshot 2024-06-26 at 1 52 40 PM


Let's discuss options.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,153 @@
...
+      - run: npm ci --ignore-scripts
+      - run: npm run build:demo
+      - run: utils/scripts/deploy.mjs
+        env:
+          NETLIFY_TOKEN: ${{ secrets.NETLIFY_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+
+  deploy-gh-pages:
+    # if: contains(needs.initialize.outputs.teams, 'Writers') && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/next')
```
ze-flo, line 94: To clarify, this will only trigger following a `push` event. 

For `push` events, `github.ref` looks something like this `refs/heads/{branch}`

For `pull_request` events, `github.ref` > `refs/pull/{PR#}/merge`

jzempel (commented):
Nice work @ze-flo. I'll plan to continue on with this PR in prep for final transition away from CircleCI

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: './demo'
+      - id: deployment
+        uses: actions/deploy-pages@v4
+
+  publish:
+    needs: [initialize, lint, test]
+    runs-on: ubuntu-latest
+    if: contains(needs.initialize.outputs.teams, 'Maintainers') && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/next')
```
jzempel, line 128: This condition is fine for now, but we'll need to allow version maintenance branches (i.e. `v7`, `v8`, etc) as well, so may need to weigh the idea of removing and relying solely on Lerna's `allowBranch` config.

jzempel (approved):

geotrev (approved):

PR #1837, opened 2024-06-13
chore: add fix-dist-tags script

ze-flo:
## Description

This small script facilitates setting the `latest` [dist-tag](https://docs.npmjs.com/cli/v10/commands/npm-dist-tag) to the version of your choice or the version specified in the package's `package.json` (default).

This is useful to quickly resolve mistakes when accidentally setting the `latest` dist-tag to an incorrect release (ex: `next`)

```
npm run fix-dist-tags // set dist-tag to latest using package.json's version

npm run fix-dist-tags -- -v 99.9.10 // using a version of your choice
```
> [!NOTE]
> Make sure to have the auth token set in `~/.mpnrc` first before running.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "lerna run build --stream",
     "build:demo": "storybook build -o ./demo",
+    "fix-dist-tags": "utils/scripts/fix-dist-tags.mjs",
     "format": "prettier-package-json --write && npm run -- format:package_json --write && npm run -- format:js --write && npm run -- format:md --write",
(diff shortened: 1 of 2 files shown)


jzempel (changes requested):
In general we can't accept on the basis of two criteria:
1. Planned Garden scripts get consolidated under https://github.com/zendeskgarden/scripts
2. We don't add package scripts for operations that require [zendesk-garden](https://github.com/zendesk-garden) token access

ze-flo:
> In general we can't accept on the basis of two criteria:
> 
> 1. Planned Garden scripts get consolidated under https://github.com/zendeskgarden/scripts
> 2. We don't add package scripts for operations that require [zendesk-garden](https://github.com/zendesk-garden) token access

Makes sense. Thanks for the clarification. Closing.

PR #1842, opened 2024-06-18
feat(notifications): new light/dark mode colors

geotrev:
## Description

Components updated in this PR:

- [x] `Notification`
    - [x] `Notification.Close`
    - [x] `Notification.Paragraph`
    - [x] `Notification.Title`
- [x] `Alert`
    - [x] `Alert.Close`
    - [x] `Alert.Paragraph`
    - [x] `Alert.Title`
- [x] `Well`
    - [x] `Well.Paragraph`
    - [x] `Well.Title`
- [x] `GlobalAlert`
    - [x] `GlobalAlert.Button`
    - [x] `GlobalAlert.Close`
    - [x] `GlobalAlert.Content`
    - [x] `GlobalAlert.Title`

## Detail

Internal prop spreads have been converted to transient props, with the exception of props used by non-converted components (e.g., `GlobalAlert.Button` retains all props from its base `Button`).

## Checklist

### DIFF
--- a/packages/notifications/src/elements/content/Close.tsx
+++ b/packages/notifications/src/elements/content/Close.tsx
@@ -19,10 +19,10 @@ import XStrokeIcon from '@zendeskgarden/svg-icons/src/12/x-stroke.svg';
 export const Close = React.forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLButtonElement>>(
   (props, ref) => {
     const ariaLabel = useText(Close, props, 'aria-label', 'Close');
-    const hue = useNotificationsContext();
+    const type = useNotificationsContext();
 
     return (
-      <StyledClose ref={ref} hue={hue} aria-label={ariaLabel} {...props}>
+      <StyledClose ref={ref} $type={type} aria-label={ariaLabel} {...props} focusInset size="small">
         <XStrokeIcon />
       </StyledClose>
     );
--- a/packages/notifications/src/elements/content/Close.tsx
+++ b/packages/notifications/src/elements/content/Close.tsx
@@ -9,7 +9,7 @@ import React, { ButtonHTMLAttributes } from 'react';
 import { StyledClose } from '../../styled';
 import { useNotificationsContext } from '../../utils/useNotificationsContext';
 import { useText } from '@zendeskgarden/react-theming';
-import XStrokeIcon from '@zendeskgarden/svg-icons/src/12/x-stroke.svg';
+import XStrokeIcon from '@zendeskgarden/svg-icons/src/16/x-stroke.svg';
 
 /**
  * @deprecated use `Alert.Close` or `Notification.Close` instead
--- a/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
+++ b/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
@@ -14,7 +14,7 @@ import SuccessIcon from '@zendeskgarden/svg-icons/src/16/check-circle-stroke.svg
 
 import { TYPE, IGlobalAlertProps } from '../../types';
 import { StyledGlobalAlert, StyledGlobalAlertIcon } from '../../styled';
-import { GlobalAlertContext } from './utility';
+import { GlobalAlertContext } from '../../utils/useGlobalAlertContext';
 import { GlobalAlertButton } from './GlobalAlertButton';
 import { GlobalAlertClose } from './GlobalAlertClose';
 import { GlobalAlertContent } from './GlobalAlertContent';
@@ -38,8 +38,8 @@ const GlobalAlertComponent = forwardRef<HTMLDivElement, IGlobalAlertProps>(
       <GlobalAlertContext.Provider value={useMemo(() => ({ type }), [type])}>
         {/* [1] */}
         {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
-        <StyledGlobalAlert ref={ref} role="status" alertType={type} {...props}>
-          <StyledGlobalAlertIcon>{icon}</StyledGlobalAlertIcon>
+        <StyledGlobalAlert ref={ref} role="status" $alertType={type} {...props}>
+          <StyledGlobalAlertIcon $alertType={type}>{icon}</StyledGlobalAlertIcon>
           {props.children}
         </StyledGlobalAlert>
       </GlobalAlertContext.Provider>
--- a/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
+++ b/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
@@ -5,8 +5,9 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { forwardRef, useContext, useMemo } from 'react';
+import React, { forwardRef, useMemo } from 'react';
 import PropTypes from 'prop-types';
+import { ThemeProvider } from 'styled-components';
 import InfoIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
 import ErrorIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
 import WarningIcon from '@zendeskgarden/svg-icons/src/16/alert-warning-stroke.svg';
@@ -19,8 +20,6 @@ import { GlobalAlertButton } from './GlobalAlertButton';
 import { GlobalAlertClose } from './GlobalAlertClose';
 import { GlobalAlertContent } from './GlobalAlertContent';
 import { GlobalAlertTitle } from './GlobalAlertTitle';
-import { DefaultTheme, ThemeContext, ThemeProvider } from 'styled-components';
-import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
 /**
  * 1. Global Alert always renders with light theme colors
@@ -36,15 +35,10 @@ const GlobalAlertComponent = forwardRef<HTMLDivElement, IGlobalAlertProps>(
       warning: <WarningIcon />,
       info: <InfoIcon />
     }[type];
-    const theme = useContext(ThemeContext) || DEFAULT_THEME;
-    /* [1] */
-    const lightTheme: DefaultTheme = useMemo(
-      () => ({ ...theme, colors: { ...theme.colors, base: 'light' } }),
-      [theme]
-    );
 
     return (
-      <ThemeProvider theme={lightTheme}>
+      /* [1] */
+      <ThemeProvider theme={theme => ({ ...theme, colors: { ...theme.colors, base: 'light' } })}>
         <GlobalAlertContext.Provider value={useMemo(() => ({ type }), [type])}>
           {/* [2] */}
           {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
--- a/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
+++ b/packages/notifications/src/elements/global-alert/GlobalAlert.tsx
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { forwardRef, useMemo } from 'react';
+import React, { forwardRef, useContext, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import InfoIcon from '@zendeskgarden/svg-icons/src/16/info-stroke.svg';
 import ErrorIcon from '@zendeskgarden/svg-icons/src/16/alert-error-stroke.svg';
@@ -19,9 +19,12 @@ import { GlobalAlertButton } from './GlobalAlertButton';
 import { GlobalAlertClose } from './GlobalAlertClose';
 import { GlobalAlertContent } from './GlobalAlertContent';
 import { GlobalAlertTitle } from './GlobalAlertTitle';
+import { DefaultTheme, ThemeContext, ThemeProvider } from 'styled-components';
+import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
 /**
- * 1. role='status' on `div` is valid WAI-ARIA usage in this context.
+ * 1. Global Alert always renders with light theme colors
+ * 2. role='status' on `div` is valid WAI-ARIA usage in this context.
  *    https://www.w3.org/TR/wai-aria-1.1/#status
  */
 
@@ -33,16 +36,24 @@ const GlobalAlertComponent = forwardRef<HTMLDivElement, IGlobalAlertProps>(
       warning: <WarningIcon />,
       info: <InfoIcon />
     }[type];
+    const theme = useContext(ThemeContext) || DEFAULT_THEME;
+    /* [1] */
+    const lightTheme: DefaultTheme = useMemo(
+      () => ({ ...theme, colors: { ...theme.colors, base: 'light' } }),
+      [theme]
+    );
 
     return (
-      <GlobalAlertContext.Provider value={useMemo(() => ({ type }), [type])}>
-        {/* [1] */}
-        {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
-        <StyledGlobalAlert ref={ref} role="status" $alertType={type} {...props}>
-          <StyledGlobalAlertIcon $alertType={type}>{icon}</StyledGlobalAlertIcon>
-          {props.children}
-        </StyledGlobalAlert>
-      </GlobalAlertContext.Provider>
+      <ThemeProvider theme={lightTheme}>
+        <GlobalAlertContext.Provider value={useMemo(() => ({ type }), [type])}>
+          {/* [2] */}
+          {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
+          <StyledGlobalAlert ref={ref} role="status" $alertType={type} {...props}>
+            <StyledGlobalAlertIcon $alertType={type}>{icon}</StyledGlobalAlertIcon>
+            {props.children}
+          </StyledGlobalAlert>
+        </GlobalAlertContext.Provider>
+      </ThemeProvider>
     );
   }
 );
--- a/packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
+++ b/packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
@@ -23,12 +23,6 @@ export const GlobalAlertButton = forwardRef<HTMLButtonElement, IGlobalAlertButto
         ref={ref}
         $alertType={type}
         {...props}
-        focusInset={false}
-        isDanger={false}
-        isLink={false}
-        isNeutral={false}
-        isPill={false}
-        isStretched={false}
         size="small"
         isPrimary={!isBasic}
         isBasic={isBasic}
--- a/packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
+++ b/packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
@@ -7,10 +7,9 @@
 
 import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
-
 import { IGlobalAlertButtonProps } from '../../types';
 import { StyledGlobalAlertButton } from '../../styled';
-import { useGlobalAlertContext } from './utility';
+import { useGlobalAlertContext } from '../../utils/useGlobalAlertContext';
 
 /**
  * @extends ButtonHTMLAttributes<HTMLButtonElement>
@@ -22,8 +21,15 @@ export const GlobalAlertButton = forwardRef<HTMLButtonElement, IGlobalAlertButto
     return (
       <StyledGlobalAlertButton
         ref={ref}
-        alertType={type}
+        $alertType={type}
         {...props}
+        focusInset={false}
+        isDanger={false}
+        isLink={false}
+        isNeutral={false}
+        isPill={false}
+        isStretched={false}
+        size="small"
         isPrimary={!isBasic}
         isBasic={isBasic}
       />
--- a/packages/notifications/src/styled/StyledBase.ts
+++ b/packages/notifications/src/styled/StyledBase.ts
@@ -61,7 +61,7 @@ const colorStyles = ({ theme, $type, $isFloating }: IStyledBaseProps) => {
         fgVariable = 'foreground.warning';
         break;
       case validationTypes.info:
-        bgVariable = 'background.raised';
+        bgVariable = 'background.subtle';
         borderVariable = 'border.default';
         fgVariable = 'foreground.subtle';
         break;
--- a/packages/notifications/src/styled/StyledBase.ts
+++ b/packages/notifications/src/styled/StyledBase.ts
@@ -28,7 +28,7 @@ const colorStyles = ({ theme, $type, $isFloating }: IStyledBaseProps) => {
   let borderVariable;
   let fgVariable;
 
-  if (!$isFloating && $type && Object.keys(validationTypes).includes($type)) {
+  if (!$isFloating && $type && !!(validationTypes as any)[$type]) {
     switch ($type) {
       case validationTypes.success:
         bgVariable = 'background.success';
--- a/packages/notifications/src/styled/StyledBase.ts
+++ b/packages/notifications/src/styled/StyledBase.ts
@@ -22,23 +22,8 @@ export interface IStyledBaseProps extends ThemeProps<DefaultTheme> {
   $type?: Type;
 }
 
-const boxShadow = (props: IStyledBaseProps) => {
-  const { theme } = props;
-  const { space, shadows, opacity } = theme;
-  const offsetY = `${space.base * 5}px`;
-  const blurRadius = `${space.base * 7}px`;
-  const color = getColor({
-    hue: 'neutralHue',
-    shade: 1200,
-    light: { transparency: opacity[200] },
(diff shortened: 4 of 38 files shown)


packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -7,72 +7,145 @@
...
-      foregroundColor = getColorV8('successHue', 100, props.theme);
-      anchorHoverColor = props.theme.palette.white;
-      anchorActiveColor = props.theme.palette.white;
-      focusColor = 'successHue';
+  let focusVariable;
+
+  switch ($alertType) {
+    case 'success': {
+      borderColor = getColor({
+        variable: 'border.successEmphasis',
```
geotrev, line 44: Interested in thoughts around this use of emphasis vs non-emphasis. In general, I tried to use non-emphasis with foreground and emphasis with background. Does this align with design intent?

geotrev: Note this doesn't apply across all components, and at some points foreground is using emphasis (see StyledAlert). 

I'd like to ensure that when a consumer updates a color variable, it cascades through these components in the least confusing way. 

Also, let me know if I'm overthinking it. 😛

steue:
thanks @geotrev!! for posterity from our IC sync
1. default well in dark mode should have "default" styling, then "raised" styling when isFloating 
image

~~2. design is in agreement to "fix" button styling so warning button treatment uses yellow-700.~~ looks like you tackled that already!

other than that it looks good!

packages/notifications/src/elements/content/Close.tsx
```diff
@@ -19,10 +19,10 @@ import XStrokeIcon from '@zendeskgarden/svg-icons/src/12/x-stroke.svg';
 export const Close = React.forwardRef<HTMLButtonElement, ButtonHTMLAttributes<HTMLButtonElement>>(
   (props, ref) => {
     const ariaLabel = useText(Close, props, 'aria-label', 'Close');
-    const hue = useNotificationsContext();
+    const type = useNotificationsContext();
 
     return (
-      <StyledClose ref={ref} hue={hue} aria-label={ariaLabel} {...props}>
+      <StyledClose ref={ref} $type={type} aria-label={ariaLabel} {...props} focusInset size="small">
         <XStrokeIcon />
```
jzempel, line 26: Do we get better results with the 16px? The 12px is looking unintentionally heavy here.

packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
```diff
@@ -22,8 +21,15 @@ export const GlobalAlertButton = forwardRef<HTMLButtonElement, IGlobalAlertButto
...
         ref={ref}
-        alertType={type}
+        $alertType={type}
         {...props}
+        focusInset={false}
+        isDanger={false}
+        isLink={false}
+        isNeutral={false}
+        isPill={false}
+        isStretched={false}
```
jzempel, line 31: Are these `false` props required? Just preventing `Button` prop leaking?

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -6,40 +6,76 @@
...
-  type?: Type;
+const COMPONENT_ID = 'notifications.base_container';
+
+export interface IStyledBaseProps extends ThemeProps<DefaultTheme> {
+  $isFloating?: boolean;
+  $type?: Type;
 }
 
-const boxShadow = (props: ThemeProps<DefaultTheme>) => {
+const boxShadow = (props: IStyledBaseProps) => {
```
jzempel, line 25: Recommend getting rid of the `boxShadow` function .. per [API docs](https://github.com/zendeskgarden/react-components/blob/main/docs/api.md#rules), this should just be part of `colorStyles`.

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -55,17 +91,22 @@ const padding = (props: ThemeProps<DefaultTheme>) => {
...
-export const StyledBase = styled.div`
+export const StyledBase = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
+})<IStyledBaseProps>`
   position: relative;
   border: ${props => props.theme.borders.sm};
   border-radius: ${props => props.theme.borderRadii.md};
-  box-shadow: ${props => props.isFloating && boxShadow};
+  box-shadow: ${props => props.$isFloating && boxShadow};
```
jzempel, line 101: Per previous comment, move `box-shadow` to `colorStyles`.

packages/notifications/src/styled/StyledIcon.ts
```diff
@@ -5,16 +5,63 @@
...
+    case validationTypes.error:
+      color = getColor({ variable: 'foreground.danger', theme });
+      break;
+    case validationTypes.warning:
+      color = getColor({ variable: 'foreground.warning', theme });
+      break;
+    case validationTypes.info:
+      color = getColor({ variable: 'foreground.subtle', theme });
+      break;
+  }
```
jzempel, line 48: Is it worth just capturing the variable string as you've done other places and just call `getColor` once?

packages/notifications/src/styled/StyledWell.ts
```diff
@@ -22,9 +22,11 @@ export const StyledWell = styled(StyledBase).attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
 })<IStyledWellProps>`
-  background-color: ${props => props.isRecessed && getColorV8('neutralHue', 100, props.theme)};
-  color: ${props => getColorV8('neutralHue', 600, props.theme)}
-    ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  background-color: ${p =>
+    p.$isRecessed && getColor({ variable: 'background.recessed', theme: p.theme })};
+  color: ${p => getColor({ variable: 'foreground.subtle', theme: p.theme })};
```
jzempel, line 27: Feels like there's enough going on here to warrant a more readable `colorStyles` with good variable declarations.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -7,72 +7,145 @@
...
 
 interface IStyledGlobalAlertProps {
-  alertType: IGlobalAlertProps['type'];
+  $alertType: IGlobalAlertProps['type'];
 }
 
+const lightDarkOptions = (lightOffset: number, darkOffset: number) => ({
+  light: { offset: lightOffset },
+  dark: { offset: darkOffset }
+});
```
jzempel, line 28: I thought the design directive for `GlobalAlert` was to go with one set of colors, regardless of light/dark mode. Hopefully the color logic below can be simplified.

packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts
```diff
@@ -6,65 +6,91 @@
...
   }
 
-  let backgroundColor;
-  let hoverBackgroundColor;
-  let activeBackgroundColor;
-  let focusColor;
+  let bgVariable;
+  let offsetOptions: OffsetOptions = { light: { offset: 200 }, dark: { offset: 300 } };
+  let offsetHoverOptions: OffsetOptions = { light: { offset: 300 }, dark: { offset: 400 } };
+  let offsetActiveOptions: OffsetOptions = { light: { offset: 400 }, dark: { offset: 500 } };
```
jzempel, line 41: Repeating from the `GlobalAlert` comment, I don't think we should be seeing light/dark offsets. The `GlobalAlert.Button` functions the same regardless of mode.

packages/notifications/src/styled/global-alert/StyledGlobalAlertClose.ts
```diff
@@ -6,53 +6,91 @@
...
+      });
+      activeBackgroundColor = getColor({
+        variable: 'background.warningEmphasis',
+        transparency: theme.opacity[200],
+        theme
+      });
+      activeForegroundColor = getColor({
+        variable: 'foreground.warningEmphasis',
+        light: { offset: 300 },
+        dark: { offset: 600 },
```
jzempel, line 86: Ditto re: modeless on `GlobalAlert.Close`.

packages/notifications/src/styled/global-alert/StyledGlobalAlertIcon.ts
```diff
@@ -30,6 +36,52 @@ const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
...
+        dark: { offset: -100 },
+        theme
+      });
+      break;
+    case 'error':
+      // red/300
+      color = getColor({
+        variable: 'foreground.danger',
+        light: { offset: -400 },
+        dark: { offset: -100 },
```
jzempel, line 60: Also modeless for the `GlobalAlert` icon.

packages/notifications/src/styled/global-alert/StyledGlobalAlertTitle.ts
```diff
@@ -6,31 +6,39 @@
...
-      color = getColorV8('warningHue', 900, props.theme);
+      color = getColor({ variable: 'foreground.warningEmphasis', dark: { offset: 600 }, theme });
       break;
 
     case 'info':
-      color = getColorV8('primaryHue', 800, props.theme);
+      color = getColor({
+        variable: 'foreground.primary',
+        light: { offset: 200 },
+        dark: { offset: 300 },
```
jzempel, line 39: Ditto re: modeless for `GlobalAlert.Title`.

jzempel (changes requested):

packages/notifications/src/utils/icons.ts
```diff
@@ -18,9 +18,9 @@ export const validationIcons: Record<Type, Record<string, unknown>> = {
...
-export const validationHues: Record<Type, string> = {
-  success: 'successHue',
-  error: 'dangerHue',
-  warning: 'warningHue',
-  info: 'neutralHue'
+export const validationTypes: Record<Type, string> = {
+  success: 'success',
+  error: 'error',
+  warning: 'warning',
+  info: 'info'
```
ze-flo, line 25: ❤️

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -55,17 +91,22 @@ const padding = (props: ThemeProps<DefaultTheme>) => {
   return `${paddingVertical} ${paddingHorizontal}`;
 };
 
-export const StyledBase = styled.div`
+export const StyledBase = styled.div.attrs({
+  'data-garden-id': COMPONENT_ID,
+  'data-garden-version': PACKAGE_VERSION
```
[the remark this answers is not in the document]
geotrev: I remember we had some back and forth originally, and we decided to leave out the details of the base component because we expected all other icon components (like this one) to define the `COMPONENT_ID` AND `retrieveComponentStyles` etc.

packages/notifications/src/utils/useNotificationsContext.ts
```diff
@@ -6,10 +6,9 @@
  */
 
 import { createContext, useContext } from 'react';
+import { Type } from '../types';
 
-export type Hue = 'successHue' | 'warningHue' | 'dangerHue' | 'neutralHue';
-
-export const NotificationsContext = createContext<Hue | undefined>(undefined);
+export const NotificationsContext = createContext<Type | undefined>(undefined);
```
ze-flo, line 11: [nit] The `props.type` is [required](https://garden.zendesk.com/components/alerts#api), thus should always defined. 

To align the context API with the props', should we specify a default rather than using `undefined`?

```suggestion
export const NotificationsContext = createContext<Type | undefined>('info');
```

Doing so will make things a bit clearer in that file:

https://github.com/zendeskgarden/react-components/pull/1842/files#diff-30a97b0e497db8834422a4c3a59fd93ac89d538cb7fc5a65d34f0f9be4850c73R17

https://github.com/zendeskgarden/react-components/pull/1842/files#diff-30a97b0e497db8834422a4c3a59fd93ac89d538cb7fc5a65d34f0f9be4850c73R29

packages/notifications/src/elements/content/Close.tsx
geotrev: For comparison:

12px
Screenshot 2024-06-20 at 3 33 56 PM

16px
Screenshot 2024-06-20 at 3 34 06 PM

packages/notifications/src/elements/global-alert/GlobalAlertButton.tsx
geotrev: That was my interpretation. The values of these props are all `false` by default so we could omit them and expect consumers not to abuse it. I was preserving the props from the underlying view, where these were defined previously.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
geotrev: Can you clarify what you are hoping to see / what is aligning for you? I'm not sure how to proceed.

Steph, Lucijan and I had a brief exchange about using variables here instead of the raw colors as the semantic importance of them fits well with GlobalAlert (success, error, warning, etc).

packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts
geotrev: Similar to my comment on the other comment, the offsets are there to ensure the underlying shade remains consistent between light/dark. If that isn't going to work here, we should definitely discuss in more detail.

packages/notifications/src/utils/useNotificationsContext.ts
geotrev: I think you caught a gap in color handling actually. 😅

`Close` is used by both Notifications and Alert, but in the former, `type` isn't required, and so the container [falls back](https://github.com/zendeskgarden/react-components/pull/1842/files#diff-b1e6dc5cf32c92184cebb6274c425a9e471f4e2ed272a0ea024c003eb028da0fR69-R72) to `default` and `subtle` for all its fg/bg/border colors.

So, looking at it again, I think the close icon would need a `default` case for this to be correct, rather than the context forcing the type. Hopefully that makes sense!

geotrev: I think I need to look into this more. I'm suspicious the issue is actually because Alert and Notifications share this view component, but only Alert requires type, so it being `undefined` might be a valid scenario. I'll check it out.

geotrev: OK, I confirmed that in all places where `type` is not required (e.g., subcomponents), there are [fall](https://github.com/zendeskgarden/react-components/pull/1842/files#diff-4cd6e77e1b13972c91aa6d3a6fea266c8e7f7c937204f518eeb7a7b6b9ae61f8R42) ]back](https://github.com/zendeskgarden/react-components/pull/1842/files#diff-30a97b0e497db8834422a4c3a59fd93ac89d538cb7fc5a65d34f0f9be4850c73R40). The rest of the components seem to handle the required prop appropriately.

ze-flo: Thanks for clarifying. 💯

packages/notifications/src/styled/content/StyledClose.ts
```diff
@@ -5,71 +5,79 @@
...
+        break;
+      default:
+        variable = 'foreground.subtle';
+        break;
+    }
+
+    color = getColor({ variable, theme });
+    hoverColor = getColor({ variable, light: { offset: 100 }, dark: { offset: -100 }, theme });
+    activeColor = getColor({ variable, light: { offset: 200 }, dark: { offset: -200 }, theme });
+  }
```
ze-flo, line 48: This is related to this [comment](https://github.com/zendeskgarden/react-components/pull/1842/files#r1649101016). I think we should let the `default` case handle a missing `$type` otherwise we're returning CSS with `undefined` colors. In other words, removing the `if($type)` statement.

```suggestion
switch ($type) {
  case validationTypes.warning:
    variable = 'foreground.warning';
    break;
  case validationTypes.error:
    variable = 'foreground.danger';
    break;
  case validationTypes.success:
    variable = 'foreground.success';
    break;
  default:
    variable = 'foreground.subtle';
    break;
}

color = getColor({ variable, theme });
hoverColor = getColor({ variable, light: { offset: 100 }, dark: { offset: -100 }, theme });
activeColor = getColor({ variable, light: { offset: 200 }, dark: { offset: -200 }, theme });

```

geotrev: Ah, yes, good call out!

packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts
jzempel: The direction I heard from @lucijanblagonic and @steue is that the components on the surface of a `GlobalAlert` do not change color (or transition hover/active state) differently in light & dark modes. I'm sure we want to use variables. But I wasn't expecting to see the `light` and `dark` parameters in use for `getColor` here.

geotrev: Oh, I see what you mean. The reason is that the values used in those variables between light and dark differs, so they need to be offset to meet the consistent shade in either mode. Hopefully I'm understanding correctly.

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -76,8 +61,21 @@ const colorStyles = ({ theme, $type, $isFloating }: IStyledBaseProps) => {
...
+    light: { transparency: opacity[200] },
+    dark: { transparency: opacity[1000] },
+    theme
+  });
+
+  const boxShadow = shadows.lg(offsetY, blurRadius, color);
+
   return css`
     border-color: ${borderColor};
+    box-shadow: ${$isFloating && boxShadow};
```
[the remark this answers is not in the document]
geotrev: Fixed :) [ec91f22](https://github.com/zendeskgarden/react-components/pull/1842/commits/ec91f227f486c32e35f735cad9049969b42d5f1b)

packages/notifications/src/elements/global-alert/GlobalAlert.tsx
```diff
@@ -33,16 +36,24 @@ const GlobalAlertComponent = forwardRef<HTMLDivElement, IGlobalAlertProps>(
...
     return (
-      <GlobalAlertContext.Provider value={useMemo(() => ({ type }), [type])}>
-        {/* [1] */}
-        {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
-        <StyledGlobalAlert ref={ref} role="status" $alertType={type} {...props}>
-          <StyledGlobalAlertIcon $alertType={type}>{icon}</StyledGlobalAlertIcon>
-          {props.children}
-        </StyledGlobalAlert>
-      </GlobalAlertContext.Provider>
+      <ThemeProvider theme={lightTheme}>
```
ze-flo, line 47: Fun fact: the `theme` prop of nested `ThemeProviders` have access to the parent theme. 😉 

https://styled-components.com/docs/api#themeprovider

```suggestion
      <ThemeProvider theme={theme => ({...theme, colors: { ...theme.colors, base: 'light' })}>
```

geotrev: Hey, I completely forgot about that!

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -7,72 +7,65 @@
...
+      $alertType === 'success' ? 'foreground.successEmphasis' : 'foreground.dangerEmphasis';
+    anchorHoverColor = theme.palette.white;
+    anchorActiveColor = theme.palette.white;
+  } else {
+    const borderVariable =
+      $alertType === 'warning' ? 'border.warningEmphasis' : 'border.primaryEmphasis';
+    const backgroundVariable =
+      $alertType === 'warning' ? 'background.warningEmphasis' : 'background.primaryEmphasis';
+    const foregroundVariable =
+      $alertType === 'warning' ? 'foreground.warning' : 'foreground.primary';
```
jzempel, line 57: Are these ternaries buying us much? Feels like the previous `switch` was more readable.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -7,72 +7,65 @@
...
+
+    borderColor = getColor({ variable: borderVariable, light: { offset: -300 }, theme });
+    backgroundColor = getColor({ variable: backgroundVariable, light: { offset: -400 }, theme });
+    foregroundColor = getColor({ variable: foregroundVariable, light: { offset: 100 }, theme });
+    anchorHoverColor = getColor({ variable: foregroundVariable, light: { offset: 200 }, theme });
+    anchorActiveColor = getColor({ variable: foregroundVariable, light: { offset: 300 }, theme });
```
jzempel, line 63: I think it would be better to use the `offset: xx` parameter throughout rather than `light: { offset: xx }` as the intent is to adjust theme-agnostic (which we know is now always forced to `light`).

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
```diff
@@ -119,7 +112,7 @@ const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
 export const StyledGlobalAlert = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
-})`
+})<IStyledGlobalAlertProps>`
```
jzempel, line 115: Is this type parameter required?

packages/notifications/src/styled/global-alert/StyledGlobalAlertButton.ts
```diff
@@ -6,65 +6,98 @@
...
   }
 
-  let backgroundColor;
-  let hoverBackgroundColor;
-  let activeBackgroundColor;
-  let focusColor;
+  let bgVariable;
+  let offsetOptions: OffsetOptions;
+  let offsetHoverOptions: OffsetOptions;
+  let offsetActiveOptions: OffsetOptions;
```
jzempel, line 42:
```suggestion
  let offset: number;
  let offsetHover: number;
  let offsetActive: number;
```

...again, I think just using the `getColor({ theme, offset: <number> })` is the better way to go, and easier to read/maintain.

packages/notifications/src/styled/global-alert/StyledGlobalAlertClose.ts
```diff
@@ -6,54 +6,66 @@
...
+      theme
+    });
+    activeForegroundColor = getColor({
+      variable: foregroundVariable,
+      light: { offset: 300 },
+      theme
+    });
+  } else {
+    focusVariable = 'foreground.primary';
   }
```
jzempel, line 69: ditto re: ternaries .. the previous `switch` feels more readable/straightforward to process.

packages/notifications/src/styled/global-alert/StyledGlobalAlertIcon.ts
```diff
@@ -30,6 +36,49 @@ const sizeStyles = (props: ThemeProps<DefaultTheme>) => {
...
+        variable: 'foreground.success',
+        light: { offset: -400 },
+        theme
+      });
+      break;
+    case 'error':
+      // red/300
+      color = getColor({
+        variable: 'foreground.danger',
+        light: { offset: -400 },
```
jzempel, line 58:
```suggestion
        offset: -400,
        theme
      });
      break;
    case 'error':
      // red/300
      color = getColor({
        variable: 'foreground.danger',
        offset: -400,
```

steue:
thanks for the fixes! visually looks good to me!

packages/notifications/src/styled/StyledBase.ts
```diff
@@ -6,42 +6,76 @@
...
+  const { space, shadows, opacity } = theme;
+  let bgVariable;
+  let borderVariable;
+  let fgVariable;
 
-  if (props.hue) {
-    backgroundColor = getColorV8(props.hue, 100, props.theme);
-    borderColor = getColorV8(props.hue, 300, props.theme);
-    foregroundColor = getColorV8(props.hue, props.type === 'info' ? 600 : 700, props.theme);
+  if (!$isFloating && $type && Object.keys(validationTypes).includes($type)) {
```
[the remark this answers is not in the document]
geotrev: Ah yes, the Object.keys is definitely not needed anymore.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
geotrev: Just so the `$alertType` prop is valid in the `elements` JSX side.

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
geotrev: Reverted. :)

packages/notifications/src/styled/global-alert/StyledGlobalAlert.ts
geotrev: Updated

jzempel (approved):
🙌

ze-flo (approved):
So many colors! 🤩 🟥 🟢 🟨 🟦 🌈

PR #1843, opened 2024-06-21
feat(tabs): new light/dark mode colors

ze-flo:
## Description

Adds new light/dark mode colors to `Tabs`
## Detail
- Adds new light/dark mode colors to `Tabs`
- Refactors using transient props

## Checklist

### DIFF
--- a/packages/tabs/src/styled/StyledTab.ts
+++ b/packages/tabs/src/styled/StyledTab.ts
@@ -17,14 +17,14 @@ import { stripUnit } from 'polished';
 const COMPONENT_ID = 'tabs.tab';
 
 interface IStyledTabProps {
-  isSelected?: boolean;
-  isVertical?: boolean;
+  $isSelected?: boolean;
+  $isVertical?: boolean;
 }
 
 const colorStyles = ({
   theme,
-  isSelected,
-  isVertical
+  $isSelected,
+  $isVertical
 }: IStyledTabProps & ThemeProps<DefaultTheme>) => {
   const borderColor = $isSelected ? 'currentcolor' : 'transparent';
   const borderBlockEndColor = $isVertical ? undefined : borderColor;
@@ -59,19 +59,18 @@ const colorStyles = ({
 
     &[aria-disabled='true'] {
       border-color: transparent;
-      color: ${props => getColorV8('neutralHue', 400, props.theme)};
       color: ${disabledColor};
     }
   `;
 };
 
-const sizeStyles = ({ theme, isVertical }: IStyledTabProps & ThemeProps<DefaultTheme>) => {
+const sizeStyles = ({ theme, $isVertical }: IStyledTabProps & ThemeProps<DefaultTheme>) => {
   const borderWidth = theme.borderWidths.md;
   const focusHeight = `${theme.space.base * 5}px`;
   let marginBottom;
   let padding;
 
-  if (isVertical) {
+  if ($isVertical) {
     marginBottom = `${theme.space.base * 5}px`;
     padding = `${theme.space.base}px ${theme.space.base * 2}px`;
   } else {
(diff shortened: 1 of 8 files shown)


packages/tabs/src/styled/StyledTab.ts
```diff
@@ -8,31 +8,36 @@
...
+  $isVertical
 }: IStyledTabProps & ThemeProps<DefaultTheme>) => {
-  const borderColor = isSelected ? 'currentcolor' : 'transparent';
-  const selectedColor = getColorV8('primaryHue', 600, theme);
+  const borderColor = $isSelected ? 'currentcolor' : 'transparent';
+  const borderBlockEndColor = $isVertical ? undefined : borderColor;
+  const borderInlineColor = $isVertical ? borderColor : undefined;
+
+  const selectedColor = getColor({ theme, variable: 'foreground.primary' });
+  const foregroundColor = $isSelected ? selectedColor : 'inherit';
```
ze-flo, line 34: @steue @lucijanblagonic The default foreground color was originally set to inherit the parent container's. In most cases (including the Storybook stories) that would be `foreground.default`, which has the benefit of showing a clear distinction between default and disabled tabs.

Screenshot 2024-06-20 at 4 33 14 PM
Screenshot 2024-06-20 at 4 33 02 PM

Should we keep it as such or use `foreground.subtle` as shown in Figma instead?

CC: @jzempel

geotrev: Inheriting feels OK here as long as the intent is preserved by the variable (e.g., disabled vs selected vs. default (inherited)). Hopefully I'm understanding this correctly. haha

lucijanblagonic: @ze-flo Yes, the `foreground.default` is the preferred color here. We corrected this in design, and we left a note for making the change in the inventory.

lucijanblagonic:
There seems to be a flash of black, when the selection indicator is changed. This is only visible for users who rely on keyboard to change the tab. I am assuming that the `foreground.primary` and `hover` mask the black for pointer users.

tabs-flash

image

jzempel (approved):

jzempel:
> There seems to be a flash of black, when the selection indicator is changed. This is only visible for users who rely on keyboard to change the tab. I am assuming that the `foreground.primary` and `hover` mask the black for pointer users. In dark mode, the flash is in white, `foreground.default` probably.

The only reason we're not seeing this in the [current](https://zendeskgarden.github.io/react-components/?path=/story/packages-tabs-tabs--uncontrolled) is due to the fact that the contrast between the inherited foreground and primary blue is not as high. We might need to question whether the border `transition` ease is serving us well here :shrug: (or could a very slight `transition-delay` fix things?).

lucijanblagonic (approved):
Everything else looks OK.

ze-flo:
> > There seems to be a flash of black, when the selection indicator is changed. This is only visible for users who rely on keyboard to change the tab. I am assuming that the `foreground.primary` and `hover` mask the black for pointer users. In dark mode, the flash is in white, `foreground.default` probably.
> 
> The only reason we're not seeing this in the [current](https://zendeskgarden.github.io/react-components/?path=/story/packages-tabs-tabs--uncontrolled) is due to the fact that the contrast between the inherited foreground and primary blue is not as high. We might need to question whether the border `transition` ease is serving us well here 🤷 (or could a very slight `transition-delay` fix things?).

The last [commit](https://github.com/zendeskgarden/react-components/pull/1843/commits/093f9c6f5fb8836c5506a85f7bcb20dd698ab4d5) removes the color flash when using the keyboard. ✅

jzempel (approved):
nice 🙌

geotrev (approved):

steue (approved):
looks good! ty!

PR #1852, opened 2024-07-10
feat(avatars): adds light/dark mode colors

geotrev:
## Description

Updates `Avatar` and `StatusIndicator` with new color palette.

## Detail

Also updates pattern examples to use [Menu](https://garden.zendesk.com/components/menu).

## Checklist

### DIFF
--- a/packages/avatars/demo/~patterns/stories/MenuStory.tsx
+++ b/packages/avatars/demo/~patterns/stories/MenuStory.tsx
@@ -5,93 +5,80 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import React, { useState } from 'react';
-import { Story } from '@storybook/react';
-import Icon from '@zendeskgarden/svg-icons/src/16/chevron-down-stroke.svg';
-import { PALETTE } from '@zendeskgarden/react-theming';
+import React, { useCallback, useState } from 'react';
+import { StoryFn } from '@storybook/react';
+import { getColor } from '@zendeskgarden/react-theming';
 import { Grid } from '@zendeskgarden/react-grid';
-import {
-  Dropdown,
-  Trigger,
-  Menu,
-  MediaItem,
-  MediaFigure,
-  MediaBody,
-  ItemMeta
-} from '@zendeskgarden/react-dropdowns.legacy';
-import { Button } from '@zendeskgarden/react-buttons';
-import { Avatar } from '@zendeskgarden/react-avatars';
+import { Menu, Item } from '@zendeskgarden/react-dropdowns';
+import { Avatar, IAvatarProps } from '@zendeskgarden/react-avatars';
+import { useTheme } from 'styled-components';
 
-export const MenuStory: Story = ({ isCompact }) => {
-  const [highlightedItem, setHighlightedItem] = useState<number | null>();
-  const [isOpen, setIsOpen] = useState<boolean | undefined>();
+const items: {
+  value: string;
+  label: string;
+  avatarProps: IAvatarProps;
+}[] = [
+  {
+    value: 'linden',
+    label: 'Linden',
+    avatarProps: {
+      status: 'away',
+      badge: undefined
+    }
+  },
+  {
+    value: 'reed',
+    label: 'Reed',
+    avatarProps: {
+      status: 'available',
+      badge: undefined
+    }
+  },
+  {
+    value: 'sage',
+    label: 'Sage',
+    avatarProps: {
+      status: undefined,
+      badge: '3'
+    }
+  }
+];
+
+export const MenuStory: StoryFn = ({ isCompact }) => {
+  const [highlightedValue, setHighlightedValue] = useState<string | null>();
+  const theme = useTheme();
+  const surfaceColor = getColor({ variable: 'background.raised', theme });
+  const highlightedColor = getColor({ variable: 'background.primary', theme });
+
+  const onChange = useCallback(({ focusedValue }: { focusedValue?: string | null }) => {
+    focusedValue !== undefined && setHighlightedValue(focusedValue);
+  }, []);
 
   return (
     <Grid>
       <Grid.Row style={{ height: 'calc(100vh - 80px)' }}>
         <Grid.Col textAlign="center" alignSelf="center">
-          <Dropdown
-            onStateChange={changes => {
-              setHighlightedItem(changes.highlightedIndex);
-              Object.prototype.hasOwnProperty.call(changes, 'isOpen') && setIsOpen(changes.isOpen);
-            }}
-          >
-            <Trigger>
-              <Button size={isCompact && 'small'}>
-                Demo
-                <Button.EndIcon isRotated={isOpen}>
-                  <Icon />
-                </Button.EndIcon>
-              </Button>
-            </Trigger>
-            <Menu isCompact={isCompact}>
-              <MediaItem value="linden">
-                <MediaFigure>
+          <Menu button="Demo" isCompact={isCompact} onChange={onChange}>
+            {items.map(item => (
+              <Item
+                key={item.value}
+                value={item.value}
+                icon={
                   <Avatar
                     size={isCompact ? 'extraextrasmall' : 'small'}
-                    status="away"
-                    surfaceColor={highlightedItem === 0 ? PALETTE.blue[100] : undefined}
+                    status={item.avatarProps.status}
+                    badge={item.avatarProps.badge}
+                    surfaceColor={highlightedValue === item.value ? highlightedColor : surfaceColor}
                   >
-                    <img alt="Linden" src="images/avatars/linden.png" />
+                    <img alt={item.label} src={`images/avatars/${item.value}.png`} />
                   </Avatar>
-                </MediaFigure>
-                <MediaBody>
-                  Linden
-                  <ItemMeta>linden@zendesk.garden</ItemMeta>
-                </MediaBody>
-              </MediaItem>
-              <MediaItem value="reed">
-                <MediaFigure>
-                  <Avatar
-                    size={isCompact ? 'extraextrasmall' : 'small'}
-                    status="available"
-                    surfaceColor={highlightedItem === 1 ? PALETTE.blue[100] : undefined}
-                  >
-                    <img alt="Reed" src="images/avatars/reed.png" />
-                  </Avatar>
-                </MediaFigure>
-                <MediaBody>
-                  Reed
-                  <ItemMeta>reed@zendesk.garden</ItemMeta>
-                </MediaBody>
-              </MediaItem>
-              <MediaItem value="sage">
-                <MediaFigure>
-                  <Avatar
-                    size={isCompact ? 'extraextrasmall' : 'small'}
-                    badge="3"
-                    surfaceColor={highlightedItem === 2 ? PALETTE.blue[100] : undefined}
-                  >
-                    <img alt="Sage" src="images/avatars/sage.png" />
-                  </Avatar>
-                </MediaFigure>
-                <MediaBody>
-                  Sage
-                  <ItemMeta>sage@zendesk.garden</ItemMeta>
-                </MediaBody>
-              </MediaItem>
-            </Menu>
-          </Dropdown>
+                }
+              >
+                {item.label}
+                <Item.Meta>{item.value}@zendesk.garden</Item.Meta>
+              </Item>
+            ))}
+          </Menu>
         </Grid.Col>
       </Grid.Row>
     </Grid>
--- a/packages/avatars/src/elements/Avatar.spec.tsx
+++ b/packages/avatars/src/elements/Avatar.spec.tsx
@@ -7,12 +7,10 @@
 
 import React from 'react';
 import { render, cleanup } from 'garden-test-utils';
-import { DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, PALETTE } from '@zendeskgarden/react-theming';
 import { Avatar } from './Avatar';
 
-const activeBoxShadow = DEFAULT_THEME.shadows.sm(
-  getColor({ hue: 'crimson', shade: 700, theme: DEFAULT_THEME })!
-);
+const activeBoxShadow = DEFAULT_THEME.shadows.sm(PALETTE.crimson[700]);
 
 describe('Avatar', () => {
   afterEach(cleanup);
--- a/packages/avatars/src/elements/Avatar.spec.tsx
+++ b/packages/avatars/src/elements/Avatar.spec.tsx
@@ -7,10 +7,12 @@
 
 import React from 'react';
 import { render, cleanup } from 'garden-test-utils';
-import { getColorV8, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
 import { Avatar } from './Avatar';
 
-const activeBoxShadow = DEFAULT_THEME.shadows.sm(getColorV8('crimson', 400)!);
+const activeBoxShadow = DEFAULT_THEME.shadows.sm(
+  getColor({ hue: 'crimson', shade: 700, theme: DEFAULT_THEME })!
+);
 
 describe('Avatar', () => {
   afterEach(cleanup);
--- a/packages/avatars/src/styled/StyledAvatar.spec.tsx
+++ b/packages/avatars/src/styled/StyledAvatar.spec.tsx
@@ -11,7 +11,6 @@ import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
 import { StyledAvatar } from './StyledAvatar';
 import { StyledStatusIndicator } from './StyledStatusIndicator';
-import { css } from 'styled-components';
 
 describe('StyledAvatar', () => {
   it('renders the expected element', () => {
@@ -37,7 +36,7 @@ describe('StyledAvatar', () => {
       const { container } = render(<StyledAvatar $status="away" $surfaceColor="red" />);
 
       expect(container.firstChild).toHaveStyleRule('color', 'red', {
-        modifier: css`&&`
+        modifier: '&&'
       });
     });
 
--- a/packages/avatars/src/styled/StyledAvatar.spec.tsx
+++ b/packages/avatars/src/styled/StyledAvatar.spec.tsx
@@ -11,6 +11,7 @@ import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 
 import { StyledAvatar } from './StyledAvatar';
 import { StyledStatusIndicator } from './StyledStatusIndicator';
+import { css } from 'styled-components';
 
 describe('StyledAvatar', () => {
   it('renders the expected element', () => {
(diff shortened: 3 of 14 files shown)


geotrev:
[Link to deployment](https://668eeecf185af55b8c0ad452--zendeskgarden-react-components.netlify.app/?path=/story/packages-avatars-avatar--avatar)

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
```diff
@@ -67,19 +67,24 @@ const sizeStyles = (props: IStyledStatusIndicatorProps) => {
   `;
 };
 
-const colorStyles = (props: IStyledStatusIndicatorProps) => {
-  let backgroundColor = getStatusColor(props.type, props.theme);
+const colorStyles = ({ theme, type }: IStyledStatusIndicatorProps) => {
+  const foregroundColor = getColor({
+    light: { hue: 'white' },
+    dark: { hue: 'neutralHue', shade: 1100 },
```
geotrev, line 73: Feels clunky, but it was in the design... waiting for their feedback to confirm better alternatives.

packages/avatars/demo/~patterns/stories/MenuStory.tsx
```diff
@@ -5,93 +5,80 @@
...
+      badge: '3'
+    }
+  }
+];
+
+export const MenuStory: StoryFn = ({ isCompact }) => {
+  const [highlightedValue, setHighlightedValue] = useState<string | null>();
+  const theme = useTheme();
+  const surfaceColor = getColor({ variable: 'background.raised', theme });
+  const highlightedColor = getColor({ variable: 'background.primary', theme });
```
jzempel, line 51: This doesn't appear to be correct for dark mode as the surface doesn't match the backing focus color.

packages/avatars/src/elements/Avatar.spec.tsx
```diff
@@ -7,10 +7,12 @@
 
 import React from 'react';
 import { render, cleanup } from 'garden-test-utils';
-import { getColorV8, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, getColor } from '@zendeskgarden/react-theming';
 import { Avatar } from './Avatar';
 
-const activeBoxShadow = DEFAULT_THEME.shadows.sm(getColorV8('crimson', 400)!);
+const activeBoxShadow = DEFAULT_THEME.shadows.sm(
+  getColor({ hue: 'crimson', shade: 700, theme: DEFAULT_THEME })!
```
jzempel, line 14:
```suggestion
  PALETTE.crimson[700]
```

Prefer `PALETTE` over `getColor` for testing clarity/maintenance.

packages/avatars/src/styled/StyledAvatar.spec.tsx
```diff
@@ -26,58 +27,60 @@ describe('StyledAvatar', () => {
...
   });
 
   describe('color', () => {
     it('renders surface color as expected', () => {
-      const { container } = render(<StyledAvatar status="away" surfaceColor="red" />);
+      const { container } = render(<StyledAvatar $status="away" $surfaceColor="red" />);
 
-      expect(container.firstChild).toHaveStyleRule('color', 'red');
+      expect(container.firstChild).toHaveStyleRule('color', 'red', {
+        modifier: css`&&`
```
jzempel, line 40: Interesting .. does `'&&'` not work here?

packages/avatars/src/styled/StyledStatusIndicator.spec.tsx
```diff
@@ -7,7 +7,7 @@
 
 import React from 'react';
 import { render } from 'garden-test-utils';
-import { getColorV8, DEFAULT_THEME } from '@zendeskgarden/react-theming';
+import { getColor, DEFAULT_THEME } from '@zendeskgarden/react-theming';
```
jzempel, line 10: Prefer `PALETTE` throughout for testing.

packages/avatars/src/styled/utility.ts
```diff
@@ -16,40 +16,40 @@ const [xxs, xs, s, m, l] = SIZE;
...
+  readonly $type?: IAvatarProps['status'] | 'active';
 }
 
 export function getStatusColor(
-  type?: IStyledStatusIndicatorProps['type'],
-  theme?: IStyledStatusIndicatorProps['theme']
+  theme: IStyledStatusIndicatorProps['theme'],
+  type?: IStyledStatusIndicatorProps['$type']
 ): string {
   switch (type) {
```
jzempel, line 27: Should this `switch` lookup be replaced with a mapping object?

packages/avatars/src/styled/StyledStatusIndicator.ts
```diff
@@ -60,22 +62,26 @@ const sizeStyles = (props: IStatusIndicatorProps & ThemeProps<DefaultTheme>) =>
...
+    $surfaceColor ||
+      ($type
+        ? getColor({ variable: 'background.default', theme })
         : (theme.palette.white as string))
   );
 
-  if (size === xxs) {
+  if ($size === xxs) {
     boxShadow = boxShadow.replace(theme.shadowWidths.sm, '1px');
   }
```
jzempel, line 81: Potential performance improvement by setting `let boxShadow;` undefined and moving the `theme.shadows.sm` to a conditional `else` clause here.

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
```diff
@@ -67,19 +67,20 @@ const sizeStyles = (props: IStyledStatusIndicatorProps) => {
   `;
 };
 
-const colorStyles = (props: IStyledStatusIndicatorProps) => {
-  let backgroundColor = getStatusColor(props.type, props.theme);
+const colorStyles = ({ theme, $type }: IStyledStatusIndicatorProps) => {
+  const foregroundColor = getColor({ variable: 'foreground.onEmphasis', theme });
+  let backgroundColor = getStatusColor(theme, $type);
   let borderColor = backgroundColor;
```
jzempel, line 73:
```suggestion
  let backgroundColor;
  let borderColor;
```

...and move assignments to a conditional `else` clause below.

jzempel (changes requested):

packages/avatars/demo/~patterns/stories/MenuStory.tsx
geotrev: I used this color because any color with opacity won't look correct as a `surfaceColor` (it will just look like no surface color is applied). Attaching a screenshot of how the color looks if I match the focused item background exactly. `background.primary` was the closest color given the limitaiton. Unless you see another option that I missed?

PS. Ignore the spacing bug, fix is WIP.

Screenshot 2024-07-12 at 11 48 20 AM

geotrev: I used this color because any color with opacity won't look correct as a `surfaceColor` (it will just look like no surface color is applied). Attaching a screenshot of how the color looks if I match the focused item background exactly. `background.primary` was the closest color given the limitaiton. Unless you see another option that I missed?

PS. Ignore the spacing bug, fix is WIP.

Screenshot 2024-07-12 at 11 48 20 AM

geotrev: I used this color because any color with opacity won't look correct as a `surfaceColor` (it will just look like no surface color is applied). Attaching a screenshot of how the color looks if I match the focused item background exactly. `background.primary` was the closest color given the limitaiton. Unless you see another option that I missed?

PS. Ignore the spacing bug, fix is WIP.

Screenshot 2024-07-12 at 11 48 20 AM

Screenshot 2024-07-12 at 11 51 11 AM

packages/avatars/src/styled/StyledAvatar.spec.tsx
geotrev: Ah, it does. I had assumed it wouldn't given the transform provided by the `css` function.

packages/avatars/src/styled/utility.ts
geotrev: Do you mean something more like this?

```
export function getStatusColor(
  theme: IStyledStatusIndicatorProps['theme'],
  type?: IStyledStatusIndicatorProps['$type']
): string {
  return (
    {
      active: getColor({ hue: 'crimson', light: { shade: 700 }, dark: { shade: 600 }, theme }),
      available: getColor({ hue: 'mint', light: { shade: 500 }, dark: { shade: 400 }, theme }),
      away: getColor({ hue: 'orange', light: { shade: 500 }, dark: { shade: 400 }, theme }),
      transfers: getColor({ hue: 'azure', light: { shade: 500 }, dark: { shade: 400 }, theme }),
      offline: getColor({ hue: 'grey', light: { shade: 500 }, dark: { shade: 400 }, theme })
    }[type as string] || 'transparent'
  );
}
```

packages/avatars/src/styled/StyledStatusIndicator.ts
geotrev: Not sure I'm seeing the gain. We calculate the box shadow either way, don't we? 🤔

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
geotrev: This feels like a nit, but I guess I'm not seeing a gain in clarity with moving the assignment down a few lines when the variables can only be one of two values anyway?

packages/avatars/src/styled/StyledStatusIndicator.ts
geotrev: On second thought, I'm not actually entirely sure what you mean by moving the `theme.shadows.sm` to an `else` clause. Can you elaborate?

packages/avatars/demo/~patterns/stories/MenuStory.tsx
jzempel: Ok, it's good to know that this is the best we can do under the circumstances.

packages/avatars/src/styled/StyledStatusIndicator.ts
jzempel: Ah, I missed the `replace`. That's pretty odd treatment for our typical goto. I think the code should follow normative Garden conventions and look more like this...

```
const boxShadowColor = surfaceColor || getColor({ theme, variable: 'background.default' })
let boxShadow;

if (size === xxs) {
  boxShadow = theme.shadows.xs(boxShadowColor);
} else {
  boxShadow = theme.shadows.sm(boxShadowColor);
}
```

Stepping away from the theme like the current code does isn't ideal.

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
jzempel: Could be a [nit], but in practice Garden tends to micro-optimize by deferring conditional assignment (and associated function call expense) until they are determined to be necessary. More like...

```
let backgroundColor;
let borderColor;

if ($type === 'offline') {
  backgroundColor = ...;
  borderColor = ...;
} else {
  backgroundColor = ...;
  borderColor = ...;
}
```

packages/avatars/src/styled/StyledStatusIndicator.ts
jzempel: Either that or make the `boxShadow` a simple ternary...

```
const boxShadow = size === xxs ? theme.shadows.xs(boxShadowColor) : theme.shadows.sm(boxShadowColor);
```

packages/avatars/src/styled/utility.ts
```diff
@@ -16,40 +16,35 @@ const [xxs, xs, s, m, l] = SIZE;
...
-      return getColorV8('mint', 400, theme)!;
-    case 'away':
-      return getColorV8('orange', 400, theme)!;
-    case 'transfers':
-      return getColorV8('azure', 400, theme)!;
-    case 'offline':
-      return getColorV8('grey', 500, theme)!;
-    default:
-      return 'transparent';
-  }
```
ze-flo, line 40: Def a [nit] and related to the micro-optimization comment above, the switch statement is slightly more performant then calculating all colors then picking the right one.

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
geotrev: I guess that's the nit I'm trying to understand. Whether the value is assigned at the initial assignment aka default value (`let thing = value`) or at the `else` (which is acting as the default value), this feels like a stylistic preference, right? I just want to make sure I understand that that's what we're talking about, because I'm not seeing any real optimization in this scenario. 🙂

packages/avatars/src/styled/StyledStatusIndicator.ts
geotrev: That makes total sense! Thanks for clarifying.

steue (approved):
visually looks good to me! thnx for working on this so meticulously! 🔥 🚀

packages/avatars/src/styled/StyledStatusIndicatorBase.ts
geotrev: Okay, I think I understand now after looking at it for a minute. I didn't see initially that `getColor` would be called extraneously in current scenario. I'll update and push it up!

packages/avatars/src/styled/utility.ts
geotrev: Let me know if I captured the intent of this optimization in my latest commit. I think I did but after reading your comment I'm not entirely sure.

jzempel: Yeah, I was more thinking the map would live in a one-time constant. And the function call would be more along the lines of https://github.com/zendeskgarden/react-components/blob/816c29f24a9353b2abc320c0c1019db83fa1b018/packages/chrome/src/styled/utils.ts#L24-L25

jzempel (approved):
All good on my end, pending resolution to @ze-flo's optimization comment

packages/avatars/src/styled/utility.ts
ze-flo: The one time constant would def work. 

What I wanted to emphasize is that when using a map 5 values are calculated every time `getStatusColor` is called but only one color is returned. With a switch statement, however, there's one calculation for one returned color. It's more efficient.

packages/avatars/src/styled/utility.ts
```diff
@@ -16,40 +16,35 @@ const [xxs, xs, s, m, l] = SIZE;
...
+  return (
+    {
+      active: getColor({ hue: 'crimson', light: { shade: 700 }, dark: { shade: 600 }, theme }),
+      available: getColor({ hue: 'mint', light: { shade: 500 }, dark: { shade: 400 }, theme }),
+      away: getColor({ hue: 'orange', light: { shade: 500 }, dark: { shade: 400 }, theme }),
+      transfers: getColor({ hue: 'azure', light: { shade: 500 }, dark: { shade: 400 }, theme }),
+      offline: getColor({ hue: 'grey', light: { shade: 500 }, dark: { shade: 400 }, theme })
+    }[type as string] || 'transparent'
+  );
```
ze-flo, line 35:
```suggestion

  switch (type) {
    case 'active':
      return getColor({ hue: 'crimson', light: { shade: 700 }, dark: { shade: 600 }, theme });
    case 'available':
      return getColor({ hue: 'mint', light: { shade: 500 }, dark: { shade: 400 }, theme });
    case 'away':
      return getColor({ hue: 'mint', light: { shade: 500 }, dark: { shade: 400 }, theme });
    case 'transfers':
      return getColor({ hue: 'orange', light: { shade: 500 }, dark: { shade: 400 }, theme });
    case 'offline':
      return getColor({ hue: 'grey', light: { shade: 500 }, dark: { shade: 400 }, theme });
    default:
      return 'transparent';
  }
  
```

packages/avatars/src/styled/utility.ts
geotrev: Can you be more specific @jzempel ? These functions feel very different to me.

packages/avatars/src/styled/utility.ts
geotrev: Ah, that's what I had before.

I feel like there are two different trains of thought here. I ended up doing something like this: 403bf68d9c3fd10076501bf31c9501cc99c84de2

packages/avatars/src/styled/utility.ts
geotrev: This was my take on it: 403bf68d9c3fd10076501bf31c9501cc99c84de2

Thoughts?

ze-flo (approved):
Very nice! 🚀 :shipit:

packages/avatars/src/styled/utility.ts
ze-flo: I dig it! 💯

jzempel: Hey @geotrev maybe something like this?

```ts
const STATUS_COLOR_OPTIONS: Record<
  NonNullable<IStyledStatusIndicatorProps['$type']>,
  Omit<ColorParameters, 'theme'>
> = {
  active: { hue: 'crimson', light: { shade: 700 }, dark: { shade: 600 } },
  available: { hue: 'mint', light: { shade: 500 }, dark: { shade: 400 } },
  away: { hue: 'orange', light: { shade: 500 }, dark: { shade: 400 } },
  transfers: { hue: 'azure', light: { shade: 500 }, dark: { shade: 400 } },
  offline: { hue: 'grey', light: { shade: 500 }, dark: { shade: 400 } }
};

export function getStatusColor(
  theme: IStyledStatusIndicatorProps['theme'],
  type?: IStyledStatusIndicatorProps['$type']
): string {
  return type ? getColor({ ...STATUS_COLOR_OPTIONS[type], theme }) || 'transparent' : 'transparent';
}
```

jzempel: Haha, looks like we arrived at a similar result :wink:

geotrev: Two peas 😎

packages/avatars/src/styled/utility.ts
```diff
@@ -16,40 +16,41 @@ const [xxs, xs, s, m, l] = SIZE;
...
-      return getColorV8('grey', 500, theme)!;
-    default:
-      return 'transparent';
+  if (type === undefined) {
+    return 'transparent';
   }
+
+  const colorArgs = StatusColorParams[type];
+
+  return getColor({ ...colorArgs, theme });
```
jzempel, line 41:
```suggestion
  return getColor({ ...colorArgs, theme }) || 'transparent';
```

The only thing is that there might be an unprotected path where a consumer provides a rogue status value :shrug:. Might not be worth sealing off.

packages/avatars/src/styled/utility.ts
geotrev: I like retaining the `'transparent'` fallback in case the color enum lookup returns `undefined`, so I added that in. :)

packages/avatars/src/styled/utility.ts
geotrev: You beat me to it. Just pushed that up. I did something similar, but I think both could be valid. I'm OK switching to you what here if you think that's worth it.

jzempel (approved):
i dig it :tada:

PR #1860, opened 2024-07-16
feat(datepickers): adds light/dark mode colors

geotrev:
## Description

Applies new colors to all date picker components + transient props.

Screenshot 2024-07-17 at 4 01 37 PM

Screenshot 2024-07-17 at 4 02 02 PM

## Checklist


ze-flo:
The calendar grid background doesn't match Figma's. Is that intended?

### Storybook
Screenshot 2024-07-16 at 11 33 34 AM

### Figma
Screenshot 2024-07-16 at 11 34 11 AM

geotrev:
@ze-flo good catch. I believe it should be `background.raised`. Thanks!

ze-flo:
> @ze-flo good catch. I believe it should be `background.raised`. Thanks!

Might be a design question... Would a `transparent` background give us more flexibility?

geotrev:
> > @ze-flo good catch. I believe it should be `background.raised`. Thanks!
> 
> 
> 
> Might be a design question... Would a `transparent` background give us more flexibility?

Hah, I was thinking exactly that earlier. V8 DateRangePicker uses the old `background` color key. We could retain this but I'll ask about transparency just to confirm.

packages/datepickers/src/styled/StyledDay.ts
```diff
@@ -6,76 +6,106 @@
...
   return css`
     background-color: ${backgroundColor};
-    color: ${color};
+    color: ${foreground};
 
-    ${!isSelected &&
-    !isDisabled &&
-    `
+    ${!$isSelected &&
+    !$isDisabled &&
```
jzempel, line 77: Pure CSS is definitely preferred here. We should be able to query against

```
:not([aria-selected]),
:not([aria-disabled]) { ...
```

packages/datepickers/src/elements/DatePicker/components/Calendar.tsx
```diff
@@ -94,13 +94,13 @@ export const Calendar = forwardRef<HTMLDivElement, ICalendarProps>(
...
           <StyledDay
-            isToday={isCurrentDate}
-            isPreviousMonth={isPreviousMonth}
-            isSelected={isSelected}
-            isDisabled={isDisabled}
-            isCompact={isCompact!}
+            $isToday={isCurrentDate}
+            $isPreviousMonth={isPreviousMonth}
+            $isSelected={isSelected}
+            $isDisabled={isDisabled}
```
jzempel, line 102:
```suggestion
            [aria-selected]={isSelected || undefined}
            [aria-disabled]={isDisabled || undefined}
```

See https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/. Even though we won't be a11y compliant here, this will give us better CSS query support.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -6,37 +6,63 @@
...
-  isCompact,
-  theme
-}: { isCompact?: boolean } & ThemeProps<DefaultTheme>) => {
+const sizeStyles = ({ $isCompact, theme }: { $isCompact?: boolean } & ThemeProps<DefaultTheme>) => {
   let size = theme.space.base * 10;
 
-  if (isCompact) {
+  if ($isCompact) {
     size = theme.space.base * 8;
   }
```
jzempel, line 16:
```suggestion
  const size = theme.space.base * ($isCompact ? 8 : 10);
```

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -6,37 +6,63 @@
...
 
   return css`
     width: ${size}px;
     height: ${size}px;
+
+    svg {
+      width: ${props => `${props.theme.iconSizes.md}`};
+      height: ${props => `${props.theme.iconSizes.md}`};
```
jzempel, line 24: Move to a local `iconSize` variable rather than invoking the `props =>` function.

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -45,25 +71,19 @@ const COMPONENT_ID = 'datepickers.header_paddle';
 export const StyledHeaderPaddle = styled.div.attrs({
   'data-garden-id': COMPONENT_ID
 })<{
-  isCompact: boolean;
-  isHidden?: boolean;
+  $isCompact: boolean;
+  $isHidden?: boolean;
```
jzempel, line 75: This inherited code feels a bit lazy. Can we declare a `IStyledHeaderPaddle` interface, consistent with other view components?

packages/datepickers/src/styled/StyledHeaderPaddle.ts
```diff
@@ -45,25 +71,19 @@ const COMPONENT_ID = 'datepickers.header_paddle';
...
 }>`
   display: flex;
   align-items: center;
   justify-content: center;
   transform: ${props => props.theme.rtl && 'rotate(180deg)'};
-  visibility: ${props => props.isHidden && 'hidden'};
+  visibility: ${props => props.$isHidden && 'hidden'};
```
jzempel, line 81: This smells like it should be `[aria-hidden]` CSS rather than passing in a `$isHidden` prop.

packages/datepickers/src/styled/StyledMenuWrapper.ts
```diff
@@ -17,28 +17,28 @@ import { Placement } from '@floating-ui/react-dom';
 const COMPONENT_ID = 'datepickers.menu_wrapper';
 
 interface IStyledMenuWrapperProps {
-  isHidden?: boolean;
-  isAnimated?: boolean;
-  zIndex?: number;
-  placement: Placement;
+  $isHidden?: boolean;
```
jzempel, line 20: Can we use `[aria-hidden]` instead?

jzempel (changes requested):

packages/datepickers/src/styled/StyledDay.ts
geotrev: I'm not seeing `aria-selected` used on this component. Are you suggesting I add it / replace `$isSelected` with `aria-selected`? 👀

packages/datepickers/src/elements/DatePicker/components/Calendar.tsx
geotrev: Since the calendar isn't keyboard accessible in any form, I'm conflicted on if a11y updates like this give us any real value. Thoughts?

packages/datepickers/src/styled/StyledMenuWrapper.ts
geotrev: Indeed we can

packages/datepickers/src/elements/DatePickerRange/components/Month.tsx
```diff
@@ -121,11 +121,11 @@ export const Month = forwardRef<HTMLDivElement, IMonthProps>(
...
         return (
-          <StyledCalendarItem key={`day-${itemsIndex}`} isCompact={isCompact}>
+          <StyledCalendarItem key={`day-${itemsIndex}`} $isCompact={isCompact}>
             <StyledDay
-              isCompact={isCompact!}
-              isPreviousMonth
-              isDisabled
+              $isCompact={isCompact!}
+              $isPreviousMonth
+              aria-disabled
```
jzempel, line 128: Does this render `aria-disabled="true"` as expected?

jzempel (approved):

packages/datepickers/src/elements/DatePickerRange/components/Month.tsx
geotrev: It does. I tested the result in Storybook, and also see these propagate in tests as expected. If being more explicit is preferred (`"true"` / `"false"` / `undefined`), we can do that instead though.

ze-flo (approved):
Great work! 💯 🔥

steue:
> > > @ze-flo good catch. I believe it should be `background.raised`. Thanks!
> > 
> > 
> > Might be a design question... Would a `transparent` background give us more flexibility?
> 
> Hah, I was thinking exactly that earlier. V8 DateRangePicker uses the old `background` color key. We could retain this but I'll ask about transparency just to confirm.

great call out! this makes sense! i think this is similar to our approach to tables, correct? dropping the opaque background to now transparency to accommodate the raised treatment.

steue:
noticed something with the hover state when a date is set to `today` in dark theme. 
Export-1721418551014
the hover treatment overlays the text. is there a way we can make sure the `today` fg treatment supersedes the hover so it maintains the bright treatment?
image

sieeeeeenna:
other than what steph noted, everything looks good to me! and agree with the transparent background, that aligns with updates we're making on the design side like Steph said

geotrev:
@steue Just pushed up the fix. Good catch. :)

geotrev:
Note for visibility: the package-lock had some issues after a rebase, so I checked out the current lock on `next` and reinstalled for this branch.

geotrev:
Pushed up a minor package lock fix from the rebase. Checked out the lock file from `next` and rebuilt from there.

steue:
noice! ty for this!

PR #1862, opened 2024-07-17
feat(colorpickers): adds light/dark mode colors

ze-flo:
## Description

Adds dark / light mode support to `@zendeskgarden/react-colorpickers`

## Detail
- Adds dark / light mode support to `ColorPicker`, `ColorPickerDialog`, `ColorSwatch`, and `ColorSwatchDialog`
- Simplify CSS with `isOpaque` option
- Use transient props to avoid invalid DOM attributes

Screenshot 2024-07-17 at 7 39 56 AM

Screenshot 2024-07-17 at 7 40 29 AM
Screenshot 2024-07-17 at 7 40 44 AM

## Checklist

### DIFF
--- a/packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
+++ b/packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
@@ -8,7 +8,7 @@
 import styled, { DefaultTheme, ThemeProps } from 'styled-components';
 import { DEFAULT_THEME, getCheckeredBackground } from '@zendeskgarden/react-theming';
 import { getTrackHeight, getTrackMargin, StyledRange } from '../common/StyledRange';
-import { IRGBColorProps } from '../../types';
+import { IRGBColorProps } from '../types';
 
 const COMPONENT_ID = 'colorpickers.colorpicker_alpha';
 
--- a/packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
+++ b/packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
@@ -8,14 +8,14 @@
 import styled, { DefaultTheme, ThemeProps } from 'styled-components';
 import { DEFAULT_THEME, getCheckeredBackground } from '@zendeskgarden/react-theming';
 import { getTrackHeight, getTrackMargin, StyledRange } from '../common/StyledRange';
-import { IRGBColor } from '../../types';
+import { IRGBColorProps } from '../../types';
 
 const COMPONENT_ID = 'colorpickers.colorpicker_alpha';
 
-const background = (props: IRGBColor & ThemeProps<DefaultTheme>) => {
+const background = (props: IRGBColorProps & ThemeProps<DefaultTheme>) => {
   const direction = `to ${props.theme.rtl ? 'left' : 'right'}`;
-  const fromColor = `rgba(${props.red}, ${props.green}, ${props.blue}, 0)`;
-  const toColor = `rgb(${props.red}, ${props.green}, ${props.blue})`;
+  const fromColor = `rgba(${props.$red}, ${props.$green}, ${props.$blue}, 0)`;
+  const toColor = `rgb(${props.$red}, ${props.$green}, ${props.$blue})`;
   const positionY = getTrackMargin(props);
   const height = getTrackHeight(props);
   const overlay = `linear-gradient(${direction}, ${fromColor}, ${toColor}) 0 ${positionY}px / 100% ${height}px no-repeat`;
@@ -29,14 +29,14 @@ const background = (props: IRGBColor & ThemeProps<DefaultTheme>) => {
   });
 };
 
-export const StyledAlphaRange = styled(StyledRange as 'input').attrs<IRGBColor>(props => ({
+export const StyledAlphaRange = styled(StyledRange as 'input').attrs<IRGBColorProps>(props => ({
   style: {
     backgroundSize: 'auto',
     background: background(props)
   },
   'data-garden-id': COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION
-}))<IRGBColor>`
+}))<IRGBColorProps>`
   /* stylelint-disable no-empty-source */
 `;
 
--- a/packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
+++ b/packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
@@ -5,14 +5,14 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { getTrackHeight, getTrackMargin } from '../common/StyledRange';
 
 const COMPONENT_ID = 'colorpickers.colorpicker_sliders';
 
 interface IStyledSlidersProps {
-  isOpaque?: boolean;
+  $isOpaque?: boolean;
 }
 
 export const StyledSliders = styled.div.attrs({
@@ -24,19 +24,24 @@ export const StyledSliders = styled.div.attrs({
   margin-${props => (props.theme.rtl ? 'right' : 'left')}: ${props => props.theme.space.base * 2}px;
   width: 100%;
 
-  & > * {
-    position: absolute;
-    width: 100%;
-    height: ${props => getTrackMargin(props) * 2 + getTrackHeight(props)}px;
-  }
+  ${props =>
+    !props.$isOpaque &&
+    css`
+      & > * {
+        position: absolute;
+        width: 100%;
+        height: ${getTrackMargin(props) * 2 + getTrackHeight(props)}px;
+      }
 
-  & > :first-child {
-    top: -${getTrackMargin}px;
-  }
+      & > :first-child {
+        top: -${getTrackMargin}px;
+      }
 
-  & > :last-child {
-    bottom: -${getTrackMargin}px;
-  }
+      & > :last-child {
+        bottom: -${getTrackMargin}px;
+      }
+    `}
+  
 
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
--- a/packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
+++ b/packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
@@ -5,7 +5,7 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled, { css } from 'styled-components';
+import styled, { DefaultTheme, ThemeProps } from 'styled-components';
 import { retrieveComponentStyles, DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { getTrackHeight, getTrackMargin } from '../common/StyledRange';
 
@@ -14,6 +14,29 @@ const COMPONENT_ID = 'colorpickers.colorpicker_sliders';
 interface IStyledSlidersProps {
   $isOpaque?: boolean;
 }
+
+const sizeStyles = (props: IStyledSlidersProps & ThemeProps<DefaultTheme>) => {
+  if (props.$isOpaque) return undefined;
+
+  const trackHeight = getTrackHeight(props);
+  const trackMargin = getTrackMargin(props);
+
+  return `
+    & > * {
+      position: absolute;
+      width: 100%;
+      height: ${trackMargin * 2 + trackHeight}px;
+    }
+
+    & > :first-child {
+      top: -${trackMargin}px;
+    }
+
+    & > :last-child {
+      bottom: -${trackMargin}px;
+    }
+  `;
+};
 
 export const StyledSliders = styled.div.attrs({
   'data-garden-id': COMPONENT_ID,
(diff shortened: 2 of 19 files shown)


packages/colorpickers/src/styled/common/StyledRange.ts
```diff
@@ -60,24 +60,38 @@ const trackLowerStyles = (styles: string, modifier = '') => {
...
+  const thumbActiveBorderColor = getColor({
+    theme,
+    variable: 'border.primaryEmphasis'
+  });
+  const thumbFocusBorderColor = thumbActiveBorderColor;
+  const thumbHoverBackgroundColor = getColor({ theme, variable: 'background.subtle' });
   const thumbHoverBorderColor = thumbHoverBackgroundColor;
 
-  return `
-    border-color: ${props.isOpaque && getColorV8('background', 600 /* default shade */, props.theme)};
```
ze-flo, line 73: With the `isOpaque` option, the border was only used for positioning. The `border-color` must then match with the background color of the the container the ColorPicker with `isOpaque` renders into. That can be problematic in some cases.

Screenshot 2024-07-17 at 7 21 54 AM

So I refactored the CSS to remove it.

geotrev: If this ends up having no issues with the final interaction/element then sounds good 👍

packages/colorpickers/src/types/index.ts
```diff
@@ -39,6 +39,13 @@ export interface IRGBColor {
   alpha?: number;
 }
 
+export interface IRGBColorProps {
+  $red: number;
+  $green: number;
+  $blue: number;
+  $alpha?: number;
+}
```
geotrev, line 47: 🔥

geotrev (approved):
Looking good!

packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
```diff
@@ -29,14 +29,14 @@ const background = (props: IRGBColor & ThemeProps<DefaultTheme>) => {
   });
 };
 
-export const StyledAlphaRange = styled(StyledRange as 'input').attrs<IRGBColor>(props => ({
+export const StyledAlphaRange = styled(StyledRange as 'input').attrs<IRGBColorProps>(props => ({
   style: {
     backgroundSize: 'auto',
     background: background(props)
   },
```
geotrev, line 36: Out of curiosity, why can't this live in the style block below?

ze-flo: I think this approach avoids dynamically creating a new class every time the color props change. `styled-component` will warn you when that happens.

packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
```diff
@@ -24,19 +24,24 @@ export const StyledSliders = styled.div.attrs({
...
-  & > :first-child {
-    top: -${getTrackMargin}px;
-  }
-
-  & > :last-child {
-    bottom: -${getTrackMargin}px;
-  }
+  ${props =>
+    !props.$isOpaque &&
+    css`
```
jzempel, line 29: Is `css` necessary here?

packages/colorpickers/src/styled/ColorPicker/StyledAlphaRange.ts
geotrev: 💀

packages/colorpickers/src/styled/common/StyledRange.ts
```diff
@@ -60,24 +60,38 @@ const trackLowerStyles = (styles: string, modifier = '') => {
   `;
 };
 
-const colorStyles = (props: IStyledRangeProps & ThemeProps<DefaultTheme>) => {
-  const thumbBackgroundColor = getColorV8('neutralHue', 100, props.theme);
+const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
+  const thumbBackgroundColor = getColor({ theme, variable: 'background.default' });
```
jzempel, line 64: @lucijanblagonic maybe just me, but I'm finding these dark mode thumbs very difficult to see. Somehow I don't have the same problem with shadowed white over white background.

packages/colorpickers/src/types/index.ts
jzempel, line 47: Can this move to a `types.ts` file within the `styled` directory? We just wouldn't want to mistake the fact that this interface should never be publicly exported.

jzempel (commented):
[nit] should the data for the `ColorSwatch` demos be updated to show all 12 colors from the primary (blue, green, red, yellow) palette?

packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
ze-flo: This provides proper placement for the sliders when `isOpaque` is `false`.

jzempel: yep, I get that. More wondering whether this can be a simple template literal rather than invoking the `css` function?

ze-flo: I tried the template literal first but it didn't parse correctly. I'll give it another shot.

jzempel: ...as I understand it, `styled` already handles this, making `css` redundant...

> during flattening, styled-components ignores interpolations that evaluate to undefined, null, false, or an empty string ("")

https://styled-components.com/docs/advanced#tagged-template-literals

packages/colorpickers/src/styled/common/StyledRange.ts
geotrev: +1, feels like an accessibility concern for sure.

ze-flo: @steue @lucijanblagonic For clarity, we're referring to this:

Screenshot 2024-07-17 at 7 40 13 AM

steue: great observation yall! curious about your thoughts of bumping up the contrast by setting the thumbs to `bg/raised` with a -100 offset? (looks brighter)
image

jzempel: > curious about your thoughts of bumping up the contrast by setting the thumbs to bg/raised with a -100 offset? (looks brighter)

@steue for me, that color tweak works great 🙌

steue (approved):
sans the `thumb` treatment, everything looks good to me! great work!

packages/colorpickers/src/styled/ColorPicker/StyledSliders.ts
```diff
@@ -15,6 +15,29 @@ interface IStyledSlidersProps {
   $isOpaque?: boolean;
 }
 
+const sizeStyles = (props: IStyledSlidersProps & ThemeProps<DefaultTheme>) => {
+  if (props.$isOpaque) return undefined;
```
jzempel, line 19:
```suggestion
  if (props.$isOpaque) {
    return undefined;
  }
```

Conditions in Garden should always be blocks. We used to lint for this, but [the rule](https://eslint.org/docs/latest/rules/brace-style) was deprecated. Probably need to bring it back with a plugin since prettier doesn't cover us.

jzempel (changes requested):

packages/colorpickers/src/styled/common/StyledRange.ts
ze-flo: Updated. ✅ 

Screenshot 2024-07-19 at 1 29 57 PM
Screenshot 2024-07-19 at 1 29 44 PM

lucijanblagonic:
The changes here look good, thanks @steue for jumping in. The only thing that seems off is the hover behavior of the handle. I would expect the hover to lighten on hover (in dark mode), to be in line with other interactive controls. What do you think?

steue:
> The changes here look good, thanks @steue for jumping in. The only thing that seems off is the hover behavior of the handle. I would expect the hover to lighten on hover (in dark mode), to be in line with other interactive controls. What do you think?

agreed!

ze-flo:
> > The changes here look good, thanks @steue for jumping in. The only thing that seems off is the hover behavior of the handle. I would expect the hover to lighten on hover (in dark mode), to be in line with other interactive controls. What do you think?
> 
> agreed!

Great! In dark mode, the slider's thumb is using `background.raised` with an `offset: -100` for its default state. @steue What offset would you recommend for the hover state?

geotrev:
Not sure the best place to put this so I'll throw it here. 😅 

I'm still worried about this contrast. Even with the box shadow, the thumb blends in with the background especially at the edges, which is the default state for this component.

Screenshot 2024-07-22 at 11 51 12 AM

Screenshot 2024-07-22 at 12 03 29 PM

I'm not sure what the best approach is (border, or maybe flipping the thumb to be white for dark mode and dark grey for light mode). Maybe this isn't a problem we can solve in this PR, which is also fair. But wanted to at least put it our there for posterity.

I also asked Jen over in accessibility and she agreed, FWIW.

steue:
> What offset would you recommend for the hover state?

i would imagine following our default on hover, so another -100 (effectively -200)?

steue:
@geotrev valid concern! color pickers have interesting a11y considerations as colors themselves pose inherent issues. we can get into the nitty gritty of it, but where is the contrast compared against? is it against the colors in the slider itself or the BG? you're right, though the the thumb might not always achieve the 3:1 contrast ratio. 

that said, the overall design does incorporate alternative methods for controlling color that are accessible. (i.e. thumbs / sliders controllable via keyboard and input fields to control HSL/RGB etc) i do think we could explore treatment as another treatment

ze-flo:
> > What offset would you recommend for the hover state?
> 
> i would imagine following our default on hover, so another -100 (effectively -200)?


Following @lucijanblagonic 's [comment](https://github.com/zendeskgarden/react-components/pull/1862#issuecomment-2243076642):
> I would expect the hover to lighten on hover (in dark mode)

@steue If the default state is `background.raised` with an offset: -100, should we increase the offset by 100 (essentially just `background.raised`) for hover instead?

geotrev:
> @geotrev valid concern! color pickers have interesting a11y considerations as colors themselves pose inherent issues. we can get into the nitty gritty of it, but where is the contrast compared against? is it against the colors in the slider itself or the BG? you're right, though the the thumb might not always achieve the 3:1 contrast ratio.

The contrast is with everything behind it, but primarily the dark grey background. Currently the background and the thumb have less than a 2:1 contrast ratio, which is concerning in the cases I mentioned. Especially since we can't depend on the thumb being positioned inside the slider at all times.

The issue of the contrast not always being perfect could be a good reason to use a border. A single solid color will present persistence issues like you mention.

> that said, the overall design does incorporate alternative methods for controlling color that are accessible. (i.e. thumbs / sliders controllable via keyboard and input fields to control HSL/RGB etc) i do think we could explore treatment as another treatment

That is true. There are input fields as well. I bring it up as a medium-high concern mainly because the bulk of the picker UI contains visual controls.

lucijanblagonic:
I don't think there is a difference in light or dark mode when it comes to accessibility in this case. Having a `thumb` in the color where it blends with the background shares the same challenge in light and dark. Here is an example with the `thumb` changes in the slider and in the color well. 

* Thumb in color well has a white `2px` inner border and `1px` outer black border. 
* Thumb in the slider has `background/default` and then `border/input` applied. That way the stroke helps to distinguish the handle a bit.

image

geotrev:
That makes sense and would definitely remediate the concerns. Not a blocker for this PR of course, we can set aside time to do it after.

ze-flo:
@lucijanblagonic @steue The latest [preview](https://66a06ae7b6b7428eb06a618e--zendeskgarden-react-components.netlify.app/?path=/story/packages-colorpickers-colorpicker--uncontrolled) increases the thumbs contrast based on the last suggestion. The border treatment of the range thumbs matches with the inputs' below.

Screenshot 2024-07-23 at 4 34 56 PM
Screenshot 2024-07-23 at 4 34 34 PM

I think this looks good. Hope this works for all too! 🤞

packages/colorpickers/src/styled/common/StyledRange.ts
```diff
@@ -60,24 +60,51 @@ const trackLowerStyles = (styles: string, modifier = '') => {
...
+  const thumbBackgroundColor = getColor({
+    theme,
+    variable: 'background.default'
+  });
+  const thumbBorderColor = getColor({
+    theme,
+    variable: 'border.default',
+    dark: { offset: -100 },
+    light: { offset: 100 }
+  });
```
jzempel, line 73: [nit] now that the thumbs have a visible border, can we check that the `transition` animation is updated to include the border?

lucijanblagonic (approved):
Tested and looks good, feel free to merge after resolving @jzempel 's comment for transition.

packages/colorpickers/src/styled/common/StyledRange.ts
ze-flo: Transition animations appear to be inherited from `Range`:

https://github.com/zendeskgarden/react-components/blob/e1bac19bbdbf99ec6a8131ce6a12c1db6e9888b4/packages/forms/src/styled/range/StyledRangeInput.ts#L142-L144

Screenshot 2024-07-24 at 6 31 22 AM

jzempel: Ok, my bad. I wasn't viewing the `:hover` state.

jzempel (approved):

PR #1871, opened 2024-07-24
feat(buttons): underline `Anchor` by default

jzempel:
## Description

The same default styling applies to `<Button isLink>`. Due to expected usage, the link-styled `Button` does not offer a capability to remove the underline. Note that existing hover/focus/active styling remains unchanged.

## Checklist

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -52,6 +52,12 @@ consider additional positioning prop support on a case-by-case basis.
 
 - Removed `ButtonGroup`: UI no longer recommended by Garden
 - Removed `IIconProps` type export. Use `IButtonStartIconProps` or `IButtonEndIconProps` instead.
+- `Anchor`: renders with an underline for improved accessibility. The same
+  treatment applies to `<Button isLink>`. The default can be removed with
+  `<Anchor isUnderlined={false}>` for word-wrapped or redundant UI where the
+  underline is considered to be a visual distraction. While technically not a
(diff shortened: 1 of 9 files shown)


docs/migration.md
```diff
@@ -52,6 +52,12 @@ consider additional positioning prop support on a case-by-case basis.
 
 - Removed `ButtonGroup`: UI no longer recommended by Garden
 - Removed `IIconProps` type export. Use `IButtonStartIconProps` or `IButtonEndIconProps` instead.
+- `Anchor`: renders with an underline for improved accessibility. The same
+  treatment applies to `<Button isLink>`. The default can be removed with
+  `<Anchor isUnderlined={false}>` for word-wrapped or redundant UI where the
+  underline is considered to be a visual distraction. While technically not a
+  breaking change, the migration guide highlights this change for upgrade cases
+  that may render unexpected styling.
```
ze-flo, line 60: Should we mention that `<Button isLink />` will always render with an underline?

jzempel: My instinct is not to highlight that aspect too hard, and see if consumers need the non-default styling over buttons. If so, we can add a `isLinkUnderlined` (or similar) prop.

ze-flo (approved):

docs/migration.md
geotrev: I suspect as long as the designs match this new default, we won't get too much humbuzz :D cc: @lucijanblagonic

PR #1872, opened 2024-07-24
fix(dropdowns.next): `Combobox` with `isBare` option never shows horiz scrollbar to preserve styling

ze-flo:
## Description
The styling of `<Combobox isBare>` would break for users with OS settings set to always show scrollbars.

Screenshot 2024-07-24 at 10 35 45 AM

This update fixes it by setting the `overflow-y` of the input wrapper to `visible` when `isBare` is enabled. After fix:

#### `isBare` 
Screenshot 2024-07-26 at 8 40 46 AM

#### `isBare` and `isMultiSelectable` 
Screenshot 2024-07-26 at 8 39 42 AM

## Checklist

- [ ] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- [x] :globe_with_meridians: demo is up-to-date (`npm start`)
- [ ] ~~:arrow_left: renders as expected with reversed (RTL) direction~~
- [ ] ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] ~~:wheelchair: tested for WCAG 2.1 AA accessibility compliance~~
- [x] :memo: tested in Chrome, Firefox, Safari, and Edge

PR #1873, opened 2024-07-25
feat: applies color-scheme to align garden and system themes

geotrev:
## Description

Currently, system controls (scrollbars) are forced to their default color even in Garden dark mode (`normal`).

This PR sets the `color-scheme` property throughout several components to ensure system controls switch between light and dark theme alongside Garden's theme.

## Detail

To ensure the theme is retained regardless of system settings, the value is set using the `only` keyword ([see spec](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme#syntax)).

To test, I include the following snippet inside the content of the affected component in dark theme (and then swap the value to `dark` when testing light theme).
```
<ThemeProvider theme={pt => ({ ...pt, colors: { ...pt.colors, base: 'light' } })}>
  Nested content
</ThemeProvider>
```

I then applied `overflowY` or `overflowX` with `auto` to containers that can be scrollable. To create constrain, `maxHeight` and `maxWidth` were added to either the component itself OR a containing `div`. These testing parameters help mimic real world scenarios where a consumer could reach for either option to achieve constrained content dimensions.

For example, compare direct component constrain (note `Tabs` and `Tab.TabPanel`):

```
<Tabs {...args} style={{ maxWidth: 250 }}>
  <Tabs.TabList style={{ overflowY: 'auto' }}>
    {tabs.map(tab => (
      <Tabs.Tab key={tab.value} item={tab.value} disabled={tab.disabled}>
        {tab.value}
      </Tabs.Tab>
    ))}
  </Tabs.TabList>
  {tabs.map(tab => (
    <Tabs.TabPanel key={tab.value} item={tab.value} style={{ overflowY: 'auto', maxHeight: 200 }}>
      {tab.panel}
    </Tabs.TabPanel>
  ))}
</Tabs>
```

With container-based constrain using `div`, meant to simulate page layout constraints:

```
<div style={{ maxWidth: 250, maxHeight: 300 }}>
  <Tabs {...args}>
    <Tabs.TabList style={{ overflowY: 'auto' }}>
      {tabs.map(tab => (
        <Tabs.Tab key={tab.value} item={tab.value} disabled={tab.disabled}>
          {tab.value}
        </Tabs.Tab>
      ))}
    </Tabs.TabList>
      <Tabs.TabPanel key={tab.value} item={tab.value} style={{ overflowY: 'auto' }}>
        {tab.panel}
      </Tabs.TabPanel>
  </Tabs>
</div>
```

---

List of components + screenshots below. 

### Chrome (`Content` + `Sheet`)

Screenshot 2024-07-25 at 10 12 43 AM

### `Grid`

Screenshot 2024-07-25 at 10 10 17 AM

### `Modal`

Screenshot 2024-07-25 at 10 08 43 AM

### `Drawer`

Screenshot 2024-07-25 at 10 09 24 AM

### `TooltipModal`

Screenshot 2024-07-25 at 10 08 24 AM

### `Menu`

Screenshot 2024-07-25 at 10 11 46 AM

#### Multiselectable scrolling

Screenshot 2024-07-25 at 2 40 29 PM

#### Listbox

Screenshot 2024-07-25 at 10 11 39 AM

### Pane (`PaneProvider`)

Screenshot 2024-07-25 at 10 11 29 AM

### Tabs (`TabList` + `TabPanel`)

Screenshot 2024-07-25 at 11 07 13 AM

## Exclusions

- **Accordions**: Scrolling in accordion content could be seen as an anti-pattern. I didn't see much precedence for this in the wild either. Given we need to support the animation for expand/collapse, anything other than a static height would break animation behavior anyway.
- **Tables**: The `table` element itself is not recommended to receive scrolling, and instead can use an external wrapper [as suggested](https://garden.zendesk.com/components/table#scroll) in Garden's API docs. In this case, consumers should set `color-scheme` on the wrapper. We can update the website example to include it.

## Checklist

### DIFF
--- a/packages/chrome/src/styled/body/StyledContent.tsx
+++ b/packages/chrome/src/styled/body/StyledContent.tsx
@@ -40,6 +40,7 @@ export const StyledContent = styled.div.attrs({
   'data-garden-version': PACKAGE_VERSION
 })<IStyledContentProps>`
   display: flex;
+  color-scheme: only ${p => p.theme.colors.base};
   color: ${props => getColor({ theme: props.theme, variable: 'foreground.default' })};
 
   ${sizeStyles};
--- a/packages/chrome/src/styled/sheet/StyledSheetBody.ts
+++ b/packages/chrome/src/styled/sheet/StyledSheetBody.ts
@@ -17,6 +17,7 @@ export const StyledSheetBody = styled.div.attrs({
   flex: 1;
   overflow-y: auto;
   padding: ${props => props.theme.space.base * 5}px;
+  color-scheme: only ${p => p.theme.colors.base};
 
   ${props => retrieveComponentStyles(COMPONENT_ID, props)};
 `;
--- a/packages/dropdowns/src/views/combobox/StyledTrigger.ts
+++ b/packages/dropdowns/src/views/combobox/StyledTrigger.ts
@@ -81,6 +81,7 @@ const colorStyles = ({
   `;
 
   return css`
+    color-scheme: only ${theme.colors.base};
     border-color: ${isLabelHovered ? hoverBorderColor : borderColor};
     background-color: ${backgroundColor};
     color: ${foregroundColor};
--- a/packages/grid/src/styled/StyledGrid.ts
+++ b/packages/grid/src/styled/StyledGrid.ts
@@ -12,20 +12,25 @@ import { IGridProps } from '../types';
 
 const COMPONENT_ID = 'grid.grid';
 
-const colorStyles = ({ theme }: IStyledGridProps) => {
-  const borderColor = getColor({
-    theme,
-    hue: 'crimson',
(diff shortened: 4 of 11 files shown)


jzempel (commented):
Are we missing the `Combobox` trigger?

geotrev:
> Are we missing the `Combobox` trigger?

@jzempel Good catch, updated!

geotrev:
I can wait to test Combobox when #1872 merges, assuming there are any differences in behavior.

jzempel (approved):

ze-flo (approved):
Nice description! 🖼️ 🖼️ 🖼️ 🖼️ 🖼️

steue (approved):
thanks for the explanation. lgtm!

PR #1883, opened 2024-08-08
fix: updates displayNames to match subcomponent properties

geotrev:
## Description

Catches several components' `displayName` properties to match corresponding subcomponent mapping. Follows [migration guide](https://github.com/zendeskgarden/react-components/blob/next/docs/migration.md) for targeted components.

## Detail

Separately:

- Removed export on `OffsetPaginationComponent`
- Copies `Close`, `Title`, and `Paragraph` subcomponents under `react-notifications` to be applied in `Well`, `Alert` and `Notification` with unique/non-shared `displayName` values. Folder structure mirrors existing packages with similar subcomponent mappings, namely `react-modals`.
  - Leaves existing deprecation text for V8 standalone subcomponent exports 

## Checklist

- ~~:globe_with_meridians: demo is up-to-date (`npm start`)~~


jzempel (approved):
Whoa, I guess there was more here than we expected 😅. Reminder to run `npx garden cmd-docgen --pretty` against key files as an extra check to verify that `@extends` clause and naming are lining up as expected.

geotrev:
> Whoa, I guess there was more here than we expected 😅. Reminder to run `npx garden cmd-docgen --pretty` against key files as an extra check to verify that `@extends` clause and naming are lining up as expected.

Good call, from what I can tell the new files look correct.

You mentioned `OffsetPagination` was having issues before with docgen – but I'm not seeing a difference after removing the `export` on `OffsetPaginationComponent`. Just checking if I should be using another command to validate that issue is fixed/

PR #1893, opened 2024-08-19
fix: React peer dependency ranges

jzempel:
## Description

This PR gets the peer dependency ranges on `react` and `react-dom` dialed-in for v9. The previous `>=16.8.0` notation [accurately](https://semver.npmjs.com/) includes everything since React hooks. However, some package managers incorrectly report React 18 (for example) as an invalid peer. The updated `^16.8.0 || ^17.0.0 || ^18.0.0` _or_ will satisfy all package managers. More importantly, the top end of the range is capped at v18 so Garden can manage the addition of future React versions _after_ testing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31226,6 +31226,48 @@
       "integrity": "sha512-eZxlbI8GZscaGS7kkc/trHTT5xgrjH3/1n2JDwusC9iahPKWMRvRjJSAN5mCXviuTGQ/lHnhvv8Q1YTpnfz9gA==",
       "dev": true
     },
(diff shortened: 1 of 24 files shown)


ze-flo (approved):
Should [react-containers](https://github.com/zendeskgarden/react-containers/blob/main/packages/accordion/package.json#L28) be updated as well?

jzempel:
> Should [react-containers](https://github.com/zendeskgarden/react-containers/blob/main/packages/accordion/package.json#L28) be updated as well?

Yeah, good call

PR #1897, opened 2024-08-20
fix(avatars): use default background surfaceColor with statuses

geotrev:
## Description

Fixes bug introduced in #1885 where `Avatar` with `status` defaulted to transparent background.

Before vs. avatar:

Screenshot 2024-08-20 at 9 36 53 AM
Screenshot 2024-08-20 at 9 36 18 AM

## Checklist

- ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- ~~:globe_with_meridians: demo is up-to-date (`npm start`)~~
- ~~:arrow_left: renders as expected with reversed (RTL) direction~~
- ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- ~~:wheelchair: tested for WCAG 2.1 AA accessibility compliance~~
- ~~:memo: tested in Chrome, Firefox, Safari, and Edge~~

PR #1885, opened 2024-08-14
feat: allow color-related props to receive color variable keys in addition to hex values

jzempel:
## Description

Updates related prop documentation and sets the following props to be able to receive color variable key strings:

- Timeline.Item `surfaceColor`
- Avatar
  - `backgroundColor`
  - `foregroundColor`
  - `surfaceColor`
- Span `hue`

## Detail

`Tag` is a description update only as the primary `hue` colors already received intended light/dark treatment.

## Checklist

- [ ] :ok_hand: ~design updates will be Garden Designer approved (add the designer as a reviewer)~
- [x] :globe_with_meridians: demo is up-to-date (`npm start`)
- [ ] :arrow_left: ~renders as expected with reversed (RTL) direction~
- [ ] :metal: ~renders as expected with Bedrock CSS (`?bedrock`)~
- [x] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] :wheelchair: ~tested for WCAG 2.1 AA accessibility compliance~
- [ ] :memo: ~tested in Chrome, Firefox, Safari, and Edge~

PR #1901, opened 2024-08-21
fix(avatars): rectify `StatusIndicator` sizing and position

ze-flo:
## Description
Fix `StatusIndicator` styling to match [v7](https://606fa12ea4f2996a9cfd97aa--zendeskgarden.netlify.app/avatars/#text)

## Detail
- Allows `StatusIndicator` to stretch up to a `max-size`
- Fix `StatusIndicator` positioning for large `Avatar`

Screenshot 2024-08-20 at 7 59 01 PM
 
Screenshot 2024-08-20 at 7 58 46 PM
Screenshot 2024-08-20 at 7 59 36 PM
Screenshot 2024-08-20 at 7 59 23 PM

## Checklist

### DIFF
--- a/packages/avatars/src/styled/StyledAvatar.ts
+++ b/packages/avatars/src/styled/StyledAvatar.ts
@@ -80,11 +80,12 @@ const colorStyles = ({
   }
 
   if ($status) {
-    surfaceColor = $surfaceColor?.includes('.')
-      ? getColor({ variable: $surfaceColor, theme })
-      : $surfaceColor || getColor({ variable: 'background.default', theme });
+    surfaceColor =
+      $surfaceColor && /^\w+\.\w+$/u.test($surfaceColor)
+        ? getColor({ variable: $surfaceColor, theme })
+        : $surfaceColor || getColor({ variable: 'background.default', theme });
   } else {
-    surfaceColor = $surfaceColor || 'transparent';
+    surfaceColor = 'transparent';
   }
 
   return css`
--- a/packages/avatars/src/styled/StyledAvatar.ts
+++ b/packages/avatars/src/styled/StyledAvatar.ts
@@ -80,10 +80,9 @@ const colorStyles = ({
   }
 
   if ($status) {
-    surfaceColor =
-      $surfaceColor && /^\w+\.\w+$/u.test($surfaceColor)
-        ? getColor({ variable: $surfaceColor, theme })
-        : $surfaceColor || getColor({ variable: 'background.default', theme });
+    surfaceColor = $surfaceColor?.includes('.')
+      ? getColor({ variable: $surfaceColor, theme })
+      : $surfaceColor || getColor({ variable: 'background.default', theme });
   } else {
     surfaceColor = 'transparent';
   }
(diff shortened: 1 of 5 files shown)


packages/avatars/src/styled/StyledStatusIndicatorBase.ts
```diff
@@ -41,7 +41,6 @@ const sizeStyles = (props: IStyledStatusIndicatorProps) => {
   return css`
     border: ${offset} ${props.theme.borderStyles.solid};
     border-radius: ${size};
-    width: ${size};
```
ze-flo, line 44: Matches with [v7](https://github.com/zendeskgarden/css-components/blob/%40zendeskgarden/css-avatars%406.0.8/packages/avatars/src/_badge.css) CSS.

jzempel (dismissed):
I think we can see from the Menu [patterns] story that I forgot to deal with the status indicator (`figcaption`) `box-shdow` in #1885. Also the #1897 update seems to yield unpredictable results if variable `surfaceColor` (ex "background.raised") is applied to the `Avatar` story before any other controls are set. Can we nab those here?

Screenshot 2024-08-21 at 6 07 11 PM

Screenshot 2024-08-21 at 6 16 50 PM

geotrev:
> I think we can see from the Menu [patterns] story that I forgot to deal with the status indicator (`figcaption`) `box-shdow` in #1885. Also the #1897 update seems to yield unpredictable results if variable `surfaceColor` (ex "background.raised") is applied to the `Avatar` story before any other controls are set. Can we nab those here?
> 
> Screenshot 2024-08-21 at 6 07 11 PM Screenshot 2024-08-21 at 6 16 50 PM

The gift that keeps on giving

packages/avatars/src/styled/StyledAvatar.ts
```diff
@@ -80,11 +80,12 @@ const colorStyles = ({
   }
 
   if ($status) {
-    surfaceColor = $surfaceColor?.includes('.')
-      ? getColor({ variable: $surfaceColor, theme })
-      : $surfaceColor || getColor({ variable: 'background.default', theme });
+    surfaceColor =
+      $surfaceColor && /^\w+\.\w+$/u.test($surfaceColor)
```
jzempel, line 84: Is there a reason we moved away from the simpler `includes('.')` conditional here?

ze-flo: Ooops, I forgot to clean up after myself. 🤦 This is what I wanted to bring up yesterday.

Will any of the [components](https://github.com/zendeskgarden/react-components/pull/1885) accepting a variable key as a user-defined color ever need to support `rgba` values created by using `getColor` with a `transparency` option?

jzempel: Ah dang, that's the sneaky notation I hadn't thought of. Let's defer to address more holistically. (I recall @geotrev suggesting some kind if `isColorVariable` utility added to `theming`).

jzempel (approved):
Thanks for nabbing those finicky bits and adding tests so we don't keep regressing.

PR #1882, opened 2024-08-07
fix(tooltips): renders correct arrow dimensions based on tooltip size

ze-flo:
## Description

This PR refines the `arrowStyles` logic to output deterministic arrow dimensions.

## Detail
- Simplifies Tooltip styles to avoid relying on arrow `inset`
- Ensures the rendered height of the arrow matches with the `size` argument

For this example, `ArrowStylesStory` was modified to display a box-shadow matching the arrow `size` input. 

```
box-shadow: ${p => p.hasBoxShadow && `0 0 0 ${p.size}px lavender`};
```

This preview shows the impact of changing the `size` and `inset` inputs. The arrow was recolored to better show its dimensions.

Screen Recording 2024-08-07 at 10 18 05 AM

Notice that the tip of the arrow ends where the drop shadow ends.

As a result, tooltip arrow rendered sizes match with V8:

Screenshot 2024-08-07 at 12 46 33 PM

## Arrows checklist
- [x] Renders `Tooltips` correctly
- [x] Renders `ArrowStyles correctly
- [x] Renders `Legacy Menu` correctly
- [x] Renders `Menu` correctly

## Checklist

- [ ] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~
- [x] :globe_with_meridians: demo is up-to-date (`npm start`)
- [ ] ~~:arrow_left: renders as expected with reversed (RTL) direction~~
- [ ] ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [x] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] ~~:wheelchair: tested for WCAG 2.1 AA accessibility compliance~~
- [x] :memo: **Thoroughly tested** in Chrome, Firefox, Safari, and Edge on retina and non-retina displays

PR #1902, opened 2024-08-21
fix(tooltips): fix arrow position when using `end-bottom` placement

ze-flo:
## Description
Fixes a regression created in https://github.com/zendeskgarden/react-components/pull/1882

Screenshot 2024-08-21 at 12 16 43 PM

## Detail
- Refines `defaultInset` to better render on retina and regular monitors 
- Adds ability to shift the arrow placement alongside the edge of the parent container. This addresses the case when a parent container is too short for be supported by the default arrow placement logic.

For example:

The arrow of a small tooltip using `end-bottom` placement will extend past the container's boundaries because of the border-radius

### Before
Screenshot 2024-08-21 at 12 10 51 PM

### After

Screenshot 2024-08-21 at 12 12 12 PM

## Checklist

- [ ] ~~:arrow_left: renders as expected with reversed (RTL) direction
- [ ] ~:metal: renders as expected with Bedrock CSS (`?bedrock`)~~
- [ ] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] ~~:wheelchair: tested for WCAG 2.1 AA accessibility compliance

### DIFF
--- a/packages/tooltips/src/styled/StyledTooltip.ts
+++ b/packages/tooltips/src/styled/StyledTooltip.ts
@@ -73,9 +73,15 @@ const sizeStyles = ({
   }
 
   let arrowSize;
+  let arrowShift;
 
   if (hasArrow) {
-    if (size === 'small' || size === 'medium') {
+    if (size === 'small') {
+      arrowSize = margin;
+      if (['left-start', 'left-end', 'right-start', 'right-end'].includes(placement)) {
+        arrowShift = '-4px';
+      }
+    } else if (size === 'medium') {
       arrowSize = margin;
     } else if (size === 'large') {
       margin = `${theme.space.base * 2}px`;
@@ -97,7 +103,8 @@ const sizeStyles = ({
     font-size: ${fontSize};
     overflow-wrap: ${overflowWrap};
 
-    ${hasArrow && arrowStyles(getArrowPosition(theme, placement), { size: arrowSize })};
+    ${hasArrow &&
+    arrowStyles(getArrowPosition(theme, placement), { size: arrowSize, shift: arrowShift })};
 
     ${StyledParagraph} {
       margin-top: ${paragraphMarginTop};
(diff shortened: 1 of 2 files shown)


packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -73,9 +73,15 @@ const sizeStyles = ({
...
   let arrowSize;
+  let arrowShift;
 
   if (hasArrow) {
-    if (size === 'small' || size === 'medium') {
+    if (size === 'small') {
+      arrowSize = margin;
+      if (['left-start', 'left-end', 'right-start', 'right-end'].includes(placement)) {
+        arrowShift = '-4px';
+      }
```
jzempel, line 83:
```suggestion
      if (['left-start', 'left-end', 'right-start', 'right-end'].includes(placement)) {
        arrowShift = '-4px';
      } else {
        arrowShift = margin;
      }
```

[nit] prefer clarifying assignment

packages/tooltips/src/styled/StyledTooltip.ts
```diff
@@ -73,9 +73,15 @@ const sizeStyles = ({
...
 
   let arrowSize;
+  let arrowShift;
 
   if (hasArrow) {
-    if (size === 'small' || size === 'medium') {
+    if (size === 'small') {
+      arrowSize = margin;
+      if (['left-start', 'left-end', 'right-start', 'right-end'].includes(placement)) {
+        arrowShift = '-4px';
```
jzempel, line 82: Would it be better to tie this to the theme with

```
arrowShift = `-${theme.borderRadii.md}`;
```

to better indicate the shifting rationale and respond to theming customizations that knock out the borders?

geotrev (approved):
Nice!

jzempel (approved):

PR #1905, opened 2024-08-23
refactor: remove `theme` `defaultProp`

ze-flo:
## Description
This initiates the process of removing `defaultProps` to avoid deprecation warnings, focusing first on `theme`. Since v9 depends entirely on `ThemeProvider` and its `theme` for dark mode, rendering a component without wrapping it in `ThemeProvider` should result in failure.

## Detail
- Removes `eslint-plugin-garden-local`
- `ThemeProvider`: sets the theme prop's default to `DEFAULT_THEME`
- Removes theme defaultProp in all components
- Removes unnecessary test

## Checklist

- [ ] ~~:globe_with_meridians: demo is up-to-date (`npm start`)~~
- [ ] ~~:guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)~~


jzempel (changes requested):
Unfortunately, I don't think this is the desired change. Garden element components need to continue to render even when not wrapped by a `ThemeProvider`. In other words, this change needs to occur in a non-breaking fashion. Most likely this is via argument defaults on element components which are in-turn passed down to view components. But some exploration/discussion may be necessary. Furthermore, the changes must be validated via `npx garden cmd-docgen --pretty <path>` to ensure that every affected component retains the expected prop documentation.

geotrev (commented):
+1 JZ's comment. I was under the impression we wanted to keep theme as a default prop but explore how that can be achieved with styled components first, as the changeset is pretty big (as this PR is evident).

Also I remember we talked about keeping the initial change isolated on `ThemeProvider` as its the biggest culprit of console warnings in development. 😅

ze-flo:
Here's a table comparing popular React component libraries written with `styled-components` or `emotion.js`. All support theming and light / dark mode. The table highlights whether each library can render components without first nesting them inside a `ThemeProvider`.

| Library               | Works without ThemeProvider | Notes                                                                                                                                                                                     | Docs                                                                                     | Sandbox                                               |
|-----------------------|-----------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|-------------------------------------------------------|
| @mui/material         | ✅                           | DefaultProps injected via React's context API with [defaultPropsProvider](https://github.com/mui/material-ui/blob/master/packages/mui-system/src/DefaultPropsProvider/DefaultPropsProvider.tsx#L56) | https://mui.com/material-ui/getting-started/usage/                                       | https://codesandbox.io/p/sandbox/jgtp7x               |
| @mantine/core         | ❌                           |                                                                                                                                                                                           | https://mantine.dev/guides/vite/#setup                                                   |                                                       |
| @primer/react         | ❌                           |                                                                                                                                                                                           | https://primer.style/guides/react                                                        |                                                       |
| @chakra-ui/react      | ❌                           |                                                                                                                                                                                           | https://v2.chakra-ui.com/getting-started                                                 |                                                       |
| @adobe/react-spectrum | ❌                           | Components requiring theme render null if its missing. Others render without styles.                                                                                                      | https://react-spectrum.adobe.com/react-spectrum/getting-started.html#setting-up-your-app | https://codesandbox.io/p/sandbox/admiring-wood-fqqj8h |


While `@mui/material` has a built-in mechanism to handle the absence of a `ThemeProvider`, most  tested libraries strictly require it to be set up for proper rendering and styling. The behavior varies across different libraries, with some failing to render components and others rendering unstyled components without a theme.

ze-flo:
> @ze-flo I think [this is pretty compelling evidence](https://github.com/zendeskgarden/react-components/pull/1905#issuecomment-2322170421) that you're on the right path with this PR. Can we get the `migration.md` updated with the breaking change to `theming` and either add a corresponding [ADR](https://github.com/zendeskgarden/react-components/tree/main/docs/adrs) or have a plan to add it immediately after?

I updated the migration docs and created a ticket to add the corresponding ADR in a follow-up PR.

jzempel (approved):

PR #1908, opened 2024-08-29
fix(tabs): border styling

jzempel:
## Description

This CSS update more effectively distinguishes between `border-[side]`, `border-width`, and `border-color`, allowing any one of these properties to be set independently without interfering with the others.

## Detail

Needed to handle a website example regression seen with https://garden.zendesk.com/components/theme-provider#targeting.


geotrev (approved):
Oof, doozy.

PR #1915, opened 2024-09-09
fix(typography): revert `react-prism-renderer` to v1

jzempel:
## Description

This fix addresses an issue that is particularly problematic in SSR. The asynchronous imports cause a delay with `CodeBlock` rendering, which throw off internal page links. Given that Garden has SSR consumers relying heavily on `CodeBlock`, we need to revert to a solution that a) offers synchronous imports, and b) doesn't pollute the global space with `window.PrismJS`.


ze-flo (approved):
That's a bummer. 😢

PR #1917, opened 2024-09-10
fix(tooltips): ensure correct padding for `TooltipDialog`'s inner elements if `Close` btn exists

ze-flo:
## Description

Adds end padding to first descendant of `StyledTooltipDialog` if `Close` btn exists

### With `Close` btn

Screenshot 2024-09-09 at 4 41 57 PM
Screenshot 2024-09-09 at 4 42 45 PM
Screenshot 2024-09-09 at 4 41 33 PM

### Without `Close` btn
Screenshot 2024-09-09 at 4 42 23 PM

## Checklist

- [ ] :metal: renders as expected with Bedrock CSS (`?bedrock`)~~

### DIFF
--- a/packages/modals/src/styled/StyledTooltipDialog.ts
+++ b/packages/modals/src/styled/StyledTooltipDialog.ts
@@ -27,6 +27,7 @@ export interface IStyledTooltipDialogProps
 const sizeStyles = (props: ThemeProps<DefaultTheme>) => `
   padding: ${props.theme.space.base * 5}px;
   width: 400px;
+  
   &:has(${StyledTooltipDialogClose}) > :first-child {
     padding-${props.theme.rtl ? 'left' : 'right'}: ${props.theme.space.base * 8}px;
   }
--- a/packages/modals/src/styled/StyledTooltipDialog.ts
+++ b/packages/modals/src/styled/StyledTooltipDialog.ts
@@ -5,13 +5,14 @@
  * found at http://www.apache.org/licenses/LICENSE-2.0.
  */
 
-import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


packages/modals/src/styled/StyledTooltipDialog.ts
```diff
@@ -23,14 +24,19 @@ export interface IStyledTooltipDialogProps
   transitionState?: TransitionStatus;
 }
 
+const sizeStyles = (props: ThemeProps<DefaultTheme>) => `
+  padding: ${props.theme.space.base * 5}px;
+  width: 400px;
+  &:has(${StyledTooltipDialogClose}) > :first-child {
```
geotrev, line 30: I'm so happy support for this selector is as high as it is. Didn't realize its been in safari since version 15!

geotrev, line 30:
```suggestion
  
  &:has(${StyledTooltipDialogClose}) > :first-child {
```

geotrev (changes requested):
Looks great!

I'm so happy support for `:has` selector is as high as it is. Didn't realize its been in safari since version 15!

jzempel (approved):
Nice – CSS-only FTW :tada:. Other components (Modal, Drawer, etc) could probably remove context and benefit from this simplification.

geotrev:
> Nice – CSS-only FTW 🎉. Other components (Modal, Drawer, etc) could probably remove context and benefit from this simplification.

It feels like maybe a few components can benefit from this (assuming the use of context isn't logic-driven) 😁

PR #1919, opened 2024-09-10
fix(dropdowns): forwards menu button ref correctly

geotrev:
## Description

This PR addresses a small bug with menu button ref handling, in addition to improving ref handling between `buttonProps` and `button` in a `Menu`.

## Detail

The main fix is reversing Menu's `ref` (from `forwardRef`) from merging with `triggerRef` when composing button props. Instead, the `ref` from `buttonProps` is merged.

Now a consumer could achieve something like this:

```
<Menu
  buttonProps={{ ref: buttonRef }} 
  button={(props) => <Button {...props}>Click me</Button>
>
  ...
</Menu>
```

... and the ref would be maintained.

## Checklist

- ~~:globe_with_meridians: demo is up-to-date (`npm start`)~~
- ~~:black_circle: renders as expected in dark mode~~

### DIFF
--- a/packages/dropdowns/src/elements/menu/Menu.tsx
+++ b/packages/dropdowns/src/elements/menu/Menu.tsx
@@ -68,7 +68,13 @@ export const Menu = forwardRef<HTMLUListElement, IMenuProps>(
       onChange
     });
 
-    const { onClick, onKeyDown, disabled, ...buttonProps } = _buttonProps;
+    const {
+      onClick,
+      onKeyDown,
+      disabled,
+      ref: _ref,
+      ...buttonProps
+    } = _buttonProps as IButtonProps & { ref: RefObject<HTMLButtonElement> };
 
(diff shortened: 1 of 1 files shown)


packages/dropdowns/src/elements/menu/Menu.tsx
```diff
@@ -79,7 +85,7 @@ export const Menu = forwardRef<HTMLUListElement, IMenuProps>(
         onKeyDown,
         disabled
       }),
-      ref: mergeRefs([triggerRef, ref]) as unknown as RefObject<HTMLButtonElement>
```
geotrev, line 82: This `ref` is from `Menu`, which we don't want.

ze-flo (approved):
Not relying on consumers to use `mergeRef` is a win to me. 🙌

jzempel (approved):

PR #1920, opened 2024-09-11
fix(chrome): prevent focus on closed `Sheet`

jzempel:
## Description

Applies `inert` attribute when the `Sheet` is closed in order to prevent tab focus through the sheet's interactive elements.

## Checklist

### DIFF
--- a/packages/chrome/src/elements/sheet/Sheet.tsx
+++ b/packages/chrome/src/elements/sheet/Sheet.tsx
@@ -51,6 +51,7 @@ const SheetComponent = React.forwardRef<HTMLElement, ISheetProps>(
     return (
       <SheetContext.Provider value={sheetContext}>
         <StyledSheet
+          inert={isOpen ? undefined : ''}
           isOpen={isOpen}
           isAnimated={isAnimated}
           placement={placement}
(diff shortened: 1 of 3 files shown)


packages/chrome/src/elements/sheet/Sheet.tsx
```diff
@@ -51,6 +51,7 @@ const SheetComponent = React.forwardRef<HTMLElement, ISheetProps>(
     return (
       <SheetContext.Provider value={sheetContext}>
         <StyledSheet
+          inert={isOpen ? undefined : ''}
```
ze-flo, line 54: Adding mental notes that this might break with React@19.x:

https://github.com/facebook/react/issues/17157#issuecomment-2003750544

ze-flo (approved):

packages/chrome/src/elements/sheet/Sheet.tsx
geotrev: using an empty string feels wrong anyway lol glad they're fixing it.

jzempel: the unit tests should tell us if the fix breaks Garden's code

PR #1927, opened 2024-09-24
fix(colorpickers): prevent arrow from positioning over a `ColorSwatchDialog` tooltip

jzempel:
## Description

The `z-index` fix prevents the following stacking problem with `<ColorSwatchDialog hasArrow>`...

Screenshot 2024-09-24 at 1 07 04 PM

## Detail

Slips in a `string` control fix for `<Avatar badge="...">`


ze-flo (approved):
Thanks for catching that. 💯

PR #1914, opened 2024-09-06
docs: add removal of `theme` default prop ADR

ze-flo:
## Description

Adds the "Removal of `theme` default prop" ADR. See reference PR for context: https://github.com/zendeskgarden/react-components/pull/1905

PR #1929, opened 2024-09-25
chore: v9 QA sweep

jzempel:
## Description

- expose `TooltipModal` internal-only `offset` prop to decrease space between `ColorPickerDialog`/`ColorSwatchDialog` and trigger (per designs)
- prevent render thrashing on `getColor` storybook demo
- split migration guides out of historical changelogs for improved `website` documentation

Screenshot 2024-09-25 at 3 37 57 PM

### DIFF
--- a/packages/colorpickers/src/elements/ColorPickerDialog/index.tsx
+++ b/packages/colorpickers/src/elements/ColorPickerDialog/index.tsx
@@ -14,6 +14,7 @@ import React, {
   forwardRef,
   ReactElement
 } from 'react';
+import { useTheme } from 'styled-components';
 import PropTypes from 'prop-types';
 import { Button } from '@zendeskgarden/react-buttons';
 import { PLACEMENT } from '@zendeskgarden/react-modals';
@@ -27,7 +28,7 @@ import {
   StyledTooltipBody
 } from '../../styled';
 import { IColor, IColorPickerDialogProps } from '../../types';
-import { useText } from '@zendeskgarden/react-theming';
+import { DEFAULT_THEME, useText } from '@zendeskgarden/react-theming';
 
 /**
  * @extends HTMLAttributes<HTMLDivElement>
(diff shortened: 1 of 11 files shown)


packages/colorpickers/src/elements/ColorPickerDialog/index.tsx
```diff
@@ -70,6 +71,7 @@ export const ColorPickerDialog = forwardRef<HTMLDivElement, IColorPickerDialogPr
       'aria-label',
       'Color picker'
     );
+    const theme = useTheme() || DEFAULT_THEME;
```
ze-flo, line 74: https://github.com/zendeskgarden/react-components/pull/1914 build the case that components should always be wrapped in `ThemeProvider` starting with v9. Do we still need the fallback `DEFAULT_THEME`?
```suggestion
    const theme = useTheme()
```

jzempel: My preference is to be consistent with other element-level components for now. If/when we determine to simplify, it should happen across everything.

ze-flo: Good point. 👍 We can always follow-up with a broad sweep. 🧹

packages/modals/src/types/index.ts
```diff
@@ -79,6 +79,8 @@ export interface ITooltipDialogProps extends Omit<IModalProps, 'isCentered' | 'i
    * Positions the modal relative to the provided `HTMLElement`
    */
   referenceElement?: HTMLElement | null;
+  /** @ignore Modifies the placement offset from the reference element (internal only) */
+  offset?: number;
```
ze-flo, line 83: ❤️

ze-flo (approved):
Great job reorganizing the migration docs. 💯

PR #1930, opened 2024-09-26
fix(dropdowns): ensure focus is returned to menu trigger before calling `onChange`

ze-flo:
## Description

This PR bumps `@zendeskgarden/container-menu` to `v0.5.0` with two key changes:

1. `Menu` now returns focus to the trigger before calling `onChange`, improving A11y for implementations that open modals after selection (see https://github.com/zendeskgarden/react-components/pull/1931)
2. `defaultFocusedValue` only applies when opening the menu via keyboard, aligning with [APG examples](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/)

These changes improve accessibility and user experience, especially for keyboard navigation and complex UI interactions.

## 💥 BREAKING CHANGE:

`Menu`: new `restoreFocus` prop (default: `true`) returns focus to trigger after menu interaction. When menu expansion is controlled to allow multiple item selection, set `restoreFocus={false}` and manage trigger focus manually on close.

## Detail
- Bump `@zendeskgarden/container-menu` to `v0.5.0`
- Update specs to reflect focus behavior when using `defaultFocusedValue`
- Adds breaking change detail to v9 migration doc. 

## Checklist


packages/dropdowns/src/elements/menu/Menu.spec.tsx
```diff
@@ -148,8 +148,22 @@ describe('Menu', () => {
...
+    expect(trigger).toHaveFocus();
 
+    // close menu
+    await user.click(trigger);
+    expect(item).not.toBeVisible();
+
+    // open menu with keyboard
+    trigger.focus();
+    await user.keyboard(' ');
+    // focus is on the focused item associated with `defaultFocusedValue`
```
ze-flo, line 166: https://github.com/zendeskgarden/react-containers/pull/660 changed the focus behavior when using `defaultFocusedValue`.

jzempel (changes requested):
The `Menu` API (and Storybook controls) are missing the new `restoreFocus` prop. This needs to be specified under `/types`.

docs/migration.md
```diff
@@ -119,6 +119,9 @@ consider additional positioning prop support on a case-by-case basis.
   - The `v8` version of `@zendeskgarden/react-dropdowns` is no longer maintained and is
     renamed to `@zendeskgarden/react-dropdowns.legacy` in `v9`
 - `Menu`: value `auto` is no longer valid for the `fallbackPlacements` prop.
+- `Menu`: new `restoreFocus` prop (default: `true`) returns focus to trigger
+  after menu interaction. To keep the dropdown open on selection,
+  set `restoreFocus={false}` and manage focus manually.
```
jzempel, line 124: Not sure I understand this. Wouldn't the `isExpanded` controlled prop be the mechanism for keeping the menu open?

ze-flo: I was referring to cases when a users chooses to keep the menu open after selecting an item within an `ItemGroup` of type `checkbox`. This [demo](https://66e49e3720ee94778c2a11a5--zendeskgarden-react-containers.netlify.app/?path=/story/packages-menu--controlled-managed-focus) showcases this pattern. 

It probably makes more sense to rephrase it as such:

```
`Menu`: new `restoreFocus` prop (default: `true`) returns focus to trigger
  after menu interaction. If keeping the dropdown open on selection is required,
  set `restoreFocus={false}` and manage focus manually.
```

What do you think?

jzempel: This still makes it sound like there is a causal relationship between keeping the menu open and setting `restoreFocus` to false.

jzempel, line 124:
```suggestion
- `Menu`
  - value `auto` is no longer valid for the `fallbackPlacements` prop.
  - new `restoreFocus` prop (default: `true`) returns focus to trigger
    after menu interaction. When menu expansion is controlled to allow
    multiple item selection, set `restoreFocus={false}` and manage trigger
    focus manually on close.
```

ze-flo: Couldn't have wrote it better myself. Thank you! 💯

jzempel (approved):
🚢

PR #1938, opened 2024-09-30
fix(theming)!: remove `background.primary` color variable

ze-flo:
## Description
Following @lucijanblagonic's Figma comment, `background.primary` isn't relevant to Garden consumers and should be removed.

## Details
- Removes `background.primary` from `Theme`
- Replaces instance of `background.primary` with `background.primaryEmphasis`

## Checklist

- [ ] ~~:metal: renders as expected with Bedrock CSS (`?bedrock`)~


packages/avatars/demo/~patterns/stories/MenuStory.tsx
```diff
@@ -64,7 +64,9 @@ export const MenuStory: StoryFn = ({ isCompact }) => {
                     status={item.avatarProps.status}
                     badge={item.avatarProps.badge}
                     surfaceColor={
-                      highlightedValue === item.value ? 'background.primary' : 'background.raised'
+                      highlightedValue === item.value
+                        ? 'background.primaryEmphasis'
```
jzempel, line 68: I'm waiting for the PR to deploy, but I don't think this is what we want to demonstrate with this pattern. The idea is to get the avatar's surface to match the active menu item as close as possible. So, it should be something like `getColor({ theme, hue: 'primaryHue', light: { shade: 100}, dark: { shade: 900} })` to correspond with the previous variable values.

jzempel (changes requested):

packages/avatars/src/types/index.ts
```diff
@@ -29,7 +29,7 @@ export interface IAvatarProps extends HTMLAttributes<HTMLElement> {
   /**
    * Provides surface color for an avatar placed on a non-default background.
    * Accepts a [color variable](/components/theme-object#colors) key (i.e.
-   * `background.primary`) to render based on light/dark mode, or any hex value.
+   * `background.primaryEmphasis`) to render based on light/dark mode, or any hex value.
```
jzempel, line 32:
```suggestion
   * `background.emphasis`) to render based on light/dark mode, or any hex value.
```

[nit] because we don't want to suggest that `primaryEmphasis` should be used for anything other than primary interactive elements.

jzempel (changes requested):

packages/avatars/demo/~patterns/stories/MenuStory.tsx
jzempel: Or maybe `getColor({ theme, variable: 'background.primaryEmphasis', transparency: 100 })` is a more successful match 🤷

jzempel (approved):

packages/avatars/demo/~patterns/stories/MenuStory.tsx
ze-flo: > Or maybe getColor({ theme, variable: 'background.primaryEmphasis', transparency: 100 }) is a more successful match 🤷

`getColor` will throw because the output of the ☝️  is `rgba()` but it treats it as a color variable (`$surfaceColor.includes('.')`). We would have to refine the logic everywhere this comes up.

```
ReferenceError: Error: color variable 'rgba(31,115,183,0.08)' is not defined
```

Also, with the way `Avatar` is built, the transparent `surfaceColor` would lay on top of the background color. That wouldn't result in the right color.

PR #1948, opened 2024-10-09
refactor(chrome): use transient props where appropriate

ze-flo:
## Description
This PR updates various components in `Chrome` to use [transient props](https://styled-components.com/docs/api#transient-props) where appropriate. These changes are necessary in  preparation for the upgrade to `styled-components@6.x.x` to ensure we avoid DOM violation errors after the transition.

## Checklist

### DIFF
--- a/packages/chrome/src/elements/SkipNav.tsx
+++ b/packages/chrome/src/elements/SkipNav.tsx
@@ -15,7 +15,7 @@ import { StyledSkipNav, StyledSkipNavIcon } from '../styled';
  */
 export const SkipNav = React.forwardRef<HTMLAnchorElement, ISkipNavProps>(
   ({ targetId, zIndex, children, ...props }, ref) => (
-    <StyledSkipNav href={`#${targetId}`} zIndex={zIndex} ref={ref} {...props}>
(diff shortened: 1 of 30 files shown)


jzempel (changes requested):
If this package is complete, then its `peerDependencies` can bump to:

`"styled-components": "^5.3.1 || ^6.0.0"`

jzempel (approved):
fwiw, the most recent code convention for spread parameters exists in `dropdowns` where we simply used `...props` naming (vs `...other` or `...rest`). I'm not opposed to `...other` – just pointing to the most evolved convention :wink:

ze-flo:
> fwiw, the most recent code convention for spread parameters exists in `dropdowns` where we simply used `...props` naming (vs `...other` or `...rest`). I'm not opposed to `...other` – just pointing to the most evolved convention 😉

When using `...other` (or `...rest`) instead of `...props`, it's immediately clear that we're only passing along a subset of the props. This makes the code easier to understand and review because you don't have to jump to the component's definition to figure out what's being passed. Since we've mostly used `...other` in our codebase, I'll stick with it for consistency.

PR #1949, opened 2024-10-14
refactor(buttons): use transient props where appropriate

ze-flo:
## Description
This PR updates various components in `react-buttons` to use [transient props](https://styled-components.com/docs/api#transient-props) where appropriate. These changes are necessary in  preparation for the upgrade to `styled-components@6.x.x` to ensure we avoid DOM violation errors after the transition.

## Checklist

### DIFF
--- a/packages/buttons/src/elements/Anchor.tsx
+++ b/packages/buttons/src/elements/Anchor.tsx
@@ -20,8 +20,8 @@ import { useText } from '@zendeskgarden/react-theming';
  * @extends AnchorHTMLAttributes<HTMLAnchorElement>
  */
 export const Anchor = forwardRef<HTMLAnchorElement, IAnchorProps>(
-  ({ children, isExternal, externalIconLabel, isUnderlined = true, ...otherProps }, ref) => {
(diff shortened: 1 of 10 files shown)


jzempel (approved):

jzempel (changes requested):
Oops, looks like the v6 update to `package.json` was missed

ze-flo:
> Oops, looks like the v6 update to `package.json` was missed

I think it might be too soon to confidently claim that this package is compatible with v6.

Here are my planned steps for the migration:
- Finish migrating to transient props in each package.
- Submit a final PR that includes:
  - Implementing changes recommended by the [v6 migration guide](https://styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v6).
  - Upgrading to `styled-components@6` in root and add it to `peerDependencies`.
  - Resolving all TypeScript type errors introduced by v6 (e.g., `ThemeProp<T>` removal).
  - Ensuring that all tests, including temporary snapshot tests, pass with no regressions.

After completing these steps, we can confirm each package is fully compatible with both `styled-components` v5 and v6.

jzempel (approved):

PR #1957, opened 2024-10-22
refactor(dropdowns.legacy): use transient props where appropriate

ze-flo:
## Description
This PR updates various components in `react-dropdowns.legacy` to use [transient props](https://styled-components.com/docs/api#transient-props) where appropriate. These changes are necessary in  preparation for the upgrade to `styled-components@6.x.x` to ensure we avoid DOM violation errors after the transition.

## Checklist

### DIFF
--- a/packages/forms/src/elements/faux-input/FauxInput.tsx
+++ b/packages/forms/src/elements/faux-input/FauxInput.tsx
@@ -26,6 +26,7 @@ const FauxInputComponent = forwardRef<HTMLDivElement, IFauxInputProps>(
       onFocus,
       readOnly,
       validation,
+      mediaLayout,
       ...other
     },
     ref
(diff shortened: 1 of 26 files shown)


packages/forms/src/elements/faux-input/FauxInput.tsx
```diff
@@ -26,6 +26,7 @@ const FauxInputComponent = forwardRef<HTMLDivElement, IFauxInputProps>(
       onFocus,
       readOnly,
       validation,
+      mediaLayout,
```
ze-flo, line 29: `StyledFauxInput` [sets](https://github.com/zendeskgarden/react-components/blob/4c887bef39da96479253a597d423f4f1d8abade2/packages/dropdowns.legacy/src/styled/select/StyledFauxInput.ts#L17) the `mediaLayout` to `true` and extends `FauxInput`, a React FC that returns a styled-component. In this case, a transient prop wouldn't be forwarded.

ze-flo (commented):
All including stories tests are [passing](https://github.com/zendeskgarden/react-components/pull/1959). ✅

jzempel (approved):

PR #1956, opened 2024-10-22
fix: add storybook demos to eslint config

jzempel:
## Description

This PR adds missing `demo` directories 😬 to eslint.

## Detail

I opted to run `eslint --fix` to address the `react/jsx-no-leaked-render` errors (rather than manually applying `!!condition && <RenderComponent>` as we prefer in production). This, along with a subsequent `prettier` resulted in the majority of file changes. As a result, only commits f38022bc00bde434d8695a839e86e72ca0d9d9dd and b371ead18e3ad171246fe22e9f9d9274ad267a34 contain non-automated changes.

PR #1958, opened 2024-10-22
Jzempel/demo lint

jzempel:
## Description

This PR adds missing `demo` directories 😬 to eslint.

## Detail

A less disruptive version of #1956. I learned that strategies could be selectively applied for `eslint --fix` autofixing – resulting in a lower impact PR. Material changes are:
- 9f6a4d86e4ff4825c87e72553e88a6581930d5c6 (updated config)
- c92ac40224489bc54cee99e0b1a9464394100256 (address remaining lint errors)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -56,5 +56,14 @@ export default [
       'no-console': 'off',
       'react/button-has-type': 'off'
     }
+  },
+  {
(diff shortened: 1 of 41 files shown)


ze-flo (approved):
I like this better! 💯

PR #1968, opened 2024-11-01
refactor(colorpickers, grid, loaders, notifications, tags tooltips): use transient props where appropriate 

ze-flo:
## Description
This PR updates various components in `react-colorpickers`, `react-grid`, `react-notifications`, `react-loaders`, and `react-tooltips` to use [transient props](https://styled-components.com/docs/api#transient-props) where appropriate. These changes are necessary in  preparation for the upgrade to `styled-components@6.x.x` to ensure we avoid DOM violation errors after the transition.

## Checklist


packages/loaders/src/elements/Dots.tsx
```diff
@@ -61,18 +61,18 @@ export const Dots = forwardRef<SVGSVGElement, IDotsProps>(
 
     return (
       <StyledSVG
+        $dataGardenId={COMPONENT_ID}
         ref={ref}
         fontSize={size!}
         color={color!}
         width="80"
         height="72"
```
jzempel, line 69: I don't think we want these props leaking to the DOM. Also, might need to `git-blame`, but why do we have these huge `width`/`height` values?

packages/loaders/src/elements/Inline.tsx
```diff
@@ -28,7 +28,7 @@ export const Inline = forwardRef<SVGSVGElement, IInlineProps>(({ size, color, ..
     // eslint-disable-next-line jsx-a11y/prefer-tag-over-role
     <StyledInline
       ref={ref}
-      size={size!}
+      $size={size!}
       color={color!}
```
jzempel, line 32:
```suggestion
      $color={color!}
```

packages/loaders/src/styled/StyledSVG.ts
```diff
@@ -9,13 +9,13 @@ import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
 import { getColor, retrieveComponentStyles } from '@zendeskgarden/react-theming';
 
 interface IStyledSVGProps {
-  dataGardenId: string;
+  $dataGardenId: string;
   color?: string;
   fontSize?: string | number;
   width: number | string;
   height: number | string;
```
jzempel, line 16:
```suggestion
  $color?: string;
  $fontSize?: string | number;
  $width: number | string;
  $height: number | string;
```

We can remove these from the DOM since they only need to inform the CSS

packages/loaders/src/elements/Dots.tsx
```diff
@@ -61,18 +61,18 @@ export const Dots = forwardRef<SVGSVGElement, IDotsProps>(
 
     return (
       <StyledSVG
+        $dataGardenId={COMPONENT_ID}
```
jzempel, line 64:
```suggestion
        dataGardenId={COMPONENT_ID}
```

The `data-garden-id` needs to remain intact in the DOM

packages/loaders/src/styled/StyledSVG.ts
```diff
@@ -9,13 +9,13 @@ import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
 import { getColor, retrieveComponentStyles } from '@zendeskgarden/react-theming';
 
 interface IStyledSVGProps {
-  dataGardenId: string;
+  $dataGardenId: string;
```
jzempel, line 12:
```suggestion
  dataGardenId: string;
```

jzempel (changes requested):

packages/loaders/src/elements/Dots.tsx
ze-flo: `dataGardenId` would be filtered out by [@emotion/is-prop-valid](https://www.npmjs.com/package/@emotion/is-prop-valid) used under the hood in `styled-components@v5`. I've replaced it with `data-garden-id`.

Screenshot 2024-11-06 at 8 03 16 AM

packages/dropdowns.legacy/src/styled/multiselect/StyledMultiselectInput.ts
```diff
@@ -13,22 +13,26 @@ const COMPONENT_ID = 'dropdowns.multiselect_input';
 
 interface IStyledMultiselectInputProps {
   isCompact: boolean;
```
jzempel, line 15: Should this be `$isCompact`?

packages/loaders/src/elements/Dots.tsx
```diff
@@ -61,18 +61,18 @@ export const Dots = forwardRef<SVGSVGElement, IDotsProps>(
...
     return (
       <StyledSVG
+        data-garden-id={COMPONENT_ID}
         ref={ref}
-        fontSize={size!}
-        color={color!}
+        $fontSize={size!}
+        $color={color!}
         width="80"
         height="72"
```
jzempel, line 69: I still think something about this has a bad smell. As a rule, we don't put `width`/`height` attributes on SVGs. Rather, the preference is to apply sizing via CSS. Not sure how/why these got here.

packages/loaders/src/elements/Spinner.tsx
```diff
@@ -74,19 +74,19 @@ export const Spinner = forwardRef<SVGSVGElement, ISpinnerProps>(
 
     return (
       <StyledSVG
+        $color={color}
+        $containerHeight="1em"
+        $containerWidth="1em"
+        $fontSize={size}
+        data-garden-id={COMPONENT_ID}
+        height={HEIGHT}
```
jzempel, line 82: ditto re: `height` and `width`. These should be CSS-only

packages/loaders/src/styled/StyledSVG.ts
```diff
@@ -9,17 +9,17 @@ import styled, { css, DefaultTheme, ThemeProps } from 'styled-components';
...
 
 interface IStyledSVGProps {
-  dataGardenId: string;
-  color?: string;
-  fontSize?: string | number;
+  'data-garden-id': string;
+  $color?: string;
+  $fontSize?: string | number;
   width: number | string;
   height: number | string;
```
jzempel, line 16: ditto re: `width` and `height`

packages/pagination/src/styled/OffsetPagination/StyledPageBase.ts
```diff
@@ -69,6 +69,12 @@ const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
       inset: true
     })}
 
+    &:disabled,
+    [aria-disabled='true'] {
+      background-color: transparent;
+      color: ${disabledColor};
+    }
```
jzempel, line 76: Why was this moved? The result is that `:active` styling takes precedent over `:disabled`, which is not correct.

packages/pagination/src/styled/OffsetPagination/StyledPageBase.ts
```diff
@@ -69,6 +69,12 @@ const colorStyles = ({ theme }: ThemeProps<DefaultTheme>) => {
       inset: true
     })}
 
+    &:disabled,
+    [aria-disabled='true'] {
```
jzempel, line 73:
```suggestion
    &[aria-disabled='true'] {
```

packages/tags/demo/stories/TagStory.spec.tsx
jzempel, line 1: I thought we weren't committing to snapshot testing. I don't think it has long-term value here.

packages/tags/demo/stories/__snapshots__/TagStory.spec.tsx.snap
jzempel, line 1: Recommend delete

packages/loaders/src/elements/Dots.tsx
jzempel: Hmm, bit of a bummer about that filtering.

packages/dropdowns.legacy/src/styled/multiselect/StyledMultiselectInput.ts
ze-flo: No in this case. This case is tricky to understand. Using `$isCompact` led to a snapshot mismatch.

TL&DR;

`StyledMultiselectInput` extends `StyledInput`, which in turns extends `Input`, a `React FC` that returns a `StyledTextInput`. `Input` only accepts `isCompact`, which is received by `StyledTextInput` after being renamed as transient prop `$isCompact`.

jzempel (approved):

PR #1981, opened 2024-12-02
fix(chrome): provide missing `useId` container dependency for `Sheet` component

jzempel:
## Description

While working on a project with Garden dependencies, I realized that `react-chrome` fails to install in certain environments due to a missing `react-uid` dependency. This is often masked, since `react-notifications` pulls it in. Since Garden no longer recommends `react-uid` in favor of the more stable `useId` from `@zendeskgarden/container-utilities`, I updated the missing dependency and code accordingly (eventually, we'll also replace `uid` in notifications `useToast`).

## Checklist


ze-flo (approved):
Nice catch! 💯

PR #1983, opened 2024-12-04
chore(dropdowns): fix multiselectable combobox tags

jzempel:
## Description

The addition of a child element to the story's custom tags ensures truncation behaves like it would in the default case.

## Detail

**Before**

Screenshot 2024-12-04 at 2 57 27 PM

**After**

Screenshot 2024-12-04 at 2 58 55 PM

## Checklist

- [ ] :guardsman: ~includes new unit tests. Maintain existing coverage (always >= 96%)~

### DIFF
--- a/packages/dropdowns/demo/stories/ComboboxStory.tsx
+++ b/packages/dropdowns/demo/stories/ComboboxStory.tsx
@@ -109,7 +109,7 @@ export const ComboboxStory: Story<IArgs> = ({
             <Svg />
           </span>
         </Tag.Avatar>{' '}
-        {option.tagProps?.children || toLabel(option)}
+        <span>{option.tagProps?.children || toLabel(option)}</span>
       </>
     ) : undefined;
 


packages/dropdowns/demo/stories/ComboboxStory.tsx
```diff
@@ -109,7 +109,7 @@ export const ComboboxStory: Story<IArgs> = ({
             <Svg />
           </span>
         </Tag.Avatar>{' '}
-        {option.tagProps?.children || toLabel(option)}
+        <span>{option.tagProps?.children || toLabel(option)}</span>
```
ze-flo, line 112: Let's remember to update the Website [example](https://github.com/zendeskgarden/website/blob/main/src/examples/combobox/MultiselectTags.tsx#L37) too. 😉

ze-flo (approved):

packages/dropdowns/demo/stories/ComboboxStory.tsx
jzempel: Ah, good call.

PR #1987, opened 2024-12-11
fix(Combobox): improve screen reader experience for Select-Only combobox w NVDA

ze-flo:
## Description

This update fixes how our Comboboxes interact with NVDA during animations. Previously, NVDA users might have experienced incomplete or inaccurate announcements of the selected option when the dropdown list was animating. This was due to the way NVDA processes rapid changes in the page, which sometimes led to announcements being out of sync with the visual state.

### The problem

When interacting with the Select-Only Combobox, NVDA's announcement of the selected option (e.g., "Label combo box Poppy collapsed opens list") is often interrupted by an incorrect announcement of a now-inactive option previously referenced by `aria-activedescendant` (e.g., "Poppy Lee 4 of 16").

Analysis revealed that the dropdown animation is interfering with NVDA’s event queue, causing it to announce outdated or incorrect information. Disabling the animation did resolve the issue, allowing NVDA to announce the correct selected option without interruption, but another solution was preferred.

https://github.com/user-attachments/assets/ca3aef29-cc0d-47aa-9128-1bb7d7f5fbb9

To further demonstrate the problem, the issue was [reproduced](https://codepen.io/zeFlo/pen/azoZJyK) in a modified version of the _ARIA Authoring Practices Guide_ (APG) Select-Only Combobox [example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) by adding a slide animation to the dropdown. The animation caused NVDA to fail in announcing the correct selection.

https://github.com/user-attachments/assets/558b93a0-a654-43db-8291-9e560f461306

### The solution

To address this, we now apply an aria-hidden="true" attribute to the Listbox immediately when it is set to `collapsed` and at the start of the animation sequence. This prevents screen readers from processing its contents during the animation.  Now NVDA correctly process and announce the selected option without interruptions.

https://github.com/user-attachments/assets/4a5ed730-00ed-4db0-95dc-9edaaf1029df

## Regression testing

The fix was tested across various screen readers to ensure compatibility and continued functionality:

**JAWS**: Announcements work as expected.

https://github.com/user-attachments/assets/db561d3b-780c-4bcb-8206-c8bf86853513

**VoiceOver**: Announcements work as expected.
[Video coming soon]

## Checklist

- [] ~~:ok_hand: design updates will be Garden Designer approved (add the designer as a reviewer)~~

### DIFF
--- a/packages/dropdowns/src/elements/combobox/Listbox.tsx
+++ b/packages/dropdowns/src/elements/combobox/Listbox.tsx
@@ -22,9 +22,6 @@ import { ThemeContext } from 'styled-components';
 import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { composeEventHandlers } from '@zendeskgarden/container-utilities';
 
-/**
- * 1. Hide from NVDA when collapsed to avoid incorrect / missing announcements caused by animation
- */
 export const Listbox = forwardRef<HTMLUListElement, IListboxProps>(
   (
     {
@@ -136,7 +133,10 @@ export const Listbox = forwardRef<HTMLUListElement, IListboxProps>(
           $isCompact={isCompact}
           $maxHeight={maxHeight}
           $minHeight={minHeight}
-          aria-hidden={!isExpanded} /* [1] */
+          aria-hidden={
+            // Hide from NVDA when collapsed to prevent incorrect / missing announcements caused by animation
+            !isExpanded
+          }
           onMouseDown={composeEventHandlers(onMouseDown, handleMouseDown)}
           style={{ height }}
           {...props}
--- a/packages/dropdowns/src/elements/combobox/Listbox.tsx
+++ b/packages/dropdowns/src/elements/combobox/Listbox.tsx
@@ -22,6 +22,9 @@ import { ThemeContext } from 'styled-components';
 import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { composeEventHandlers } from '@zendeskgarden/container-utilities';
 
+/**
+ * 1. Hide from NVDA when collapsed to avoid incorrect / missing announcements caused by animation
+ */
 export const Listbox = forwardRef<HTMLUListElement, IListboxProps>(
   (
     {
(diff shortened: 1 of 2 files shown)


jzempel (approved):
brilliant @ze-flo 🙌

packages/dropdowns/src/elements/combobox/Listbox.tsx
```diff
@@ -22,6 +22,9 @@ import { ThemeContext } from 'styled-components';
 import { DEFAULT_THEME } from '@zendeskgarden/react-theming';
 import { composeEventHandlers } from '@zendeskgarden/container-utilities';
 
+/**
+ * 1. Hide from NVDA when collapsed to avoid incorrect / missing announcements caused by animation
```
jzempel, line 26: [nit] this comment could be inlined – the header notation is more of a thing within CSS view components 😉

ze-flo:
@jgorfine-zendesk Any objections to merging this fix?

jgorfine-zendesk:
Nope, none from me, @ze-flo. Having watched your before & after videos, I'd say this is sounding pretty great. 🙌  

Just to confirm: there's no scenario in which there's no animation, right? i.e., we're not unsetting the animation when we detect `prefers-reduced-motion: reduce`, or anything like that?

ze-flo:
> Nope, none from me, @ze-flo. Having watched your before & after videos, I'd say this is sounding pretty great. 🙌
> 
> Just to confirm: there's no scenario in which there's no animation, right? i.e., we're not unsetting the animation when we detect `prefers-reduced-motion: reduce`, or anything like that? I imagine the `aria-hidden="true"` would still have the same effect, even if there weren't an animation, but I figured it's worth asking if it should be conditional (and kick in only if there's an animation). 👀

I'm glad you're bringing that up. Currently, we do not use `prefers-reduced-motion: reduce` to remove animations anywhere. However, it would be beneficial to add it to files like [menuStyles](https://github.com/zendeskgarden/react-components/blob/main/packages/theming/src/utils/menuStyles.ts) to improve the experience for some users. I'll raise it with the team.

For peace of mind, I've tested the current `aria-hidden` logic on all supported SR's with and without animation. All good! 🙌 ✅

PR #1986, opened 2024-12-06
feat(theming): add componentStyles utility

jzempel:
## Description

This is an improved version of `retrieveComponentStyles`, with:
- improved typing (consumer doesn't need to fuss with `data-` attribute exceptions in styled components)
- the ability to specify a custom ID attribute (defaults to `data-garden-id`)
- defaults support no-parameter implementations
- naming aligns better with other Garden view component conventions (`sizeStyles` & `colorStyles`)

## Detail

All instances of the `@deprecated retrieveComponentStyles` will be replaced in a follow-on PR.

## Checklist

- [ ] :ok_hand: ~design updates will be Garden Designer approved (add the designer as a reviewer)~
- [x] :globe_with_meridians: demo is up-to-date (`npm start`)
- [ ] :arrow_left: ~renders as expected with reversed (RTL) direction~
- [ ] :black_circle: ~renders as expected in dark mode~
- [ ] :metal: ~renders as expected with Bedrock CSS (`?bedrock`)~
- [x] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] :wheelchair: ~tested for WCAG 2.1 AA accessibility compliance~
- [ ] :memo: ~tested in Chrome, Firefox, Safari, and Edge~

PR #1989, opened 2024-12-12
chore: replace deprecated `retrieveComponentStyles` with `componentStyles` utility

jzempel:
## Description

Big find/replace follow-on from #1986

### DIFF
--- a/packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
+++ b/packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
@@ -7,7 +7,7 @@
 
 import styled from 'styled-components';
 import { Input } from '@zendeskgarden/react-forms';
-import { retrieveComponentStyles } from '@zendeskgarden/react-theming';
+import { componentStyles } from '@zendeskgarden/react-theming';
 
 const COMPONENT_ID = 'colorpickers.colorpicker_input';
 
@@ -18,5 +18,5 @@ export const StyledInput = styled(Input as any).attrs({
 })`
   text-align: center;
 
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${componentStyles};
 `;
--- a/package-lock.json
+++ b/package-lock.json
@@ -32640,48 +32640,6 @@
       "integrity": "sha512-eZxlbI8GZscaGS7kkc/trHTT5xgrjH3/1n2JDwusC9iahPKWMRvRjJSAN5mCXviuTGQ/lHnhvv8Q1YTpnfz9gA==",
       "dev": true
     },
-    "node_modules/netlify-cli/node_modules/@types/body-parser": {
-      "version": "1.19.2",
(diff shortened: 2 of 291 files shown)


packages/colorpickers/src/styled/ColorPicker/StyledInput.ts
```diff
@@ -18,5 +18,5 @@ export const StyledInput = styled(Input as any).attrs({
 })`
   text-align: center;
 
-  ${props => retrieveComponentStyles(COMPONENT_ID, props)};
+  ${componentStyles};
```
ze-flo, line 21: Food for thought... 

At first glance, one could assume that the `componentId` received by `componentStyles` is `colorpickers.colorpicker_input`. 

Sadly, that's incorrect as this [story](https://zendeskgarden.github.io/react-components/?path=/story/packages-colorpickers-colorpicker--uncontrolled)'s DOM  shows. 😞 


Screenshot 2024-12-12 at 10 51 27 AM

The `data-garden-id` is in fact coming from the component returned by `Input`: `StyledTextInput`.

https://github.com/zendeskgarden/react-components/blob/0d1817e55ae67334997485d943e2e4f7e241d826/packages/forms/src/styled/text/StyledTextInput.tsx#L24

This [PR](https://github.com/zendeskgarden/react-components/pull/1804#issuecomment-2113644493) explains why and what pattern works with great detail.

For the intended behavior, we should explicitly define `componentId`.

```suggestion
  ${(props) => componentStyles({ ...props, componentId: COMPONENT_ID })};
```

However, unless our users found the `colorpickers.colorpicker_input` component id by exploring the source code, it's unlikely they knew the `ColorPicker`'s input could be modified independently from the other `Inputs` only by viewing the DOM. Therefore I think it's acceptable keeping `${componentStyles};`. However, I would remove 
```
```

from `attrs`. It has no impact on the output and creates confusion.

Finally, we agree that how `.attrs` inherits values is confusing. This is why some cases call for ugly [logic](https://github.com/zendeskgarden/react-components/pull/1804/files#diff-5feec13542c185a9f6d9f139bdc364169a96966a91f3de4280b5b006943af075R72-R86) to get the intended result.

This can be all simplified by explicitly setting `data-garden-id` in the element component rather than `.attrs`. I hope we can get to that point in the future. 🤞

jzempel: My instinct is to leave these (false) `COMPONENT_ID` "overrides" in place. Following Garden's original implementation and expectation, these IDs were meant to provide highly targeted component styling overrides without unanticipated component side effects. When we rectify the current implementation with element-specified `data-garden-id`s (most likely in a future major) the existing `COMPONENT_ID` will serve as a breadcrumb for the original intention.

ze-flo: To be clear, the clean up work would only apply to styled components extending another styled component. 

Sounds good to me. We can follow-up on this in the future. 👍

ze-flo (approved):

PR #1990, opened 2024-12-13
chore: set `dist-tag` to `v8`

ze-flo:
## Description
Explicitly sets the `dist-tag` to `v8` instead of the [default](https://www.npmjs.com/package/@lerna/publish#--dist-tag-tag) `latest`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -21,7 +21,7 @@
   ],
   "command": {
     "publish": {
-      "preDistTag": "next",
+      "distTag": "v8",
       "registry": "https://registry.npmjs.org/"
     },
     "version": {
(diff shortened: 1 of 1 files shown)


lerna.json
```diff
@@ -21,7 +21,7 @@
   ],
   "command": {
     "publish": {
-      "preDistTag": "next",
```
jzempel, line 24: In general, we're leaving `preDistTag` intact (along with `preid` below). Not that it makes much/any difference here, but it will help clarify the compare diff between the `main` and `v8` branches.

jzempel (approved):

PR #2009, opened 2025-03-20
fix(forms): remove `renderToString` for the calendar picker icon

jzempel:
## Description

Replace the `react-dom/server` `renderToString` import with a static SVG icon string. The import causes issues in some environments and adds unwanted bloat.

## Detail

https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code


ze-flo (approved):
👍

PR #2032, opened 2025-07-14
feat(theming): allow `null` color scheme key to bypass `ColorSchemeProvider` local storage

jzempel:
## Description

While bypassing `localStorage` would typically be discouraged, there are cases (iframe'd content for example) where `ColorSchemeProvider` usage might not want to compete with a higher-level provider's local storage. This PR introduces `<ColorSchemeProvider colorSchemeKey={null}>` for these cases.

## Checklist


ze-flo (approved):
👍

PR #2033, opened 2025-07-22
feat(dots): update to use pure CSS animation

kgoggin:
## Description
While working on a new feature in Support UI, I ran into a performance issue with the `<Dots/>` and its use of the `useSchedule` hook. When trying to "un-suspend" some UI, the fact that the hook was using a `useEffect` to set state over and over seemed to cause React to need to re-compute the UI over and over and prevented it from ever moving off of the pending UI.

## Detail

The hook was there to enable the `delay` feature of the component. We can replace this with pure CSS animation logic now, so that's what I changed here. I'm animating the `visibility` of the SVG using the delay prop.

Many of the tests are no longer relevant as, to my knowledge, there's not really a good way to unit test CSS animations, so I removed them. 

## Checklist

- [ ] :globe_with_meridians: demo is up-to-date (`npm start`)
- [ ] :black_circle: renders as expected in dark mode
- [ ] :guardsman: includes new unit tests. Maintain existing coverage (always >= 96%)
- [ ] :wheelchair: tested for WCAG 2.1 AA accessibility compliance


jzempel:
> There's one a11y bug in that tab in Storybook:
> 
> ```
> <svg> elements with an img role must have an alternative text
> ```
> 
> I assume this is an existing issue as I didn't change anything about the role attribute. But, if there's a suggested fix for this I'm happy to make it!

Yeah, we see this in the existing https://zendeskgarden.github.io/react-components/?path=/story/packages-loaders-dots--dots. The typical solution would be to add a `aria-label` control to Storybook preset to "Dots" (similar to [Breadcrumbs](https://zendeskgarden.github.io/react-components/?path=/story/packages-breadcrumbs-breadcrumb--breadcrumb)). But we don't need to deal with that in this PR.

packages/loaders/src/elements/Dots.spec.tsx
jzempel, line 1: tbh, it's nice to get rid of these snapshot tests

jzempel (approved):
Nice simplification

PR #2034, opened 2025-07-24
fix(modals): change StyledModal's max-height, when viewport max-height is less than 400 CSS px

jgorfine-zendesk:
## Description

While reviewing outstanding accessibility bugs related to [WCAG 2.2 SC 1.4.4, Resize Text](https://www.w3.org/WAI/WCAG22/Understanding/resize-text.html), Zendesk's Product Accessibility team discovered that the Modal component was at the root of several bugs assigned to consuming teams. At some point, code was introduced into the Modal that causes the scrollbar within its main body — the one that normally allows users to view overflow content — to disappear, when the viewport is 400 CSS pixels or shorter.

This PR adjusts the Modal's `max-height` on shorter viewports, so that the vertical scrollbar will appear, as expected.

### Before

`max-height: none;` set on `role="dialog"` element causes Modal body's scrollbar to become hidden when viewport is less than 400px tall. This prevents users from viewing overflow content.

Garden Modal in Storybook on production, within a 2560px by 320px viewport

### After

`max-height: calc(100vh - 48px;` set on `role="dialog"` element causes Modal body's scrollbar to become visible when viewport is less than 400px tall. This allows users to view overflow content.

Garden Modal in Storybook on localhost, within a 2560px by 320px viewport

## Checklist

### DIFF
--- a/packages/modals/src/styled/StyledModal.spec.tsx
+++ b/packages/modals/src/styled/StyledModal.spec.tsx
@@ -22,6 +22,10 @@ describe('StyledModal', () => {
     expect(container.firstChild).not.toHaveStyleRule('direction');
     expect(container.firstChild).not.toHaveStyleRule('animation-duration', '0.3s');
     expect(container.firstChild).not.toHaveStyleRule('animation-timing-function', 'ease-in-out');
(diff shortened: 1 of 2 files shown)


jzempel (approved):
nice, thanks @jgorfine-zendesk

PR #2114, opened 2026-03-19
fix(accordion): restore `data-garden-id` and `data-garden-version` attributes on `Accordion.Label`

jzempel:
## Description

Restores the attribute used for styling overrides.

## Detail

The `dist/node_modules` directory was not being uploaded to the CI `dist-artifact` which ends up publishing an invalid package without necessary inline SVG references.

- [before](https://github.com/zendeskgarden/react-components/actions/runs/23313562697/job/67806974192#step:8:13) 55 files
- [after](https://github.com/zendeskgarden/react-components/actions/runs/23315096635/job/67812339390?pr=2114#step:8:13) 57 files (aka, includes the 2 missing built SVGs from `node_modules`)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,10 +63,6 @@ jobs:
           cache: pnpm
 
       - run: pnpm install
-      - uses: actions/download-artifact@v8
-        with:
-          name: dist-artifact
-          path: dist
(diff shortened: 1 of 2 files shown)


ze-flo (approved):
Good catch. Still, I haven't quite figured out how we've managed to not need it until now.

PR #2130, opened 2026-05-26
fix(modals): keep tooltip dialog within viewport bounds

bmson:
## Summary
- Add Floating UI shift middleware to TooltipDialog so it repositions horizontally when it would overflow the viewport
- Clamp TooltipDialog width with max-width: 100vw so it never exceeds window bounds
- Update TooltipDialog placement assertions and add coverage for viewport width clamping

## Dependent
https://github.com/zendesk/zendesk_console/pull/51380

###### Before
https://github.com/user-attachments/assets/f2e30320-6a1d-450e-b951-2b2cfb747615

###### After
https://github.com/user-attachments/assets/d0d7836f-d30b-4730-8105-bfc767ff6ea4


packages/modals/src/elements/TooltipDialog/TooltipDialog.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('TooltipDialog', () => {
         await user.click(getByText('open'));
       });
 
-      expect(getByRole('dialog').parentElement).toHaveStyle({ transform: 'translate(-12px, 0px)' });
+      expect(getByRole('dialog').parentElement).toHaveStyle({ transform: 'translate(0px, 0px)' });
```
ze-flo, line 67: From what I've gathered, the change from `translate(-12px, 0px)` to `translate(0px, 0px)` happens because `shift` clamps the offset back to zero in JSDOM's zero-dimension environment. This means the test no longer validates that offset is actually applied. It would pass even if offset were removed entirely.

Here are 2 suggestions for focused tests to give us better coverage: one proving offset works (by giving the trigger real dimensions away from edges), and one proving shift constrains it at edges. 

Note that I haven't tried these specs locally!

```js
it('renders LTR placement correctly', async () => {
  const { getByRole, getByText } = render(<Example placement="start" />);
  const trigger = getByText('open');

  trigger.getBoundingClientRect = () => ({
    x: 500, y: 300, width: 100, height: 40,
    top: 300, left: 500, right: 600, bottom: 340,
    toJSON: () => undefined
  });

  await act(async () => {
    await user.click(trigger);
  });

  expect(getByRole('dialog').parentElement).toHaveStyle({ transform: 'translate(-12px, 0px)' });
});

it('shift clamps offset when trigger is near viewport edge', async () => {
  const { getByRole, getByText } = render(<Example placement="start" />);
  const trigger = getByText('open');

  trigger.getBoundingClientRect = () => ({
    x: 5, y: 300, width: 100, height: 40,
    top: 300, left: 5, right: 105, bottom: 340,
    toJSON: () => undefined
  });

  await act(async () => {
    await user.click(trigger);
  });

  const { transform } = getByRole('dialog').parentElement!.style;

  expect(transform).not.toBe('translate(-12px, 0px)');
});
```

bmson: I ran this locally and both "away from edge" and "near edge" ended up being translate(0px, 0px), so it didn't test the offset behavior.

Mocking only the trigger rect wasn't not enough because Floating UI still computes layout from constrained viewport.

I think we may need Cypress tests to truly test this.

bmson: We may be able to mock `@floating-ui/react-dom`, but not sure if that is really worth it, to be honest.

ze-flo: Agreed. Let's move forward and revisit later if needed.

ze-flo (approved):

PR #2136, opened 2026-06-24
feat(tooltips): add toggletip support

jgorfine-zendesk:
## Description

Adds toggletip support to the `Tooltip` component via a new `isToggletip` prop, backed by the toggletip API in `@zendeskgarden/container-tooltip`. When enabled, the tooltip switches from hover/focus behavior to click-to-toggle: it opens and closes on click, dismisses on outside click or `Escape`, and announces its content through a `role="status"` live region. The trigger must be a `<button>` element.

## Detail

- `isToggletip` is threaded through `Tooltip` into `useTooltip`. The render tree stays unified — the toggletip path applies the live-region props to the always-mounted wrapper and gates the visible bubble on `isAnnouncementReady`, while the tooltip path keeps its hover handlers on the bubble.
- Gating the visible bubble (rather than the live region) avoids an empty-bubble flash on open and preserves screen-reader re-announcement on each open.
- The `Tooltip` story exposes `isToggletip` as a control (default `false`).
- New `[patterns]` demos: **Toggletip**, **Toggletip with anchor** (interactive content), and **Info toggletip** (icon-only `IconButton` trigger with an explicit `aria-label`).

> ⚠️ **Temporary:** this branch depends on the `@zendeskgarden/container-tooltip` `2.2.3-alpha.1` tarball (committed at the repo root) so CI can install it and build the Storybook preview for manual testing. Once the react-containers PR merges and the container is published, the tarball will be removed and the dependency will point at the released `@zendeskgarden/container-tooltip`.

## Checklist


packages/tooltips/package.json
```diff
@@ -22,7 +22,7 @@
   "types": "dist/typings/index.d.ts",
   "dependencies": {
     "@floating-ui/react-dom": "^2.0.0",
-    "@zendeskgarden/container-tooltip": "^2.2.0",
+    "@zendeskgarden/container-tooltip": "file:../../zendeskgarden-container-tooltip-2.2.3-alpha.1.tgz",
```
bmson, line 25: This points `@zendeskgarden/container-tooltip` at a committed local tarball (`file:../../zendeskgarden-container-tooltip-2.2.3-alpha.1.tgz`). If this ships, every consumer of `@zendeskgarden/react-tooltips` inherits a dependency on a path that doesn't exist in the published package → broken install.

But seems like you called that out above and it will be fixed before PR, is merged.

packages/tooltips/demo/~patterns/patterns.mdx
```diff
@@ -14,3 +14,35 @@ menu. The tooltip is designed to hide when the popup is expanded.
 <Canvas>
   <Story of={Stories.Example} />
 </Canvas>
+
+## Toggletip
+
+Setting the `isToggletip` prop switches the `Tooltip` into a toggletip: it opens
+and closes on click rather than hover/focus, dismisses on outside click or
```
bmson, line 21: There's no close path on trigger re-click and the outside-mousedown handler explicitly excludes the trigger. So clicking the trigger a second time re-announces but leaves the toggletip open, with `aria-expanded` stuck at `"true"`. Since the trigger exposes `aria-expanded`/`aria-controls`, a screen-reader user activating the button to collapse it will find it won't close.

Not sure if this is a valid issue

packages/tooltips/demo/~patterns/patterns.mdx
```diff
@@ -14,3 +14,35 @@ menu. The tooltip is designed to hide when the popup is expanded.
...
+`<button>` element. The following example renders rich content with
+`Tooltip.Title` and `Tooltip.Paragraph`.
+
+<Canvas>
+  <Story of={Stories.Toggletip} />
+</Canvas>
+
+## Toggletip with interactive content
+
+Because a toggletip is dismissed on click rather than blur, its content can
```
bmson, line 32: This demo's premise — interactive content that stays open while you interact, doesn't hold under a **mouse** click, given the bundled container's outside dismiss handler:

```js
const clickedTrigger = triggerRef.current?.contains(target);
if (!clickedTrigger) { closeTooltip(0); }   // never checks tooltipRef
```

`tooltipRef` is never consulted, so a mousedown on interactive content *inside* the toggletip counts as an outside click and dismisses it.

The `Anchor` here happens to look fine only because clicking a link is terminal (it navigates as it closes). Any non-navigational interactive content (input, copy button, toggle) will dismiss on mousedown. The outside handler should probably exclude `tooltipRef.current?.contains(target)`. Fix is in the container, but this demo asserts the behavior, so we should probably fix that

packages/tooltips/package.json
jgorfine-zendesk: Definitely. My understanding is that this is the typical way we test container changes before they're merged in. 🙂 👍

packages/tooltips/demo/~patterns/patterns.mdx
jgorfine-zendesk: Hey @bmson, this particular behavior comes from the [Inclusive Components tutorial for toggletips](https://inclusive-components.design/tooltips-toggletips/#inclusivetoggletips), and it's intentional. The idea is that pressing the button again keeps the tooltip open — and refreshes the screen reader announcement — if the tooltip is already open. Screen reader users would hear about the collapsed state of the UI if they put focus on the trigger button, and the tooltip isn't visible.

All that being said: `aria-expanded` should only be stuck on `true` as long as the tooltip is open/visible —every time the trigger button is pressed. If it remains `true` once someone either a) presses the Escape key, or b) blurs the trigger button, tooltip, or tooltip content, then that would be a bug.

packages/tooltips/demo/~patterns/patterns.mdx
jgorfine-zendesk: Hey @bmson, that's a totally fair point, and I appreciate you flagging it.  

Reading all of that, my question is: can/should we try and restrict the kinds of children a toggletip bubble can have, so that it's not even possible to render interactive elements (especially non-navigational interactive content) inside of them? 🤔 

I ask for a few reasons:

1. Having interactive content inside of a toggletip sorta goes against the idea of what a toggletip is: _"toggletips exist to reveal information balloons, and serve no other purpose."_ —Inclusive Components 
2. [Live regions don’t handle rich text](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/#live-regions-don%E2%80%99t-handle-rich-text)
3. [Live regions are not suitable for notifications with interactive elements](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/#live-regions-are-not-suitable-for-notifications-with-interactive-elements)

If we're being honest, even having links inside of toggletips is a bit of a stretch...but I've seen designs with tooltips + links, out in the wild, and I wanted to try and be accommodating. 👀

bmson: Yea, I've seen us have links in the tooltip 😆 
It's probably fine, since it's not really a regression.

I'll leave this one up to you as the expert on this subject.

bmson (approved):
This looks really good 👍🏼 

Should we wait for `@zendeskgarden/container-tooltip` to release, since we've fixed the NPM issue? So we can just ship this packages fully complete?

jgorfine-zendesk:
Hey @bmson, thanks for the approval! NGL, it's been a while, since I've looked at this...I can't remember exactly where I left it, but I believe I was still working on incorporating [this piece of feedback from you](https://github.com/zendeskgarden/react-components/pull/2136#discussion_r3539390030). 👀 Let's not merge it until I've had a chance to revisit it.

PR #2146, opened 2026-08-25
feat(datepickers): align DatePicker and DatePickerRange with the APG combobox patterns

jgorfine-zendesk:
## Description

Brings `DatePicker` and `DatePickerRange` into alignment with the [WAI-ARIA APG combobox-datepicker pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/), closing accessibility gaps around keyboard operability (unfocusable month-nav paddles and day cells), missing name/role/state on calendar controls, and the calendar popover lacking dialog structure — plus a number of visual/interaction bugs found and fixed along the way.

## Detail

**`DatePicker`**
- Input is a real ARIA combobox (`role="combobox"`, `aria-haspopup="dialog"`, `aria-expanded`, `aria-controls`); opens via Down Arrow/Alt+Down Arrow and on pointer click, closes on Escape (returning focus to the input) or on blur outside the widget.
- Calendar popover has real dialog structure (`role="dialog"`, `aria-modal="false"`, labelled by the trigger button).
- Day grid is a real `<table role="grid">` with button-in-cell days, roving tabindex, and full keyboard navigation (arrows, Home/End, PageUp/PageDown, Shift+Page for year).
- Month/year navigation is a real `role="toolbar"` of labeled, focusable buttons (previously unfocusable `div`s).
- `onValueSettled` reports validity (including malformed/out-of-range/required reasons) on blur, Escape, and calendar selection, without disturbing whatever the user typed.

**`DatePickerRange`**
- `Start`/`End` inputs are ARIA comboboxes controlling the shared calendar; day cells are keyboard-operable buttons with the same grid/roving-tabindex/keyboard-navigation model as `DatePicker`, spanning both months.
- `onValueSettled` mirrors `DatePicker`'s, with a `field: 'start' | 'end'` discriminator and an `'out-of-order'` reason when a typed or picked value conflicts with the other field — the conflicting field is now preserved (not silently cleared), so a cleared start date can be re-picked backward without losing the end date.
- Month paddles have accessible names; a redundant, always-hidden inner paddle was removed.
- Fixed the range highlight's day-cell background rendering: it's now the cell's own background rather than a covering overlay, shows the correct half-tint/solid fill at range boundaries, keeps its hover tint visible when the cursor is anywhere in the cell (not just the button), falls back to no tint instead of a disconnected gradient when a hovered boundary candidate lands on the first/last day of its row, and no longer stays stuck highlighted after the cursor moves onto an already-selected boundary day.
- Fixed day-button color states (selected fill, unselected foreground, hover tint) to match production.

## Localization

- Fixed `getStartOfWeek()`'s locale → first-day-of-week fallback, after this bug was reported: the calendar didn't respect a locale's real first day of the week. It now prefers the browser's native `Intl.Locale.prototype.getWeekInfo()`, falling back to a static region/language table when that API is unavailable or the locale string doesn't parse. That static table had itself drifted from real CLDR data — e.g. mapping Japanese (`ja`) to Monday when Japan's first day of the week is Sunday — so 9 stale entries were corrected as part of this fix. Shared by `DatePicker` and `DatePickerRange` via `calendar-utils.ts`, so both benefit.

## Future extraction into `@zendeskgarden/react-containers`

The intention is to eventually extract `useDatePicker`/`useDatePickerRange`'s headless state and prop-getter logic into `@zendeskgarden/react-containers`, alongside the existing `useSelection`/`useMenu` container hooks. As prep for that, this PR also:
- Turns `datepickerReducer`/`datepickerRangeReducer` into plain, pure `(state, action) => state` functions (no closures over host props), matching how `useSelection`/`useMenu` structure their reducers.
- Colocates `useDatePicker.ts` with its reducer under `elements/DatePicker/utils/`, matching `useDatePickerRange`'s existing layout.

No behavior change from this — just structure, so the hook/reducer pairs are easier to lift out as self-contained units later. Flagging it so the reducer/file-layout diffs make sense in review.

## Demo

Added pattern stories demonstrating `onValueSettled`-driven validation, invalid date-range states, and composing `DatePickerRange`'s calendar inside a consumer-managed dialog (mirroring `DatePicker`'s own open/close UX).

## Checklist

- [ ] :black_circle: renders as expected in dark mode
- [ ] :wheelchair: tested for WCAG 2.1 AA accessibility compliance

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bmson (approved):
This PR seems rock solid!

zendesk-lzingarelli:
Hi,  @jgorfine-zendesk 
Here are a few accessibility notes I discovered while testing the Datepicker:
After opening the calendar with the down arrow key, focus currently leaks out of the dialog when tabbing through it there's no focus trap in place. If we add one so focus cycles/wraps within the dialog (matching the W3C datepicker-dialog pattern), it would also solve the next point below.
Right now the Previous/Next buttons can only be reached with Shift+Tab, since focus opens directly on a day cell and skips past the toolbar. That's easy for a screen reader user to miss. With focus cycling in place, Tab would naturally wrap back around to the Previous/Next buttons, making them discoverable without needing Shift+Tab.

jgorfine-zendesk:
Hey @zendesk-lzingarelli, thanks so much for that feedback. The APG pattern I was following _did_ call for `aria-modal="true"`, but I didn't think it was necessary. Now I understand why it is! Whoops. 🙃 Going to make the changes you recommended now.

PR #2148, opened 2026-08-27
feat(forms): add isUnified InputGroup variant and ClearableInput component

jgorfine-zendesk:
## Description

Adds an `isUnified` variant to `InputGroup` that renders as a single bordered/focused field (like the Garden `Input`) instead of the classic overlapping-border segmented look, plus a `focusInset` prop for insetting that container's own focus ring (e.g. when nested inside another group). The unified container now also:

- colors its border/focus ring by a descendant input's validation state, via a new `data-validation` attribute on `StyledTextInput` selected with `:has()` - no `validation` prop needed on `InputGroup` itself
- transitions and highlights its border on hover, matching `Input`, and transitions its focus box-shadow at the same timing as `MediaInput`
- centers its children by default, with only the text input stretching to the container's full height
- sizes any icon button (`IconButton`, `ToggleIconButton`, …) to fit the container by targeting react-buttons' public `data-garden-id` hooks via `:has()`/descendant selectors — no new `size` variant is added to `IconButton`; in the compact case the button is shrunk further, while its icon glyph stays at `iconSizes.md`

Also fixes `InputGroup` to omit its auto-derived `aria-labelledby` when an explicit `aria-label` is provided.

Adds a new `ClearableInput` component built on the unified `InputGroup`: an `Input` paired with a clear button that appears once there's a value to clear. Supports both controlled and uncontrolled usage, returns focus to the input when the clear button is activated (even though the button unmounts as a result), and defaults the clear button's `aria-label` to "Clear" (the icon itself is hidden from assistive technology).

Adds a "Date field" pattern story demonstrating a nested `InputGroup` layout (an inner unified `ClearableInput` alongside a separate bordered calendar `IconButton`) as a step toward the upcoming `DatePicker` component, including a required-field validation example that only flags the field once focus leaves the whole group, not just the inner input.

## Checklist


packages/buttons/src/types/index.ts
```diff
@@ -9,6 +9,9 @@ import { AnchorHTMLAttributes, ButtonHTMLAttributes, SVGAttributes } from 'react
 
 export const SIZE = ['small', 'medium', 'large'] as const;
 
+/** "smallest" is only meaningful for an icon-only button, whose icon glyph never shrinks below `iconSizes.md` */
+export const ICON_BUTTON_SIZE = ['smallest', ...SIZE] as const;
```
ze-flo: Introducing `smallest` deviates from our design specs and clears the possibility for someone to use it incorrectly outside of an InputGroup. Instead I would recommend that an IconButton found inside an InputGroup is correctly sized with just CSS.

packages/forms/src/elements/input-group/InputGroup.tsx
```diff
@@ -8,19 +8,36 @@
...
-        <StyledInputGroup ref={ref} $isCompact={isCompact} {...other} />
+        {/* eslint-disable-next-line jsx-a11y/prefer-tag-over-role */}
+        <StyledInputGroup
+          aria-labelledby={labelId}
+          ref={ref}
+          $isCompact={isCompact}
+          $isUnified={isUnified}
+          $focusInset={focusInset}
+          {...other}
+          role="group"
```
ze-flo: User-defined role (like role="presentation") will be silently dropped. `{...other}` should be placed after `role="group"` otherwise this is a breaking change.

packages/forms/src/styled/input-group/StyledInputGroup.ts
```diff
@@ -48,56 +241,59 @@ const positionStyles = (props: ThemeProps<DefaultTheme> & IStyledInputGroupProps
...
+      & > button[aria-pressed='true'],
+      & > button[aria-pressed='mixed'] {
+        z-index: 1;
+      }
+    `}
 
     & > button:disabled {
       border-top-width: 0;
       border-bottom-width: 0;
     }
```
ze-flo, line 318: This rule exists to hide the horizontal borders of disabled buttons that overlap neighbors in the classic segmented variant. In a unified group, children no longer overlap. Disabling it removes the top/bottom borders, leaving two floating vertical lines.

```suggestion
${!props.$isUnified &&
css`
  & > button:disabled {
    border-top-width: 0;
    border-bottom-width: 0;
  }
`}
```

packages/forms/src/elements/ClearableInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+          id={inputId}
+          data-garden-id="forms.clearable_input"
+        />
+        {hasValue && !disabled && !readOnly ? (
+          <IconButton
+            aria-label={ariaLabel}
+            aria-controls={inputId}
+            isBasic
+            {...buttonProps}
+            onClick={composeEventHandlers(onClick, onClearButtonClick, onClear)}
```
ze-flo: Mouse-clear currently blurs the input before the value is gone.

Here's what happens today:
1. Input is focused, value is "hello".
2. Mousedown on the Clear button. The input blurs. Consumer `onBlur` runs with "hello".
3. The click occurs. The value becomes ''. `onChange` fires. Focus moves back to the input.

Callbacks that track the whether the field has been touched on blur will will see the uncleared value, then a change. Keyboard users already blurred when they tabbed to the button, which is expected. Mouse users shouldn't get a blur in the middle of the clear.

The usual fix is `preventDefault` on the button's `mousedown` so it never takes focus. I recommend composing with `buttonProps.onMouseDown` and keep the consumer first so we don't skip their handler.

Also I'd suggest dropping `onClear`. It duplicates `onChange` and `buttonProps.onClick` (the click). Clearing should just be an `onChange` to `''`. If a consumer needs the click itself, they already have `buttonProps.onClick`.

```suggestion
            onMouseDown={composeEventHandlers(onMouseDown, (event: MouseEvent<HTMLButtonElement>) => {
    event.preventDefault();
  })}
            onClick={composeEventHandlers(onClick, onClearButtonClick)}
```

Let's also add a spec that uses fireEvent.mouseDown then `fireEvent.click` and checks that `onBlur` didn't fire.

packages/forms/src/types/index.ts
```diff
@@ -88,6 +95,23 @@ export interface IInputProps extends IRadioProps {
...
+export interface IClearableInputProps extends Omit<IInputProps, 'value' | 'defaultValue'> {
+  /** Sets the input's value */
+  value?: string;
+  /** Sets the input's default value */
+  defaultValue?: string;
+  /** Fires when the clear button is activated */
+  onClear?: () => void;
+  /** Defines the clear button's `aria-label` */
+  clearButtonLabel?: string;
+  /** Applies props to the wrapping [InputGroup](#inputgroup) element */
```
ze-flo: Going forward, let's not follow the old convention and avoid using section anchors only found in the Garden website. It breaks in Storybook.

packages/forms/src/types/index.ts
```diff
@@ -88,6 +95,23 @@ export interface IInputProps extends IRadioProps {
...
+  wrapperProps?: Omit<IInputGroupProps, 'children'>;
```
ze-flo: Users could break the layout if unsupported props are passed.

```suggestion
  wrapperProps?: Omit<IInputGroupProps, 'children' | 'isUnified'>
```

packages/forms/src/elements/ClearableInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+          >
+            <ClearIcon aria-hidden="true" />
+          </IconButton>
+        ) : null}
+      </InputGroup>
+    );
+  }
+);
+
+ClearableInput.displayName = 'ClearableInput';
```
ze-flo, line 122: Proptypes are missing:


```suggestion
ClearableInput.displayName = 'ClearableInput';

ClearableInput.propTypes = {
  isCompact: PropTypes.bool,
  isBare: PropTypes.bool,
  focusInset: PropTypes.bool,
  validation: PropTypes.oneOf(VALIDATION),
  clearButtonLabel: PropTypes.string,
  wrapperProps: PropTypes.object,
  wrapperRef: PropTypes.any,
  buttonProps: PropTypes.object
};
```

packages/forms/src/elements/ClearableInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { composeEventHandlers, useId } from '@zendeskgarden/container-utilities';
+import { useText } from '@zendeskgarden/react-theming';
+import { IconButton } from '@zendeskgarden/react-buttons';
+import ClearIcon from '@zendeskgarden/svg-icons/src/16/x-stroke.svg';
+import { IClearableInputProps } from '../types';
+import useFieldContext from '../utils/useFieldContext';
+import { Input } from './Input';
+import { InputGroup } from './input-group/InputGroup';
+
+export const ClearableInput = React.forwardRef<HTMLInputElement, IClearableInputProps>(
```
ze-flo, line 23: To follow the repo convention:

```suggestion
/**
 * @extends InputHTMLAttributes<HTMLInputElement>
 */
export const ClearableInput = React.forwardRef<HTMLInputElement, IClearableInputProps>(
```

packages/forms/src/elements/ClearableInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+      }
+    };
+
+    return (
+      <InputGroup
+        isUnified
+        focusInset={false}
+        isCompact={isCompact}
+        {...wrapperProps}
+        ref={wrapperRef}
```
ze-flo, line 91: `wrapperProps` can accidentally replace the props we want to control.

Also `MediaInput` forwards wrapper clicks to the input. ClearableInput does not. Clicks on padding/gap do nothing.

```suggestion
        focusInset={false}
        {...wrapperProps}
        isUnified
        isCompact={isCompact}
        onClick={composeEventHandlers(wrapperProps?.onClick, () => {
    inputRef.current?.focus();
  })}
        ref={wrapperRef}
```

packages/forms/src/types/index.ts
```diff
@@ -88,6 +95,23 @@ export interface IInputProps extends IRadioProps {
   validation?: Validation;
 }
 
+export interface IClearableInputProps extends Omit<IInputProps, 'value' | 'defaultValue'> {
+  /** Sets the input's value */
+  value?: string;
+  /** Sets the input's default value */
+  defaultValue?: string;
+  /** Fires when the clear button is activated */
+  onClear?: () => void;
```
ze-flo: I recommend removing `onClear`. It duplicates `onChange` and `buttonProps.onClick`, with a weaker type (`() => void`)

Remove the prop, the compose slot, the specs that assert it, and the JSDoc. After the drop, `composeEventHandlers(onClick, onClearButtonClick)` is enough. `preventDefault()` on `buttonProps.onClick` still cancels the clear (Garden convention). Document that on `buttonProps`.

```suggestion
```

packages/forms/src/styled/text/StyledTextInput.tsx
```diff
@@ -232,9 +232,10 @@ export interface IStyledTextInputProps {
 }
 
 export const StyledTextInput = styled.input.attrs<IStyledTextInputProps>(props => ({
-  'data-garden-id': COMPONENT_ID,
+  'data-garden-id': (props as any)['data-garden-id'] || COMPONENT_ID,
   'data-garden-version': PACKAGE_VERSION,
-  'aria-invalid': isInvalid(props.$validation)
+  'aria-invalid': isInvalid(props.$validation),
+  'data-validation': props.$validation
```
ze-flo: I'd recommend not shipping the `data-validation` attribute. Garden's only documented DOM hooks are `data-garden-id` and `data-garden-version`. A generic `data-validation` rendered by every StyledTextInput-based component (Input, Textarea, Select, MediaInput, FauxInput) turns into public styling API the day it ships. It can also overwrite a consumer's own `data-*` attribute and we can't remove it later without a breaking change.

I have a [branch](https://github.com/zendeskgarden/react-components/compare/jgorfine/datepicker-bug-bash/1.5/input-group-variant...ze-flo/input-group-patch?expand=1) with the alternative. The Input registers its `validation` with the nearest unified group over the existing `InputGroupContext`, and the group styles itself from a transient `$validation` prop. Same colors, same focus ring, nothing new on the DOM. Cherry-pick `d12ef6c4dff476086004454aa75b27b4a1c7ff8b` whenever you're ready.

Happy to walk through it if easier.

packages/forms/src/styled/input-group/StyledInputGroup.ts
```diff
@@ -6,18 +6,287 @@
...
+      ${focusStyles({
+        theme,
+        inset: $focusInset,
+        color: { variable: VALIDATION_BORDER_VARIABLE[validation] },
+        selector: `&:focus-within:has(${selector}):not(:has(button:focus-visible))`,
+        styles: { borderColor }
+      })}
+    `;
+  });
+};
```
ze-flo: See this [comment](https://github.com/zendeskgarden/react-components/pull/2148/changes#r3919513036).

packages/buttons/src/types/index.ts
jgorfine-zendesk: Hey @ze-flo, I'm going to mark this comment as resolved, since this change has already been incorporated (and the new size variant removed.)

packages/forms/src/types/index.ts
jgorfine-zendesk: Hey @ze-flo, which part of this is the section anchor?

packages/forms/src/types/index.ts
jgorfine-zendesk: Hey @ze-flo, are you saying we wouldn't want to allow users to use a `ClearableInput` with the `isUnified` styles? 🤔

packages/forms/src/types/index.ts
ze-flo: Syntax like `[InputGroup](#inputgroup)` refers to a section anchor only present on the Garden website.

packages/forms/src/types/index.ts
ze-flo: I may have misunderstood since the `ClearableInput` story doesn't let the user control the `isUnified` prop. We can support an `isUnified={false}`  override if Design allows it. It might be worth checking with them.

packages/forms/src/elements/input-group/InputGroup.tsx
jgorfine-zendesk: Hey @ze-flo, it's worth noting that, if someone does override this role with something like `role="presentation"` or `role="none"`, it puts them at risk of violating [WCAG 2.2 SC 1.3.1, Info and Relationships](https://aaardvarkaccessibility.com/wcag-plain-english/1-3-1-info-and-relationships/). Is that a gap we still want to leave open? 

The only other role I could imagine being a good override here is [search](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/search.html). What if we allow someone to pass that, but not other roles that wouldn't make sense?

ze-flo: I'd prefer we keep the override.

`role="group"` is the right default and I agree `presentation`/`none` can fail 1.3.1 if someone uses `InputGroup` as the actual grouping.

Locking `role` (or allowing only `search`) is still a silent breaking change. The component already `@extends HTMLAttributes<HTMLDivElement>`, and today it doesn't set a role.

That's also the pattern we use elsewhere. For ex, `Alert` defaults to `alert` and still lets you change or remove it. `Tiles` does `role="radiogroup"` then `{...props}`.

Restricting the override to `search` still drops every other role people can pass today. `search` is also a landmark, so it usually belongs on a parent `<search>` or form, not on the `InputGroup`.

This PR also nests InputGroups (`ClearableInput` inside the date field). Both would get `role="group"` and the same Field `aria-labelledby`. `role="none"` on the inner wrapper is a valid way to flatten that tree for simplification.

I'd keep:

```
role="group"
{...other}
```

The docs can say `group` is the intended role. If someone overrides it, that's on them, same as any other HTML attribute we forward.

jgorfine-zendesk: Ok! Sounds reasonable to me. Thanks for walking me through it.

packages/forms/src/styled/input-group/StyledInputGroup.ts
jgorfine-zendesk: Fixed 👍

packages/forms/src/types/index.ts
jgorfine-zendesk: Oh, got it! Great catch. Will remove.

packages/forms/src/types/index.ts
jgorfine-zendesk: Ah, wait, sorry, I'm getting turned around. I believe we do NOT want someone to change the `ClearableInput` to resemble the original, segmented `InputGroup`. (I was getting this mixed up with `focusInset`.) I'll lock that down.

ze-flo (approved):
Great win for our consumers needing an a11y compliant `ClearableInput`. Thanks for addressing the PR review comments. Great job! 🔥 🚀

